PR #321, opened 2020-01-27
[MNG-5669] same pom.xml is read multiple times

rfscholte:

       pulling in other changes.
       commit message.

### DIFF
--- a/maven-model-builder/src/main/java/org/apache/maven/model/building/DefaultModelBuilder.java
+++ b/maven-model-builder/src/main/java/org/apache/maven/model/building/DefaultModelBuilder.java
@@ -19,10 +19,11 @@ package org.apache.maven.model.building;
  * under the License.
  */
 
-
+import org.apache.maven.artifact.versioning.ArtifactVersion;
 import org.apache.maven.artifact.versioning.DefaultArtifactVersion;
 import org.apache.maven.artifact.versioning.InvalidVersionSpecificationException;
 import org.apache.maven.artifact.versioning.VersionRange;
+import org.apache.maven.building.Source;
 import org.apache.maven.model.Activation;
 import org.apache.maven.model.Build;
 import org.apache.maven.model.Dependency;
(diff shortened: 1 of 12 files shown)


maven-model-builder/src/main/java/org/apache/maven/model/building/DefaultModelBuilder.java
```diff
@@ -627,9 +644,71 @@ private Model readModel( ModelSource modelSource, File pomFile, ModelBuildingReq
...
         return model;
     }
 
+    private Model getModelFromCache( ModelSource modelSource, ModelCache cache )
+    {
+        Model model;
+        if ( modelSource instanceof ArtifactModelSource )
+        {
+            ArtifactModelSource artifactModelSource = ( ArtifactModelSource ) modelSource;
+            ModelData modelData = getCache( cache, artifactModelSource.getGroupId(),
```
mthmulders: The method name `getCache` suggests that it returns a cache. But one of its parameters is _named_ "cache" and it returns a `ModelData` instance. Does the name `getCache` match with what the method does?

maven-model-builder/src/main/java/org/apache/maven/model/building/DefaultModelBuilder.java
```diff
@@ -1352,6 +1449,19 @@ private void importDependencyManagement( Model model, ModelBuildingRequest reque
         return null;
     }
 
+    private <T> T getCache( ModelCache modelCache, Source source, ModelCacheTag<T> tag )
```
mthmulders: The method name `getCache` suggests that it returns a cache. But one of its parameters is _named_ "cache" and it returns a `<T>` instance. Does the name `getCache` match with what the method does?

rfscholte: Somebody else introduced `getCache()` and `putCache()` based on GAV. My guess: he had to choose between `get()`/`put()` and `getFromCache()`/`putIntoCache()` and chose a little of both( poldermodel ;) ), resulting in probably the worst, especially for `getChache()`.
I went for consistency, but since these are private methods we should be safe to rename them.

rfscholte:
Tests are covered by the maven-integration-testing project, and especially this code is too hard to unittest. Almost every Maven integrationtest needs a pom, hence this code is already touched a lot.

rfscholte:
Fixed in https://github.com/apache/maven/commit/716cc1fe02661897232a7cc3e4c1bb3b3df3b832

jira-importer:
Resolve #6664

PR #329, opened 2020-02-16
[MNG-5760] resume from without args

MartinKanters:

 - [x] Make sure there is a [JIRA issue](https://issues.apache.org/jira/browse/MNG-5760) filed 
       pulling in other changes.
       commit message.

This PR solves MNG-5760 which allows for resuming the build from the last failed module.
It works by saving a file in the build directory of the root pom including that module as file content.

Relevant integration test:
https://github.com/apache/maven-integration-testing/pull/56

### DIFF
--- a/maven-core/src/main/java/org/apache/maven/execution/DefaultMavenExecutionRequest.java
+++ b/maven-core/src/main/java/org/apache/maven/execution/DefaultMavenExecutionRequest.java
@@ -302,16 +302,14 @@ public class DefaultMavenExecutionRequest
         return excludedProjects;
     }
 
+    /**
+     * @return the project where the build should be resumed from,
+     *  or empty when it should start from the last failed project.
+     */
     @Override
     public String getResumeFrom()
     {
         return resumeFrom;
-    }
-
-    @Override
-    public boolean isResumeFromLastFailedProject()
-    {
-        return resumeFromLastFailedProject;
     }
 
     @Override
@@ -606,18 +604,15 @@ public class DefaultMavenExecutionRequest
         return this;
     }
 
+    /**
+     * @param project The project where the build should be started from,
+     *                or empty string when it should resume from the last failed project.
+     * @return this MavenExecutionRequest
+     */
     @Override
     public MavenExecutionRequest setResumeFrom( String project )
     {
         this.resumeFrom = project;
-
-        return this;
-    }
-
-    @Override
-    public MavenExecutionRequest setResumeFromLastFailedProject()
-    {
-        this.resumeFromLastFailedProject = true;
 
         return this;
     }
--- a/maven-core/src/main/java/org/apache/maven/execution/DefaultMavenExecutionRequest.java
+++ b/maven-core/src/main/java/org/apache/maven/execution/DefaultMavenExecutionRequest.java
@@ -118,6 +118,8 @@ public class DefaultMavenExecutionRequest
 
     private String resumeFrom;
 
+    private boolean resumeFromLastFailedProject = false;
+
     private String makeBehavior;
 
     private Properties systemProperties;
@@ -304,6 +306,12 @@ public class DefaultMavenExecutionRequest
     public String getResumeFrom()
     {
         return resumeFrom;
+    }
+
+    @Override
+    public boolean isResumeFromLastFailedProject()
+    {
+        return resumeFromLastFailedProject;
     }
 
     @Override
@@ -602,6 +610,14 @@ public class DefaultMavenExecutionRequest
     public MavenExecutionRequest setResumeFrom( String project )
     {
         this.resumeFrom = project;
+
+        return this;
+    }
+
+    @Override
+    public MavenExecutionRequest setResumeFromLastFailedProject()
+    {
+        this.resumeFromLastFailedProject = true;
 
         return this;
     }
--- a/maven-core/src/main/java/org/apache/maven/graph/DefaultGraphBuilder.java
+++ b/maven-core/src/main/java/org/apache/maven/graph/DefaultGraphBuilder.java
@@ -344,6 +344,12 @@ public class DefaultGraphBuilder
                         "Could not determine build directory for root project", request.getPom() ) );
 
         Path resumeFromCacheFile = Paths.get( buildDirectory, "resume-from-cache" );
+        if ( !resumeFromCacheFile.toFile().isFile() )
+        {
+            throw new MavenExecutionException( "--resume-from without args expects a file called resume-from-cache "
+                    + "in the build directory with the last failed project descriptor.", request.getPom() );
+        }
+
         try ( Stream<String> allLines = Files.lines( resumeFromCacheFile ) )
         {
             return allLines.findFirst().orElseThrow(
--- a/maven-core/src/main/java/org/apache/maven/graph/DefaultGraphBuilder.java
+++ b/maven-core/src/main/java/org/apache/maven/graph/DefaultGraphBuilder.java
@@ -20,6 +20,10 @@ package org.apache.maven.graph;
  */
 
 import java.io.File;
+import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.Paths;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
@@ -29,6 +33,7 @@ import java.util.Iterator;
 import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Map;
+import java.util.stream.Stream;
 
 import org.apache.maven.DefaultMaven;
 import org.apache.maven.MavenExecutionException;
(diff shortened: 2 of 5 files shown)


MartinKanters:
I'm interested in knowing your opinions on what the behavior should be when someone tries to run `-rf` (without args) on a project which does not have the cached file there. Currently it fails hard, but I think it might be better to just include all projects in the reactor and have a warning log say that `-rf` did not work.

Also I'm interested in a good place to store the constant for the filename (currently called resume-from-cache) which is currently copied around in a couple places.

maven-core/src/main/java/org/apache/maven/execution/DefaultMavenExecutionRequest.java
```diff
@@ -118,6 +118,8 @@
 
     private String resumeFrom;
```
rfscholte: I don't think we should split the resume-from into 2 fields.
Just add a clear javadoc saying: `null` is unused, empty unspecified.

maven-core/src/main/java/org/apache/maven/graph/DefaultGraphBuilder.java
```diff
@@ -282,15 +287,42 @@ else if ( StringUtils.isNotEmpty( request.getMakeBehavior() ) )
...
-            String selector = request.getResumeFrom();
+            String selector;
+
+            if ( StringUtils.isNotEmpty( request.getResumeFrom() ) )
+            {
+                selector = request.getResumeFrom();
+            }
+            else
+            {
+                String buildDirectory = projects.stream()
```
mthmulders: I would suggest to refactor this into a dedicated method, something like `determineWhereToResumeBuild()`.

maven-embedder/src/main/java/org/apache/maven/cli/CLIManager.java
```diff
@@ -134,7 +134,7 @@ public CLIManager()
         options.addOption( Option.builder( FAIL_FAST ).longOpt( "fail-fast" ).desc( "Stop at first failure in reactorized builds" ).build() );
         options.addOption( Option.builder( FAIL_AT_END ).longOpt( "fail-at-end" ).desc( "Only fail the build afterwards; allow all non-impacted builds to continue" ).build() );
         options.addOption( Option.builder( FAIL_NEVER ).longOpt( "fail-never" ).desc( "NEVER fail the build, regardless of project result" ).build() );
-        options.addOption( Option.builder( RESUME_FROM ).longOpt( "resume-from" ).hasArg().desc( "Resume reactor from specified project" ).build() );
+        options.addOption( Option.builder( RESUME_FROM ).longOpt( "resume-from" ).hasArg().optionalArg( true ).desc( "Resume reactor from last failed project or specified project" ).build() );
```
mthmulders: @rfscholte, should the message read "last failed **project**" or "last failed **module**"? I still find myself struggling with terminology every now and then...

maven-embedder/src/main/java/org/apache/maven/cli/MavenCli.java
```diff
@@ -1079,6 +1084,33 @@ private String getResumeFrom( List<MavenProject> mavenProjects, MavenProject fai
         return ":" + failedProject.getArtifactId();
     }
 
+    private boolean createResumeFromCacheFile( MavenExecutionResult result, String resumeFromProject )
+    {
+        String buildDirectory = result.getProject().getBuild().getDirectory();
+        File buildDirectoryFile = new File( buildDirectory );
+        boolean isBuildDirectoryPresent = ( buildDirectoryFile.exists() || buildDirectoryFile.mkdirs() );
```
mthmulders: Nitpicking: no brackets needed.

maven-embedder/src/main/java/org/apache/maven/cli/MavenCli.java
```diff
@@ -1079,6 +1084,33 @@ private String getResumeFrom( List<MavenProject> mavenProjects, MavenProject fai
...
+            catch ( IOException e )
+            {
+                slf4jLogger.debug( "Failed writing last failed project to resume-from-cache file at: {}",
+                        resumeFromCache, e );
+                return false;
+            }
+        }
+        else
+        {
+            slf4jLogger.debug( "Failed creating build directory for root project to store resume-from-cache file" );
```
mthmulders: Wouldn't it make sense to use a little more severe message? Most uses will not see them, so they will not know they are missing out on this great new feature :-)

mthmulders:
I'm impressed by how you implemented this feature, @MartinKanters. Nice work!

As for the file name, you might be able to introduce a constant for it. I'm all for failing "hard" when a user does `-rf` without a selector. It's an invalid use of the switch.

maven-embedder/src/main/java/org/apache/maven/cli/MavenCli.java
```diff
@@ -1079,6 +1084,33 @@ private String getResumeFrom( List<MavenProject> mavenProjects, MavenProject fai
...
+    {
+        String buildDirectory = result.getProject().getBuild().getDirectory();
+        File buildDirectoryFile = new File( buildDirectory );
+        boolean isBuildDirectoryPresent = ( buildDirectoryFile.exists() || buildDirectoryFile.mkdirs() );
+        if ( isBuildDirectoryPresent )
+        {
+            Path resumeFromCache = Paths.get( buildDirectory, "resume-from-cache" );
+            try
+            {
+                Files.write( resumeFromCache, resumeFromProject.getBytes() );
```
rfscholte: Will work for now, but the MavenExecutionRequest might contain more information worth storing

maven-embedder/src/main/java/org/apache/maven/cli/CLIManager.java
rfscholte: Module might cause more confusion due to JPMS. However it is a Maven MultiModule project... sorry, didn't answer your question ;)

maven-core/src/main/java/org/apache/maven/execution/DefaultMavenExecutionRequest.java
MartinKanters: Alright! I've just changed this behavior.

maven-core/src/main/java/org/apache/maven/graph/DefaultGraphBuilder.java
MartinKanters: I have refactored this in a method now, but I called it `getLastFailedProject`. What do you think?

maven-embedder/src/main/java/org/apache/maven/cli/CLIManager.java
MartinKanters: 'Project' was already there, I've at least wanted to keep it consistent. Let me know if we want module instead, I'll change it gladly

maven-embedder/src/main/java/org/apache/maven/cli/MavenCli.java
MartinKanters: Sure, I was doubting about it. Anyway, I made them warnings now. @rfscholte do you think this is the proper logging severity?

maven-embedder/src/main/java/org/apache/maven/cli/MavenCli.java
MartinKanters: Right, do you think we should make it in a more extensible format like json already? Or should we refactor it when we see the need for other bits of info?

maven-core/src/main/java/org/apache/maven/graph/DefaultGraphBuilder.java
rfscholte: Yes, I prefer the `getLastFailedProject`

maven-embedder/src/main/java/org/apache/maven/cli/MavenCli.java
rfscholte: If we say that it should be possible to run a build without warnings, then this message should be info, because it doesn't effect the build itself, only the efficiency of the (next) build. (I know this  message is only shown after an error, but still...)

maven-embedder/src/main/java/org/apache/maven/cli/MavenCli.java
rfscholte: Maven doesn't have a json parser. Properties would make more sense here.

maven-embedder/src/main/java/org/apache/maven/cli/CLIManager.java
mthmulders: In that case I would vote to stay with "project".

maven-embedder/src/main/java/org/apache/maven/cli/MavenCli.java
MartinKanters: I like that reasoning, I'll change it to info then!

maven-embedder/src/main/java/org/apache/maven/cli/MavenCli.java
MartinKanters: Ok, but do you think we should implement this already, or only when we need more data there?

MartinKanters:
Some changes are needed: https://issues.apache.org/jira/browse/MNG-5571. Please do not merge yet.

MartinKanters:
Will deliver a new merge request since we decided to go for a different approach.

PR #331, opened 2020-03-02
[MNG-6873] algin junit version to 4.13

calvinhkf:
align library versions...


calvinhkf:
@slachiewicz  I was trying to keep junit version consistent as the issue reports.  Updated two junit dependencies to 4.13.

slachiewicz:
maybe more places to fix, like here (searched for <artifactId>junit</
maven-embedder\src\examples\simple-project\pom.xml
maven-core\src\test\resources\org\apache\maven\lifecycle\pom.xml

calvinhkf:
@slachiewicz hi, I have  re-run the dependency analysis and have included another 5 files that are missed previously and junit versions are all upgrated to 4.13...

slachiewicz:
Change for Junit merged to master - thx!

PR #341, opened 2020-05-02
[MNG-6863] Support --also-make flag in combination with --resume-from. - Refactoring

MartinKanters:
Refactoring the DefaultGraphBuilder class by mainly using the Java 8 streams API.

       pulling in other changes.
       commit message.


maven-core/src/main/java/org/apache/maven/graph/DefaultGraphBuilder.java
```diff
@@ -144,43 +148,27 @@
...
+            File reactorDirectory = getReactorDirectory( request );
 
-            File reactorDirectory = null;
-            if ( request.getBaseDirectory() != null )
-            {
-                reactorDirectory = new File( request.getBaseDirectory() );
-            }
-
-            Collection<MavenProject> selectedProjects = new LinkedHashSet<>( projects.size() );
+            Collection<MavenProject> selectedProjects = new LinkedHashSet<>( request.getSelectedProjects().size() );
```
mthmulders, line 153: Why change the initial size of the `selectedProjects`? It used to be `projects.size()`, now it's `request.getSelectedProjects().size()`.

mthmulders:
I love how concise the code gets by applying Streams here. Looks good to me, one small question left.

maven-core/src/main/java/org/apache/maven/graph/DefaultGraphBuilder.java
MartinKanters: Since this set will only get as large as the amount of selected projects (-pl). I should maybe go for this amount * 2 in order for Java not unnecessarily expanding the list. What do you think?

mthmulders: I've seen this slight over-allocation a lot in the Maven codebase. Personally, it feels like premature optimisation, and arguing that it'll never grow _over_ `selectedProjects.size()` makes perfect sense to me. Although you should specify the `loadFactor` to be 1.0 as well in that case, otherwise it'll still grow since the default `loadFactor` is 0.75.

MartinKanters: Sounds like a good solution! Since we are initializing the set with an initial capacity, I think setting the loadfactor to 1 makes perfect sense. Committed the change!

rfscholte:
Fixed in https://github.com/apache/maven/commit/1e217703a8f6d6378e8a92c57878f53368b289af

PR #344, opened 2020-05-10
Maven metrics runtime Proof-of-Concept playground

eolivelli:
This is a demo of how the Maven Metrics API can be used to instrument Maven Core.


gnodet:
I think this could be revived and adapted to master which has changed quite a bit.

PR #351, opened 2020-05-24
Refactor MavenCli.populateRequest

mthmulders:
While working on the _resume_ feature (#342) I noticed that the [`MavenCli#populateRequest` method](https://github.com/apache/maven/blob/c6c7311713b0d60e445b2ffe32c5588791edf7ac/maven-embedder/src/main/java/org/apache/maven/cli/MavenCli.java#L1322) is very long and a bit unstructured. It follows all types of patterns for parsing and processing command line arguments and populating the `MavenExecutionRequest`.

I've refactored this method step-by-step. Hence there are a lot of small(ish) commits but I hope this makes it easier to review the changes and verify they don't break anything. The method itself is not easily testable so I couldn't rely on that. Where possible I created a few unit tests.

There are some bits & pieces left:

* The `determineProjectActivation` and `determineProfileActivation` methods are very much alike. We could merge them but it would lose quite some semantics, so I initially decided not to do so.
* The `determineLocalRepositoryPath` only uses information from the `MavenExecutionRequest` itself. I wonder if we could move it into the `DefaultMavenExecutionRequest`, but I'm unsure where / how exactly.

## Checklist

- [ ] ~~Make sure there is a [JIRA issue](https://issues.apache.org/jira/browse/MNG) filed for the change (usually before you start working on it).  Trivial changes like typos do not require a JIRA issue.  Your pull request should address just this issue, without pulling in other changes.~~
- [ ] ~~Format the pull request title like `[MNG-XXX] - Fixes bug in ApproximateQuantiles`, where you replace `MNG-XXX` with the appropriate JIRA issue. Best practice is to use the JIRA issue title in the pull request title and in the first line of the commit message.~~

I have signed the Individual Contributor License Agreement, and my employer signed the Corporate Contributor License Agreement.

### DIFF
--- a/maven-core/src/main/java/org/apache/maven/execution/DefaultMavenExecutionRequest.java
+++ b/maven-core/src/main/java/org/apache/maven/execution/DefaultMavenExecutionRequest.java
@@ -64,9 +64,9 @@ public class DefaultMavenExecutionRequest
 
     private boolean interactiveMode = true;
 
-    private boolean cacheTransferError;
+    private boolean cacheTransferError = false;
 
-    private boolean cacheNotFound;
+    private boolean cacheNotFound = false;
 
     private List<Proxy> proxies;
 
@@ -159,7 +159,7 @@ public class DefaultMavenExecutionRequest
      *
      * @issue MNG-2681
      */
-    private boolean noSnapshotUpdates;
+    private boolean noSnapshotUpdates = false;
 
     private boolean useLegacyLocalRepositoryManager = false;
 
--- a/maven-core/src/main/java/org/apache/maven/execution/DefaultMavenExecutionRequest.java
+++ b/maven-core/src/main/java/org/apache/maven/execution/DefaultMavenExecutionRequest.java
@@ -692,7 +692,7 @@ public class DefaultMavenExecutionRequest
     {
         if ( pomFilename != null )
         {
-            this.setPom( new File( pomFilename ) );
+            pom = new File( pomFilename );
         }
 
         return this;
@@ -702,11 +702,6 @@ public class DefaultMavenExecutionRequest
     public MavenExecutionRequest setPom( File pom )
     {
         this.pom = pom;
-
-        if ( this.pom != null && this.pom.getParentFile() != null )
-        {
-            this.setBaseDirectory( this.pom.getParentFile() );
-        }
 
         return this;
     }
--- a/maven-core/src/main/java/org/apache/maven/execution/DefaultMavenExecutionRequest.java
+++ b/maven-core/src/main/java/org/apache/maven/execution/DefaultMavenExecutionRequest.java
@@ -692,7 +692,7 @@ public class DefaultMavenExecutionRequest
     {
         if ( pomFilename != null )
         {
-            pom = new File( pomFilename );
+            this.setPom( new File( pomFilename ) );
         }
 
         return this;
@@ -702,6 +702,11 @@ public class DefaultMavenExecutionRequest
     public MavenExecutionRequest setPom( File pom )
     {
         this.pom = pom;
+
+        if ( this.pom != null && this.pom.getParentFile() != null )
+        {
+            this.setBaseDirectory( this.pom.getParentFile() );
+        }
 
         return this;
     }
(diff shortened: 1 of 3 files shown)


maven-core/src/main/java/org/apache/maven/execution/DefaultMavenExecutionRequest.java
```diff
@@ -687,6 +687,11 @@ public MavenExecutionRequest setPom( File pom )
     {
         this.pom = pom;
 
+        if ( this.pom != null && this.pom.getParentFile() != null )
```
rfscholte, line 690: IIRC this will go wrong. There are a couple of plugins that want to rewrite the pom (maven-shade-plugin, flatten-maven-plugin). This pom will be in the target directory, so it can be cleaned up. However, the basedir should stay the same.

maven-embedder/src/main/java/org/apache/maven/cli/MavenCli.java
```diff
@@ -19,6 +19,7 @@
  * under the License.
  */
 
+import com.google.common.annotations.VisibleForTesting;
```
rfscholte, line 22: This looks like a guava class. We want to avoid guava in our codebase, and only keep it as a transitive dependency for sisu(?). A comment explaining its visibility is good enough

maven-embedder/src/main/java/org/apache/maven/cli/MavenCli.java
```diff
@@ -1318,103 +1322,165 @@ private MavenExecutionRequest populateRequest( CliRequest cliRequest )
...
-        String[] deprecatedOptions = { "up", "npu", "cpu", "npr" };
-        for ( String deprecatedOption : deprecatedOptions )
-        {
-            if ( commandLine.hasOption( deprecatedOption ) )
-            {
-                slf4jLogger.warn( "Command line option -{} is deprecated and will be removed in future Maven versions.",
-                        deprecatedOption );
-            }
-        }
+        warnAboutDeprecatedOptionsUsed( commandLine );
```
rfscholte, line 1334: IMO the name is too long and too explicit. `handleDeprecatedOptions` should be sufficient and leaves enough room for changes, like breaking the build.

maven-core/src/main/java/org/apache/maven/execution/DefaultMavenExecutionRequest.java
mthmulders: Reverted that commit.

maven-embedder/src/main/java/org/apache/maven/cli/MavenCli.java
mthmulders: Replaced the annotation with a comment.

maven-embedder/src/main/java/org/apache/maven/cli/MavenCli.java
mthmulders: Accepted naming suggestion.

rfscholte:
Can't apply this patch anymore, please rebase

mthmulders:
> Can't apply this patch anymore, please rebase

Just rebased against current state of master branch.

maven-embedder/src/main/java/org/apache/maven/cli/MavenCli.java
```diff
@@ -1331,103 +1335,172 @@ private MavenExecutionRequest populateRequest( CliRequest cliRequest )
...
-        boolean showErrors = cliRequest.showErrors;
+        request.setShowErrors( cliRequest.showErrors ); // default: false
+        File baseDirectory = new File( workingDirectory, "" ).getAbsoluteFile();
 
-        String[] deprecatedOptions = { "up", "npu", "cpu", "npr" };
-        for ( String deprecatedOption : deprecatedOptions )
+        handleDeprecatedOptions( commandLine );
+
+        disableOnPresentOption( commandLine, CLIManager.BATCH_MODE, request::setInteractiveMode );
+        enableOnPresentOption( commandLine, CLIManager.SUPRESS_SNAPSHOT_UPDATES, request::setNoSnapshotUpdates );
```
MartinKanters, line 1350: The old functionality sets the `noSnapshotUpdates` to false (or true if the option is there).
Because we do not set it now, it would become implicitly false (because it's a primitive boolean). Did you consider setting the default value explicitly?

mthmulders: Thanks for the suggestion. I just did that. Also, I rebased against the master branch to be sure the integration test would succeed.

MartinKanters (approved):
LGTM! @rfscholte were you done reviewing as well? I can merge it then tomorrow (or @mthmulders if he wants to)

rfscholte (approved):

mthmulders:
Merged in ac80f5c.

PR #356, opened 2020-06-01
[MNG-5868] Adding serval times the same artifact via MavenProjectHelp…

slachiewicz:
…er (attachArtifact) does not produce a failure

o Updated to restore the behavior consensus had been reached in MNG-5387.
o Kept 'MavenProject.getAttachedArtifacts' to return an unmodifiable list.
o Updated to log an informational message when an existing artifact got replaced.

---
This change is [Reviewable](https://reviewable.io/reviews/apache/maven/356)


jira-importer:
Resolve #7565

PR #363, opened 2020-06-21
[MNG-5760] --resume feature (part 2)

mthmulders:
This pull request is a follow-up on #342, in which two little bugs had sneaked in. Since the feature hasn't been released yet, I haven't added new JIRA tickets for these bugs.

1. Resolution of **resume.properties** file didn't work as expected. I have fixed it and added a unit test to illustrate how it should work.
1. When the `excludedProjects` property in the  **resume.properties** file was empty, the build would break (since it added an empty project to the list of excluded projects). Of course, this shouldn't happen, so that's fixed too.

 - [X] Format the pull request title like `[MNG-XXX] - Fixes bug in ApproximateQuantiles`, where you replace `MNG-XXX` with the appropriate JIRA issue. Best practice is to use the JIRA issue title in the pull request title and in the first line of the commit message.
 - [X] You have run the [Core IT][core-its] successfully. Also, the previously added test case for this feature is now added to the suite.

**ICLA has been signed by @MartinKanters and myself. CCLA has been signed by our employer, Info Support.**

### DIFF
--- a/maven-core/src/main/java/org/apache/maven/execution/DefaultBuildResumptionDataRepository.java
+++ b/maven-core/src/main/java/org/apache/maven/execution/DefaultBuildResumptionDataRepository.java
@@ -109,7 +109,7 @@ public class DefaultBuildResumptionDataRepository implements BuildResumptionData
     private Properties loadResumptionFile( Path rootBuildDirectory )
     {
         Properties properties = new Properties();
-        Path path = rootBuildDirectory.resolve( Paths.get( RESUME_PROPERTIES_FILENAME ) );
+        Path path = rootBuildDirectory.resolve( RESUME_PROPERTIES_FILENAME );
         if ( !Files.exists( path ) )
         {
             LOGGER.warn( "The {} file does not exist. The --resume / -r feature will not work.", path );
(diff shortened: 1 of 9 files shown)


maven-core/src/main/java/org/apache/maven/execution/DefaultBuildResumptionDataRepository.java
```diff
@@ -105,7 +109,7 @@ public void removeResumptionData( MavenProject rootProject )
     private Properties loadResumptionFile( Path rootBuildDirectory )
     {
         Properties properties = new Properties();
-        Path path = Paths.get( RESUME_PROPERTIES_FILENAME ).resolve( rootBuildDirectory );
+        Path path = rootBuildDirectory.resolve( Paths.get( RESUME_PROPERTIES_FILENAME ) );
```
kwin, line 112: You can use `Path.resolve(String)` instead of first explicitly converting the argument to Path.

mthmulders: Thanks, didn't know that!

MartinKanters:
Merged in [117cfde44e34a45f2b38929aa164e61650681aeb](https://github.com/apache/maven/commit/117cfde44e34a45f2b38929aa164e61650681aeb)

jira-importer:
Resolve #7909

PR #364, opened 2020-06-24
[MCHECKSTYLE-393] Upgrade Doxia Version to 1.9.2

santik:
[MCHECKSTYLE-393] Upgrade Doxia Version to 1.9.2


santik:
> I don't think this is fixing what you think it's fixing. I'm not sure these need an upgrade.

I don't think it is affects what you think it should fix. It is always good to have recent versions of dependencies anyways.

elharo:
These are integration test project data, not model or even test code. Not sure about these in particular, but I often write integration tests to depend on very specific versions of dependencies. What's being tested here how Maven processes these particular poms so we don't want or need to change them.

PR #365, opened 2020-06-28
[MNG-6949] Add a GitHub Actions Workflow for building and integration…

MartinKanters:
… testing changes

 - [x] Make sure there is a [JIRA issue](https://issues.apache.org/jira/browse/MNG-6949) filed 
       pulling in other changes.
       commit message.

### DIFF
--- a/.github/workflows/maven.yml
+++ b/.github/workflows/maven.yml
@@ -1,0 +1,116 @@
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License.  You may obtain a copy of the License at
+#
+#       http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+
+name: Java CI
+
+on: [push, pull_request]
+
+jobs:
+  build:
+    strategy:
+      matrix:
+        os: [ubuntu-latest, windows-latest, macOS-latest]
+        java: [8, 11, 14]
+      fail-fast: false
+
+    runs-on: ${{ matrix.os }}
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+
+      - name: Set up cache for ~./m2/repository
+        uses: actions/cache@v1
+        with:
+          path: ~/.m2/repository
+          key: maven-${{ matrix.os }}-java${{ matrix.java }}-${{ hashFiles('**/pom.xml') }}
+          restore-keys: |
+            maven-${{ matrix.os }}-java${{ matrix.java }}-
+            maven-${{ matrix.os }}-
+
+      - name: Set up JDK
+        uses: actions/setup-java@v1
+        with:
+          java-version: ${{ matrix.java }}
+
+      - name: Build with Maven
+        run: mvn verify -e -B -V -DdistributionFileName=apache-maven
+
+      - name: Upload built Maven
+        uses: actions/upload-artifact@v2
+        if: ${{ matrix.os == 'ubuntu-latest' && matrix.java == '11' }}
+        with:
+          name: built-maven
+          path: apache-maven/target/
+
+      - name: Upload built Maven Wrapper
+        uses: actions/upload-artifact@v2
+        if: ${{ matrix.os == 'ubuntu-latest' && matrix.java == '11' }}
+        with:
+          name: built-maven-wrapper
+          path: maven-wrapper/target/maven-wrapper.jar
+
+  integration-test:
+    needs: build
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Collect environment context variables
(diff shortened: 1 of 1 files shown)


.github/workflows/maven.yml
```diff
@@ -0,0 +1,116 @@
...
+name: Java CI
+
+on: [push, pull_request]
+
+jobs:
+  build:
+    strategy:
+      matrix:
+        os: [ubuntu-latest, windows-latest, macOS-latest]
+        java: [8, 11, 14]
```
eolivelli, line 27: We can keep just 8 and 11 for github actions.
We already have Jenkins with the full matrix.

.github/workflows/maven.yml
```diff
@@ -0,0 +1,116 @@
...
+      - name: Upload built Maven Wrapper
+        uses: actions/upload-artifact@v2
+        if: ${{ matrix.os == 'ubuntu-latest' && matrix.java == '11' }}
+        with:
+          name: built-maven-wrapper
+          path: maven-wrapper/target/maven-wrapper.jar
+
+  integration-test:
+    needs: build
+    runs-on: ubuntu-latest
```
mthmulders, line 69: Would it be possible (and useful) to run the integration tests on other platforms as well? Say Windows or macOS? I can foresee it will increase the build duration so maybe it doesn't fit in the idea of a "sanity check". But if so, then why do we _build_ it on those platforms?

.github/workflows/maven.yml
```diff
@@ -0,0 +1,116 @@
...
+      - name: Checkout
+        uses: actions/checkout@v2
+
+      - name: Set up cache for ~./m2/repository
+        uses: actions/cache@v1
+        with:
+          path: ~/.m2/repository
+          key: maven-${{ matrix.os }}-java${{ matrix.java }}-${{ hashFiles('**/pom.xml') }}
+          restore-keys: |
+            maven-${{ matrix.os }}-java${{ matrix.java }}-
```
mthmulders, line 42: You might want to consider using a version number in the cache key. If - for any reason - the current caches become unsuitable, you can bump the version number and thus manually invalidate it.

.github/workflows/maven.yml
MartinKanters: Thanks for the feedback! 
On the one hand I'm fine with removing 14 and only keeping the LTSes. On the other hand by keeping the latest non-LTS version we would reduce the risk of it failing on Jenkins eventually. The only downside I see is that we would have to upgrade this version number every half year. Am I missing other disadvantages?

.github/workflows/maven.yml
MartinKanters: I think we could definitely try and enable this on the other platforms. We'll need to make a pre-integration-test job then which at least determines the integration-test repo (either the forked repo or from apache), but perhaps we can do more.

.github/workflows/maven.yml
MartinKanters: Hmm, interesting. How it currently works is that it will try to find a cache exactly matching the key: `maven-${{ matrix.os }}-java${{ matrix.java }}-${{ hashFiles('**/pom.xml') }}`. If that fails it will try to find any of the `restore-keys` caches (which are prefixes of caches). So it is designed to always try to find any maven repository cache.

There is no way to manually clear the cache as far as I know, so if we really want this to work, we should remove the `restore-keys` list and always hit on exactly the cache key. If we would need to clear the cache, we can make a simple change in a pom file which changes the hash in the caching key. It's a bit cumbersome in my opinion.

Alternatively we could also keep it like this and make a change once needed. We can always remove the caching job in this file, when needed. Also I think that the GitHub Actions are only advisory (they won't block PRs), so we could always fallback to local runs.

.github/workflows/maven.yml
mthmulders: We don't make that much cost by adding the latest non-LTS here :-). I'd argue to keep it.

.github/workflows/maven.yml
mthmulders: Hmm, I see, apparently the caching mechanism works differently than it does in CircleCI (where I have often seen the practice of prefixing a cache with a version identifier. Let's keep it as it is.

.github/workflows/maven.yml
MartinKanters: After discussing this offline with @mthmulders and checking the ASF Jenkins build, we chose to run the build only on Java 8 on all platforms. We think the most value lies in running the integration tests against a wide variety of platforms and Java versions, so we introduced the matrix at the integration tests job. Interested in your thoughts on this @eolivelli .

.github/workflows/maven.yml
MartinKanters: With the latest commit, we have enabled the integration tests on the three major platforms and Java versions. Introducing a pre-integration test job was not necessary, since the bash job runs fine on Windows anyway.

.github/workflows/maven.yml
olamy: 11 sounds good to have as well

eolivelli:
Let me merge this change today.
We have enough binding approvals

MartinKanters:
Great to hear guys, thanks for the quick reviews!

sormuras:
Seconds too late to the approval party. ;-)

Looks all green, so far: https://github.com/apache/maven/actions/runs/153922419

slachiewicz:
Awesome, thx for your work

MartinKanters:
> Seconds too late to the approval party. ;-)
> 
> Looks all green, so far: https://github.com/apache/maven/actions/runs/153922419

It's still very much appreciated :) 
It looks like the build was green, except for the macOS integration-test builds. @mthmulders sent a mail on the dev list about this.

PR #373, opened 2020-09-09
[MNG-6118] submodule invocation

MartinKanters:

 - [x] Make sure there is a [JIRA issue](https://issues.apache.org/jira/browse/MNG-6118) filed 
       pulling in other changes.
       commit message.
 - [x] You have run the [Core IT][core-its] successfully. (see https://github.com/infosupport/maven/actions)

---

This pull request enables resolving inter-module dependencies of the whole multi module project while building a part of a multi module project (by using `-f` or by navigating to a subdirectory). 
This makes the following scenarios consistent:
- mvn <goal> -pl module-a
- mvn <goal> -f module-a/pom.xml
- cd module-a && mvn <goal>


maven-core/src/main/java/org/apache/maven/graph/ProjectCollector.java
```diff
@@ -0,0 +1,37 @@
...
+
+import java.io.File;
+import java.util.List;
+
+/**
+ * Facade to collect projects for a given set of pom.xml files.
+ */
+public interface ProjectCollector
+{
+    String HINT = "projectCollector";
```
rfscholte, line 34: No need use hints anymore. If you see "hint", it was used by the Java 1.4  compatible plexus container. We now use @Inject / @Named

rfscholte: Shouldn't this be called `ProjectsCollector`

maven-core/src/main/java/org/apache/maven/graph/MultiModuleCollectionStrategy.java
```diff
@@ -0,0 +1,181 @@
...
+
+    private File getMultiModuleProjectPomFile( MavenExecutionRequest request )
+    {
+        if ( request.getPom().getParentFile().equals( request.getMultiModuleProjectDirectory() ) )
+        {
+            return request.getPom();
+        }
+        else
+        {
+            File multiModuleProjectPom = new File( request.getMultiModuleProjectDirectory(), "pom.xml" );
```
rfscholte, line 101: This might not work with [polyglot-maven](https://github.com/takari/polyglot-maven). IIRC there's a PomLocator that handles the right extension.

maven-core/src/main/java/org/apache/maven/graph/MultiModuleCollectionStrategy.java
```diff
@@ -0,0 +1,181 @@
...
+ */
+@Named( "MultiModuleCollectionStrategy" )
+@Singleton
+public class MultiModuleCollectionStrategy implements ProjectCollectionStrategy
+{
+    private final Logger logger;
+    private final ProjectCollector projectCollector;
+
+    @Inject
+    public MultiModuleCollectionStrategy( Logger logger, ProjectCollector projectCollector )
```
rfscholte, line 55: No need to inject Loggers anymore, instead simply do `private static final Logger LOGGER = LoggerFactory...`

maven-core/src/main/java/org/apache/maven/graph/DefaultProjectCollector.java
```diff
@@ -0,0 +1,99 @@
...
+    {
+        this.logger = logger;
+        this.projectBuilder = projectBuilder;
+    }
+
+    @Override
+    public void collectProjects( List<MavenProject> projects, List<File> files, MavenExecutionRequest request )
+            throws ProjectBuildingException
+    {
+        // TODO [mthmulders] Refactor so the result of the method is returned instead of passed to the method argument.
```
rfscholte, line 60: TODO still valid?

maven-core/src/main/java/org/apache/maven/graph/DefaultGraphBuilder.java
```diff
@@ -361,76 +406,35 @@ private File getReactorDirectory( MavenExecutionRequest request )
...
-                    + result.getProject().getId() );
-
-                for ( ModelProblem problem : result.getProblems() )
-                {
-                    String loc = ModelProblemUtils.formatLocation( problem, result.getProjectId() );
-                    logger.warn( problem.getMessage() + ( StringUtils.isNotEmpty( loc ) ? " @ " + loc : "" ) );
-                }
-
-                problems = true;
+                logger.debug( "Successfully collected projects using the " + strategy.getClass().getSimpleName() );
```
rfscholte, line 423: Doesn't look like a helpful message anymore

maven-core/src/main/java/org/apache/maven/graph/DefaultGraphBuilder.java
```diff
@@ -361,76 +406,35 @@ private File getReactorDirectory( MavenExecutionRequest request )
...
+        List<ProjectCollectionStrategy> projectCollectionStrategies = Arrays.asList(
+                projectlessCollectionStrategy, // 1. Collect project for invocation without a POM.
+                multiModuleCollectionStrategy, // 2. Collect projects for all modules in the multi-module project.
+                requestPomCollectionStrategy   // 3. Collect projects for explicitly requested POM.
+        );
 
-        // We have no POM file.
-        //
-        if ( request.getPom() == null )
+        for ( ProjectCollectionStrategy strategy : projectCollectionStrategies )
```
rfscholte, line 417: This looks weird to me. Now all stragegies are executes, but I would expect that only one really applies.
Instead I would expect to see a swich/if-elseif codeblock selecting the right strategy per case.

rfscholte (changes requested):

maven-core/src/main/java/org/apache/maven/graph/MultiModuleCollectionStrategy.java
mthmulders: Should that be the SLF4J Logger + Factory?

maven-core/src/main/java/org/apache/maven/graph/MultiModuleCollectionStrategy.java
mthmulders: I found a `DefaultModelLocator` but its default implementation does exactly this:

```java
return new File( projectDirectory, "pom.xml" );
```

I found `PomFinder` but that's definitely doing something else.

So maybe there's another class for this? Any hints for its name?

maven-core/src/main/java/org/apache/maven/graph/MultiModuleCollectionStrategy.java
rfscholte: yes

maven-core/src/main/java/org/apache/maven/graph/MultiModuleCollectionStrategy.java
rfscholte: So that means that if we have an AtomModelLocator, it'll select a different pom, so I guess that's the right class. Keep in mind, that if there's a non-default ModelLocator, that's the one that will be picked, otherwise Maven will fall back to this default.

maven-core/src/main/java/org/apache/maven/graph/DefaultGraphBuilder.java
MartinKanters: I think it does (although the diff looks terrible in GitHub..). This change introduces three strategies of collecting projects. This debug log would be the only way to find out which strategy was used in the end (without attaching a debugger).

maven-core/src/main/java/org/apache/maven/graph/MultiModuleCollectionStrategy.java
mthmulders: Just to make sure I understand correctly: the `ModelLocator` interface is the one we need here?

rfscholte: Yes, you will need the ModelLocator Component, the right implementation will be serverd for you.

maven-core/src/main/java/org/apache/maven/graph/ProjectCollector.java
mthmulders: It should, and it now is.

maven-core/src/main/java/org/apache/maven/graph/DefaultGraphBuilder.java
MartinKanters: This logic has indeed changed, with the main reasoning being that it is possible that the second strategy [1] fails and that we will have to fallback on the third strategy [2]. The second strategy can fail when it finds that the multiModuleProjectDirectory is incorrect; aka [not linking to the actual root of the multi-module project](https://github.com/apache/maven/pull/373/files#diff-cdc90272b007188972c28fe364870fb1R73). It can also fail when a module is collected which [requires a plugin in the same multi-module](https://github.com/apache/maven/pull/373/files#diff-cdc90272b007188972c28fe364870fb1R133).

The for loop will break when it found the last strategy to be successful (meaning: one or more projects have been collected).

But maybe the for loop structuring does not improve readability, so we could rework it into an if-else structure as you suggest. WDYT @mthmulders ?

[1] The second strategy is most of what this PR is about: collecting projects based on the multiModuleProjectDirectory/pom.xml.
[2] The third strategy is collecting projects based on the pom in the request (basically the current 3.6.3 behavior).

maven-core/src/main/java/org/apache/maven/graph/DefaultProjectCollector.java
```diff
@@ -42,7 +42,7 @@
  */
 @Named
 @Singleton
-public class DefaultProjectCollector implements ProjectCollector
+public class DefaultProjectCollector implements ProjectsCollector
```
MartinKanters, line 45: I think we should call this class DefaultProjectsCollector.

MartinKanters: I took the liberty to pick this up, hope you are fine with it.

maven-core/src/main/java/org/apache/maven/graph/DefaultGraphBuilder.java
mthmulders: I thought I replied to this one already... Can't find it though.

In fact, the old code already had strategies, except they weren't implemented in separate classes, just all in this class. It wasn't very obvious from the old code, but in fact, it already employed two strategies: if there's no POM (say when you run `archetype:generate`) or when there is a POM. If the former would fail, the second would take over. Now that we're adding a third strategy, we felt we'd make this structure more explicit, including the fact that the strategies are executed one after each other.

rfscholte: I don't mind moving the logic to separate classes, that'll probably improve readability and testability.
I really prefer to make the choice of the strategy explicit instead of magically breaking out of a for-loop.
It will also make it easier to identify the awkward usecase, making the discussion easier to drop this support in the future.

MartinKanters: @mthmulders and I discussed it offline, and we agree. Let's go for more readability and explictness. 
Can you please check the new commit whenever you have some time?

maven-core/src/main/java/org/apache/maven/graph/DefaultProjectCollector.java
mthmulders: Yep, but we didn't want to complicate this PR even further...

mthmulders: It'll be a separate PR

mthmulders (approved):

mthmulders:
[Jenkins running](https://ci-builds.apache.org/job/Maven/job/maven-box/job/maven/job/mng-6118-submodule-invocation/)

maven-core/src/main/java/org/apache/maven/graph/DefaultGraphBuilder.java
```diff
@@ -71,15 +63,23 @@
 public class DefaultGraphBuilder
     implements GraphBuilder
 {
+    private final Logger logger = LoggerFactory.getLogger( getClass() );
```
rfscholte, line 66: private static final Logger logger = LoggerFactory.getLogger( DefaultGraphBuilder.class );

maven-core/src/main/java/org/apache/maven/graph/DefaultProjectsCollector.java
```diff
@@ -0,0 +1,99 @@
...
+    public DefaultProjectsCollector( ProjectBuilder projectBuilder )
+    {
+        this.projectBuilder = projectBuilder;
+    }
+
+    @Override
+    public void collectProjects( List<MavenProject> projects, List<File> files, MavenExecutionRequest request )
+            throws ProjectBuildingException
+    {
+        // TODO [mthmulders] Refactor so the result of the method is returned instead of passed to the method argument.
```
rfscholte, line 60: TODO?

maven-core/src/main/java/org/apache/maven/graph/ProjectCollectionStrategy.java
```diff
@@ -0,0 +1,41 @@
...
+public interface ProjectCollectionStrategy
+{
+    /**
+     *
+     * @param request
+     * @return
+     * @throws ProjectBuildingException
+     */
+    List<MavenProject> collectProjects( final MavenExecutionRequest request )
+            throws ProjectBuildingException;
```
rfscholte, line 40: Shouldn't this be an exception in the same package?

maven-core/src/main/java/org/apache/maven/graph/ProjectlessCollectionStrategy.java
```diff
@@ -0,0 +1,67 @@
...
+import javax.inject.Inject;
+import javax.inject.Named;
+import javax.inject.Singleton;
+import java.util.ArrayList;
+import java.util.List;
+
+/**
+ * Strategy to collect projects for building when the Maven invocation is not in a folder that contains a pom.xml.
+ */
+@Named( "ProjectlessCollectionStrategy" )
```
rfscholte, line 40: `PomlessCollectionStrategy` I guess, because e.g. `help:system` can be executed anywhere, it will provide the standalone project.

maven-core/src/main/java/org/apache/maven/graph/ProjectsCollector.java
```diff
@@ -0,0 +1,43 @@
...
+{
+    /**
+     * Collect Maven projects from a list of POM files.
+     * @param projects List that will be filled with the found projects.
+     * @param files List of POM files.
+     * @param request The {@link MavenExecutionRequest}
+     * @throws ProjectBuildingException In case the POMs are not used.
+     */
+    void collectProjects( List<MavenProject> projects, List<File> files, MavenExecutionRequest request )
+            throws ProjectBuildingException;
```
rfscholte, line 42: Shouldn't this be a an exception in the same package (or should ProjectsCollector be moved to `org.apache.maven.project`?

rfscholte (changes requested):
Some additional remarks

maven-core/src/main/java/org/apache/maven/graph/DefaultGraphBuilder.java
MartinKanters: Done.

maven-core/src/main/java/org/apache/maven/graph/DefaultProjectsCollector.java
MartinKanters: I've removed it, hopefully we won't forget about it, though :)

maven-core/src/main/java/org/apache/maven/graph/ProjectCollectionStrategy.java
MartinKanters: I've moved the Collector related classes to a new package within the `project` package. Is that better?

maven-core/src/main/java/org/apache/maven/graph/ProjectlessCollectionStrategy.java
MartinKanters: Right! I renamed it.

maven-core/src/main/java/org/apache/maven/graph/ProjectsCollector.java
MartinKanters: Same as above: I've moved the Collector related classes to a new package within the project package.

maven-core/src/main/java/org/apache/maven/graph/DefaultProjectsCollector.java
rfscholte: Why not refactor it?

maven-core/src/main/java/org/apache/maven/graph/ProjectCollectionStrategy.java
rfscholte: I think that was indeed the best option (new exception versus move to project package)

maven-core/src/main/java/org/apache/maven/graph/DefaultProjectsCollector.java
mthmulders: We thought this PR is big enough in itself, and prefer to perform the refactoring in a separate PR.

rfscholte (approved):

eolivelli (approved):
+1 but please consider this approval as non binding. I checked the code and it looks good but I don't know very much this part of Maven. I am not sure we are not breaking something.
The fact that integration tests are passing is a good index that this patch is good.

I like this feature. Thanks for providing it

MartinKanters:
> +1 but please consider this approval as non binding. I checked the code and it looks good but I don't know very much this part of Maven. I am not sure we are not breaking something.
> The fact that integration tests are passing is a good index that this patch is good.
> 
> I like this feature. Thanks for providing it

Thanks @eolivelli ! Although non-binding, we are grateful you like the code and the feature. I was confident enough to merge the changes.

jira-importer:
Resolve #7324

PR #384, opened 2020-10-15
[MNG-6957] Build/consumer should work with reverse module order in aggregator

rfscholte:


rfscholte:
I guess this was closed by accident...

PR #386, opened 2020-10-21
[MNG-6999] Chained (consumer) XMLFilters can result in "floating" comments

rfscholte:


maven-core/src/test/java/org/apache/maven/internal/aether/ConsumerModelSourceTransformerTest.java
```diff
@@ -0,0 +1,69 @@
...
+import org.apache.maven.model.building.TransformerContext;
+import org.junit.Test;
+import org.xmlunit.assertj.XmlAssert;
+
+public class ConsumerModelSourceTransformerTest
+{
+    private ConsumerModelSourceTransformer transformer = new ConsumerModelSourceTransformer();
+
+    @Test
+    public void test() throws Exception
```
mthmulders, line 37: Would it be possible to come up with a name for the test that better describes what it verifies?

maven-xml/src/test/java/org/apache/maven/xml/sax/filter/ConsumerPomXMLFilterTest.java
```diff
@@ -231,5 +234,22 @@ public void licenseHeader() throws Exception {
         String actual = transform( input );
         assertThat( actual ).and( expected ).areIdentical();
     }
+    
+    @Test
+//    @Ignore
```
mthmulders, line 239: I think this comment can be removed :-)

maven-xml/src/test/java/org/apache/maven/xml/sax/filter/ConsumerPomXMLFilterTest.java
```diff
@@ -231,5 +234,22 @@ public void licenseHeader() throws Exception {
...
+    public void lexicalHandler() throws Exception
```
mthmulders, line 240: Should this test method not describe what it verifies, i.e. "removes modules" or something like that?

mthmulders (commented):
I left a few remarks but I don't oversee what this code does, nor what the effect of the proposed changes would be.

maven-xml/src/test/java/org/apache/maven/xml/sax/filter/ConsumerPomXMLFilterTest.java
rfscholte: It tests that the lexical handler is used in both SaxSource and StreamResult. If one of both doesn't provide the lexical handler, all comments will be ignored. By default the SaxSource has no LexicalHandler, so the XMLParser will simply get elements, attributes, etc.

But yes, this is now testing 2 different things. I will add a test with a minimal pom + license, that should be enough to verify if the lexicalHander is properly chained.

maven-core/src/test/java/org/apache/maven/internal/aether/ConsumerModelSourceTransformerTest.java
rfscholte: of course, my bad practice when starting a new class with only one method, to simply start with `test`

maven-core/src/test/java/org/apache/maven/internal/aether/ConsumerModelSourceTransformerTest.java
```diff
@@ -0,0 +1,69 @@
...
+    {
+        Path beforePomFile = Paths.get( "src/test/resources/projects/transform/before.pom").toAbsolutePath();
+        Path afterPomFile = Paths.get( "src/test/resources/projects/transform/after.pom").toAbsolutePath();
+
+        try( InputStream in = transformer.transform( beforePomFile, new NoTransformerContext() ) )
+        {
+            XmlAssert.assertThat( in ).and( Files.newInputStream( afterPomFile ) ).areIdentical();
```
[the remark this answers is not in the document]
rfscholte: no it does, good find!

maven-xml/src/test/java/org/apache/maven/xml/sax/filter/ConsumerPomXMLFilterTest.java
rfscholte: true

rfscholte:
if there are no objections, I'll merge this tomorrow

MartinKanters (commented):
I think the code looks good in general, but I added some very small comments, so it's up to you whether you want to process them or not.

maven-core/src/test/java/org/apache/maven/project/ProjectBuilderTest.java
```diff
@@ -249,9 +249,9 @@ public void testReadInvalidPom()
         catch ( ProjectBuildingException ex )
         {
             assertEquals( 1, ex.getResults().size() );
-            MavenProject project = ex.getResults().get( 0 ).getProject();
-            assertNotNull( project );
-            assertNotSame( 0, ex.getResults().get( 0 ).getProblems().size() );
+            assertNotNull( ex.getResults().get( 0 ).getPomFile() );
```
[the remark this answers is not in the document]
rfscholte: Noted. Now that we are using Java 8 it should be quite to catch these leaks by using assertThrows. If you hit such a test class, just fix it as part of continuous maintenance.

PR #400, opened 2020-11-26
[MNG-7035] Migrate all tests to JUnit 4

gnodet:
Migration of all tests to Unit 4 without any refactoring beyond the bare minimum (i.e. add @Test, @Before and @After) methods.
For plexus powered tests, I had to copy the migrated `PlexusTestCase` as there is currently no shared test jar.

### DIFF
--- a/maven-compat/src/test/java/org/apache/maven/PlexusTestCase.java
+++ b/maven-compat/src/test/java/org/apache/maven/PlexusTestCase.java
@@ -1,0 +1,317 @@
+package org.apache.maven;
+
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+/*
+ * Copyright 2001-2006 Codehaus Foundation.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import java.io.File;
+import java.io.InputStream;
+
+import org.codehaus.plexus.ContainerConfiguration;
+import org.codehaus.plexus.DefaultContainerConfiguration;
+import org.codehaus.plexus.DefaultPlexusContainer;
+import org.codehaus.plexus.PlexusContainer;
+import org.codehaus.plexus.PlexusContainerException;
+import org.codehaus.plexus.component.repository.exception.ComponentLifecycleException;
+import org.codehaus.plexus.component.repository.exception.ComponentLookupException;
+import org.codehaus.plexus.configuration.PlexusConfiguration;
+import org.codehaus.plexus.context.Context;
+import org.codehaus.plexus.context.DefaultContext;
+import org.junit.After;
+import org.junit.Before;
+
+/**
+ * This is a slightly modified version of the original plexus class
+ * available at https://raw.githubusercontent.com/codehaus-plexus/plexus-containers/master/plexus-container-default/src/main/java/org/codehaus/plexus/PlexusTestCase.java
+ * in order to migrate the tests to JUnit 4.
+ *
+ * @author Jason van Zyl
+ * @author <a href="mailto:trygvis@inamo.no">Trygve Laugst&oslash;l</a>
+ * @author <a href="mailto:michal@codehaus.org">Michal Maczka</a>
+ * @author Guillaume Nodet
+ */
+public abstract class PlexusTestCase
+{
+    private PlexusContainer container;
+
+    private static String basedir;
+
+    @Before
+    public void setUp()
+            throws Exception
+    {
+        basedir = getBasedir();
+    }
+
+    @SuppressWarnings("ResultOfMethodCallIgnored")
+    protected void setupContainer()
+    {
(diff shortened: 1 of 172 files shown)


slachiewicz (approved):

slachiewicz:
please also look into #346 where I made a related changes

maven-compat/src/test/java/org/apache/maven/PlexusTestCase.java
```diff
@@ -0,0 +1,294 @@
...
+import org.codehaus.plexus.PlexusContainer;
+import org.codehaus.plexus.PlexusContainerException;
+import org.codehaus.plexus.component.repository.exception.ComponentLifecycleException;
+import org.codehaus.plexus.component.repository.exception.ComponentLookupException;
+import org.codehaus.plexus.context.Context;
+import org.codehaus.plexus.context.DefaultContext;
+import org.junit.After;
+import org.junit.Before;
+
+public class PlexusTestCase {
```
slachiewicz, line 38: is it ok to copy code from codehaus-plexus repo without credits?

gnodet: I've fixed the commit to add the copyright, thx !

gnodet:
> please also look into #346 where I made a related changes

This is mostly related to JSR330 afaik.
I think the blocking issue for migrating to JSR330 is [MNG-6898](https://issues.apache.org/jira/browse/MNG-6898) though it should be possible to migrate the tests and most of core and keep that one.
I'm willing to investigate the JSR330 issue later...

gnodet:
Is anybody willing to look at this PR ?
I can update it, but don't want to spend more time if no-one cares...

maven-compat/src/test/java/org/apache/maven/PlexusTestCase.java
```diff
@@ -0,0 +1,317 @@
...
+ * This is a slightly modified version of the original plexus class
+ * available at https://raw.githubusercontent.com/codehaus-plexus/plexus-containers/master/plexus-container-default/src/main/java/org/codehaus/plexus/PlexusTestCase.java
+ * in order to migrate the tests to JUnit 4.
+ *
+ * @author Jason van Zyl
+ * @author <a href="mailto:trygvis@inamo.no">Trygve Laugst&oslash;l</a>
+ * @author <a href="mailto:michal@codehaus.org">Michal Maczka</a>
+ * @author Guillaume Nodet
+ */
+public abstract class PlexusTestCase
```
rfscholte, line 64: Would it make sense to turn this into a rule like done for https://github.com/apache/maven-plugin-testing/blob/master/maven-plugin-testing-harness/src/main/java/org/apache/maven/plugin/testing/MojoRule.java

gnodet: Sure, but this would require lots of additional refactoring to migrate from a parent class to a rule.  Given how big this PR is (and already conflicting), I'd rather to by steps.  This is the same problem than for tackling https://github.com/apache/maven/pull/346.  Having PR reviewed and merged takes time, so I'd rather handle issues one by one.

gnodet:
I've rebased the PR on top of master.

gnodet:
Rebased and fix all conflicts again after MNG-6909.

cstamas (approved):

slachiewicz:
with Sisu we have also org.eclipse.sisu.launch.InjectedTest available.

One comment - new class org.apache.maven.PlexusTestCase should be based on org.codehaus.plexus.PlexusTestCase from org.eclipse.sisu.plexus artefact

gnodet:
> with Sisu we have also org.eclipse.sisu.launch.InjectedTest available.
> 
> One comment - new class org.apache.maven.PlexusTestCase should be based on org.codehaus.plexus.PlexusTestCase from org.eclipse.sisu.plexus artefact

Well, the goal is not have have ties to junit3 and the PlexusTestCase is not junit4 ready, so that kinda removes the purpose of this PR if we want to stick with the junit3 base class in all tests. Did I miss something ?

maven-compat/src/test/java/org/apache/maven/PlexusTestCase.java
```diff
@@ -0,0 +1,317 @@
...
+import org.codehaus.plexus.component.repository.exception.ComponentLookupException;
+import org.codehaus.plexus.configuration.PlexusConfiguration;
+import org.codehaus.plexus.context.Context;
+import org.codehaus.plexus.context.DefaultContext;
+import org.junit.After;
+import org.junit.Before;
+
+/**
+ * This is a slightly modified version of the original plexus class
+ * available at https://raw.githubusercontent.com/codehaus-plexus/plexus-containers/master/plexus-container-default/src/main/java/org/codehaus/plexus/PlexusTestCase.java
```
slachiewicz, line 56: this should refer to https://github.com/eclipse/sisu.plexus/blob/master/org.eclipse.sisu.plexus/src/org/codehaus/plexus/PlexusTestCase.java

slachiewicz:
I am on your side, I am in favor of switching to Junit4 - especially that it would be worthwhile to switch to Junit5.

rmannibucau:
Guess the question is do we switch to junit4 ot junit5 directly today. I think in both cases we need to write the BaseTest as an extension (junit5) or rule (junit4) so the effort seems kind of the same to move to junit5 to me at the moment. Did you evaluate junit5 directly? Is there any issue?

maven-compat/src/test/java/org/apache/maven/PlexusTestCase.java
gnodet: Not really, IIRC, I purposedly used the older codehaus version to avoid the EPL in source form which could cause a licensing problem with the ASL at the ASF afaik.

cstamas:
I'd prefer "baby steps", so this PR as is is okay for me, and let have later go with other changes...

gnodet:
> Guess the question is do we switch to junit4 ot junit5 directly today. I think in both cases we need to write the BaseTest as an extension (junit5) or rule (junit4) so the effort seems kind of the same to move to junit5 to me at the moment. Did you evaluate junit5 directly? Is there any issue?

You're right.  When rewriting the base class, it would be a good time to jump to unit 5.
I did upgrade Camel to junit 5.  Most of it is painless (though the arguments order of the assertions has been changed, so you need some regex skill...), but the pain comes when using specific unit 4 rules.
I'm currently working on another PR to move the 3 duplicated `PlexusTestCase` classes to a new `maven-test-support` module so that they can be shared.

gnodet:
> > Guess the question is do we switch to junit4 ot junit5 directly today. I think in both cases we need to write the BaseTest as an extension (junit5) or rule (junit4) so the effort seems kind of the same to move to junit5 to me at the moment. Did you evaluate junit5 directly? Is there any issue?
> 
> You're right. When rewriting the base class, it would be a good time to jump to unit 5.
> I did upgrade Camel to junit 5. Most of it is painless (though the arguments order of the assertions has been changed, so you need some regex skill...), but the pain comes when using specific unit 4 rules.
> I'm currently working on another PR to move the 3 duplicated `PlexusTestCase` classes to a new `maven-test-support` module so that they can be shared.

Following PR that may be split also between the first commits (cleanup) and the last one (upgrade to junit 5) : https://github.com/apache/maven/pull/431

gnodet:
Incorporated in https://github.com/apache/maven/pull/431

PR #402, opened 2020-11-27
DefaultProjectBuilder enhancements

gnodet:
While initially working on [MNG-7027](https://issues.apache.org/jira/projects/MNG/issues/MNG-7027), I came across several issues, so I'm raising this PR in order to gather feedback.
I've split the changes in small commits in order to make reviewing them easier.
There are 2 big important (and non trivial) commits though: https://github.com/apache/maven/commit/331e402b1a7f6e158444984d1272aeff18f54549 and https://github.com/apache/maven/commit/aa285f20bc2c4aef9898e42e297dd43430ac2c11.  Those have comments in the commit, but I can explain more if needed.

This PR is layered on top of #399 and #391 to avoid conflicts.


gnodet:
On top of that, I'm going to check if caching of *effective* models can be done.  This requires tracking which properties / profiles are used, but it looks doable.  It may not be very interesting for stock maven, but it would certainly be for `mvnd`.

gnodet:
@rfscholte could you have a quick look and see if that's something that could be merged at some point ? I don't want to waste time if there's no interest at all.
The 2 big and non trivial commits are 331e402 and aa285f2. 
I'm willing to rebase if there's a chance it could be included.

gnodet:
@rfscholte the commit I pointed you at and on which you (commented)[https://github.com/apache/maven/commit/aa285f20bc2c4aef9898e42e297dd43430ac2c11#r45727670] yesterday was outdated.  I've rebased on top of the recent changes in master.
The related commit is now https://github.com/apache/maven/pull/402/commits/22582b662401afbb1ee648b5d63be49a8076dbb0.  The answer to the question should be in the commit comment:
```
The checks were previously done after the cache hit, but they are now a duplication of the checks performed in readParentLocally.
Doing caching is useless as the first time, the version range will be resolved and stored in the Parent and caching happens during the call to readRawModel anyway.
```

gnodet:
I'll investigate the integration test failures asap.

rfscholte:
Probably related is MNG-7063: if trying to read the same file multiple times (in this case {{dependency-reduced-pom.xml}}, it'll be picked up from cache. In this case the model shouldn't have been cached.
Would that be solved with these changes?

rfscholte:
I've written a unittest for MNG-7063. With this PR it succeeds again. This looks very promising.

gnodet:
I'll get back on this PR next week.

gnodet:
@rfscholte I've rebased and fix the integration tests.

rfscholte:
I just had another look at this. It looks to me that this commit has become too big. We should split it into 2 pieces: 
* rewriting JUnit tests using maven-test-support. 
* Remove double cache, which would fix MNG-7063.

gnodet:
> I just had another look at this. It looks to me that this commit has become too big. We should split it into 2 pieces:
> 
> * rewriting JUnit tests using maven-test-support.
> * Remove double cache, which would fix [MNG-7063](https://issues.apache.org/jira/browse/MNG-7063).

Done.
Fwiw, https://github.com/apache/maven/pull/432 is about the junit 5.

rfscholte:
ok, let me check that one first

rfscholte:
Merged with https://github.com/apache/maven/commit/619973b91cff5af7b5695bda11761a982a159349 which fixes MNG-7063. Thanks for this PR!

PR #407, opened 2020-12-02
Revert "[MNG-5728] Switch the default checksum policy from "warn" to "fail""

rfscholte:
Reverts apache/maven#405


jira-importer:
Resolve #6581

PR #414, opened 2020-12-23
[MNG-4660] Increase usefulness of logging

mthmulders:
When the user resumes a build (MNG-4660), they could see a message "Packaged artifact is not up-to-date compared to the build output directory". This could even be logged multiple times for one build. It was unclear why it was logged and what a user could do to address it. This change adds a little bit of detail to that message so it becomes more clear to the user what has happened and where to look for solutions.

The [integration test](https://github.com/apache/maven-integration-testing/pull/90) is updated as well.

ICLA and CCLA signed.

### DIFF
--- a/maven-core/src/main/java/org/apache/maven/ReactorReader.java
+++ b/maven-core/src/main/java/org/apache/maven/ReactorReader.java
@@ -267,8 +267,8 @@ class ReactorReader
                 if ( outputFileLastModified > artifactLastModified )
                 {
                     LOGGER.warn(
-                            "Packaged artifact for {} is not up-to-date compared to the build output directory; " +
-                            "file {} is more recent than {}.",
+                            "Packaged artifact for {} is not up-to-date compared to the build output directory; "
+                          + "file {} is more recent than {}.",
                             project.getArtifactId(),
                             relativizeOutputFile( outputFile ), relativizeOutputFile( packagedArtifactFile.toPath() )
                     );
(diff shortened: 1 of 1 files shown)


eolivelli (approved):
Lgtm

mthmulders:
Merged in 7e21a52.

maven-core/src/main/java/org/apache/maven/ReactorReader.java
```diff
@@ -266,7 +266,12 @@ private boolean isPackagedArtifactUpToDate( MavenProject project, File packagedA
                 long outputFileLastModified = Files.getLastModifiedTime( outputFile ).toMillis();
                 if ( outputFileLastModified > artifactLastModified )
                 {
-                    logger.warn( "Packaged artifact is not up-to-date compared to the build output directory" );
+                    LOGGER.warn(
+                            "Packaged artifact for {} is not up-to-date compared to the build output directory; "
```
rfscholte, line 270: One other thing I was missing is: which one is used?
Could this be compressed to something like: 

    For ARTIFACTID the jar is older than o.a.m.some.File.class: using target/classes.

mthmulders: Since this PR is already merged, I've created a follow-up where we can discuss the actual message: https://github.com/apache/maven/pull/416.

jira-importer:
Resolve #6417

PR #415, opened 2020-12-23
[MNG-7032] do not print colours for --version when in batch mode.

bmarwell:
This PR will fix MNG-7032: ANSI will not be used when using both `--batch-mode` and `--version`.
This is a bug because with batch mode, ANSI colours should not be used.

The tests works on a char level: By comparing the stripped and the non-stripped version, the test would fail if they differed. Or put in another way: If the sysout-version contained ANSI colour information, it would not match the stripped version.

---

       pulling in other changes.
       commit message.

### DIFF
--- a/maven-embedder/src/main/java/org/apache/maven/cli/MavenCli.java
+++ b/maven-embedder/src/main/java/org/apache/maven/cli/MavenCli.java
@@ -425,6 +425,8 @@ public class MavenCli
             throw e;
         }
 
+        applyColorMode( cliRequest );
+
         if ( cliRequest.commandLine.hasOption( CLIManager.HELP ) )
         {
             cliManager.displayHelp( System.out );
@@ -433,8 +435,6 @@ public class MavenCli
 
         if ( cliRequest.commandLine.hasOption( CLIManager.VERSION ) )
         {
-            // MNG-7032: Also disable colours if in batch mode
-            disableColorsInLogfileOrBatch( cliRequest );
             System.out.println( CLIReportingUtils.showVersion() );
             throw new ExitException( 0 );
         }
@@ -522,7 +522,7 @@ public class MavenCli
         }
         else
         {
-            disableColorsInLogfileOrBatch( cliRequest );
+            applyColorMode( cliRequest );
         }
 
         // LOG STREAMS
@@ -575,7 +575,7 @@ public class MavenCli
      * or {@link CLIManager#LOG_FILE} option was given (or both). In those cases, ANSI output is never feasible.
      * @param cliRequest the arguments as request pojo.
      */
-    private void disableColorsInLogfileOrBatch( CliRequest cliRequest )
+    private void applyColorMode( CliRequest cliRequest )
     {
         if ( cliRequest.commandLine.hasOption( CLIManager.BATCH_MODE )
                 || cliRequest.commandLine.hasOption( CLIManager.LOG_FILE ) )
--- a/maven-embedder/src/main/java/org/apache/maven/cli/MavenCli.java
+++ b/maven-embedder/src/main/java/org/apache/maven/cli/MavenCli.java
@@ -433,6 +433,8 @@ public class MavenCli
 
         if ( cliRequest.commandLine.hasOption( CLIManager.VERSION ) )
         {
+            // MNG-7032: Also disable colours if in batch mode
+            disableColorsInLogfileOrBatch( cliRequest );
             System.out.println( CLIReportingUtils.showVersion() );
             throw new ExitException( 0 );
         }
@@ -518,10 +520,9 @@ public class MavenCli
             throw new IllegalArgumentException( "Invalid color configuration option [" + styleColor
                 + "]. Supported values are (auto|always|never)." );
         }
-        else if ( cliRequest.commandLine.hasOption( CLIManager.BATCH_MODE )
-            || cliRequest.commandLine.hasOption( CLIManager.LOG_FILE ) )
+        else
         {
-            MessageUtils.setColorEnabled( false );
+            disableColorsInLogfileOrBatch( cliRequest );
         }
 
         // LOG STREAMS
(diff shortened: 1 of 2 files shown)


bmarwell:
Huh, I need some help with the tests.
I do not think any IT should fail because of my change, but they did fail: https://ci-builds.apache.org/job/Maven/job/maven-box/job/maven/job/MNG-7032_versioncolours/2/testReport/

On the other hand, I get an NPE locally – in `MavenCliTest.testStyleColors:340`. I really cannot explain what is going on here. Is `cli` being nulled? If so, why?

maven-embedder/src/main/java/org/apache/maven/cli/MavenCli.java
```diff
@@ -433,6 +433,8 @@ void cli( CliRequest cliRequest )
 
         if ( cliRequest.commandLine.hasOption( CLIManager.VERSION ) )
         {
+            // MNG-7032: Also disable colours if in batch mode
```
michael-o, line 436: This comment is redundant because we have commit messages and blames.

bmarwell: Hi @michael-o – are you sure it is "unwanted"? There are numerous similar comments…

Also, I think it is helpful for readability. You do not always have all commits on your second screen. Blame might not help if there was a slight change in a later version, as blame only shows the LATEST commit of a line.

2020-12-27T11:46:22_0001_screenshot

michael-o: I now that there are, I think this really belongs to an SCM unless you are describing the behavior of the code. Getting "MNG-7032" git be done with 'git blame', so this information is redundant for me. Like people adding dates to their comments.

maven-embedder/src/main/java/org/apache/maven/cli/MavenCli.java
```diff
@@ -433,6 +433,8 @@ void cli( CliRequest cliRequest )
...
+            disableColorsInLogfileOrBatch( cliRequest );
```
rfscholte, line 437: Please move this line to line 427. I can imagine we're going to add colors to the help output too. And maybe there will be more options in the future. This disabling a global thing, should not be restricted to `version`.

maven-embedder/src/main/java/org/apache/maven/cli/MavenCli.java
rfscholte: Only if there's a huge comment required to explain what's happening based on a ticket, it might be worth adding the issue number for easy access to the discussion. In this case it is quite clear what's happing, especially if the comment is almost the same as the methodname.

maven-embedder/src/main/java/org/apache/maven/cli/MavenCli.java
```diff
@@ -569,6 +570,20 @@ else if ( cliRequest.commandLine.hasOption( CLIManager.BATCH_MODE )
         }
     }
 
+    /**
+     * Disables the colour output in the case that the {@link CLIManager#BATCH_MODE} option
+     * or {@link CLIManager#LOG_FILE} option was given (or both). In those cases, ANSI output is never feasible.
+     * @param cliRequest the arguments as request pojo.
+     */
+    private void disableColorsInLogfileOrBatch( CliRequest cliRequest )
```
rfscholte, line 578: I'd prefer a more general method name: applyColorMode. This way we can add more logic when necessary

maven-embedder/src/main/java/org/apache/maven/cli/MavenCli.java
bmarwell: Thanks, that is helpful insight. I now see why the comment is redundant.

michael-o:
I will happily go through. I was never really happy that Jansi is installed by default even if batch or file mode were activated.

maven-embedder/src/main/java/org/apache/maven/cli/MavenCli.java
```diff
@@ -518,10 +520,9 @@ else if ( !"auto".equals( styleColor ) )
             throw new IllegalArgumentException( "Invalid color configuration option [" + styleColor
                 + "]. Supported values are (auto|always|never)." );
         }
-        else if ( cliRequest.commandLine.hasOption( CLIManager.BATCH_MODE )
-            || cliRequest.commandLine.hasOption( CLIManager.LOG_FILE ) )
+        else
```
rfscholte, line 523: this else-statement can now be removed, because the color mode has already been applied

michael-o (changes requested):
This looks OK, but incomplete. When if you at `org.apache.maven.cli.MavenCli.logging(CliRequest)` in the block marked with `// LOG COLOR` you see duplicate code now. It is probably time to move that code out of `logging()` into `cli()`.

I wonder what will happen when one does the following:
```
mvn -v -Dstyle.color=always | less -R
```

I would expect to see colors here also stdout is not attached to a tty. I think we need to move the use property to a CLI option to make it more obvious.

I am quite used to: `grep --color=always ... | less -R`

bmarwell:
> This looks OK, but incomplete. When if you at `org.apache.maven.cli.MavenCli.logging(CliRequest)` in the block marked with `// LOG COLOR` you see duplicate code now. It is probably time to move that code out of `logging()` into `cli()`.
> 
> I wonder what will happen when one does the following:
> 
> ```
> mvn -v -Dstyle.color=always | less -R
> ```
> 
> I would expect to see colors here also stdout is not attached to a tty. I think we need to move the use property to a CLI option to make it more obvious.
> 
> I am quite used to: `grep --color=always ... | less -R`

Which one? This? 
https://github.com/apache/maven/blob/496b9bd65d0ddd0fb19c8c9f89058555f458790e/maven-embedder/src/main/java/org/apache/maven/cli/MavenCli.java#L508-L522

I do not see this code duplicated somewhere else. There are other duplicated parts, however.

2020-12-28T16:35:36_0001_screenshot

michael-o:
My bad, I did not notice that you have moved the code. But still, The property needs to be in the new method now for consistency reasons.

bmarwell:
> My bad, I did not notice that you have moved the code. But still, The property needs to be in the new method now for consistency reasons.

I see. I could just move that block into the new method `applyColorMode( CliRequest cliRequest )` and update its Javadoc.

michael-o:
Can this be closed? MNG-7032 has been addressed with another PR.

PR #421, opened 2020-12-28
Artifact.getPath() and .setPath()

mkarg:
Simplifying usage of Java NIO API (since Java 7), so the *calling* code does not need to convert to / from Path ON ITS OWN again and again.

By using `default` interface methods, this should not imply any side effects to any existing code.


maven-artifact/src/main/java/org/apache/maven/artifact/Artifact.java
```diff
@@ -86,6 +87,16 @@
 
     void setFile( File destination );
 
+    default Path getPath()
+    {
+        return this.getFile().toPath();
```
rfscholte, line 92: If `getFile()` is `null`, it should not throw an NPE

mkarg: Good catch! Just fixed it using force-push. If you prefer, I can replace by `Optional`.

rmannibucau:
Quick side note: using a default method is a breaking change so should likely happen only in maven >= 4 (it breaks java lang proxies).

mkarg:
> Quick side note: using a default method is a breaking change so should likely happen only in maven >= 4 (it breaks java lang proxies).

What can break here?

rmannibucau:
@mkarg java lang proxies don't know how to invoke the default method so you will not get the expected behavior. This is a common breaking change since default methods were introduced. Some workaround for 2 java versions can look like https://github.com/Talend/component-runtime/blob/master/component-runtime-impl/src/main/java/org/talend/sdk/component/runtime/reflect/Defaults.java#L34 but this is really a workaround. jsonp got this breakage BTW even if it had been warned :(.

> Isn't master only be used for Maven >= 4 ?

Yes, point is more that if kept, it must not be backported and that if for maven 4 it can be added as a not default method too ;).

mkarg:
I have no strong feelings whether we should keep it as `default` methods or whether to force all implementations to adopt it on their own. Somebody has to decide.

rmannibucau:
From a quick review it sounds hurting less to add it and enforce impl to have it (for impl using proxies it will be hurtless) and others likely use our artifact impl so sounds a bit better even if still not 100%. Just requires to add it in maven-artifact-resolver too probably. Wdyt?

mkarg:
Well, in the first place it means more work for me while it is unclear how many proxies exist for `Artifact`. ;-) But if the majority wants me to change my solution to that more complex one just for the sake of proxies, I would be happy to do that. So please vote. :-)

slachiewicz (commented):
To use this default methods in plugin, someone would need to require Maven 4 as prerequisite. So far in many places we are using prerequisite 2.2.1 and slowly migrate to 3.1.1. so potential benefits for now - only in maven core. 

Changing internal field to Path would be more interesting for me.

rmannibucau:
Path is slower than File, it is only interesting when the filesystem can be abstracted IIRC. Also changing the internal will not help much in regards of what is done or we need to change it everywhere at the same time (to benefit from the abstraction).
That said I understand Markus "caller need" very much so guess we should tackle that at some point.
An alternative is to provide another resolver method which returns NioArtifacts. Here nothing would be ambiguous probably and a plugin could even test if it exists or not.
Lastly we must start somewhere to provide a more user friendly API, if we stick to current prerequisites it will never happen ;).

mkarg:
If the majority wants me to replace `default` in the interface by an actual implementation in `DefaultArtifact`, then I could solve it like this: Both `setField` and `getField` do store *both* variants in *two* internal fields, i. e. we have internally a `File` *and* a `Path`, so we spare the repeated conversion-on-read while only having an overhead of one additional reference. As typically read happens much more than write, this should speed up **both** code paths (the IO path and the NIO path) in an *optimal* way, as the getters always use the specific optimal field. That would satisfy @slachiewicz **and** take care of @rmannibucau's performance concerns. WDYT? Shall I change the PR in such a way? Votes please, as I like to get this done within this week.

slachiewicz:
-1 

Your initial proposal is good to me, no need to change anything for me ;-)

mkarg:
So if there is nobody voting -1 I would kindly ask someone to merge this, as I am not a committer. :-)

rmannibucau:
Ok for me when all *our* impl will have implemented it and do not rely on the default method.

mkarg:
> Ok for me when all _our_ impl will have implemented it and do not rely on the default method.

Done in https://github.com/apache/maven/pull/421/commits/5c7edad17aa2bc8b545522dba275b1d0521f6948, by either implementing it or relying on the implementation of `DefaultArtifact` (not *default method*). :-)

mkarg:
There are no more open requests to change anything since six days. What do I have to do next to get this PR committed actually (I am not a committer)?

rmannibucau:
@mkarg im away from keyboard for a week but probably ping dev@maven about it saying you have 2 +1, should be enough to get a merge.

mkarg:
> @mkarg im away from keyboard for a week but probably ping dev@maven about it saying you have 2 +1, should be enough to get a merge.

Romain, thanks a lot for this information! So **finally** I learned how to trigger things in the Maven community! :-)

maven-artifact/src/main/java/org/apache/maven/artifact/DefaultArtifact.java
```diff
@@ -207,6 +208,16 @@ public File getFile()
...
+    public Path getPath()
+    {
+        return this.getFile() == null ? null : this.getFile().toPath();
+    }
+
+    public void setPath( Path path )
+    {
+        this.setFile ( path == null ? null : path.toFile() );
+    }
+
```
slawekjaranowski, line 220: did you override default implementation from interface?

maven-core/src/main/java/org/apache/maven/project/artifact/ActiveProjectArtifact.java
```diff
@@ -65,6 +66,12 @@ public File getFile()
         return project.getArtifact().getFile();
     }
 
+    /** {@inheritDoc} */
+    public Path getPath()
+    {
+        return this.getFile() == null ? null : this.getFile().toPath();
+    }
+
```
slawekjaranowski, line 74: did you override default implementation from interface?

maven-core/src/main/java/org/apache/maven/project/artifact/ActiveProjectArtifact.java
```diff
@@ -120,6 +127,12 @@ public void setFile( File destination )
         project.getArtifact().setFile( destination );
     }
 
+    /** {@inheritDoc} */
+    public void setPath( Path path )
+    {
+        this.setFile( path == null ? null : path.toFile() );
+    }
+
```
slawekjaranowski, line 135: did you override default implementation from interface?

slawekjaranowski (changes requested):

maven-artifact/src/main/java/org/apache/maven/artifact/DefaultArtifact.java
mkarg: I do not understand the question (or why you are asking it). Actually that above code *is* effectively overriding the default implementation of the interface, and it was explicitly done because @rmannibucau asked for this.

maven-core/src/main/java/org/apache/maven/project/artifact/ActiveProjectArtifact.java
mkarg: Yes, implicitly, as this class extends `DefaultArtifact`.

maven-core/src/main/java/org/apache/maven/project/artifact/ActiveProjectArtifact.java
mkarg: Yes, implicitly, as this class extends `DefaultArtifact`.

maven-artifact/src/main/java/org/apache/maven/artifact/DefaultArtifact.java
slawekjaranowski: I asking because body of method in default method in interface is the same as overridden one.
So maybe body in interface is redundant for first look, and I ask.
Ok, I have read previous comments - I didn't know about proxy problems.

slawekjaranowski (approved):

elharo (changes requested):
Default interface methods require java 8. Maven still supports Java 7.

rfscholte:
https://issues.apache.org/jira/browse/MNG-6399 , so default interfaces should be used

mkarg:
> Default interface methods require java 8. Maven still supports Java 7.

Maven `master` branch does *not* support Java 7 anymore.

rmannibucau:
It is all good for master with or without default methods. Just need a run of IT repo to be sure but code is more than fine now even for proxy case.

elharo (changes requested):
All these methods require unit tests.

mkarg:
> All these methods require unit tests.

I just pushed new unit tests covering all new methods.

maven-artifact/src/test/java/org/apache/maven/artifact/ArtifactTest.java
```diff
@@ -0,0 +1,43 @@
...
+import java.io.File;
+import java.nio.file.Path;
+
+import org.junit.jupiter.api.Test;
+
+public class ArtifactTest {
+
+    @Test
+    public void testPathYieldsFile() {
+        final Artifact artifact = spy( Artifact.class );
```
elharo, line 37: This is testing implementation. Spies aren't needed here. Just test that calling setPath changes the return value of getFile and vice versa. And you can use actual Paths/Files instead of mocks.

elharo (changes requested):

elharo (changes requested):
Test failures are related:

Error:  COMPILATION ERROR : 
[INFO] -------------------------------------------------------------
Error:  /home/runner/work/maven/maven/maven-artifact/src/test/java/org/apache/maven/artifact/ArtifactTest.java:[38,31] cannot find symbol
  symbol:   method of(java.lang.String)
  location: interface java.nio.file.Path
Error:  /home/runner/work/maven/maven/maven-artifact/src/test/java/org/apache/maven/artifact/ArtifactTest.java:[41,27] cannot find symbol
  symbol:   method of(java.lang.String)
  location: interface java.nio.file.Path
Error:  /home/runner/work/maven/maven/maven-artifact/src/test/java/org/apache/maven/artifact/DefaultArtifactTest.java:[158,31] cannot find symbol
  symbol:   method of(java.lang.String)
  location: interface java.nio.file.Path
Error:  /home/runner/work/maven/maven/maven-artifact/src/test/java/org/apache/maven/artifact/DefaultArtifactTest.java:[161,27] cannot find symbol
  symbol:   method of(java.lang.String)
  location: interface java.nio.file.Path

maven-artifact/src/test/java/org/apache/maven/artifact/ArtifactTest.java
mkarg: This seems to me like personal preference. Actually the Spy is used to test the default implementation, which is a pretty valid use of spies.

mkarg: This code *is* creating actual Paths/Files and no mocks of those classes *explicitly*. It is the implementation's decision how it works internally, so it is even fine that Mockito creates mocks internally.

rmannibucau: Agree using mocks is like not writing a test so if possible just use impls and temporaryfolder rule, this is enough IMHO

pzygielo:
What's the purpose of new empty file `maven-core/src/test/java/org/apache/maven/git`? RAT does not like it.

maven-artifact/src/test/java/org/apache/maven/artifact/ArtifactTest.java
mkarg: Why is this like "not writing a test"? Can you please elaborate on that? In fact, the test is perfectly well and proofs exactly what it should, without the need of implementing lots and lots of unused methods.

mkarg:
> What's the purpose of new empty file `maven-core/src/test/java/org/apache/maven/git`? RAT does not like it.

Seems this was added by incident. I will remove it. Good catch!

maven-artifact/src/test/java/org/apache/maven/artifact/ArtifactTest.java
rmannibucau: Right but the mock is useless and you test the mock not prod code, this is what i meant by useless. It costs nothing to write a test using 100% of prod code there, so it is worth it imho if we add a test for that addition.

mkarg: This is not true, I am not testing the mock at all, and it is anything but useless: I am testing the *original* default methods of the interface, as this is a *spy* (if it would be a *mock*, then the test would *fail* -- maybe you did not read the actual code lines I wrote?), I use the mocked methods just for fulfilling the interface without the need of creating an *unnecessary* explicit implementation class. And "nothing" is far from the truth: The *spy* (it is *not* a mock) costed me exactly one minute, while implementing an real Artifact costs me much more time, provides *useless* code lines, and forces future maintenance of them!

In fact, I wrote that code *by intention* as it is *exactly* what I want to proof: Is the *default* method of the *interface* working correctly? I do *not* want to test if an *actual implementation* is working correctly!

mkarg:
> Error: COMPILATION ERROR :

Good catch! Accidentially I pushed a Java 11 code variant, while actually I wanted Java 8. Fixed by latest push, now CI is passing fine. :-)

maven-artifact/src/test/java/org/apache/maven/artifact/ArtifactTest.java
mkarg: Apparently some of you misunderstood what this code does. So let me explain with comments:
```java
final Artifact artifact = spy( Artifact.class ); // if you have ANY implementation that does NOT overwrite the default methods
artifact.setPath( Paths.get( "testPath" ) ); // when invoking setPATH
verify( artifact ).setFile( new File( "testPath" ) ); // then getFILE must get invoked in turn with exactly the same argument
when( artifact.getFile() ).thenReturn( new File( "testFile" ) );
assertEquals( Paths.get( "testFile" ), artifact.getPath() ); // when someone call getPATH it must get taken from getFILE (see line above)
```

rmannibucau: I get that but you precisely do what i say: not test what you think and test the mock.
To prove it, do a proxy based test, it will fail.
Personally i think the defauklt artifact test is more than sufficient since this code must never be used in practise but it is important to add useful tests validating prod code IMHO.

mkarg: So actually it is OK if there is *no* test for the *default methods*?

mkarg: Even for @elharo ?

elharo: I might not be seeing it, but I don't think this is doing what you claim it is doing. A test for the default method would:

1. create a class that does not override the default method
2. call setPath on that class
3. verify that getPath returns the new value

It does not need mocks and it certainly does not need spies.

mkarg: @elharo So in contrast to @rmannibucau you do want to keep the test for the default methods, or why else do you continue this thread?

So for clarity:
1. The result of calling `spy()` *is exactly* a class that implements the interface but does not override the default methods.
2. The test *does* call `setPath` on that class
3. No, the test *shall not* test whether `getPath` returns what `setPath` has set, as it is *not* a test for some arbitrary implementation, but is a test for *the default method* itself and ONLY for itself. This IS how pure interface get tested, as otherwise you would NOT test the interface but test THE SUBCLASS's ability to store between get/set, which is NOT the intention of the test. It actually shall proof that `setPath` effectively calls `setFile`, as it tests the default methods ONLY, and that is exactly what the default methods shall do, and that is exactly what the test does, and that is why the test is named `testPathYieldsFile` but NOT `testGetPathYieldsSetPath`.

rmannibucau: Side note: 1 is not strictly true, it overrides it even if it delegates it at the end with mock handlers so not exactly the same.

mkarg: ...yes, but which is completely irrelevant for the sake of this actual test!

rmannibucau: it does not test the default method so I think it is relevant if @elharo insists to keep the test, the use case he described is the only one to test about the interface and is not tested by the test you wrote.

maven-artifact/src/main/java/org/apache/maven/artifact/Artifact.java
```diff
@@ -86,6 +87,16 @@
 
     void setFile( File destination );
 
+    default Path getPath()
```
elharo, line 90: You can test these two new default methods with a subclass that implements getFile and setFile but not getPath and setPath. You don't need mocks, and you certainly don't need spies to do that. Focus on the values returned by getPath, not which methods getPath and setPath call.

mkarg: Yes I *can* do that but I do not see any need to do that. My test tests exctly what I want to test, so I do not see *why* I should change it. I *do not want* to test what would happen *if* someone implements that, I *only* want to proof what the default methods *actually* are like.

mkarg: The reason is that it is *necessary* to test that `setPath` actually invokes `setFile` as this is *what I have coded* in the default method and what get executed one you do *not* implement `setPath`. And only THAT case is what I MUST test, everything is something *differen* that you can add in *another* test ontop.

maven-artifact/src/test/java/org/apache/maven/artifact/ArtifactTest.java
mkarg: This is not true. It *does* implement the default method. There is a simple proof: Change the default implementation of `setPath` in a away that *does not* invoke `setFile` and the test will fail. THAT and ONLY THAT is what this test SHALL proof and nothing else.

mkarg: The use case he describes is NOT to be tested, as it is would not test the default method, but is about testing the actual implementing class. There is no need for such a test.

mkarg:
Guys, don't get me wrong, but for the sake of finding a concensus, maybe we should agree on two things:

(1) We do *not* insist on personally wanted changes *must* be done unless there is a technical fault or *actual* (not *assumed*) risk that *we all* agree that it really exists. Which is not the case here so far, as @elharo insists on KEEPING a test while @rmannibucau plead for DISCARDING it. So apparently, it is NOT ESSENTIAL to have that particular test, so we should not keep it.

(2) We separate the discussion about what is *wrong* (i. e. *proven* to fail with the currrent code) from what one thinks *could* be tested *in addition* (like testing particular use cases instead of the actual contributed code). The proof @elharo insists upon is NOT ESSENTIAL, so it CANNOT be showstopper for this PR, but it just can be a good and valuable *proposal for addition ontop* in a further PR.

I respect all you opinions, but this whole discussion seems to be more about personal visions what we could add *ontop* of my PR, or person vision what valid use cases for mocks and spies are, but not about the PR's actual status itself. Or at least, I have not yet understood what is actually and proven *wrong* in my PR.

elharo:
It is not necessary to test that setPath actually invokes setFile. Indeed it is a **bad idea** to test that setPath  invokes setFile. Testing the implementation instead of the behavior leads to brittle tests that are tightly coupled to private implementation details.

mkarg:
> It is not necessary to test that setPath actually invokes setFile. Indeed it is a **bad idea** to test that setPath invokes setFile. Testing the implementation instead of the behavior leads to brittle tests that are tightly coupled to private implementation details.

This is just your personal opinion. I think it is a pretty good idea to test whether a default methods does what I *want* it to do. Testing the implementation that solely was written for the sake of the test, makes no sense at all, as *any other* implementation would behave *differently* and possibly *wrong*. So the test would proof *nothing*.

rmannibucau:
@mkarg Agree with @elharo , you dont test the contract (which is 100%) of what a default method is but its impl. There are other ways to impl it correctly and your test would fail whereas it shouldnt which tend to proove it is a test of mocks and not of contract.
Since it is the last discussion point I think the fastest is to fix the test instead of discussing it is worth or not which can end up depending on people - will take you 2mn or so to fix it properly ;) - then everything will be green and it can be merged IMHO.

mkarg:
@rmannibucau Talking about speeding things up, I think it is best if you just author a commit ontop that replaces my "wrong" code by your "correct" code, as apparently we have completely different visions what we *want* to test. Just open a PR against my fork, and I will be happy to merge it, so we can make actual progress.

mkarg:
(just as a side note: My test does *not strive* to test whether "getPath" returns "setPath" as this contract is not defined anywhere)

rmannibucau:
@mkarg hmm, for me the contract was that getPath returns either setPath or setFile value (converted with toPath if needed or the opposite). If there is no contract then don't add the methods. Also note that default has a huge bug for me - this is why I think testing default is pointless, at the end it is not a default method but an abstract method - if you impl *now* an Artifact you will impl setPath/getPath - ok, strictly speaking 50% of cases will be like that but it is tempting since it looks more modern - and not getFile/setFile....and nothing will work at all with your impl until it is an internal usage only. This is why I don't push too much this, I have no issue integrating it but I don't think it helps to have anything else than getPath *only* - but once again I have nothing against if people want.

slachiewicz:
FYI: https://issues.apache.org/jira/browse/MNG-6444

mkarg:
The contract is:
* There is *one single* property, which is named `file`, and it has the getters and setters `getFile` and `setFile`, and the contract is that `getFile` returns `setFile`. This already existed and is tested, so there is no need to test *that* contract again.
* There is **no** separate property named `path`. `getPath` simply is *an alias* for `getFile`, and `setPath` is *an alias* for `setFile`, working on the existing (and tested) property `File`, but certainly doing conversion on the fly. This is **by intention** because the idea is **not** to have *two* properties, but to provide a smooth transition phase from the IO-era to the NIO-era, i. e. to **replace** `File` some fine day in the future.

Hence it makes no sense to tests *again* whether the property works as expected, but it makes much sense to check if the alias really works as an alias.

In addition, my new test *solely* tests the code *that I added*, and it only tests it for *what I itend to*. Hence, it proofs whether the new alias methods are there, and whether they really act as an *alias*. I deliberate *do not* test if `getPath` returns `setPath` as this would imply testing the `file` property (which is already tests).

Also it is irrelevent if *some other* implementation of the interface would fail the tests, as I deliberate *do not* tests *some* implementation, but in fact, I *only* test the default alias implementation (not *any* alias implementation). Testing *other* implementations, or testing *other* scenario simply is out of scope of this PR, which *solely* has the intention to add the alias methods to the interface, if you remember.

So if there is anything unclear about the intention of this PR, or if somebody has really a huge problem with this, then please tell now and clearly **what** problem does **actually** exist. Does this PR provide harm? No. Is it beneficial? Yes.

Don't let us squander valuable time arguing about things *beyond* the scope of this PR. Let's focus on the merits of **this PR**, and let's move on, even if we have different likes or dislikes.

Otherwise I would really like to kindly ask you to merge this PR (I am not a committer, someone of you need to do it). Thanks.

slachiewicz:
to see more benefits - could you review our Maven Core and use this new method to improve it with Your proposed changes?
Maybe we can even do something more for https://issues.apache.org/jira/browse/MNG-6444

PR #428, opened 2021-01-08
[EXPERIMENT] Update resolver and use newly introduced method

cstamas:
Update resolver and make use of newly introduced method in https://github.com/apache/maven-resolver/pull/87

This PR is NOT meant for merge as is, is more just an example.


slachiewicz:
I'm really interested in how this will work (ie use the newer resolver method) in plugins - that run in older maven 3.6.3 (as a replacement to m-artifact-transfer.

cstamas:
Was m-a-t initial goal to "hide" org.sonatype vs org.eclipse resolver diffs? So to say maven 3.0 vs maven 3.1. Also, it is sitting between maven (core) and resolver (core), but m-a-t is not in core? -- correct me if am wrong.

PR #433, opened 2021-01-21
[MNG-7079] Upgrade to jansi 2.2.0

gnodet:
This PR depends on https://github.com/apache/maven-shared-utils/pull/69 and the release of `maven-shared-utils`.

PR #436, opened 2021-01-21
[MNG-7080] Add a --color option to simplify color support

gnodet:
This PR is layered on top of #433 

       pulling in other changes.
       commit message.

### DIFF
--- a/maven-embedder/src/main/java/org/apache/maven/cli/CLIManager.java
+++ b/maven-embedder/src/main/java/org/apache/maven/cli/CLIManager.java
@@ -107,6 +107,8 @@ public class CLIManager
 
     public static final String NO_TRANSFER_PROGRESS = "ntp";
 
+    public static final String COLOR = "color";
+
     protected Options options;
 
     @SuppressWarnings( "checkstyle:linelength" )
@@ -149,6 +151,7 @@ public class CLIManager
         options.addOption( Option.builder( LEGACY_LOCAL_REPOSITORY ).longOpt( "legacy-local-repository" ).desc( "Use Maven 2 Legacy Local Repository behaviour, ie no use of _remote.repositories. Can also be activated by using -Dmaven.legacyLocalRepo=true" ).build() );
         options.addOption( Option.builder( BUILDER ).longOpt( "builder" ).hasArg().desc( "The id of the build strategy to use" ).build() );
         options.addOption( Option.builder( NO_TRANSFER_PROGRESS ).longOpt( "no-transfer-progress" ).desc( "Do not display transfer progress when downloading or uploading" ).build() );
+        options.addOption( Option.builder().longOpt( COLOR ).hasArg().desc( "Defines the color mode of the output. Available options are auto/always/never" ).build() );
     }
 
     public CommandLine parse( String[] args )
--- a/maven-embedder/src/main/java/org/apache/maven/cli/MavenCli.java
+++ b/maven-embedder/src/main/java/org/apache/maven/cli/MavenCli.java
@@ -119,6 +119,7 @@ import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 
 import static java.util.Comparator.comparing;
+import static org.apache.maven.cli.CLIManager.COLOR;
 import static org.apache.maven.cli.ResolveFile.resolveFile;
 import static org.apache.maven.shared.utils.logging.MessageUtils.buffer;
 
@@ -504,6 +505,7 @@ public class MavenCli
 
         // LOG COLOR
         String styleColor = cliRequest.getUserProperties().getProperty( STYLE_COLOR_PROPERTY, "auto" );
+        styleColor = cliRequest.commandLine.getOptionValue( COLOR, styleColor );
         if ( "always".equals( styleColor ) )
         {
             MessageUtils.setColorEnabled( true );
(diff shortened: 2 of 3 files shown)


pom.xml
```diff
@@ -319,12 +319,12 @@ under the License.
       <dependency>
         <groupId>org.apache.maven.shared</groupId>
         <artifactId>maven-shared-utils</artifactId>
-        <version>3.2.1</version>
+        <version>3.3.4-SNAPSHOT</version>
```
michael-o, line 322: We can't do this. When bisecting down to this commit it will be virtually impossible to reproduce. I went through with Maven Artifact Transfer.

gnodet: The only relevant commit is https://github.com/apache/maven/pull/436/commits/690aff3ffb08fc5366f72736f668775e86ba67e9.  The two ones before are part of https://github.com/apache/maven/pull/436/commits/690aff3ffb08fc5366f72736f668775e86ba67e9 which itself depends on https://github.com/apache/maven-shared-utils/pull/69.
Things have to be done step by step...

olamy (approved):

maven-embedder/src/main/java/org/apache/maven/cli/CLIManager.java
```diff
@@ -149,6 +151,7 @@ public CLIManager()
         options.addOption( Option.builder( LEGACY_LOCAL_REPOSITORY ).longOpt( "legacy-local-repository" ).desc( "Use Maven 2 Legacy Local Repository behaviour, ie no use of _remote.repositories. Can also be activated by using -Dmaven.legacyLocalRepo=true" ).build() );
         options.addOption( Option.builder( BUILDER ).longOpt( "builder" ).hasArg().desc( "The id of the build strategy to use" ).build() );
         options.addOption( Option.builder( NO_TRANSFER_PROGRESS ).longOpt( "no-transfer-progress" ).desc( "Do not display transfer progress when downloading or uploading" ).build() );
+        options.addOption( Option.builder().longOpt( COLOR ).hasArg().desc( "Color support mode (auto/always/never)" ).build() );
```
MartinKanters, line 154: Could you please be a bit more elaborate on what this flag entails? I fear not everyone will know what it will do.

MartinKanters (changes requested):
Sorry, didn't review this PR before. I have a small comment.

maven-embedder/src/main/java/org/apache/maven/cli/CLIManager.java
mthmulders, line 154: Out of curiosity, why only a "long" option? IIRC, this would be the first switch that only has a "long" version and no shorthand.

MartinKanters: If it's because both `-c` and `-C` are already taken, we could also go for `-cm` and `--color-mode`

MartinKanters: After discussing offline with @rfscholte and @mthmulders , I would prefer `--color-mode` anyway. `--color` suggests that you can put in a color there (as in 'red' on 'green').

michael-o: I urge to retain `--color` because many many other tools use the same option.

michael-o: @MartinKanters This is a manifestation of the sytem property. No logical changes.

gnodet: I agree with @michael-o , the goal of this PR is to make things easier for users, rather than looking for a non trivial `-Dstyle.color=always` option, so the `--color` option similar to the gnu `ls` or ` grep` seems a better choice to me.

michael-o: Exactly, when I reviewed the PR for `-Dstyle.color` I requested to apply the `grep` logic to no to invent new terms.

MartinKanters: I understand what it does, and I agree with adding the flag. My point was that people will find this option in `mvn help` and might not know what it entails. For example, `ls` has this as help text:
> --color[=WHEN] colorize the output; WHEN can be 'always' (default if omitted), 'auto', or 'never'; more info below

So perhaps we can change the description to something like:
```suggestion
        options.addOption( Option.builder().longOpt( COLOR ).hasArg().desc( "Defines the color mode of the output. Available options are auto/always/never" ).build() );
```

MartinKanters: Ah, I was not aware of this. Then `--color` makes sense to me. 

Also just checked to be sure, the help text looks good even without a short character:
image

MartinKanters: Also, just checked how the help output looks without having a shorthand flag. Looks good to me:
image

gnodet: The suggestion looks good to me.

MartinKanters: Alright, could you please take and squash it? Afterwards, I'll make sure to test it on Jenkins and take care of merging.

gnodet: @MartinKanters done

MartinKanters (approved):
Looks great to me! Running on [jenkins](https://ci-builds.apache.org/job/Maven/job/maven-box/job/maven/job/MNG-7080/1/console).

PR #450, opened 2021-02-21
[MNG-7097] Plugin Dependency Resolution Improvement

cstamas:
Implements a helper component for plugin dependency resolution that provider
filter and selector that leaves out core artifacts from collection
and resolution.

### DIFF
--- a/maven-core/src/main/java/org/apache/maven/extension/internal/CoreExportsProvider.java
+++ b/maven-core/src/main/java/org/apache/maven/extension/internal/CoreExportsProvider.java
@@ -27,7 +27,8 @@ import org.codehaus.plexus.PlexusContainer;
 import org.eclipse.sisu.Nullable;
 
 /**
- * CoreExportsProvider
+ * CoreExportsProvider, that CANNOT be <tty>javax.inject.Provider</tty> as ctor would have circular dep.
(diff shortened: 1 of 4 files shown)


maven-core/src/main/java/org/apache/maven/extension/internal/CoreExportsProvider.java
```diff
@@ -27,7 +27,8 @@
 import org.eclipse.sisu.Nullable;
 
 /**
- * CoreExportsProvider
+ * CoreExportsProvider, that CANNOT be <tty>javax.inject.Provider</tty> as ctor would have circular dep.
```
pzygielo, line 30: `javadoc: error: unknown tag: tty`

cstamas: Fixed, tx

cstamas:
Something is strange: for example this one fails (on GH but also locally as well):
```
mng7045 DropUselessAndOutdatedCdiApi.it().................... FAILURE (0.3 s)
```
**but**, when I run the POM in IT from CLI (is simple, uses GMavenPlus plugin and a script), it passes ok...

jira-importer:
Resolve #8359

PR #464, opened 2021-04-04
[MNG-7133] Allow to inline collection/array items within plugin root configuration

slawekjaranowski:

       pulling in other changes.
       commit message.


maven-core/src/test/projects/lifecycle-executor/mojo-inline-configuration/pom.xml
```diff
@@ -0,0 +1,71 @@
...
+        <artifactId>maven-it-plugin</artifactId>
+        <version>0.1</version>
+        <executions>
+          <execution>
+            <goals>
+              <goal>xpp3-reader</goal>
+            </goals>
+            <phase>generate-sources</phase>
+            <configuration>
+              <models>item1</models>
```
rfscholte, line 63: This scares me. `models` sounds like plural, wheres is contains only one item. And with plexus you can define how elements should be merged. Here I would expect that items2 overrides item1.
To this will make the configuration less clear, so I'm -1

slawekjaranowski: Do you think about plural noun in this test example or about possibility use such configuration at all?

Currently we can do it on Bean level https://issues.apache.org/jira/browse/MNG-5033

jira-importer:
Resolve #8316

PR #472, opened 2021-05-14
[MNG-6915] Adapt the logging width to the terminal width, including sensible limits.

MartinKanters:
This PR uses the newly introduced Jansi's `getTerminalWidth()` to make the terminal a little bit more responsive.
We used sensible defaults to ensure the logging output makes sense.
This is especially useful for bigger screens.

By the way, this does not fix issues around characters which are larger than one column. That is still working the same as in master. 

 - [x] Make sure there is a [JIRA issue](https://issues.apache.org/jira/browse/MNG-6915) filed 
       pulling in other changes.
       commit message.

### DIFF
--- a/maven-embedder/src/main/java/org/apache/maven/cli/event/ExecutionEventLogger.java
+++ b/maven-embedder/src/main/java/org/apache/maven/cli/event/ExecutionEventLogger.java
@@ -39,6 +39,7 @@ import org.apache.maven.plugin.MojoExecution;
 import org.apache.maven.plugin.descriptor.MojoDescriptor;
 import org.apache.maven.project.MavenProject;
 import org.apache.maven.shared.utils.logging.MessageBuilder;
+import org.apache.maven.shared.utils.logging.MessageUtils;
 import org.codehaus.plexus.util.StringUtils;
 import org.slf4j.ILoggerFactory;
 import org.slf4j.Logger;
@@ -53,9 +54,15 @@ public class ExecutionEventLogger extends AbstractExecutionListener
 {
     private final Logger logger;
 
-    private static final int LINE_LENGTH = 72;
+    private static final int MAX_LOG_PREFIX_SIZE = 8; // "[ERROR] "
+    private static final int PROJECT_STATUS_SUFFIX_SIZE = 20; // "SUCCESS [  0.000 s]"
+    private static final int MIN_TERMINAL_WIDTH = 60;
+    private static final int MAX_TERMINAL_WIDTH = 130;
+    private static final int TERMINAL_WIDTH =
+            Math.min( MAX_TERMINAL_WIDTH, Math.max( MessageUtils.getTerminalWidth(), MIN_TERMINAL_WIDTH ) );
+    private static final int LINE_LENGTH = TERMINAL_WIDTH - MAX_LOG_PREFIX_SIZE;
     private static final int MAX_PADDED_BUILD_TIME_DURATION_LENGTH = 9;
-    private static final int MAX_PROJECT_NAME_LENGTH = 52;
+    private static final int MAX_PROJECT_NAME_LENGTH = LINE_LENGTH - PROJECT_STATUS_SUFFIX_SIZE;
 
     private int totalProjects;
     private volatile int currentVisitedProjectCount;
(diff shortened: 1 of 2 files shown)


MartinKanters:
@gnodet @rfscholte Could you please take a look at this?

maven-embedder/src/main/java/org/apache/maven/cli/event/ExecutionEventLogger.java
```diff
@@ -53,9 +54,15 @@
 {
     private final Logger logger;
 
-    private static final int LINE_LENGTH = 72;
+    private static final int MAX_LOG_PREFIX_SIZE = 8; // "[ERROR] "
+    private static final int PROJECT_STATUS_SUFFIX_SIZE = 20; // "SUCCESS [  0.000 s]"
+    private static final int MIN_TERMINAL_WIDTH = 60;
+    private static final int MAX_TERMINAL_WIDTH = 130;
+    private static final int TERMINAL_WIDTH =
+            Math.min( MAX_TERMINAL_WIDTH, Math.max( MessageUtils.getTerminalWidth(), MIN_TERMINAL_WIDTH ) );
```
gnodet, line 62: I don't think this is a good idea to use a constant for the terminal width.  In the case where the maven class loader is reused, the terminal could change its width.  What about computing this variable and caching it for the `MavenSession` ?

mthmulders: If we understand correctly, this would make sure that if a terminal gets resized _during_ a Maven build, the output width would be recomputed accordingly?
If that is the case, we don't understand how storing this value in a `MavenSession` instance rather than a constant would solve this. As far as we know, the `MavenSession` lifetime roughly matches the lifetime of the JVM process, and as such, of the classloader.

gnodet: > If we understand correctly, this would make sure that if a terminal gets resized _during_ a Maven build, the output width would be recomputed accordingly?
> If that is the case, we don't understand how storing this value in a `MavenSession` instance rather than a constant would solve this. As far as we know, the `MavenSession` lifetime roughly matches the lifetime of the JVM process, and as such, of the classloader.

I was thinking about embedded maven in Eclipse and Mvnd, where the maven class loader and classes are reused, but a new session is created for a given build.  I don't think adjusting to the terminal size inside a given build is a good idea.

gnodet: Actually, this use case it not really supported yet.  For this to work, we'd need the information about the terminal width to be set publicly on the MavenSession.  The `MavenCLI` would set it based on a call to `MessageUtils.getTerminalWidth()`.

MartinKanters: Ah, I didn't realize that tools would create new `MavenSession`'s while reusing the classloaders, thanks for explaining!

I assume the terminal width would have to be publicly available on the session, in order for `mvnd` to create new sessions with the same width. For Eclipse, I assume that they would have to set the width to their internal terminal/logger window.
I'm just not sure whether `MavenSession` is the right class for this, because afaik it's currently cli-unaware. Perhaps there is a better place, but I don't know that by heart.

gnodet: @MartinKanters maybe the `MavenExecutionRequest` would be a better place as it contains already a bunch of things somewhat related to cli output: loggingLevel, showError, etc...

MartinKanters: Hmm, it's definitely better, but I think it would be best if we could keep this term width state in `maven-embedder`.
What if we store it as a regular field in `ExecutionEventLogger`? Or are you reusing the same instance of that as well in separate runs? Otherwise we could find some way to recalculate it every time `projectDiscoveryStarted()` is called or introduce a new method for it..

gnodet: Then reusing the cache from the `MavenSession` simply to store the value might be the easiest way.  It does not have to leak outside the `ExecutionEventLogger`.

gnodet:
I propose a slightly modified version at https://github.com/gnodet/maven/commit/10cf52c69c175863b0382098a5bf840c4dd74862
The reason is that the `ExecutionEventLogger` is not a singleton as I thought, so we can simply compute the values depending on the terminal width in fields.  Way easier and will work great with `mvnd` (where the terminal width is actually passed from the client to the daemon).

mthmulders:
Superseded by #473.

PR #473, opened 2021-05-19
[MNG-6915] Adapt the logging width to the terminal width, including s…

gnodet:
…ensible limits.

Slightly modified version of https://github.com/apache/maven/pull/472 / https://github.com/apache/maven/pull/472/commits/e47d647eee04888713a1875ddadf2ae1f28bc94e

       pulling in other changes.
       commit message.

### DIFF
--- a/maven-embedder/src/main/java/org/apache/maven/cli/event/ExecutionEventLogger.java
+++ b/maven-embedder/src/main/java/org/apache/maven/cli/event/ExecutionEventLogger.java
@@ -39,6 +39,7 @@ import org.apache.maven.plugin.MojoExecution;
 import org.apache.maven.plugin.descriptor.MojoDescriptor;
 import org.apache.maven.project.MavenProject;
 import org.apache.maven.shared.utils.logging.MessageBuilder;
+import org.apache.maven.shared.utils.logging.MessageUtils;
 import org.codehaus.plexus.util.StringUtils;
 import org.slf4j.ILoggerFactory;
 import org.slf4j.Logger;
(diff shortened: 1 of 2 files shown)


MartinKanters:
Looks good to me! 

Didn't think of this last time, but I'm wondering, when Jansi cannot determine the width, whether we should fallback the width of the terminal equal to the 'old' terminal width of 80 (I believe?). Then we are backwards-compatible for non-jansi terminals and probably would not have to change any ITs and unit tests. What do you think? And @mthmulders?

gnodet:
@MartinKanters it seems to work.  I've forced pushed the change.

Fwiw, there is a minor problem with the terminal width which can be improved later: ideally, we'd make the difference between the `System.out` and `System.err` streams.  Currently, `MessageUtils.getTerminalWidth()` relies on `AnsiConsole.getTerminalWidth()` which first tries the get the width of the terminal using `System.out` and then try with `System.err`.  The downside is that if the output is redirected using `mvn > out.txt` or even simply `mvn -l out.txt`, then the computed terminal width is not relevant.  It's no big deal though.  This also makes we wonder if the `MavenCli` should use `System.err` to print non standard output (i.e. version, help, etc...) instead of `System.out`...

MartinKanters:
> @MartinKanters it seems to work. I've forced pushed the change.

Nice, I definitely prefer defaulting to the old width. I see that you are not targeting any customized ITs and removed the changes in the unit tests. To me that proves that it's backwards compatible for non-ansi terminals.
And now that we have a constructor where we can put in the values, we can create some unit tests which verify that the terminal width is set correctly in all scenarios. I'm sorry that I'm asking for more changes on a PR that @mthmulders and I initially delivered... :)
 
> Fwiw, there is a minor problem with the terminal width which can be improved later: ideally, we'd make the difference between the `System.out` and `System.err` streams. Currently, `MessageUtils.getTerminalWidth()` relies on `AnsiConsole.getTerminalWidth()` which first tries the get the width of the terminal using `System.out` and then try with `System.err`. The downside is that if the output is redirected using `mvn > out.txt` or even simply `mvn -l out.txt`, then the computed terminal width is not relevant. It's no big deal though. This also makes we wonder if the `MavenCli` should use `System.err` to print non standard output (i.e. version, help, etc...) instead of `System.out`...

Right, that indeed sounds like something that can be dealt with later. About the last point we can also create a thread on the dev mailing list.

gnodet:
> > @MartinKanters it seems to work. I've forced pushed the change.
> 
> Nice, I definitely prefer defaulting to the old width. I see that you are not targeting any customized ITs and removed the changes in the unit tests. To me that proves that it's backwards compatible for non-ansi terminals.
> And now that we have a constructor where we can put in the values, we can create some unit tests which verify that the terminal width is set correctly in all scenarios. I'm sorry that I'm asking for more changes on a PR that @mthmulders and I initially delivered... :)

I've added a test to check various terminal widths.

MartinKanters (approved):
LGTM! Let's keep it open for a bit to see if @mthmulders agrees. 
I can merge it afterwards.

mthmulders (approved):
Thanks @gnodet!

MartinKanters:
[Running the build on Jenkins](https://ci-builds.apache.org/job/Maven/job/maven-box/job/maven/job/MNG-6915/2/). First try failed because of an infra issue, I think.

jira-importer:
Resolve #8616

PR #475, opened 2021-06-01
[MNG-7157] [MNG-6843] Provide a better design for obtaining artifacts…

gnodet:
… from a MavenProject inside a mojo

Follow-up on #413 ...

### DIFF
--- a/maven-core/src/main/java/org/apache/maven/plugin/PluginParameterExpressionEvaluator.java
+++ b/maven-core/src/main/java/org/apache/maven/plugin/PluginParameterExpressionEvaluator.java
@@ -22,7 +22,9 @@ package org.apache.maven.plugin;
 import java.io.File;
 import java.util.Properties;
 
+import org.apache.maven.artifact.resolver.filter.ArtifactFilter;
 import org.apache.maven.execution.MavenSession;
+import org.apache.maven.lifecycle.internal.MojoExecutor;
 import org.apache.maven.plugin.descriptor.MojoDescriptor;
 import org.apache.maven.plugin.descriptor.PluginDescriptor;
 import org.apache.maven.project.MavenProject;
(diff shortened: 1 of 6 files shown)


maven-core/src/main/java/org/apache/maven/project/MavenProject.java
```diff
@@ -347,84 +351,62 @@ public void addTestCompileSourceRoot( String path )
         return testCompileSourceRoots;
     }
 
+    @Deprecated
```
famod, line 354: It would be nice to document what to use instead (and why) via JavaDoc `@deprecated ...`.

gnodet:
> Is this really a follow-up? Instead, to me it seems like an alternative, more comprehensive solution than #413?

Yes, the wording is wrong.  I originally went to only extend and provide an alternative api, but then refactored your original proposal a bit, so it's now more an extended alternative....

maven-core/src/main/java/org/apache/maven/project/MavenProject.java
gnodet: I'm willing to provide more doc once there's a consensus.
Honestly, I'm not even sure we need such an API, unless there's a clear path for the deprecation and for moving to a more immutable `MavenProject`.   What are the deprecation rules ? Will 4.0 remove things deprecated in 3.x ? If they stay forever, there's really no point in deprecating and trying to find slightly better APIs...  I'm just asking here, but 3.0 is very old I think there are quite a few things that are deprecated since early 3.x versions, which looks like 10 years or so.

maven-core/src/main/java/org/apache/maven/plugin/PluginParameterExpressionEvaluator.java
```diff
@@ -244,6 +246,10 @@ else if ( "mojoExecution".equals( expression ) )
         {
             value = mojoExecution;
         }
+        else if ( "artifactFilter".equals( expression ) )
```
rfscholte, line 249: We should investigate if we can solve this with MPLUGIN-302 instead of exposing this a new expression.

jira-importer:
Resolve #8297

PR #476, opened 2021-06-02
[MNG-7156] Parallel build can cause issues between clean and forked g…

gnodet:
…oals

       pulling in other changes.
       commit message.

### DIFF
--- a/maven-core/src/main/java/org/apache/maven/execution/LockingEventSpy.java
+++ b/maven-core/src/main/java/org/apache/maven/execution/LockingEventSpy.java
@@ -46,22 +46,9 @@ import org.slf4j.LoggerFactory;
 public class LockingEventSpy extends AbstractEventSpy
 {
 
-    private final Logger logger = LoggerFactory.getLogger( getClass() );
-
     private static final Object LOCKS_KEY = new Object();
 
-    @SuppressWarnings( { "unchecked", "rawtypes" } )
-    private Lock getLock( ExecutionEvent event )
-    {
-        SessionData data = event.getSession().getRepositorySession().getData();
-        Map<MavenProject, Lock> locks = ( Map ) data.get( LOCKS_KEY );
-        if ( locks == null )
-        {
-            data.set( LOCKS_KEY, null, new ConcurrentHashMap<>() );
-            locks = ( Map ) data.get( LOCKS_KEY );
-        }
-        return locks.computeIfAbsent( event.getProject(), p -> new ReentrantLock() );
-    }
+    private final Logger logger = LoggerFactory.getLogger( getClass() );
 
     @Override
     public void onEvent( Object event ) throws Exception
@@ -77,7 +64,7 @@ public class LockingEventSpy extends AbstractEventSpy
                 if ( !lock.tryLock() )
                 {
                     logger.warn( "Suspending concurrent execution of project " + ee.getProject() );
-                    getLock( ee ).lockInterruptibly();
+                    lock.lockInterruptibly();
                     logger.warn( "Resuming execution of project " + ee.getProject() );
                 }
                 break;
@@ -91,6 +78,19 @@ public class LockingEventSpy extends AbstractEventSpy
                 break;
             }
         }
+    }
+
+    @SuppressWarnings( { "unchecked", "rawtypes" } )
+    private Lock getLock( ExecutionEvent event )
+    {
+        SessionData data = event.getSession().getRepositorySession().getData();
+        Map<MavenProject, Lock> locks = ( Map ) data.get( LOCKS_KEY );
+        if ( locks == null )
+        {
+            data.set( LOCKS_KEY, null, new ConcurrentHashMap<>() );
+            locks = ( Map ) data.get( LOCKS_KEY );
+        }
+        return locks.computeIfAbsent( event.getProject(), p -> new ReentrantLock() );
     }
 
 }
--- a/maven-core/src/main/java/org/apache/maven/execution/LockingEventSpy.java
+++ b/maven-core/src/main/java/org/apache/maven/execution/LockingEventSpy.java
@@ -63,9 +63,9 @@ public class LockingEventSpy extends AbstractEventSpy
                 Lock lock = getLock( executionEvent );
                 if ( !lock.tryLock() )
                 {
-                    logger.warn( "Suspending concurrent execution of project '{}'", executionEvent.getProject() );
+                    logger.debug( "Suspending concurrent execution of project '{}'", executionEvent.getProject() );
                     lock.lockInterruptibly();
-                    logger.warn( "Resuming execution of project '{}'", executionEvent.getProject() );
+                    logger.debug( "Resuming execution of project '{}'", executionEvent.getProject() );
                 }
                 break;
             case ProjectSucceeded:
--- a/maven-core/src/main/java/org/apache/maven/execution/LockingEventSpy.java
+++ b/maven-core/src/main/java/org/apache/maven/execution/LockingEventSpy.java
@@ -1,0 +1,96 @@
+package org.apache.maven.execution;
+
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *  http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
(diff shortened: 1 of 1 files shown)


bmarwell:
Will this also resolve `-DdeployAtEnd=true`?

gnodet:
> Will this also resolve `-DdeployAtEnd=true`?

I don't understand.  This only adds a locking mechanism to avoid conflicts caused by parallel execution between the normal lifecycle and forked lifecycles, which can cause the same project to be build twice concurrently.  This has nothing to do with `deployAtEnd` afaik.

maven-core/src/main/java/org/apache/maven/execution/LockingEventSpy.java
```diff
@@ -0,0 +1,96 @@
...
+import org.apache.maven.eventspy.EventSpy;
+import org.apache.maven.project.MavenProject;
+import org.eclipse.aether.SessionData;
+import org.eclipse.sisu.Typed;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * EventSpy implementation that provides a per-project locking mechanism
+ * to make sure a given project can not be build twice concurrently.
```
MartinKanters, line 40:
```suggestion
 * to make sure a given project can not be built concurrently.
```

maven-core/src/main/java/org/apache/maven/execution/LockingEventSpy.java
```diff
@@ -0,0 +1,96 @@
...
+            switch ( ee.getType() )
+            {
+            case ProjectStarted:
+            case ForkedProjectStarted:
+                Lock lock = getLock( ee );
+                if ( !lock.tryLock() )
+                {
+                    logger.warn( "Suspending concurrent execution of project " + ee.getProject() );
+                    getLock( ee ).lockInterruptibly();
+                    logger.warn( "Resuming execution of project " + ee.getProject() );
```
MartinKanters, line 81:
```suggestion
                    logger.warn( "Suspending concurrent execution of project '{}'", ee.getProject() );
                    getLock( ee ).lockInterruptibly();
                    logger.warn( "Resuming execution of project '{}'", ee.getProject() );
```

maven-core/src/main/java/org/apache/maven/execution/LockingEventSpy.java
```diff
@@ -0,0 +1,96 @@
...
+        }
+        return locks.computeIfAbsent( event.getProject(), p -> new ReentrantLock() );
+    }
+
+    @Override
+    public void onEvent( Object event ) throws Exception
+    {
+        if ( event instanceof ExecutionEvent )
+        {
+            ExecutionEvent ee = ( ExecutionEvent ) event;
```
MartinKanters, line 71: Can we come up with a more descriptive name than this? Perhaps just `executionEvent`?

maven-core/src/main/java/org/apache/maven/execution/LockingEventSpy.java
```diff
@@ -0,0 +1,96 @@
...
+                Lock lock = getLock( ee );
+                if ( !lock.tryLock() )
+                {
+                    logger.warn( "Suspending concurrent execution of project " + ee.getProject() );
+                    getLock( ee ).lockInterruptibly();
+                    logger.warn( "Resuming execution of project " + ee.getProject() );
+                }
+                break;
+            case ProjectSucceeded:
+            case ProjectFailed:
```
MartinKanters, line 85: I think we might be missing ProjectSkipped right now, did your leave this out on purpose?

maven-core/src/main/java/org/apache/maven/execution/LockingEventSpy.java
```diff
@@ -0,0 +1,96 @@
...
+
+/**
+ * EventSpy implementation that provides a per-project locking mechanism
+ * to make sure a given project can not be build twice concurrently.
+ * This case can happen when running parallel builds with forked lifecycles.
+ */
+@Singleton
+@Named
+@Typed( EventSpy.class )
+public class LockingEventSpy extends AbstractEventSpy
```
MartinKanters, line 46: I think it's quite an elegant solution, but I'm not sure if using an `EventSpy` for this purpose is how it should be done in `maven-core`. I don't think there are other implementations of it in core.

MartinKanters (changes requested):
Nice proposal, I left some comments. Also, I think it's worth it to create at least an integration test. It might be annoying to build an IT, but covering this just in unit tests will be much harder.

maven-core/src/main/java/org/apache/maven/execution/LockingEventSpy.java
gnodet: My understanding is the `ProjectSkipped` is send after a failure and in such a case, the project is... skipped, so there's no need to lock / unlock, especially as the `ProjectStarted` event is not sent in such cases.

maven-core/src/main/java/org/apache/maven/execution/LockingEventSpy.java
gnodet: Another possibility would be to change the `MavenCli` to support multiple `ExecutionListener`.  The problem (minor) is that there's no discovery mechanism, so they are hard coded, which I wanted to avoid...  Another possibility would be to define a component to hold the locks and inject it into the `LifeycleModuleBuilder` and `MojoExecutor` where those events are fired initially.  I don't really mind, but the current solution was the less intrusive.

maven-core/src/main/java/org/apache/maven/execution/LockingEventSpy.java
gnodet: Sure, sounds good.

maven-core/src/main/java/org/apache/maven/execution/LockingEventSpy.java
```diff
@@ -0,0 +1,96 @@
...
+    private static final Object LOCKS_KEY = new Object();
+
+    @SuppressWarnings( { "unchecked", "rawtypes" } )
+    private Lock getLock( ExecutionEvent event )
+    {
+        SessionData data = event.getSession().getRepositorySession().getData();
+        Map<MavenProject, Lock> locks = ( Map ) data.get( LOCKS_KEY );
+        if ( locks == null )
+        {
+            data.set( LOCKS_KEY, null, new ConcurrentHashMap<>() );
```
Tibor17, line 60: What mean `null` here?
Three parameters?
This line can be called concurrently. It should be treated somehow, like the semantics `trySet(): boolean`

gnodet: I don't think there's a concurrency issue here.
The code does the following:
```
        Map<MavenProject, Lock> locks = ( Map ) data.get( LOCKS_KEY );
        if ( locks == null )
        {
            data.set( LOCKS_KEY, null, new ConcurrentHashMap<>() );
            locks = ( Map ) data.get( LOCKS_KEY );
        }
```
The `data.set(k, null, v)` performs a `map.putIfAbsent(k, v)`, so this can not overwrite the value.  In case there is a concurrent access to this very method, the following line does a `get`, so I think there's no way the code can use a wrong map.

maven-core/src/main/java/org/apache/maven/execution/LockingEventSpy.java
MartinKanters: I checked the code and it looks like you are right :)

maven-core/src/main/java/org/apache/maven/execution/LockingEventSpy.java
MartinKanters: It's definitely less intrusive, but I'm not sure whether something that controls the build as much as this (temporarily pausing and resuming project builds) should be done in an `ExecutionListener` or `EventSpy`. My gut feeling says those classes should only extend the build with simple side-effects, like logging a project's phase. I would prefer your last suggestion. Then again, maybe this is the right direction. Perhaps some others can give their opinion on it?

maven-core/src/main/java/org/apache/maven/execution/LockingEventSpy.java
Tibor17: ok, if the session is implemented using `putIfAbsent` then it should be fine.

gnodet:
> Cool, I think I saw some of issues while using mvnd that should be solved...

Fwiw, it's not really urgent, as it has already been integrated into `mvnd` ...
  https://github.com/mvndaemon/mvnd/blob/master/daemon/src/main/java/org/mvndaemon/mvnd/execution/LockingEventSpy.java

gnodet:
> So this revolves around the idea that all elements subject to a project cannot be modified while the lock is obtained, so everything else concurrently is halted?
> So as far as I understand this PR is global while PR #578 is rather local?

I had a closer look at both problems and proposed solutions.  They are indeed related but for two different use cases.  The This one #476 address the use case for synchronising  between a standard lifecycle and a forked lifecycle, while #578 synchronises between a mojo execution for a given project and a mojo execution that has the _reactor_ flag set.  
Both are related because they both address concurrency execution between projects.

I'll work on a PR that addresses both problem at the same time.

gnodet:
Superseded by #628 for 3.x and #627 for master.

PR #479, opened 2021-06-14
[MNG-7173] The maven-resolver-provider:MavenResolverModule is broken

cstamas:
It is broken since 3.6.2 or so, when ModelBuilder got decorated with
JSR330 injections, that work in sisu but not in Guice as it needs
explicit bindings added.

This is an UT that proves that module is incomplete.


cstamas:
As expected, newly added UT fails:
```
Error:  Errors: 
Error:    MavenResolverModuleTest.smokeTest:31 » Creation Unable to create injector, see...
[INFO] 
Error:  Tests run: 28, Failures: 0, Errors: 1, Skipped: 0
```

cstamas:
Related: after https://issues.apache.org/jira/browse/MNG-6685 the Guice module was not updated.

cstamas:
This PR is defunct, Guice modules were removed in https://github.com/apache/maven/pull/1282

jira-importer:
Resolve #8282

PR #554, opened 2021-09-26
[MNG-6898] Drop plexus-component-metadata plugin from maven-core

cstamas:
Now that Plexus XMLs (and components) are out, no need
for this plugin in maven-core module anymore.

Changes:
* convert DefaultLifecyclePluginAnalyzer to JSR330 (last plexus component in core)
* drop dependency on plexus-component-annotations (note: it is still there but only as transitive dep of plexus-shim, needed to boot up Plexus Shim in tests)
* remove plexus-component-metadata plugin from build
* drop unused test Plexus XML

Fixes https://issues.apache.org/jira/browse/MNG-6898


cstamas:
This is now stalled until MNG-6898 is resolved (we have 2 PRs for that)

maven-core/src/test/resources/META-INF/plexus/components.xml
```diff
@@ -1,107 +0,0 @@
-<component-set>
```
gnodet, line 1: That's unrelated right ? It's just a left over from a previous commit ?

cstamas: Yes, but frankly, this XML was unused even before previous commit, it was simply a piece of garbage that I spotted when I went thru all Plexus XMLs

slachiewicz:
Looks good to me. I would only check impact on performance, because previously everything was stored in singleton/map and here we have calls for every component lookup.

cstamas:
> Looks good to me. I would only check impact on performance, because previously everything was stored in singleton/map and here we have calls for every component lookup.

Sisu caches

cstamas:
Opps

gnodet:
> > Looks good to me. I would only check impact on performance, because previously everything was stored in singleton/map and here we have calls for every component lookup.
> 
> IMO not biggie, but yes (note: even before, the injected map was not a "plain map", but a dynamic map provided by SISU)

I suppose fixing sisu so that `@Inject` would take care of plexus visibility rules has been considered and discarded ?

slachiewicz (approved):

gnodet (approved):

jira-importer:
Resolve #8610

PR #563, opened 2021-10-03
[MNG-7103] VersionScheme provider

cstamas:
This PR makes VersionScheme a component, is injected where needed
(instead of ad-hoc instantiation), but provides room for
different schemas, as GenericVersionScheme is "default"
but now nothing stops to add other schemes as well.

### DIFF
--- a/maven-resolver-provider/src/main/java/org/apache/maven/repository/internal/MavenResolverModule.java
+++ b/maven-resolver-provider/src/main/java/org/apache/maven/repository/internal/MavenResolverModule.java
@@ -34,6 +34,8 @@ import org.eclipse.aether.impl.MetadataGeneratorFactory;
 import org.eclipse.aether.impl.VersionRangeResolver;
 import org.eclipse.aether.impl.VersionResolver;
 import org.eclipse.aether.impl.guice.AetherModule;
+import org.eclipse.aether.util.version.GenericVersionScheme;
+import org.eclipse.aether.version.VersionScheme;
 
 /**
  * MavenResolverModule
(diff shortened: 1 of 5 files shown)


khmarbaise (approved):

maven-resolver-provider/src/main/java/org/apache/maven/repository/internal/MavenResolverModule.java
```diff
@@ -46,6 +47,7 @@
     protected void configure()
     {
         install( new AetherModule() );
+        bind( VersionScheme.class ).toProvider( new DefaultVersionSchemeProvider() );
```
gnodet, line 50: Just a quick question : is the explicit registration required ? Shouldn't sisu automatically wire the provider ?

cstamas: This class is NOT used in Maven, this is Guice module we provide for 3rd parties that use Guice for DI (so they can simply install this module and have "maven embedded"), as Resolver (the `AetherModule` above) itself is _incomplete_, Resolver has no notion of "maven models", have no idea how to read POM, repository metadata XML etc. This module "merely completes" resolver by adding "maven bits".

Sisu will pick up the DefaultVersionSchemeProvider from sisu-index, as Sisu Mojo generates index for this module.

Note: to use Guice modules within Sisu, you'd need a) directly add is somewhere (we do not), or b) annotate it with `@Named` (which is not) to have Sisu pick it up.

Also, see https://github.com/apache/maven/pull/479 and related issue.

gnodet (approved):

jira-importer:
Resolve #8346

PR #565, opened 2021-10-04
Mojo API update proposal

cstamas:
This is a DRAFT that started as booster for discussion about Mojo
logging. 

Re logging: I believe Maven should not get in a way to
Maven Plugin developers, but we still can offer
some "convenience" helpers and make things right
(like for example logging in ctor of Mojo).

IMHO, I distinguish 2 log-related situations:
* trivial to simple Maven Plugins (from one mojo plugin to more mojos + several "own"
  components): here developer can easily choose (and usually choose Mojo Log)
  to do the job. Maven is helping here.
* complex Maven plugins: plugins pulling in "foreign" frameworks/components/whatever,
  here logging is usually "given", but as Mojo class loader is "self first", it is really
  easy for developer to decide (and implement) which logging it wants (or must) to
  use, Maven should not interfere.

One thing IMHO remains, to make Mojo Log interface more "slf4j Logger"-lookalike.
This is just a "quick draft" adjusted to make ITs pass, but I am sure there are some (binary)
incompatibilities remaining we'd need to investigate and fix.

Other changes: ~~MavenSession (compatible) update, to make clear Mojo contexts are
ConcurrentMaps~~. And a bit of Mojo context cleanup, mostly regarding Javadoc
mentioning how should it be used.

Changes in this PR:
* un-deprecate Log and Log related methods on Mojo, AbstractMojo
* ~~introduce LogFactory **component** for those seeking for simple logging solution from Mojos and own components~~
* loggers are lazily looked up and "memoized" (thanks Romain)
* classes Mojo, AbstractMojo are mostly unchanged (Log injection here happens "manually" in DefMvnPluginMgr as before)
* ~~MavenSession plugin context change (is ConcurrentMap)~~
* Javadoc around Mojo Context

Issues:
* https://issues.apache.org/jira/browse/MNG-7084 -- indirectly fixes (does not adds injection of Log instances due simplicity, it ads a component that may be used for same thing, for example with using ctor injection)
* https://issues.apache.org/jira/browse/MNG-7083 -- TBD, make Log SLF4J Logger "alike"
* https://issues.apache.org/jira/browse/MNG-7082 -- fixes


maven-plugin-api/src/main/java/org/apache/maven/plugin/logging/LogFactory.java
```diff
@@ -0,0 +1,62 @@
...
+
+    private LogFactory()
+    {
+        // no instances of this can be created
+    }
+
+    /**
+     * Initialized Mojo LogFactory with a bridge.
+     */
+    public static void initLogFactory( ILogFactory bridge )
```
rmannibucau, line 42: shouldn't be exposed IMO and we should ensure it is called very early in maven lifecycle + calling it twice with a different value should likely fail. Also means the log factory shouldnt call it in its constructor otherwise we can't change it after, no?

maven-plugin-api/src/main/java/org/apache/maven/plugin/Mojo.java
```diff
@@ -48,25 +46,4 @@
      */
     void execute()
         throws MojoExecutionException, MojoFailureException;
-
```
rmannibucau, line 51: not sure it makes much sense to drop them, the setLog is useful to keep a lazy init but more importantly, both are used by mojo so it will reduce migration cost (which is not handlable at maven level) to not break them IMHO

cstamas:
ping @McFoggy

maven-plugin-api/src/main/java/org/apache/maven/plugin/AbstractMojo.java
```diff
@@ -144,55 +144,39 @@
     implements Mojo, ContextEnabled
 {
     /** Instance logger */
-    private Log log;
+    private final Log log = LogFactory.getLog( getClass() );
```
rmannibucau, line 147: open question: makes the lookup earlier, not sure it is good in all cases + it now prevents to reset the log which was sometimes done with extension components

maven-core/src/main/java/org/apache/maven/plugin/internal/MojoLogFactory.java
```diff
@@ -0,0 +1,56 @@
...
+{
+    public MojoLogFactory()
+    {
+        LogFactory.initLogFactory( this ); // init Mojo Logging
+    }
+
+    @Override
+    public Log getLog( final Class<?> clazz )
+    {
+        return new MojoLog( LoggerFactory.getLogger( clazz ) );
```
rmannibucau, line 48: maybe lazy lookup the actual underlying logger since SimpleLoggerFactory#loggerMap will grow more than needed since some mojo don't use logging at all. Also wonder if we should provide a cleanup callback to drop the loggers once the mojo is dropped (likely requires some reference counting but will avoid any leak), wdyt?

rmannibucau (changes requested):
A few early comments on common log factory pitfalls (multiple init, cleanup depending on classrealm etc) + a question on the compat we want for mojo to avoid everyone to be forced to rewrite to support maven.next. Hope it will help a bit.

maven-plugin-api/src/main/java/org/apache/maven/plugin/logging/LogFactory.java
cstamas: reworked, this is not here anymore

maven-plugin-api/src/main/java/org/apache/maven/plugin/AbstractMojo.java
cstamas: back to the roots (as it was)

maven-core/src/main/java/org/apache/maven/plugin/internal/MojoLogFactory.java
cstamas: Lazyness: done. Cleanup callback? meh?

maven-plugin-api/src/main/java/org/apache/maven/plugin/Mojo.java
cstamas: back to the roots (as it as)

cstamas:
Moved unrelated (to Mojo API) session changes to it's own PR https://github.com/apache/maven/pull/575

rmannibucau:
@cstamas the only agreement was for @apache/maven  project. For the API there is no slf4j agreement but the agreement to get a logging API. Since JUL is the built-in and well known java logging API and from maven standpoint we can completely control it properly for all plugins and components it makes sense to not reinvent the wheel and just reuse this *API* (we don't have to use the impl and should likely stay on slf4j until we rediscuss it). So don't read it as a "why not" but more a "the API we should use is already there".

rmannibucau:
@cstamas guess it comes from the factory which is likely considered as not needed and duplicating logger abstractions. Dropping it and 100% relying on Log is fine for me while you can inject a Log instance in a component and maybe solves this topic?

cstamas:
> When I created the PRs, I was looking for a standardized way to use & inject "Logger" (whatever the interface/impl) in different maven places. Standard plugins used to inherit from Mojo and thus access a Log interface from Mojo/AbstractMojo but **for core extensions (what I was more interested of) it was not clear what to use and how**. Looking at this PR it looks like the answer is now `inject and use LogFactory`.

@McFoggy sorry, I just reread the bits that I also highlighted, as seems I missed those at first read and response. Well, for "core extensions" is same what stands in maven core itself: slf4j. Given core extension is loaded like this: http://takari.io/book/91-maven-classloading.html

rmannibucau:
will remains external extensions which must not depend on slf4j so Log can likely become the abstraction to use for external code at least, no?

PR #575, opened 2021-10-09
[MNG-7294] Refactor MavenSession#getPluginContext to be thread safe

cstamas:
This PR introduce no API change, merely refactors MavenSession
getPluginContext method to truly thread-safe.

Also added Javadoc with explanations.

---
https://issues.apache.org/jira/browse/MNG-7294


jira-importer:
Resolve #8139

PR #603, opened 2021-11-01
[MNG-7344] track dependencyManagement import location in effective Model for MPH-183

jjkester:
The goal of [MPH-183](https://issues.apache.org/jira/browse/MPH-183) is to print via which BOM(s) a dependencyManagement dependency got in the effective POM in case of dependencyManagement import.
MNG-7344 implements the required tracking in Maven core effective model.

This information was previously not available to the help plugin. This MR:

- introduces a change to the `InputSource` model to keep a hierarchy of sources (which are POM files) by tracking the POM that imported it
- tracks dependencies through different dependency management imports, setting the hierarchy of sources on the `InputSource` objects

For now, this code has been written under the assumption that a BOM hierarchy (at least at the point the dependency management blocks are imported) is tree-shaped, without diamonds. That is, given a project `P` with dependency `D`, importing BOMs `A` and `B`, BOM `C` with dependency `D`, only `A` or `B` (not both) can import `C`.

- [ ] Check assumption

If such a structure would be possible, the end result may be incorrectly represented as `C via B via A`, which is not correct, because there is no direct relationship between `A` and `B`.

Furthermore, at the moment the original `InputSource` of the file is mutated. This makes the hierarchy valid for all imported dependencies, and potentially other parts of the effective POM. This may not be desirable. A relatively minor change, replacing `InputSource` objects that are about to be mutated, would solve this issue. In case updating the original `InputSource` is desirable, there _may_ be a better place to set it.

- [ ] Check solution strategy

Any input on the above points is highly appreciated.

A pull request for printing this information in the effective POM: https://github.com/apache/maven-help-plugin/pull/37

---

       pulling in other changes.
       commit message.


maven-model-builder/src/main/java/org/apache/maven/model/composition/DefaultDependencyManagementImporter.java
```diff
@@ -83,4 +90,43 @@ public void importManagement( Model target, List<? extends DependencyManagement>
...
+        InputSource bomSource = bomLocation.getSource();
+
+        // Skip if the dependency and bom have the same source
+        if ( hierarchicalSource == null || bomSource == null || Objects.equals( hierarchicalSource.getModelId(),
+                bomSource.getModelId() ) )
+        {
+            return;
+        }
+
+        while ( hierarchicalSource.getImportedBy() != null )
```
rfscholte, line 114: Is this loop really necessary? Would love to see an example that needs it. (maybe as an integration test)

maven-model-builder/src/main/java/org/apache/maven/model/composition/DefaultDependencyManagementImporter.java
```diff
@@ -83,4 +90,43 @@ public void importManagement( Model target, List<? extends DependencyManagement>
...
+        // We are only interested in the InputSource, so the location of the <dependency> element is sufficient
+        InputLocation dependencyLocation = dependency.getLocation( "" );
+        InputLocation bomLocation = bom.getLocation( "" );
+
+        if ( dependencyLocation == null || bomLocation == null )
+        {
+            return;
+        }
+
+        InputSource hierarchicalSource = dependencyLocation.getSource();
```
rfscholte, line 104: This is actually very interesting: it looks like it is possible to even use InputLocation instead of InputSource, whcih would give even more details of the path. How about using that instead?

rfscholte (changes requested):

maven-model/src/main/mdo/maven.mdo
```diff
@@ -3137,6 +3137,20 @@
             ]]>
           </description>
         </field>
+        <field>
+          <name>importedBy</name>
```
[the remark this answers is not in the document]
jjkester: I will look into that approach.

I did consider `pluginManagement` as well, but this approach (as-is) would be easy to replicate in the similar `DefaultPluginManagementImporter`.

In order to keep track of all references, I think it needs to be tracked elsewhere (and I have not found that location yet - but I will have a look).

In order to support keeping the type of reference, I propose we do not reference an `InputLocation` or `InputSource` (see also the discussion [here](https://github.com/apache/maven/pull/603#discussion_r745439915)) directly, but use a container type that can be extended (for example `ReferencedInputLocation`).

maven-model-builder/src/main/java/org/apache/maven/model/composition/DefaultDependencyManagementImporter.java
jjkester: I had the impression while testing that in long chains (e.g. `pom - bom1 - bom2 - bom3 - bom4`) parts of the objects defined in `bom4` are copied to `bom3`, then `bom2`, etc. This loop ensures that we set a new location at the right place, without overwriting any existing. I will however investigate this. Also, I expect this code to be quite different if we decide to go with `InputLocation` (see [here](https://github.com/apache/maven/pull/603#discussion_r745439915)) so I'll look into that first.

maven-model-builder/src/main/java/org/apache/maven/model/composition/DefaultDependencyManagementImporter.java
jjkester: That is interesting. `InputLocation` is tied to a specific line in the file, so making those hierarchical would only work for single locations. Given that a dependency has many, we either need to decide on what to track or just modify them all (the latter would of course be much more generic). Since the `InputSource` objects are per-file, that is just a quick way to get some order.

Just to check your suggestion, would your idea be that an `InputSource` refers to the `InputLocation` where it is referenced? In that case (keeping in mind a more generic solution), I think there are a few things to investigate:

- Is there a place in the code where we have access to the `InputLocation` of the reference (parent, import, ...) and the `InputSource` of the referenced POM?
- In this place in the code, is the model already a tree or is it a graph? I.e., is there only one path from the root POM to every other (indirectly) referenced POM?

A simple test case for the latter could be a project like this:

```
   pom.xml --imports-> bom1
      |                 |
   imports           imports
      |                 |
      v                 v
    parent --imports-> bom2
                        |
                     imports
                        |
                        v
                       bom3
```

There are two paths to resolve a dependency defined in `bom3`, ideally there should be a deterministic way to define the path that Maven takes. And that path should (in my opinion) be accurately reflected in the model. Currently my knowledge is too limited to determine how Maven actually resolves this.

rfscholte: For the effective pom there will be only one route, either `bom1` or `parent` wins. Saying it differently: if something in bom3 has to be changed and I have access to both bom1 and parent, I need to know which of the 2 has to be changed.

jjkester:
Hey @rfscholte, would it be appropriate to create a new ticket for this? The ticket this work is under is focused on maven-help-plugin, but this is turning out a bit more involved. I'm happy to work on it, but I think it makes sense to make the requirements, reasoning and discussion easily findable at a location where it is expected.

rfscholte:
I've created https://issues.apache.org/jira/browse/MNG-7344

api/maven-api-model/src/main/java/org/apache/maven/api/model/InputLocation.java
```diff
@@ -83,6 +95,13 @@ public Map<Object, InputLocation> getLocations() {
         return locations;
     }
 
+    /**
+     * Gets the input location that caused this model to be read.
+     */
```
hboutemy, line 100: we need an `@since` here to remember

api/maven-api-model/src/main/java/org/apache/maven/api/model/InputLocationTracker.java
```diff
@@ -20,4 +20,6 @@
 
 public interface InputLocationTracker {
     InputLocation getLocation(Object field);
+
+    InputLocation getImportedFrom();
```
hboutemy, line 24: same `@since`

api/maven-api-model/src/main/java/org/apache/maven/api/model/InputSource.java
```diff
@@ -64,6 +74,14 @@ public String getModelId() {
         return this.modelId;
     }
 
+    public InputLocation getImportedFrom() {
+        return importedFrom;
+    }
+
+    public InputSource importedFrom(InputLocation importedFrom) {
+        return new InputSource(modelId, location, importedFrom);
```
hboutemy, line 82: really create a new object and forget about it? strange (i did not really dig in depth, just giving initial feedback)

api/maven-api-model/src/main/java/org/apache/maven/api/model/InputSource.java
```diff
@@ -64,6 +74,14 @@ public String getModelId() {
         return this.modelId;
     }
 
+    public InputLocation getImportedFrom() {
```
hboutemy, line 77: same `@since`

maven-model/src/main/java/org/apache/maven/model/InputLocation.java
```diff
@@ -217,6 +223,24 @@ public InputSource getSource() {
         return this.source;
     } // -- InputSource getSource()
 
+    /**
+     * Get the imported from location.
+     *
+     * @return InputLocation
```
hboutemy, line 229: same `@since`

hboutemy (changes requested):
now I need to check MPH-183

and I'd love to have equivalent update for 3.9.x: I'll dig into it later

api/maven-api-model/src/main/java/org/apache/maven/api/model/InputLocation.java
hboutemy: "that caused this model to be read."?
I need to dig more into code to propose another description, but this one is not clear

maven-model-builder/src/main/java/org/apache/maven/model/building/DefaultModelBuilder.java
```diff
@@ -1683,13 +1684,20 @@ private void importDependencyManagement(
...
-                if (importMgmts == null) {
-                    importMgmts = new ArrayList<>();
-                }
-
+            if (request.isLocationTracking()) {
+                // Keep track of why this DependencyManagement was imported.
+                // And map model v3 to model v4 -> importMgmt(v3).getDelegate() returns a v4 object
+                importMgmts.add(
+                        org.apache.maven.api.model.DependencyManagement.newBuilder(importMgmt.getDelegate(), true)
+                                .importedFrom(dependency.getDelegate().getLocation(""))
```
hboutemy, line 1698: importedFrom should not be assigned to DependencyManagement but to each Dependency, isn't it?

api/maven-api-model/src/main/java/org/apache/maven/api/model/InputSource.java
Giovds: This is due to the immutable model, but I think we can simply remove this and make the new constructor public. We don't seem to use it (anymore) upon further inspection...

maven-model-builder/src/main/java/org/apache/maven/model/building/DefaultModelBuilder.java
juulhobert: @hboutemy you're right. It should only map it from model v3 to model v4. We're going to remove the line `.importedFrom(dependency.getDelegate().getLocation(""))`

gnodet:
The changes to `DefaultModelBuilder` and `DefaultDependencyManagementImporter` need to be moved from `maven-model-builder` to the implementations in `maven-api-impl`, as the ones from `maven-model-builder` are kept for compatibility but are not used anymore.

gnodet (changes requested):
The changes to `DefaultModelBuilder` and `DefaultDependencyManagementImporter` need to be moved from `maven-model-builder` to the implementations in `maven-api-impl`, as the ones from `maven-model-builder` are kept for compatibility but are not used anymore.

gnodet:
> The changes to `DefaultModelBuilder` and `DefaultDependencyManagementImporter` need to be moved from `maven-model-builder` to the implementations in `maven-api-impl`, as the ones from `maven-model-builder` are kept for compatibility but are not used anymore.

@Giovds I've merged, moved the code and fixed a few things, but I cannot push to your branch. 
I've pushed to https://github.com/gnodet/maven/commits/MPH-183_dependency-management-hierarchy-inputsource/ instead.
The build seems to work for Maven, but I cannot find a way to display with [MPH-183 PR](https://github.com/apache/maven-help-plugin/pull/37).

juulhobert:
@gnodet I've merged your changes into this branch. I've also verified that the output is still the same. If you would like to test it yourself, these are the reproduction steps:
- Download example project [mng-7344.zip](https://github.com/user-attachments/files/16192112/mng-7344.zip)
- Install dependencies and boms in the same sequence as in `validate.sh`
- `mvn install` help plugin branch `MPH-183_effective-pom-path-to-source`
- `mvn org.apache.maven.plugins:maven-help-plugin:3.4.1-SNAPSHOT:effective-pom -Dverbose` (replace mvn with the packaged Maven from this branch)

image

juulhobert:
@mthmulders can you please run the failed integration tests again? I expect it to give a green check mark this time (see https://github.com/infosupport/maven/actions/runs/10215678351)

gnodet:
@juulhobert I merged another PR which causes a small conflict with this one.
As I can't write to your branch, could you please merge master and push ?

juulhobert:
@gnodet I've merged the changes and pushed it

gnodet (approved):

jira-importer:
Resolve #8095

PR #604, opened 2021-11-03
Finish switch to modello

gnodet:
This fixes a possible exception when using forked lifecycles.
In my case, this happens when using the `spotbugs-maven-plugin:spotbugs` plugin which forks the lifecycle (see https://github.com/spotbugs/spotbugs-maven-plugin/blob/spotbugs/src/main/groovy/org/codehaus/mojo/spotbugs/SpotbugsViolationCheckMojo.groovy#L57).


gnodet:
@hboutemy should I commit them directly to the MNG-7129 branch ?

hboutemy:
> @hboutemy should I commit them directly to the MNG-7129 branch ?

@gnodet my idea is that you should commit directly to MNG-7129_Modello = the place where we prepare the PR that we will review all together before merging to MNG-7129

maven-core/src/main/java/org/apache/maven/lifecycle/internal/DefaultLifecycleExecutionPlanCalculator.java
```diff
@@ -554,6 +554,8 @@ private void injectLifecycleOverlay( Map<String, List<MojoExecution>> lifecycleM
 
         MojoExecution forkedExecution = new MojoExecution( forkedMojoDescriptor, forkedGoal );
 
+        forkedExecution.setLifecyclePhase( mojoExecution.getLifecyclePhase() );
```
hboutemy, line 557: how does that relate to MNG-7129 and the switch from JAXB to Modello?

gnodet: It does not relate to JAXB, however, this fixes a NPE which happens in the cache system when using forked lifecycles.  It may be fixed in a different way, but it did look like the easier fix.
My reasoning is the following : I suppose no-one ever looked for the phase of forked plugin executions and it makes sense to me to have those phases set to the phase of the original mojo execution.
The cache system does look and uses the phase in several locations, and this caused NullPointerException.

maven-core/pom.xml
```diff
@@ -235,11 +235,16 @@ under the License.
             <goals>
               <goal>java</goal>
               <goal>xpp3-reader</goal>
+              <goal>xpp3-writer</goal>
```
hboutemy, line 238: do we really need a writer for cache files?

hboutemy: ok, I found my answer: not really for domain or config, but we need it for diff and report

hboutemy:
ok, now that I reviewed and understood many parts, here my proposal:
1. the first 2 commits should be directly be done in MNG-7129 branch: they are nice little fixes independent from Modello switch
2. last 3 commits should go directly to MNG-7129_Modello branch (that will have to be rebased on MNG-7129)

and the next update I want on MNG-7129_Modello branch is the removal of all dependencies that were added for JAXB: removing them is a key benefit

the full idea behind splitting everything in many buckets is that both original developers of MNG-7129 need to be able to review/understand what is done to their original code: for usual Maven contributors, all the work is about lowering the bar to understanding the global MNG-7129 branch to be able to merge some time later to master

hboutemy:
thinking a little bit more, I just found that merging the PR to MNG-7129_Modello branch is what I just need to do
and only inject the 2 first commits to MNG-7129

I'll do that and we can continue the work together

hboutemy:
ok, I did the cherry-picks and the rebase: I think we can close this PR
I'm now cleaning up unused dependencies

maven-core/src/main/java/org/apache/maven/lifecycle/internal/DefaultLifecycleExecutionPlanCalculator.java
AlexanderAshitkin: Hi
Thanks for the change! Jaxb is gone for the good. In overall pr looks good for me.
Few notes: 
1. as i understand modello schema is located in core. if we move the cache to an extension in future - doest it mean there will be dependency on core from the extension? Looks like modello scheme actuall couples the cache and the core.
2 There should be few small jaxb related leftovers - dependency in pom itself and in licenses, but in general it looks good.

gnodet: @AlexanderAshitkin I don't see why the modello generated code could not be moved into the extension.

PR #617, opened 2021-11-23
[MNG-7129] http repo refactoring for java8

AlexanderAshitkin:
Cache repositories refactoring for java8 - using optionals instead of nullables

       pulling in other changes.
       commit message.

### DIFF
--- a/maven-caching/src/main/java/org/apache/maven/caching/CacheControllerImpl.java
+++ b/maven-caching/src/main/java/org/apache/maven/caching/CacheControllerImpl.java
@@ -40,6 +40,8 @@ import java.util.TreeSet;
 import java.util.UUID;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ConcurrentMap;
+import java.util.concurrent.Future;
+import java.util.concurrent.FutureTask;
 import java.util.function.Function;
 import java.util.regex.Pattern;
 import java.util.stream.Collectors;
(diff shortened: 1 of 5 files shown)


michael-o (commented):
I believe that this entire code needs to be rewritten to use Wagon instead of manually fiddling with `HttpClient`.

AlexanderAshitkin:
> I believe that this entire code needs to be rewritten to use Wagon instead of manually fiddling with `HttpClient`.

good point and should not be a big deal. will try to do it in following prs

gnodet:
> > I believe that this entire code needs to be rewritten to use Wagon instead of manually fiddling with `HttpClient`.
> 
> good point and should not be a big deal. will try to do it in following prs

Not necessarily instead, but a simple abstraction layer would allow to use the default http repository or use wagon as an alternative and more versatile provider.

AlexanderAshitkin:
> > > I believe that this entire code needs to be rewritten to use Wagon instead of manually fiddling with `HttpClient`.
> > 
> > 
> > good point and should not be a big deal. will try to do it in following prs
> 
> Not necessarily instead, but a simple abstraction layer would allow to use the default http repository or use wagon as an alternative and more versatile provider.

Agree. It could be reworked as we like in the future, just at this point it will be a small cleanup which doesnt have much of implications. also, it was tested, so it will be great to merge it.

Conceptually i agree with more comprehensive rework in future

gnodet:
> > > > I believe that this entire code needs to be rewritten to use Wagon instead of manually fiddling with `HttpClient`.
> > > 
> > > 
> > > good point and should not be a big deal. will try to do it in following prs
> > 
> > 
> > Not necessarily instead, but a simple abstraction layer would allow to use the default http repository or use wagon as an alternative and more versatile provider.
> 
> Agree. It could be reworked as we like in the future, just at this point it will be a small cleanup which doesnt have much of implications. also, it was tested, so it will be great to merge it.
> 
> Conceptually i agree with more comprehensive rework in future

I'm going to merge.  What's the reasoning for using an `Optional ` instead of throwing an exception ?

AlexanderAshitkin:
> I'm going to merge.  What's the reasoning for using an `Optional ` instead of throwing an exception ?

Hi. The rationale is that cache misses are natural. Optionals allow to isolate normal flow from real errors (like transport and format issues) and to avoid flood of not so useful "not found" exceptions. If it is necessary to assert a resource presence it should be done in cache logic by a caller

PR #623, opened 2021-11-26
Mng 7129/master db subtree feature

andrey-petrenko-db:
This PR provides:

1) support to change poms version (for example during release), like it is done by maven set version plugin
Version in pom files do not affect checksum now
2) support to sub tree build when project build start not from multi module root but from one of sub project.
In that case if in cache configuration present the corresponding section the logic tries to build whole project models starting from multi module root pom file. This allows to calculate checksums for some dependencies which are not in maven session (not currently building) but belong to multi module project.

For example:

If I have root (R), module (A), module (B)
And (A) depends on (B)

If I started to build (A) I don't see (B) as the project in MavenSession, hence I cannot determine whether it "my" dependency and it is possible to calculate checksum for it

If I have the following configuration
```
    <configuration>
        ...
        <multiModule>
            <discovery>
                <scanProfiles>
                    <scanProfile>all</scanProfile>
                </scanProfiles>
            </discovery>
        </multiModule>
    </configuration>
```

Then the logic tries to build whole multi module projects models which allows to find (B) and calculate checksum for it and include this value to checksum for (A)

If scanProfile not specified then it tries to build whole model with the same profiles which specified for sub tree build.
If multiModule section in configuration is not specified then this logic skipped and (B) treated as external dependency

       pulling in other changes.
       commit message.


maven-caching/src/main/java/org/apache/maven/caching/checksum/MavenProjectInput.java
```diff
@@ -743,75 +695,49 @@ private boolean isFilteredOutSubpath( Path path )
...
-            if ( currentlyBuilding )
+            String projectHash;
+            if ( dependencyProject != null ) //part of multi module
             {
-                resolved = projectArtifactsByKey.get( artifactKey );
-                if ( resolved == null )
-                {
-                    throw new DependencyNotResolvedException( "Expected dependency not resolved: " + dependency );
-                }
+                projectHash = projectInputCalculator.calculateInput( dependencyProject ).getChecksum();
```
AlexanderAshitkin, line 712: result is not cached, potentially could be recalculated many times for the same dependency in different modules

AlexanderAshitkin (approved):

maven-caching/src/main/java/org/apache/maven/caching/CacheControllerImpl.java
```diff
@@ -397,67 +415,126 @@ private RestoredArtifact restoredArtifact( org.apache.maven.artifact.Artifact pa
...
         {
-            LOGGER.debug( "Downloading artifact {}", artifact.getArtifactId() );
-            final Path artifactFile = localCache.getArtifactFile( context, cacheResult.getSource(),
-                    artifact );
-            if ( !Files.exists( artifactFile ) )
+            return artifactFile;
+        }
+
+        String currentVersion = project.getVersion();
+        File tmpJarFile = File.createTempFile( artifactFile.toFile().getName(),
```
AlexanderAshitkin, line 433: i woul better avois using tmp - on many machines tmp could be 2G or similar. All downloaded artifacts might just not fit there.  There is an [File.html#createTempFile](https://docs.oracle.com/javase/7/docs/api/java/io/File.html#createTempFile(java.lang.String,%20java.lang.String,%20java.io.File)) method which takes directory as a parameter - it is better to use it and create artifacts somethere under the target dir (in assumption that there should be enough space to create them) + easy to clean.

maven-caching/src/main/java/org/apache/maven/caching/CacheControllerImpl.java
```diff
@@ -397,67 +415,126 @@ private RestoredArtifact restoredArtifact( org.apache.maven.artifact.Artifact pa
...
-        }
-        String extension = artifactHandler.getExtension();
-        if ( !Objects.equals( extension, artifact.getType() ) && !Objects.equals( extension, packaging ) )
+        byte[] bytes = modified.getBytes( StandardCharsets.UTF_8 );
+        JarEntry newEntry = new JarEntry( entry.getName() );
+        jos.putNextEntry( newEntry );
+        jos.write( bytes );
+    }
+
+    private static boolean isArchive( File file )
```
AlexanderAshitkin, line 506: should be izZip and contains("META-INF")

maven-caching/src/main/java/org/apache/maven/caching/CacheControllerImpl.java
```diff
@@ -397,67 +415,126 @@ private RestoredArtifact restoredArtifact( org.apache.maven.artifact.Artifact pa
...
+                throw new FileNotFoundException(
+                        "Missing file for cached build, cannot restore. File: " + artifactFile );
             }
-            final MavenProjectInput inputs = new MavenProjectInput( project, projectIndex, session, cacheConfig,
-                    artifactDigestByKey, repoSystem, localCache, remoteCache );
-            return inputs.calculateChecksum();
-        }
-        catch ( Exception e )
+            LOGGER.debug( "Downloaded artifact " + artifact.getArtifactId() + " to: " + artifactFile );
+            return adjustArchiveArtifactVersion( project, originalVersion, artifactFile ).toFile();
```
AlexanderAshitkin, line 531: will be good to keep under flag - performance penalty of copying could be significant

maven-caching/src/main/java/org/apache/maven/caching/CacheControllerImpl.java
```diff
@@ -397,67 +415,126 @@ private RestoredArtifact restoredArtifact( org.apache.maven.artifact.Artifact pa
...
+
+    private static boolean isArchive( File file )
+    {
+        String fileName = file.getName();
+        if ( !file.isFile() || file.isHidden() )
         {
-            artifactDigestByKey.put( KeyUtils.getArtifactKey( artifact, extension ), projectArtifact );
+            return false;
         }
+        return fileName.endsWith( ".jar" ) || fileName.endsWith( ".zip" )
```
AlexanderAshitkin, line 513: is [StringUtils#endsWithAny](https://commons.apache.org/proper/commons-lang/apidocs/org/apache/commons/lang3/StringUtils.html#endsWithAny-java.lang.CharSequence-java.lang.CharSequence...-) avilable? 
```
 StringUtils.endsWithAny("abcxyz", new String[] {null, "xyz", "abc"}) = true
 StringUtils.endsWithAny("abcXYZ", "def", "XYZ") = true
 StringUtils.endsWithAny("abcXYZ", "def", "xyz") = false
```

maven-caching/src/main/java/org/apache/maven/caching/CacheControllerImpl.java
```diff
@@ -397,67 +415,126 @@ private RestoredArtifact restoredArtifact( org.apache.maven.artifact.Artifact pa
...
     }
 
-    private Future<File> createDownloadTask( CacheResult cacheResult, CacheContext context, MavenProject project,
-                                             Artifact artifact )
-            throws ExecutionException, InterruptedException
+    //we might store in cache artifact which was build with previous version
+    //1.0-SNAPSHOT is kept in cache but real version of project is 2.0
+    //for pom packaging this is done automatically by maven but for jar and other there might be
+    //sensitive metadata with previous version. Versions mismatch could lead errors
+    private Path adjustArchiveArtifactVersion( MavenProject project, String originalArtifactVersion, Path artifactFile )
```
AlexanderAshitkin, line 422: i would move it to an utility class just to make it more testable

maven-caching/src/main/java/org/apache/maven/caching/checksum/MavenProjectInput.java
andrey-petrenko-db: > 

the result is cached take a look at https://github.com/apache/maven/pull/623/files#diff-e4da53c029acd1f06075465c790bd5c3f8330048ee490be33cac19d49a5b9130R53

maven-caching/src/main/java/org/apache/maven/caching/CacheControllerImpl.java
andrey-petrenko-db: > 

what does zip.contains("META-INF") means, should we unpack archive and try to find META-INF inside? Can we have artifacts without META-INF? I suspect yes...

maven-caching/src/main/java/org/apache/maven/caching/CacheControllerImpl.java
andrey-petrenko-db: > 

should we avoid potential directory names clash? how to do this?

maven-caching/src/main/java/org/apache/maven/caching/CacheControllerImpl.java
AlexanderAshitkin: ZipFile.stream() doesn't unpack it afair - just gives stream of entries

maven-caching/src/main/java/org/apache/maven/caching/CacheControllerImpl.java
AlexanderAshitkin: Risk of conflicts is not higher than in any over cases for any over plugins. I don't even know if need to solve this problem. But artifacts need to stay for a time of the build so everyone of them need to fit tmp in order to succeed. It might be not convenient.

maven-caching/src/main/java/org/apache/maven/caching/DefaultProjectInputCalculator.java
```diff
@@ -0,0 +1,127 @@
...
+    }
+
+    @Override
+    public ProjectsInputInfo calculateInput( MavenProject project )
+    {
+        LOGGER.info( "Going to calculate checksum for project [groupId=" + project.getGroupId()
+                + ", artifactId=" + project.getArtifactId() + "]" );
+
+        String key = BuilderCommon.getKey( project );
+        ProjectsInputInfo projectsInputInfo = checkSumMap.get( key );
```
AlexanderAshitkin, line 82: looks exactly like computeIfAbsent

andrey-petrenko-db: > could be simplified with computeIfAbsent

recursive computeIfAbsent calls are not allowed in CHM, this could lead to IllegalStateException("Recursive update")

here is the example to reproduce

```
    private static ConcurrentMap<Integer, String> suspect = new ConcurrentHashMap<>();

    private static String recursiveUpdate( int i )
    {
        if ( i == 16 )
        {
            return String.valueOf( i );
        }
        return suspect.computeIfAbsent( i, k -> recursiveUpdate( k++ ) );
    }


    public static void main(String[] args) {
        recursiveUpdate( 0 );

        System.out.println( "Never printed" );
    }
```
In jdk8 there could be infinite loop
https://stackoverflow.com/questions/28840047/recursive-concurrenthashmap-computeifabsent-call-never-terminates-bug-or-fea

AlexanderAshitkin: Interesting, thank you. worth to leave a code comment because it is not obvious (and recursive calls too tbh) - probably i'm not the last one of thinking to refactor this.

maven-caching/src/main/java/org/apache/maven/caching/CacheControllerImpl.java
andrey-petrenko-db: > To be clear, i didn't meant to restore artifacts on original paths. I meant create temp files in a project 'target/build-cache-tmp'. Risk of conflicts in thta case is not higher than in any over cases for any over plugin. I don't even know if need to solve this problem. But in current design all artifacts need to stay for a time of the build so each of them needs to fit tmp in order to succeed. It might be not convenient and cause env specific failures.


changed destination to target/cache-build-tmp

maven-caching/src/main/java/org/apache/maven/caching/CacheControllerImpl.java
andrey-petrenko-db: > will be good to keep it under flag - performance penalty of copying/repacking could be significant and not necessary in verify scenarios

done

maven-caching/src/main/java/org/apache/maven/caching/CacheControllerImpl.java
andrey-petrenko-db: > StringUtils#endsWithAny could be handy here

changed

maven-caching/src/main/java/org/apache/maven/caching/CacheControllerImpl.java
andrey-petrenko-db: > i would move it to an utility class just to make it easier testable

moved to CacheUtils

maven-caching/src/main/java/org/apache/maven/caching/DefaultProjectInputCalculator.java
andrey-petrenko-db: > worth to leave a code comment because it is not obvious (and recursive calls not immediately visible too tbh) - probably i'm not the last one of thinking to refactor this.

added comment

PR #628, opened 2021-12-03
[3.8.x][MNG-7156][MNG-7285] Add locking in MojoExecutor

gnodet:
Same as https://github.com/apache/maven/pull/627 for for 3.8.x branch.


maven-core/src/test/java/org/apache/maven/project/ProjectBuilderTest.java
```diff
@@ -119,6 +119,18 @@ public void testResolveDependencies()
         assertEquals( 1, results.size() );
         MavenProject mavenProject = results.get( 0 ).getProject();
         assertEquals( 1, mavenProject.getArtifacts().size() );
+
+        final MavenProject project = mavenProject;
+        final int[] getArtifactsResultInAnotherThead = { 0 };
```
garydgregory, line 124: Youi'll likely want to use an AtomicInteger here instead of this classic hack (in the good sense of the word hack of course)

PR #629, opened 2021-12-03
[master][MNG-7350] Introduce a factory for ModelCache

gnodet:
`mvnd` does redefine a few maven classes in order to plugin its caching layer or change some behaviours, but this is sometime very difficult without copying some core class entirely.
One of those class is the `ProjectBuilder` and it's being reimplemented in `mvnd` only because the `ModelCache` is created using a direct call on the implementation.
This ticket aims to introduce a `ModelCacheFactory` so that `mvnd` can simply define a custom factory instead of redefining the full `ProjectBuilder`.


michael-o (approved):
This truly makes sense. @cstamas Has tackled similar issues with Resolver earlier this year.

cstamas (approved):

cstamas:
> This truly makes sense.
Actually all these "extension points" should be componentized (if not already), for easier extension.

cstamas:
Is https://github.com/apache/maven/pull/630/files#diff-50ddbca622e461d2056985a5c792fc4652574831f3a8b69da9be3a2f5af9e552R65 missing from here or I missed it?

gnodet:
> Is https://github.com/apache/maven/pull/630/files#diff-50ddbca622e461d2056985a5c792fc4652574831f3a8b69da9be3a2f5af9e552R65 missing from here or I missed it?

Nice catch, I missed it because the module was renamed. I've also added a short javadoc on the two new classes.

PR #651, opened 2022-01-06
[MNG-7264] refactor packaging bindings providers to ease maintenance

hboutemy:
same as #650 / MNG-7263 but on per-packaging plugins bindings for default lifecycle


cstamas:
I'd propose following change (to both related PRs): https://github.com/apache/maven/pull/652

maven-core/src/main/java/org/apache/maven/lifecycle/providers/packaging/AbstractLifecycleMappingProvider.java
```diff
@@ -22,50 +22,40 @@
...
-  }
-
-  @Override
-  public LifecycleMapping get()
-  {
-    return lifecycleMapping;
-  }
+    private final LifecycleMapping lifecycleMapping;
+
+    protected AbstractLifecycleMappingProvider( String[] pluginBindings )
```
mthmulders, line 40: If I'm not mistaken, all plugin bindings are a one-to-one mapping of phase -> plugin (see subclasses). Why not change this parameter to a `Map<String, String>`, and have subclasses pass the mapping as a `Map<String, String>`? It feels as if a `String[]` (where each consecutive two elements form a "pair") is the wrong abstraction in this case.

cstamas: Goal is "ease of documentation" AFAIU... as before, doco was simply pointing to XML (that was nicely formatted and totally "human readable"), I think Herve plans the same thing here, just have doco point at (trivially simple) class. So while I agree that abstraction seems not the best, keep in mind the reason of this change of his (if am not mistaken). And I agree, Java11 or Guava ImmutableMaps would really help here to build a Map in "readable" way.

hboutemy: previously, everything was pure XML parsing: Tamasz switched to Provider code, but still using old `new LifecyclePhase( String )` that parses (and splits on comma if there are multiple bindings
so we're going step by step from Plexus XML to perfectly maintainable code (and maintainable code will match the documentation objective: code before this PR is just not readable for documentation, then is not maintainable either = what is even more important than documentation)

mthmulders: Makes sense, one improvement at a time... Getting rid of the XML descriptors is a big win already!

michael-o (approved):
Frenchtastic!

maven-core/src/main/java/org/apache/maven/lifecycle/providers/packaging/AbstractLifecycleMappingProvider.java
```diff
@@ -0,0 +1,69 @@
...
+public abstract class AbstractLifecycleMappingProvider
+    implements Provider<LifecycleMapping>
+{
+    private final LifecycleMapping lifecycleMapping;
+
+    protected AbstractLifecycleMappingProvider( String[] pluginBindings )
+    {
+        requireNonNull( pluginBindings );
+        final int len = pluginBindings.length;
+        if ( len < 1 || len % 2 != 0 )
```
cstamas, line 46: do we allow "null" (empty) lifecycle? If so, remove the `len <. 1`. If not, make it `len < 2`

hboutemy: clearly, 0 does not make sense
in reality, the smallest existing packaging bindings have 2 bindings = install + deploy

jira-importer:
Resolve #8183

PR #652, opened 2022-01-06
Proposal to change abstract provider

cstamas:
Proposed changes for https://github.com/apache/maven/pull/651:
* validate input
* do not modify loop invariant within loop body

Remarks:
* IF we allow "null" life-cycles then modify validation from `if ( len < 1 || len % 2 != 0 )` to `if ( len % 2 != 0 )`
* IF we do NOT allow "null" life-cycles, then modify validation from `if ( len < 1 || len % 2 != 0 )` to `if ( len < 2 || len % 2 != 0 )`

(and in both cases adjust exception message as well)


hboutemy:
nice improvements, thank you

PR #658, opened 2022-01-14
[MNG-7339] Test if Maven can build itself on less OS'es & JDK's

mthmulders:

 - [X] Re-using the existing [Jira issue: MNG-7339](https://issues.apache.org/jira/browse/MNG-7339)
 - [ ] ~~Run `mvn clean verify` to make sure basic checks pass. A more thorough check will
       be performed on your pull request automatically.~~
 - [ ] ~~You have run the [Core IT][core-its] successfully.~~

Follow-up on a quick chat that took place on the ASF Slack, this pull request removes the following entities from the build matrix that checks if Maven can build itself:
- macOS
- JDK 11

This brings the size of the build matrix from 9 down to 4.

### DIFF
--- a/.github/workflows/maven_build_itself.yml
+++ b/.github/workflows/maven_build_itself.yml
@@ -23,8 +23,8 @@ jobs:
   build:
     strategy:
       matrix:
-        os: [ubuntu-latest, windows-latest, macOS-latest]
-        java: [8, 11, 17]
+        os: [ubuntu-latest, windows-latest]
+        java: [8, 17]
       fail-fast: false
 
     runs-on: ${{ matrix.os }}


.github/workflows/maven_build_itself.yml
```diff
@@ -23,8 +23,8 @@ jobs:
   build:
     strategy:
       matrix:
-        os: [ubuntu-latest, windows-latest, macOS-latest]
-        java: [8, 11, 17]
+        os: [ubuntu-latest, windows-latest]
```
olamy, line 26: Why not osx?

olamy (changes requested):

.github/workflows/maven_build_itself.yml
mthmulders: This is a secondary check (verifies Maven SNAPSHOT can build itself), doesn't run the full IT suite. I asked [on Slack](https://the-asf.slack.com/archives/C7Q9JB404/p1641824261011800) a couple of days ago and it seemed nobody cared for this particular check to run on OSX. [We do run the full IT's on OSX, though.](https://github.com/apache/maven/blob/master/.github/workflows/maven.yml#L53).

jira-importer:
Resolve #8084

PR #689, opened 2022-03-07
[NOT FOR MERGE] Example of Maven BOMs

cstamas:
The "old" [maven-bom](https://github.com/apache/maven/blob/1fff17398ed9738015a992502481bb79a2797ab5/maven-bom/pom.xml) introduced in MNG-6806 provides dependencyManagement only for artifacts in "org.apache.maven" groupId, that are produced by the reactor

This PR provides a new o.a.m.boms:bom-maven (new groupId and artifactId) that provides ALL the transitive dependencies, a project can import bom-all and have all the versions set to those used by given version of Maven.

Also, removed root defined and inherited dependencies,
it is not nice, and was not consistent, as some modules did redefine them. Now every module explicitly defines things it depends on.

Ugly:
- repeated depMgt in every submodule, as root POM cannot import bom-all (cycle)
  This could be avoided by pushing maven modules "level down", more restructure
  of project...

---

This DRAFT PR tries to address Konrad remarks, see https://issues.apache.org/jira/browse/MNG-6806?focusedCommentId=17499945&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-17499945  and below.

Basically, existing maven bom should be improved, get all versions under control.
This PR is just a POC, and is not quite good, as it according to current module layout, must add depMgt/import to each module....

OTOH, it revealed several issues with current setup: we have 2 test dependencies from maven root POM inherited (IMHO is wrong to inherit and hence force deps on all childs), also, some modules did redeclare those same deps, and finally, current root POM is huge and hardly overseen as it is crowded with many properties for example, that am unsure why we define in the first place (I usually define some property when it is repeating)

This kind of a layout since eons and works well
* makes module POMs simpler, as the only place where you set dep versions are BOMs, all the rest is just re-using those managed deps
* logically groups deps, I proposed one grouping, but we may end with other, this is just an example
* makes downstream users fully served, as they can choose which BOM to import, from bom-maven (equivalent of current maven-bom) to bom-all and have all the transitive deps managed


hboutemy:
what are you trying to solve, please?

kwin:
IMHO a BOM should only manage those dependencies which cannot be overridden through the plugin classloader as they are always determined from the Maven Core classloader. That is not the case for https://github.com/apache/maven/pull/689/files#diff-f1748463c332cad4071b21dbd4c295bea0422a9c9bc0dc2bd0547605a62d91d4 as none of the commons packages are exported in https://github.com/apache/maven/blob/master/maven-core/src/main/resources/META-INF/maven/extension.xml.
For me the BOM is only useful for Maven plugins, and should only manage those versions which are given by the Maven distribution.

rfscholte:
To me a BOM represents only the set of modules of a multimodule, that contain the  versions that work together. This should prevent issues with dependency depth/distance which could pull in an unexpected version.
Ideally all multimodules also provide a BOM.
Other dependencies don't belong here, as they should be upgradable/downgradable independent of the BOM.

kwin:
From a plugin developer's perspective it would be nice to have one(!) BOM available for the minimum target Maven version which includes all dependencies which cannot be overwritten as they are fully (or partially) exposed by the Core classloader.

The current proposal in https://github.com/apache/maven/blob/master/maven-bom/pom.xml doesn't contain all of those, but this proposal on the other hand, contains a lot more which can in fact be overridden by plugin developers, like https://github.com/apache/maven/pull/689/files#diff-f1748463c332cad4071b21dbd4c295bea0422a9c9bc0dc2bd0547605a62d91d4R51. So I would expect that the bom would would only manage those dependencies which are exported through https://github.com/apache/maven/blob/master/maven-core/src/main/resources/META-INF/maven/extension.xml.

rfscholte:
I guess the word BOM is causing confusion. What you're asking does not fit my definition of a BOM, as I described in a previous comment. 
All other dependencies shouldn't be an issue. e.g. Classworlds is able to let a plugin have a different version of plexus-utils than the one used by Maven itself.

kwin:
> Classworlds is able to let a plugin have a different version of plexus-utils than the one used by Maven itself.

IMHO not, you cannot use another version of it as it is part of the core classloader (at least not for the exported classes https://github.com/apache/maven/blob/631f1c76116d11ce2c6c46a801ef1fed08211908/maven-core/src/main/resources/META-INF/maven/extension.xml#L84-L87).
But my concern is not so much about the dependencies which can be overridden (to a great extent) but about the ones which can't like https://github.com/apache/maven/blob/631f1c76116d11ce2c6c46a801ef1fed08211908/maven-core/src/main/resources/META-INF/maven/extension.xml#L178 or https://github.com/apache/maven/blob/631f1c76116d11ce2c6c46a801ef1fed08211908/maven-core/src/main/resources/META-INF/maven/extension.xml#L60-L75

rfscholte:
Still that's not the purpose of a BOM. But it seems like we can't agree on that. I doubt this is a real issue to plugin developers, sounds like an edge case to me.
If you really want to solve this properly, you need to introduce new maven apis to hide this.

PR #692, opened 2022-03-14
[MNG-7391] Fix MojosExecutionStrategy lookup to be able to look into the SessionScope

gnodet:


laeubi:
This looks good even though I'm not that familiar with how providers work, will this lookup on each request (so you get the highest component freshyl on the call) or do it cache it once requested?

gnodet:
The component is cached by its scope.  When `m-build-cache-e` is not active, it's `@Singleton` and when it's active, it's the `@SessionScope`.  In both cases, that's fine, as the build cache extension does everything in the session scope anyway.

laeubi:
@gnodet @michael-o just wondering, would this maybe a good candidate for a maven-3.8.6 or are there no bug-fix releases for maven?

gnodet:
@laeubi the MojosExecutionStrategy has been introduced in 3.9.x and 4.0.x, but it's not present in 3.8.x branch.

laeubi:
Alright, I forget about this :-)

jira-importer:
Resolve #9173

PR #699, opened 2022-03-22
Refactoring

sun-qiwei:
some refactoring on maven-core components to increase the readability and reduce coupling as well as increase cohesion.


khmarbaise:
Can you please follow the contribution guide. Unfortunately I don't see any JIRA issue related to that change.

PR #706, opened 2022-04-02
[MNG-7432] Proposed UT for maven-3.9.x and 3.8.x

cstamas:
This UT makes sure that DefaultMaven sets up proper type
of WorkspaceReader in RepositorySystemSession.

For example, this UT fails with current maven-3.8.x and
maven-3.9.x branches due MNG-7432


laeubi:
@michael-o whats the process here, should the test be included (with `@Ignore`) and then the fix with enabling it?

kwin:
Just take over the UT to your branch and then this PR can be closed. That makes it easier to commit to 3.9/3.8.x.

cstamas:
Yes please, just lift it. Am currently a bit overwhelmed with $other work, otherwise would help with all of this. No need to retain author or anything (is few liner), just lift it/squash all right.

laeubi:
All right I'll try that! Thanks for the Test it looks quite straight forward but I'm not sure I would have been able to come up with such one :-)

jira-importer:
Resolve #9148

PR #712, opened 2022-04-08
[MNG-7456] Provide FileTransformer API replacement in Maven

cstamas:
To implement Consumer POM, as showcase, but this is all "maven specific", nothing to be done in resolver, so no resolver needed. Bonus, the deprecated FileTransformer is not even used anymore.

All the ITs locally pass with this build.

Effect can be seen in mvn DEBUG output (transformers are shown in-order as they are applied, order can be tuned with SISU priorities):

```
[DEBUG] Using manager EnhancedLocalRepositoryManager with priority 10.0 for /home/cstamas/.m2/repository
[DEBUG] Applying install request transformers (shown in order):
[DEBUG]  * class org.apache.maven.internal.aether.ConsumerPomTransformer
[DEBUG] 
[DEBUG] Applying deploy request transformers (shown in order):
[DEBUG]  * class org.apache.maven.internal.aether.ConsumerPomTransformer
[DEBUG] 
```

---

https://issues.apache.org/jira/browse/MNG-7456


maven-core/src/main/java/org/apache/maven/internal/aether/ConsumerPomArtifactTransformer.java
```diff
@@ -0,0 +1,94 @@
...
+    public Collection<Artifact> transformArtifact( Artifact artifact )
+    {
+        MavenSession mavenSession = mavenSessionProvider.get();
+        TransformerContext context = (TransformerContext) mavenSession
+                .getRepositorySession().getData().get( TransformerContext.KEY );
+        if ( "pom".equals( artifact.getExtension() ) && context != null )
+        {
+            // TODO: fix this, make it target rather
+            Path buildOutputDirectory = Paths.get(
+                    mavenSession.getCurrentProject().getBuild().getTestOutputDirectory() );
```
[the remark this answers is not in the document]
cstamas: is wip, nvm

cstamas:
@slawekjaranowski so, this is new proposal. The interesting thing is that now we can define a "chain" of transformers, separately, for install and for deploy. The order of transformers is currently defined by SISU only (so sisu component ordering applies, as to everything in Maven, see https://www.eclipse.org/lists/sisu-users/msg00144.html, we can later refine this if needed).

So, the point is, that now one can define for deploy only:
* consumer-pom - transforms installed/deployed pom to consumer pom
* signer - signs artifacts
* CI build provenance - attaches some extra descriptor artifact
* etc

slawekjaranowski:
Will be appreciate show example how plugin can add new transformers like signers. 
I use transformers in sign-m-p for build signatures.
https://github.com/s4u/sign-maven-plugin/blob/master/src/main/java/org/simplify4u/plugins/sign/ArtifactSigner40.java

maven-core/src/main/java/org/apache/maven/internal/aether/DefaultRepositorySystemSessionFactory.java
```diff
@@ -261,9 +261,37 @@ else if ( request.isUpdateSnapshots() )
...
+                logger.debug( "" );
+            }
+            session.getData().set(
+                    InstallRequestTransformer.KEY,
+                    new ChainedInstallRequestTransformer( installRequestTransformers )
+            );
+        }
+        if ( !deployRequestTransformers.isEmpty() )
         {
-            session.setFileTransformerManager( a -> getTransformersForArtifact( a, session.getData() ) );
```
slawekjaranowski, line 266: do we remove both `setFileTransformerManager` and `getFileTransformerManager` from `session`?
if not set what return get method?

cstamas:
@slawekjaranowski just FYI https://github.com/apache/maven-resolver/commit/1c79c4f8b6573e9693f4ae61c14bff17755aa32d

maven-core/src/main/java/org/apache/maven/internal/aether/DefaultRepositorySystemSessionFactory.java
cstamas: Do not use FileTransformer API anymore, in short. If this get's accepted, you will want to implement DeployRequestTransformer component... but let's go baby steps first, and make sure this would work for you too.

slawekjaranowski:
We will have `InstallRequestTransformer` and   `DeployRequestTransformer` so when I implement different component for it
will it possible that I will have different list of installed and deployed artifacts?

slawekjaranowski:
I see another issue `TransformerContext` is used by `ConsumerModelSourceTransformer` but leak to other Maven core components.

IMHO eache Transformer component implementation should be independent of internal core implementation.
But it can be other issue.

slawekjaranowski:
Another idea maybe we should have dedicated transformers for poms where result should be visible in reactor for other modules - m-shade-p case.
And another transformers for rest of artifacts - where result can be consumed once by install and deploy phase.
Now we do the same transformations twice.

cstamas:
> We will have `InstallRequestTransformer` and `DeployRequestTransformer` so when I implement different component for it will it possible that I will have different list of installed and deployed artifacts?

That's right and true even today, yes.

maven-core/src/main/java/org/apache/maven/artifact/transform/ConsumerPomTransformer.java
```diff
@@ -0,0 +1,140 @@
...
+                            ? Files.getLastModifiedTime( transformedPom ) : null;
+
+                    if ( !originalPomTs.equals( transformedPomTs ) )
+                    {
+                        // save it: either does not exist or TS differ
+                        Files.createDirectories( transformedPom.getParent() );
+                        try ( InputStream inputStream = new ConsumerModelSourceTransformer()
+                                .transform( originalPom, context ) )
+                        {
+                            Files.copy( inputStream, transformedPom );
```
rfscholte, line 117: One of my personal goals was preventing temporary files, as it opens a backdoor to rewrite the pom.
And I don't think there should be a need for it: passing the inputstream should be possible.

maven-core/src/main/java/org/apache/maven/artifact/transform/ConsumerPomTransformer.java
```diff
@@ -0,0 +1,140 @@
...
+import org.eclipse.aether.artifact.Artifact;
+import org.eclipse.aether.deployment.DeployRequest;
+import org.eclipse.aether.installation.InstallRequest;
+
+/**
+ * Consumer POM install and deploy transformer.
+ */
+@Named( ConsumerPomTransformer.NAME )
+@Singleton
+public class ConsumerPomTransformer implements InstallRequestTransformer, DeployRequestTransformer
```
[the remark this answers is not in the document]
cstamas: That's fine, Maven (and resolver) is already full of "internal" things, not MEANT to be extended by end users. We can make it clear here as well, but putting this code into some "internal" package instead (to not look like "API-ish thing".

maven-core/src/main/java/org/apache/maven/artifact/transform/ConsumerPomTransformer.java
cstamas: Am unsure what backdoor would be solved by not using temporary file. True, here am using component name, but the file name could be randomized instead, and effectively it would prevent any code to tamper with it.

Resolver is all about `java.io.Files`, so no, resolver does not allow passing over InputStream. Moreover, misuse of it's internal APIs (like it happened with FileTransformer), where `org.eclipse.aether.spi.connector.transport.PutTask#dataBytes` (meant and used for checksums only, by resolver itself) was used to transfer rewritten POM is just IMHO merely wrong: it misuses resolver API to prevent Maven API misuse....?

This whole "do not save the file" to me looks like some sort of "premature optimization" parallel: it is assumed, that not saving transformed POM prevents some sort of backdoor, that we are actually not even sure about what it is....  Let's solve problems when we actually hit them IMHO.

Finally, IMHO all this feature is really Maven concern, not Resolver concern. And true, maven-3.x does not have any "callback" to transform "just in time" when m-install-p or m-deploy-p calls (directly) into resolver, but IMHO this is much cleaner solution, and you are using proper APIs just to "swap out" (replace) the artifact about to be deploy to another one. Also, it works, as ITs are showing. Hence, I think this should be solved within realm of Maven, and not have Resolver involved into this at all.

rfscholte: Overall I agree and main thing is that the concept of build/consumer keeps working. 
However, I think it is worth reconsidering if resolver is really about Files and not InputStreams. With the architecture of Maven 2+3 it makes sense to use File as the base, but IMO with Maven 4+5 an InputStream would make more sense. And as a bonus I would expect less IO which should result in better performance as well.

cstamas: Not to forget upcoming Maven API, that also (may) solve this issue. Ultimate goal is to confine Resolver fully into Maven, and unlike today (just think about sonatype/eclipse package change) not have it littered into plugins, but offer API for all the use cases plugins would do, and make them "come back to core" instead, where all kind of things would become possible, and would be safe as well from future changes like possibly eclipse/apache package change of resolver for example... (not that it's about to happen tomorrow, but once in future will for sure).

gnodet (approved):
This looks much better than the other PRs previously raised to improve the situation.  I also tend to agree with @rfscholte that the maven-resolver should deal with `InputStream` rather than `File` if possible, but that's a different concern.
For the openness problem, if the signing plugin needs access to this layer, it has to be open, so hiding it in an _internal_ package does not really makes sense imho.

cstamas:
Superseded by https://github.com/apache/maven/pull/907

jira-importer:
Resolve #9296

PR #716, opened 2022-04-12
[MNG-7455] Another attempt to fix the concurrency issue

gnodet:
This is another attempt to fix the concurrency issue with minimal changes and without introducing back a thread local storage.
The downside of this proposal is that components annotated with `@SessionScoped` will always be injected the root session, which imho, is not really a problem as all derived sessions are cloned from the root one, except for the value of the `MavenSession#getCurrentProject()` which changes from one session to another.

The current scope is flawed anyway, as components are clearly not cached for all derived sessions, which was the original point of MNG-7347.

### DIFF
--- a/maven-core/src/main/java/org/apache/maven/SessionScoped.java
+++ b/maven-core/src/main/java/org/apache/maven/SessionScoped.java
@@ -31,6 +31,11 @@ import com.google.inject.ScopeAnnotation;
  * Indicates that annotated component should be instantiated before session execution starts
  * and discarded after session execution completes.
  *
+ * Note that components will be cached in the session scope and be injected with the root session.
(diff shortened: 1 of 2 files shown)


laeubi:
> is not really a problem as all derived sessions are cloned from the root one, except for the value of the `MavenSession#getCurrentProject()` which changes from one session to another.

I found especially this rather annoying, as one has to use the Legacy Support to get the "current" session.

gnodet:
> > is not really a problem as all derived sessions are cloned from the root one, except for the value of the `MavenSession#getCurrentProject()` which changes from one session to another.
> 
> I found especially this rather annoying, as one has to use the Legacy Support to get the "current" session.

I think https://github.com/apache/maven/pull/743 should solve MNG-7455 too.

PR #719, opened 2022-04-13
[MNG-7453] Upgrade Maven Resolver to 1.8.0 in maven-4.x

cstamas:
Note: current repository change in POM is ONLY to
make CI run ITs. This PR is placeholder until
maven-resolver 1.8.0 is released, but until then
let's see ITs at least.

---

https://issues.apache.org/jira/browse/MNG-7453

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -45,6 +45,19 @@ under the License.
   <url>https://maven.apache.org/ref/${project.version}/</url>
   <inceptionYear>2001</inceptionYear>
 
+  <repositories>
+    <repository>
+      <id>repository.apache.org</id>
+      <url>https://repository.apache.org/content/repositories/snapshots</url>
+      <releases>
+        <enabled>false</enabled>
+      </releases>
+      <snapshots>
+        <enabled>true</enabled>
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -45,6 +45,19 @@ under the License.
...
+      <id>repository.apache.org</id>
+      <url>https://repository.apache.org/content/repositories/snapshots</url>
+      <releases>
+        <enabled>false</enabled>
+      </releases>
+      <snapshots>
+        <enabled>true</enabled>
+      </snapshots>
+    </repository>
+  </repositories>
```
slawekjaranowski, line 59: There is in ASF pom ....
https://github.com/apache/maven-apache-parent/blob/master/pom.xml#L110-L129
Is it not the same?

cstamas: eh, missed it :) I assumed maven build does not allow snapshots for some reason (and CI builds w/ snapshots would fail)

cstamas: Now I remember what I mixed up: not snapshot artifacts, but **staged artifacts**....

PR #728, opened 2022-04-27
[MNG-7466] Align assembly XSD version with plugin used

cstamas:
As title says, XSD is ancient old 2.0.0, while all
latest assembly plugins uses 2.1.0.

Not that this matters or changes anything at all,
this is more about correctness.

---

https://issues.apache.org/jira/browse/MNG-7466


michael-o (changes requested):
Still has `http://maven.apache.org/ASSEMBLY/2.0.0`

cstamas:
> Still has `http://maven.apache.org/ASSEMBLY/2.0.0`

Fixed the missed ones.

michael-o (approved):

slachiewicz (approved):

PR #737, opened 2022-05-12
[3.9.x] [MNG-7477] Adapt the maven output to the width of the terminal

gnodet:
PR for testing.
A few things:
  * I had to redefine the `getTerminalWidth()` method to take into account the fact that the output can be redirected, in which case the previous `getTerminalWidth()` returns the width of the console using the err stream.
  *  the exact behavior when the output is redirected needs to be chosen: the behavior could depend if the output is actually padded or not, or the output could be padded only in the case the output is actually displayed on the console.  The behavior with this PR is to consider a default width of 80, thereby using the previous behavior
  *  the terminal width can be overridden by using the `maven.console.lineLength` system property

### DIFF
--- a/maven-core/src/main/java/org/apache/maven/internal/MessageHelper.java
+++ b/maven-core/src/main/java/org/apache/maven/internal/MessageHelper.java
@@ -1,0 +1,91 @@
+package org.apache.maven.internal;
+
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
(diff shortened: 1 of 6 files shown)


michael-o (commented):
Stupid question. Shouldn't that rather be part of shared which does a lot of Jansi handling?

mthmulders:
> Shouldn't that rather be part of shared which does a lot of Jansi handling?

I think that's how it's done in the `master` branch. Unless that solves a different problem than this PR does...?

gnodet:
> > Shouldn't that rather be part of shared which does a lot of Jansi handling?
> 
> I think that's how it's done in the `master` branch. Unless that solves a different problem than this PR does...?

Argh, I did not even checked if that was already done in master :-(

jira-importer:
Resolve #9127

PR #739, opened 2022-05-13
[MNG-7478] Transport selection should use config properties

cstamas:
Instead of user properties, as this allows making "permanent" the
selection by using MAVEN_OPTS and other places as well. Currently
only via command line works.

Also, do not modify existing Maven behavior, so introduce "default" branch
in selection. See comment in code.

---

https://issues.apache.org/jira/browse/MNG-7478


slawekjaranowski (approved):

slawekjaranowski:
Please look at jankis build ...

gnodet:
@cstamas does that mean you wouldn't be able to use `-Dxxx` anymore ? If so, should we support both ?

cstamas:
> @cstamas does that mean you wouldn't be able to use `-Dxxx` anymore ? If so, should we support both ?

Nope, you can still use -D at command line, but also you can use MAVEN_OPTS (not currently) etc

gnodet (approved):

PR #741, opened 2022-05-13
[MNG-7468] Check unsupported plugins parameters in configuration

slawekjaranowski:
After `MojoExecution` was created and configured all configuration will only contains parameters listed in plugin.xml
  
---

       pulling in other changes.

### DIFF
--- a/maven-core/src/main/java/org/apache/maven/lifecycle/internal/DefaultMojoExecutionConfigurator.java
+++ b/maven-core/src/main/java/org/apache/maven/lifecycle/internal/DefaultMojoExecutionConfigurator.java
@@ -20,6 +20,9 @@ package org.apache.maven.lifecycle.internal;
  */
 
 import java.util.Collection;
+import java.util.Set;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
 
 import javax.inject.Named;
 import javax.inject.Singleton;
@@ -28,9 +31,17 @@ import org.apache.maven.lifecycle.MojoExecutionConfigurator;
 import org.apache.maven.model.Plugin;
 import org.apache.maven.model.PluginExecution;
 import org.apache.maven.plugin.MojoExecution;
+import org.apache.maven.plugin.descriptor.MojoDescriptor;
+import org.apache.maven.plugin.descriptor.Parameter;
 import org.apache.maven.project.MavenProject;
+import org.apache.maven.shared.utils.logging.MessageBuilder;
+import org.apache.maven.shared.utils.logging.MessageUtils;
 import org.codehaus.plexus.util.StringUtils;
 import org.codehaus.plexus.util.xml.Xpp3Dom;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import static java.util.Arrays.stream;
 
 /**
  * @since 3.3.1, MNG-5753
@@ -40,6 +51,7 @@ import org.codehaus.plexus.util.xml.Xpp3Dom;
 public class DefaultMojoExecutionConfigurator
     implements MojoExecutionConfigurator
 {
+    private final Logger logger = LoggerFactory.getLogger( getClass() );
 
     @Override
     public void configure( MavenProject project, MojoExecution mojoExecution, boolean allowPluginLevelConfig )
@@ -76,6 +88,8 @@ public class DefaultMojoExecutionConfigurator
             mojoConfiguration = Xpp3Dom.mergeXpp3Dom( mojoExecution.getConfiguration(), mojoConfiguration );
 
             mojoExecution.setConfiguration( mojoConfiguration );
+
+            checkUnknownMojoConfigurationParameters( mojoExecution );
         }
     }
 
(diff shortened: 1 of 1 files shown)


maven-core/src/main/java/org/apache/maven/lifecycle/internal/DefaultLifecycleExecutionPlanCalculator.java
```diff
@@ -358,6 +368,61 @@ private void finalizeMojoConfiguration( MojoExecution mojoExecution )
...
+        throws PluginConfigurationException
+    {
+        if ( mojoExecution.getConfiguration() == null || mojoExecution.getConfiguration().getChildCount() == 0 )
+        {
+            return;
+        }
+
+        // first stem get parameter names of current goal
+        Set<String> parametersNames = new HashSet<>();
+        for ( Parameter p : mojoExecution.getMojoDescriptor().getParameters() )
```
MartinKanters, line 381: I unfortunately do not have the time to do a proper review, but I noticed some things at a glance:
- I believe this for loop and the one on line 404 is the same. Have you considered refactoring that to a method? 
- If you do the first point, then this method can all be done in a couple of streams instead of for-loops alternated with streams. Could be a small styling benefit. 
- typo in the comment on line 379 (step*) 
-

slawekjaranowski: Thanks for review. I will improvement code.
First I want to confirm that it is good place for such verification.

maven-core/src/main/java/org/apache/maven/lifecycle/internal/DefaultLifecycleExecutionPlanCalculator.java
```diff
@@ -358,6 +368,61 @@ private void finalizeMojoConfiguration( MojoExecution mojoExecution )
...
+
+        unknownParameters = stream( mojoExecution.getConfiguration().getChildren() )
+            .map( Xpp3Dom::getName )
+            .filter( name -> !parametersNames.contains( name ) )
+            .collect( Collectors.toSet() );
+
+        if ( !unknownParameters.isEmpty() )
+        {
+            String message = "Unknown plugin configuration parameters: " + unknownParameters;
+            throw new PluginConfigurationException( mojoExecution.getMojoDescriptor().getPluginDescriptor(), message );
```
michael-o, line 422: Important question. Consider that you have a plugin with n mojos, they share some parameters, but goals have their own. I if do not configure the plugin in general I cannot select the mojo the config should go. Will this fail on a per usecase basis?

cstamas:
-1 Maven should NEVER fail the build in this case... you simply don't know what Mojo does (it can inject MojoExecution and tamper directly with it's own config).

cstamas:
> @cstamas
> 
> * can plugin descriptor simply have empty list of parameters?
> * NEVER fail or don't check at all?
> * or check in DefaultMojoExecutionConfigurator?

* it can, if plugin does not declare any param?
* IMHO NEVER fail, as you have no idea what Mojo is up to, checking may just cause extra noise (IF Mojo does something funky)
* unsure what to say here 😄

slawekjaranowski:
- if plugin does not declare any param user should not provide any configuration - true?
- break build in 4.x according to issue comment - `It should issue a warning in 3.9.0 and break in 4.0.0.` by @michael-o 
- idea to check in `DefaultMojoExecutionConfigurator` is to avoid noise if author provide own configuration for plugin, so we will only check for `standard` configuration

cstamas:
> * if plugin does not declare any param user should not provide any configuration - true?
> * break build in 4.x according to issue comment - `It should issue a warning in 3.9.0 and break in 4.0.0.` by @michael-o
> * idea to check in `DefaultMojoExecutionConfigurator` is to avoid noise if author provide own configuration for plugin, so we will only check for `standard` configuration

* nope: plugin could get it from PlexusXmlConfig manually
* NOPE: NEVER break a build due this IMHO, not even in 4.x (unless we change how plugins are configured, but I doubt)
* ack

michael-o:
Picking up @cs comments: If we cannot tell for 100% that we can gather all information to break the build, then we cannot do it, but issue a warning at most.

slawekjaranowski:
IT tests https://github.com/apache/maven-integration-testing/pull/165

cstamas (approved):

slawekjaranowski:
@michael-o any comments
after merge I will cherry-pick to 3.9

michael-o:
Will check tomorrow, today is full

maven-core/src/main/java/org/apache/maven/lifecycle/internal/DefaultMojoExecutionConfigurator.java
```diff
@@ -40,6 +51,7 @@
 public class DefaultMojoExecutionConfigurator
     implements MojoExecutionConfigurator
 {
+    private Logger logger = LoggerFactory.getLogger( getClass() );
```
michael-o, line 54: Make it final, good habit.

maven-core/src/main/java/org/apache/maven/lifecycle/internal/DefaultMojoExecutionConfigurator.java
```diff
@@ -108,4 +122,75 @@ private PluginExecution findPluginExecution( String executionId, Collection<Plug
...
+        if ( unknownParameters.isEmpty() )
+        {
+            return;
+        }
+
+        // second step get parameter names of all plugin goals
+        Set<String> parametersNamesAll = mojoDescriptor.getPluginDescriptor().getMojos().stream()
+            .flatMap( m -> m.getParameters().stream() )
+            .flatMap( this::getParameterNames )
+            .collect( Collectors.toSet() );
```
michael-o, line 150: Why do you compare with all other goals too?

michael-o (changes requested):

maven-core/src/main/java/org/apache/maven/lifecycle/internal/DefaultMojoExecutionConfigurator.java
```diff
@@ -108,4 +122,75 @@ private PluginExecution findPluginExecution( String executionId, Collection<Plug
...
+
+        unknownParameters = getUnknownParameters( mojoExecution, parametersNamesAll );
+
+        unknownParameters.forEach(
+            name ->
+            {
+                MessageBuilder messageBuilder = MessageUtils.buffer()
+                    .warning( "Parameter '" )
+                    .warning( name )
+                    .warning( "' is unknown for plugin: '" )
```
michael-o, line 160: Single quote is unbalanced. Combining single quote and colon doesn't really make sense, no? I'd drop the colon and use single quotes.

maven-core/src/main/java/org/apache/maven/lifecycle/internal/DefaultMojoExecutionConfigurator.java
```diff
@@ -108,4 +122,75 @@ private PluginExecution findPluginExecution( String executionId, Collection<Plug
...
+                    .warning( mojoExecution.getArtifactId() ).warning( ":" )
+                    .warning( mojoExecution.getVersion() ).warning( ":" )
+                    .warning( mojoExecution.getGoal() );
```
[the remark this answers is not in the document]
slawekjaranowski: example message will be:

```
[WARNING] Parameter 'invalidValue' is unknown for plugin: maven-it-plugin-configuration:2.1-SNAPSHOT:touch (default)
```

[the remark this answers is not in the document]
michael-o: That is fine, just drop the colon after "plugin"

maven-core/src/main/java/org/apache/maven/lifecycle/internal/DefaultMojoExecutionConfigurator.java
slawekjaranowski: One plugin can have many goals with different parameters list.
This checking is done in context with specific goal, but configuration can have parameters for other goals.
I simply check if given parameters is valid for any goal in plugin.
Even parameters can be provided for goal which will be executed only be cli.

michael-o: I see, makes sense.

maven-core/src/main/java/org/apache/maven/lifecycle/internal/DefaultMojoExecutionConfigurator.java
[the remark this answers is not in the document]
slawekjaranowski: droped

maven-core/src/main/java/org/apache/maven/lifecycle/internal/DefaultMojoExecutionConfigurator.java
```diff
@@ -108,4 +122,77 @@ private PluginExecution findPluginExecution( String executionId, Collection<Plug
...
+
+        unknownParameters = getUnknownParameters( mojoExecution, parametersNamesAll );
+
+        unknownParameters.forEach(
+            name ->
+            {
+                MessageBuilder messageBuilder = MessageUtils.buffer()
+                    .warning( "Parameter '" )
+                    .warning( name )
+                    .warning( "' is unknown for plugin '" )
```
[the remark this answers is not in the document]
slawekjaranowski: It is for plugin - we check parameters for all goals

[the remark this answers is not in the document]
slawekjaranowski: https://github.com/apache/maven/pull/741#discussion_r882879288

michael-o (approved):

michael-o (approved):

jira-importer:
Resolve #8990

PR #744, opened 2022-05-17
[MNG-7360] Fix xml transformation to ensure proper context

gnodet:
As hinted by @rfscholte , the `FastForwardFilter` is supposed to handle those problems by disabling the filters when parsing  xml config nodes.  I've added a unit test and investigated in that direction, which leads to this new PR.  It looks much simpler than #697, so I think it's better to keep that one.

### DIFF
--- a/maven-model-transform/src/main/java/org/apache/maven/model/transform/pull/BufferingParser.java
+++ b/maven-model-transform/src/main/java/org/apache/maven/model/transform/pull/BufferingParser.java
@@ -41,7 +41,7 @@ public class BufferingParser implements XmlPullParser
     protected XmlPullParser xmlPullParser;
     protected Deque<Event> events;
     protected Event current;
-    protected boolean disabled;
+    protected boolean bypass;
 
     @SuppressWarnings( "checkstyle:VisibilityModifier" )
     public static class Event
@@ -404,7 +404,7 @@ public class BufferingParser implements XmlPullParser
                 throw new XmlPullParserException( "already reached end of XML input", this, null );
             }
             int currentEvent = xmlPullParser.next();
-            if ( disabled || accept() )
+            if ( bypass || accept() )
             {
                 return currentEvent;
             }
@@ -430,7 +430,7 @@ public class BufferingParser implements XmlPullParser
                 throw new XmlPullParserException( "already reached end of XML input", this, null );
             }
             int currentEvent = xmlPullParser.nextToken();
-            if ( disabled || accept() )
+            if ( bypass || accept() )
             {
                 return currentEvent;
             }
@@ -531,22 +531,13 @@ public class BufferingParser implements XmlPullParser
         return true;
     }
 
-    protected void enable()
+    public void bypass( boolean bypass )
     {
-        disabled = false;
-    }
-
-    protected void disable()
-    {
-        if ( events != null && !events.isEmpty() )
+        if ( bypass && events != null && !events.isEmpty() )
         {
             throw new IllegalStateException( "Can not disable filter while processing" );
         }
-        disabled = true;
-        if ( xmlPullParser instanceof BufferingParser )
-        {
-            ( ( BufferingParser ) xmlPullParser ).disable();
-        }
+        this.bypass = bypass;
     }
 
     protected static String nullSafeAppend( String originalValue, String charSegment )
--- a/maven-model-transform/src/main/java/org/apache/maven/model/transform/pull/BufferingParser.java
+++ b/maven-model-transform/src/main/java/org/apache/maven/model/transform/pull/BufferingParser.java
@@ -430,7 +430,7 @@ public class BufferingParser implements XmlPullParser
                 throw new XmlPullParserException( "already reached end of XML input", this, null );
             }
             int currentEvent = xmlPullParser.nextToken();
-            if ( accept() )
+            if ( disabled || accept() )
             {
                 return currentEvent;
             }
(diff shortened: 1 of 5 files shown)


maven-model-transform/src/main/java/org/apache/maven/model/transform/pull/BufferingParser.java
```diff
@@ -430,7 +430,7 @@ public int nextToken() throws XmlPullParserException, IOException
                 throw new XmlPullParserException( "already reached end of XML input", this, null );
             }
             int currentEvent = xmlPullParser.nextToken();
-            if ( accept() )
+            if ( disabled || accept() )
```
rfscholte, line 433: This looks like to conflicting states, which makes me wonder if we can always return currentEvent.
If current expression is required, we need to think of better naming.

gnodet: Maybe rename `disabled` to `bypass` ?
Or `accept()` to `canYieldCurrentEvent()` ?
Or both...

mthmulders: I like `bypass` - and I'd like it even more if the variable name would be in the same terminology as the class Javadoc. So maybe we should rephrase the Javadoc a bit and say something like "This filter will bypass all following filters and write directly to the output."

maven-model-transform/src/test/java/org/apache/maven/model/transform/ParentXMLFilterTest.java
```diff
@@ -33,30 +33,55 @@
...
 
-    protected ParentXMLFilter createFilter( XmlPullParser parser, Function<Path, Optional<RelativeProject>> pathMapper, Path projectPath ) {
+    protected XmlPullParser createFilter( XmlPullParser parser, Function<Path, Optional<RelativeProject>> pathMapper, Path projectPath ) {
         ParentXMLFilter filter = new ParentXMLFilter( parser, pathMapper, projectPath );
-        return filter;
+        return new FastForwardFilter( filter );
+    }
+
+    @Test
+    public void testWithFastForward()
```
mthmulders, line 63: Just to check: would it make sense to also add a test for the opposite case, i.e. where the ParentXMLFilter is _not_ supposed to bypass other filters?

cstamas:
Restarted GH ITs as I broke and @slawekjaranowski fixed ITs 😄

cstamas (approved):

mthmulders:
Although the changes look sane, I'm not sure we're on the right path. I've tried building this branch of Maven manually (succeeded) and then use that `mvn` to build Camel. Still fails:

```bash
mvn -V validate
Apache Maven 4.0.0-alpha-1-SNAPSHOT (c02b5f38d1dbc39db5b5360bed5c5807c3c6fa2a)
Maven home: /tmp/maven-gnodet/apache-maven-4.0.0-alpha-1-SNAPSHOT
Java version: 17.0.3, vendor: Eclipse Adoptium, runtime: /Library/Java/JavaVirtualMachines/temurin-17.jdk/Contents/Home
Default locale: en_US, platform encoding: UTF-8
OS name: "mac os x", version: "12.3.1", arch: "x86_64", family: "mac"
[INFO] BuildTimeEventSpy is registered.
[ERROR]
[ERROR]   The project (/Users/maarten/Code/open-source/camel/pom.xml) has 1 error
[ERROR]     Unable to transform pom: Not a regular file: /Users/maarten/Code/open-source/pom.xml
[ERROR]
[ERROR]
```

gnodet:
Thx for the check ! I'll work on an IT and a proper fix on monday.

gnodet:
IT added with https://github.com/apache/maven-integration-testing/pull/167

gnodet:
> Although the changes look sane, I'm not sure we're on the right path. I've tried building this branch of Maven manually (succeeded) and then use that `mvn` to build Camel. Still fails:
> 
> ```shell
> mvn -V validate
> Apache Maven 4.0.0-alpha-1-SNAPSHOT (c02b5f38d1dbc39db5b5360bed5c5807c3c6fa2a)
> Maven home: /tmp/maven-gnodet/apache-maven-4.0.0-alpha-1-SNAPSHOT
> Java version: 17.0.3, vendor: Eclipse Adoptium, runtime: /Library/Java/JavaVirtualMachines/temurin-17.jdk/Contents/Home
> Default locale: en_US, platform encoding: UTF-8
> OS name: "mac os x", version: "12.3.1", arch: "x86_64", family: "mac"
> [INFO] BuildTimeEventSpy is registered.
> [INFO] Scanning for projects...
> [ERROR] Some problems were encountered while processing the POMs
> [ERROR] The build could not read 1 project -> [Help 1]
> [ERROR]
> [ERROR]   The project (/Users/maarten/Code/open-source/camel/pom.xml) has 1 error
> [ERROR]     Unable to transform pom: Not a regular file: /Users/maarten/Code/open-source/pom.xml
> [ERROR]
> [ERROR] To see the full stack trace of the errors, re-run Maven with the '-e' switch
> [ERROR] Re-run Maven using the '-X' switch to enable verbose output
> [ERROR]
> [ERROR] For more information about the errors and possible solutions, please read the following articles:
> [ERROR] [Help 1] http://cwiki.apache.org/confluence/display/MAVEN/ProjectBuildingException
> ```
> 
> I've attached a debugger, but it seems the `FastForwardFilter` is never hit.

I've added an IT and fixed the behavior.

gnodet:
@mthmulders would you mind approving the PR for the record ?

mthmulders:
> @mthmulders would you mind approving the PR for the record ?

My pleasure!

mthmulders (approved):

PR #757, opened 2022-06-18
[MNG-7353] support CLI "mvn pluginPrefix:version:goal"

hboutemy:
https://issues.apache.org/jira/browse/MNG-7353

### DIFF
--- a/maven-core/src/main/java/org/apache/maven/lifecycle/internal/MojoDescriptorCreator.java
+++ b/maven-core/src/main/java/org/apache/maven/lifecycle/internal/MojoDescriptorCreator.java
@@ -176,18 +176,31 @@ public class MojoDescriptorCreator
         }
         else if ( numTokens == 3 )
         {
-            // We have everything that we need except the version
-            //
-            // org.apache.maven.plugins:maven-remote-resources-plugin:???:process
-            //
-            // groupId
-            // artifactId
-            // ???
-            // goal
-            //
-            plugin = new Plugin();
-            plugin.setGroupId( tok.nextToken() );
-            plugin.setArtifactId( tok.nextToken() );
+            // groupId:artifactId:goal or pluginPrefix:version:goal (since Maven 3.9.0)
+
+            String firstToken = tok.nextToken();
+            // groupId or pluginPrefix? heuristics: groupId contains dot (.) but not pluginPrefix
+            if ( firstToken.contains( "." ) )
+            {
+                // We have everything that we need except the version
+                //
+                // org.apache.maven.plugins:maven-remote-resources-plugin:???:process
+                //
+                // groupId
+                // artifactId
+                // ???
+                // goal
+                //
+                plugin = new Plugin();
+                plugin.setGroupId( firstToken );
+                plugin.setArtifactId( tok.nextToken() );
+            }
+            else
+            {
+                // pluginPrefix:version:goal, like remote-resources:3.5.0:process
+                plugin = findPluginForPrefix( firstToken, session );
+                plugin.setVersion( tok.nextToken() );
+            }
             goal = tok.nextToken();
         }
         else if ( numTokens <= 2 )
--- a/maven-core/src/main/java/org/apache/maven/lifecycle/internal/DefaultLifecycleTaskSegmentCalculator.java
+++ b/maven-core/src/main/java/org/apache/maven/lifecycle/internal/DefaultLifecycleTaskSegmentCalculator.java
@@ -96,7 +96,7 @@ public class DefaultLifecycleTaskSegmentCalculator
         {
             if ( isGoalSpecification( task ) )
             {
-                // "pluginPrefix:goal" or "groupId:artifactId[:version]:goal"
+                // "pluginPrefix[:version]:goal" or "groupId:artifactId[:version]:goal"
 
                 lifecyclePluginResolver.resolveMissingPluginVersions( session.getTopLevelProject(), session );
 


maven-core/src/main/java/org/apache/maven/lifecycle/internal/MojoDescriptorCreator.java
```diff
@@ -176,18 +176,31 @@ public MojoDescriptor getMojoDescriptor( String task, MavenSession session, Mave
...
-            // goal
-            //
-            plugin = new Plugin();
-            plugin.setGroupId( tok.nextToken() );
-            plugin.setArtifactId( tok.nextToken() );
+            // groupId:artifactId:goal or pluginPrefix:version:goal (since Maven 3.9)
+
+            String firstToken = tok.nextToken();
+            // groupId or pluginPrefix? heuristics: groupId contains . but not pluginPrefix
+            if ( firstToken.contains( "." ) )
```
michael-o, line 183: I don't think we can make this assumption for sure since for testing or other reasong the groupId could be: `foo`, `bar-baz`

mthmulders (commented):
I like this feature a lot. Would it be possible to add an IT for it?

maven-core/src/main/java/org/apache/maven/lifecycle/internal/MojoDescriptorCreator.java
hboutemy: if you want to test a plugin for such simplified run, just don't define such a groupId
at least, it seems to work for real world plugins, that's the first intent

of course, if you find a better heuristics, or even an algorithm, I'm all ears open

hboutemy:
IT added, that checks every CLI goal invocation format (with or without executionId)
please review

mthmulders:
Thanks for the IT. I've learned about `mvn <pluginPrefix>:<goal>@<id>` today :-)

Since there's no PR for the IT, let me ask one question here: I noticed the name of the test methods all start with `test`, but there's no `@Test` annotation. I thought we were on JUnit 4, with JUnit 5 [underway](https://github.com/apache/maven-integration-testing/pull/176). Wouldn't it make more sense to write the tests in JUnit 4 style? Also, if we were on JUnit 5 already, you could even turn it into a parameterised test, but I think that's not possible at the moment.

mthmulders:
> (notice: PR CI build does not seem to detect the associate core ITs branch...)

Looking at the [CI job for this PR](https://github.com/apache/maven/runs/7169900218?check_suite_focus=true), I see it detected the branch in m-i-t correctly and ran the new tests correctly:

image

michael-o:
Will test this week...

maven-core/src/main/java/org/apache/maven/lifecycle/internal/MojoDescriptorCreator.java
mthmulders: I gave it some thought, but I can't find a better heuristic either. The only thing more or less foolproof would be to look at the first token, assuming it is a plugin prefix. If that lookup yields nothing, it would be a groupId.

Even that is not 100% safe, though. Imagine you're not connected to the network, so lookup fails, so we decide it's a groupId -- this certainly doesn't make sense. But more importantly, it's probably quite costly compared to a `String.contains` :-).

hboutemy:
on Junit 4 vs 5: I'm not a purist, I just copied from others, it's so basic :)

maven-core/src/main/java/org/apache/maven/lifecycle/internal/MojoDescriptorCreator.java
```diff
@@ -176,18 +176,31 @@ public MojoDescriptor getMojoDescriptor( String task, MavenSession session, Mave
...
-            //
-            // groupId
-            // artifactId
-            // ???
-            // goal
-            //
-            plugin = new Plugin();
-            plugin.setGroupId( tok.nextToken() );
-            plugin.setArtifactId( tok.nextToken() );
+            // groupId:artifactId:goal or pluginPrefix:version:goal (since Maven 3.9)
```
michael-o, line 179: 3.9.0

maven-core/src/main/java/org/apache/maven/lifecycle/internal/MojoDescriptorCreator.java
```diff
@@ -176,18 +176,31 @@ public MojoDescriptor getMojoDescriptor( String task, MavenSession session, Mave
...
+
+            String firstToken = tok.nextToken();
+            // groupId or pluginPrefix? heuristics: groupId contains . but not pluginPrefix
```
michael-o, line 182: ...contains dot (.), but ...

michael-o (changes requested):

maven-core/src/main/java/org/apache/maven/lifecycle/internal/MojoDescriptorCreator.java
hboutemy: yes, but what?

maven-core/src/main/java/org/apache/maven/lifecycle/internal/MojoDescriptorCreator.java
hboutemy: and do you really think that we'll have any single word groupId in the future, given Maven Central explicitely asks for DNS matching?
heuristics is heuristics: in theory there could be edge cases, but in real world, it simply works

maven-core/src/main/java/org/apache/maven/lifecycle/internal/MojoDescriptorCreator.java
michael-o: It must be clear that this isn't a typo. That's why the symbol name and itself in parens.

hboutemy: I don't understand the sentence, then I can't transform it into any action, sorry

hboutemy: oh, I now understood...
this is the type of fixes where adding a commit is easier than commenting...

michael-o: Good

michael-o (approved):
All good. I have already rebased and cleaned up for you. You just need to merge! Allez !

maven-core/src/main/java/org/apache/maven/lifecycle/internal/MojoDescriptorCreator.java
kelunik: This indeed breaks real-world projects. We're using a single word groupId internally and invoking maven using `mvn groupId:artifactId:goal@executionId` no longer works starting with Maven 3.8.8.

PR #788, opened 2022-08-22
run clean before building maven and integration tests

hgschmie:
Explicitly run clean before (re-)building the maven distribution
and integration tests. Currently the tests seem to be a bit unstable
for 3.9.x and master; I can reproduce this locally and the errors
go away when explicitly cleaning the workspace before building.

This is a non-code, build only change.

### DIFF
--- a/.github/workflows/maven.yml
+++ b/.github/workflows/maven.yml
@@ -37,7 +37,7 @@ jobs:
           cache: 'maven'
 
       - name: Build with Maven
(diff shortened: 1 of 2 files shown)


slawekjaranowski (changes requested):
As I known each jobs running by GitHub is done on new fresh virtual machine, 
so `clean` is not what your are searching for.

PR #789, opened 2022-08-22
[MNG-7529] Maven resolver makes bad repository choices

hgschmie:
Testing integration tests

### DIFF
--- a/maven-resolver-provider/src/main/java/org/apache/maven/repository/internal/DefaultVersionRangeResolver.java
+++ b/maven-resolver-provider/src/main/java/org/apache/maven/repository/internal/DefaultVersionRangeResolver.java
@@ -69,6 +69,8 @@ public class DefaultVersionRangeResolver
 
     private static final String MAVEN_METADATA_XML = "maven-metadata.xml";
 
+    private static final String SNAPSHOT = "SNAPSHOT";
+
     private final MetadataResolver metadataResolver;
     private final SyncContextFactory syncContextFactory;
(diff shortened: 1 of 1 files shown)


hgschmie:
got the integration tests to work.

maven-resolver-provider/src/main/java/org/apache/maven/repository/internal/DefaultVersionRangeResolver.java
```diff
@@ -195,6 +199,18 @@ private Map<String, ArtifactRepository> getVersions( RepositorySystemSession ses
...
     }
 
+    private boolean isEnabled( RemoteRepository remoteRepository, String version )
+    {
+        if ( remoteRepository == null )
+        {
+            return true;
+        }
+
+        boolean snapshot = version != null && version.endsWith( SNAPSHOT );
```
cstamas, line 209: What about timestamped snapshots?

hgschmie: Hi @cstamas ,

Thank you for looking at this. I am not I understand this comment. The check is similar to the code in the DefaultVersionResolver (where it determines whether a local file is a SNAPSHOT version by checking that it ends with 'SNAPSHOT' (https://github.com/apache/maven/blob/master/maven-resolver-provider/src/main/java/org/apache/maven/repository/internal/DefaultVersionResolver.java#L208).

Can you explain a bit more what you mean by "timestamped snapshots" ?

PR #799, opened 2022-09-09
[MNG-7537] Point to the old IT branch when building 3.8.x

gnodet:


michael-o (approved):
Good catch...

slawekjaranowski:
We should do the same for Jenkins .... but it looks more complicated - now Jenkins always use master

slawekjaranowski (approved):

PR #811, opened 2022-10-05
[MNG-6801] MavenXpp3Writer doesn't retain order of properties

gnodet:
See https://issues.apache.org/jira/browse/MNG-6801


michael-o (approved):
LGTM, but commit message needs to change. I have modified in the ticket since this applied to properties and not the entire file:

    MavenXpp3Writer doesn't retain order of properties

jira-importer:
Resolve #8481

PR #813, opened 2022-10-05
Make sure RAT plugin ignores files in the target folder

gnodet:
When switching branches, left-over `target` folder *often* causes issues with the `rat-maven-plugin`.
Ignoring those folders does help from a day-to-day pov.

See #814 and #815

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,4 @@
-target/
+**/target/**
 .project
 .classpath
 .settings/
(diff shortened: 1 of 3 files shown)


gnodet:
@michael-o @cstamas I gave it another try
The rat plugin does read the `.gitignore` file to add excludes, but the syntax is slightly different in that it does not assume `**` before and after the pattern, so I fixed the `target/` into `**/target/**` so that it works for both.
If you add a `repo/` and use that directory as the local repo, it should be ignore too.
I've also disabled the execution in child modules.

PR #816, opened 2022-10-05
[MNG-7617] Small optimisations and cleanup in the project/model building

gnodet:
Extracted from #803

### DIFF
--- a/maven-model-builder/src/main/java/org/apache/maven/model/interpolation/BuildTimestampValueSource.java
+++ b/maven-model-builder/src/main/java/org/apache/maven/model/interpolation/BuildTimestampValueSource.java
@@ -23,17 +23,19 @@ import java.util.Map;
 import org.codehaus.plexus.interpolation.AbstractValueSource;
 
 class BuildTimestampValueSource extends AbstractValueSource {
-    private final MavenBuildTimestamp mavenBuildTimestamp;
+    private final Date startTime;
+    private final Map<String, String> properties;
 
     BuildTimestampValueSource(Date startTime, Map<String, String> properties) {
         super(false);
-        this.mavenBuildTimestamp = new MavenBuildTimestamp(startTime, properties);
+        this.startTime = startTime;
+        this.properties = properties;
     }
 
     @Override
     public Object getValue(String expression) {
         if ("build.timestamp".equals(expression) || "maven.build.timestamp".equals(expression)) {
-            return mavenBuildTimestamp.formattedTimestamp();
+            return new MavenBuildTimestamp(startTime, properties).formattedTimestamp();
         }
         return null;
     }
--- a/maven-model-builder/src/main/java/org/apache/maven/model/profile/DefaultProfileActivationContext.java
+++ b/maven-model-builder/src/main/java/org/apache/maven/model/profile/DefaultProfileActivationContext.java
@@ -18,7 +18,6 @@
  */
 package org.apache.maven.model.profile;
 
-import static java.util.stream.Collectors.collectingAndThen;
 import static java.util.stream.Collectors.toMap;
 
 import java.io.File;
@@ -26,6 +25,7 @@ import java.util.Collections;
 import java.util.List;
 import java.util.Map;
 import java.util.Properties;
+import java.util.stream.Collectors;
 
 /**
  * Describes the environmental context used to determine the activation status of profiles.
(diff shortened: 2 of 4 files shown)


maven-model-builder/src/main/java/org/apache/maven/model/interpolation/BuildTimestampValueSource.java
```diff
@@ -23,17 +23,19 @@
 import org.codehaus.plexus.interpolation.AbstractValueSource;
 
 class BuildTimestampValueSource extends AbstractValueSource {
-    private final MavenBuildTimestamp mavenBuildTimestamp;
+    private final Date startTime;
+    private final Map<String, String> properties;
 
     BuildTimestampValueSource(Date startTime, Map<String, String> properties) {
```
mthmulders, line 29: Now that we've switched to Java 8, would it make sense to refactor `Date` to a JSR-310 type?

maven-model-builder/src/main/java/org/apache/maven/model/profile/DefaultProfileActivationContext.java
```diff
@@ -194,15 +162,30 @@ public Map<String, String> getProjectProperties() {
...
-                            toMap(k -> String.valueOf(k.getKey()), v -> String.valueOf(v)),
-                            Collections::unmodifiableMap));
-        } else {
-            this.projectProperties = Collections.emptyMap();
-        }
+        return setProjectProperties(toMap(projectProperties));
+    }
+
+    public DefaultProfileActivationContext setProjectProperties(Map<String, String> projectProperties) {
+        this.projectProperties = unmodifiable(projectProperties);
```
mthmulders, line 169: Previously, a `null` for the `projectProperties` argument would result in `this.projectProperties` being an empty `Map`. With this patch, it would result in `this.projectProperties` being `null`. Are you sure that's an intended change?

maven-model-builder/src/main/java/org/apache/maven/model/interpolation/BuildTimestampValueSource.java
```diff
@@ -23,17 +23,19 @@
...
         super(false);
-        this.mavenBuildTimestamp = new MavenBuildTimestamp(startTime, properties);
+        this.startTime = startTime;
+        this.properties = properties;
```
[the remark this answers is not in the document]
gnodet: We could, but the only usage is when passed to the `MavenBuildTimestamp` which does support `null`.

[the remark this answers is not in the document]
mthmulders: All right, that makes sense.

maven-model-builder/src/main/java/org/apache/maven/model/interpolation/BuildTimestampValueSource.java
gnodet: The problem is that the only usage is when using the `ModelBuildingRequest.getBuildStartTime()`, which comes from `ProjectBuildingRequest.getBuildStartTime()` and `MavenExecutionRequest.getStartTime()`.  Without modifying those, which are public interfaces, there's no real value in changing this internal class.
Maybe when the core switches to the v4 api, as the start time is available as an `Instant` on the `Session`, then it will make sense to switch.  For now, it will only add `Date` -> `Instant` conversions for no real value imho.

maven-model-builder/src/main/java/org/apache/maven/model/profile/DefaultProfileActivationContext.java
gnodet: > Previously, a `null` for the `projectProperties` argument would result in `this.projectProperties` being an empty `Map`. With this patch, it would result in `this.projectProperties` being `null`. Are you sure that's an intended change?

I don't think that's true.  `toMap` can return `null`, but `unmodifiable(x)` returns an empty list/map, so `this.projectProperties` can never be `null` afaik.

maven-model-builder/src/main/java/org/apache/maven/model/interpolation/BuildTimestampValueSource.java
gnodet: Another possibility is to get rid of the `MavenBuildTimestamp` and simply inline the call inside `BuildTimestampValueSource.getValue()`

maven-model-builder/src/main/java/org/apache/maven/model/profile/DefaultProfileActivationContext.java
mthmulders: Ah, I see, we now have *two* methods named `setProjectProperties`. The `null` guarding is now split over those two methods, where it was previously in one method.

maven-model-builder/src/main/java/org/apache/maven/model/interpolation/BuildTimestampValueSource.java
mthmulders: AFAICS, the `MavenBuildTimestamp` class takes care of properly applying the `maven.build.timestamp.format` value, too. I think it's a bit too much to inline all of that here. Let's leave it as is.

mthmulders (approved):
Last question: would we need a JIRA ticket for this?

jira-importer:
Resolve #8836

PR #817, opened 2022-10-06
Simplify build / consumer

gnodet:
This commit optimizes and simplifies the build/consumer feature.

For the _build_ -> _raw_ transformation, a context containing all raw models from the reactor is needed. The relative path version mapping will infer the version of the parent while the dependency key version mapping will infer the version of a dependency, but in both cases, all the _file_ models from the reactor have to be parsed in order to actually know those versions. This actually forces a multi-step loading of the projects, first to read _file_ models, then then to re-read them with the known context to obtain _raw_ models.  This is ineffective and actually makes implementing the _build -> raw_ transformation quite complicated, using filters at the xml level. This commit simply changes the transformation from the xml level to the model level : the transformers now modify the parsed Model directly.

For the _raw -> consumer_ transformation, this commit currently keeps the transformation at the xml level, but it could be changed too.  In addition, the maven-resolver transformer api has been deprecated (https://github.com/apache/maven-resolver/commit/1c79c4f8b6573e9693f4ae61c14bff17755aa32d), so we'd better find another way to compute the consumer model.  This can be done using the v4 api inside the [`ArtifactInstaller`](https://github.com/apache/maven/blob/master/api/maven-api-core/src/main/java/org/apache/maven/api/services/ArtifactInstaller.java) and [`ArtifactDeployer`](https://github.com/apache/maven/blob/master/api/maven-api-core/src/main/java/org/apache/maven/api/services/ArtifactDeployer.java) services, but that would be limited to using the v4 version of the `maven-install-plugin` and `maven-deploy-plugin`.  At a small (compared to the deployment cost itself) performance cost, the transformation at the `Model` level could also be used which would further simplify the code.


gnodet:
The whole PR is based on a wrong assumption, so closing it.

PR #818, opened 2022-10-06
v4 api enhancements

gnodet:
- Fix DefaultArtifactManager to allow using both Artifact and Dependency
- Add a TODO for an EventSpy related api
- Add simple string/model conversion on the XmlFactory
- Decouple services creation

       pulling in other changes.

### DIFF
--- a/maven-core/src/main/java/org/apache/maven/internal/impl/DefaultArtifactInstaller.java
+++ b/maven-core/src/main/java/org/apache/maven/internal/impl/DefaultArtifactInstaller.java
@@ -21,6 +21,7 @@ package org.apache.maven.internal.impl;
 
 import javax.inject.Inject;
 import javax.inject.Named;
+import javax.inject.Singleton;
 
 import org.apache.maven.api.annotations.Nonnull;
 
@@ -42,6 +43,7 @@ import static org.apache.maven.internal.impl.Utils.cast;
 import static org.apache.maven.internal.impl.Utils.nonNull;
 
 @Named
+@Singleton
 public class DefaultArtifactInstaller implements ArtifactInstaller
 {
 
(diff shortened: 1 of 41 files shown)


maven-core/src/main/java/org/apache/maven/internal/impl/DefaultArtifactInstaller.java
```diff
@@ -38,11 +41,13 @@
 import static org.apache.maven.internal.impl.Utils.cast;
 import static org.apache.maven.internal.impl.Utils.nonNull;
 
+@Named
```
cstamas, line 44: and not singleton?

maven-core/src/main/java/org/apache/maven/internal/impl/DefaultArtifactResolver.java
```diff
@@ -43,6 +44,7 @@
 import static org.apache.maven.internal.impl.Utils.cast;
 import static org.apache.maven.internal.impl.Utils.nonNull;
 
+@Named
```
cstamas, line 47: singleton?

cstamas:
general remark: all those "named" classes are not singletons? is this on purpose?

gnodet:
> general remark: all those "named" classes are not singletons? is this on purpose?

They are stateless beans, so they could be moved to the `@Singleton` scope.  The question is: should they ?

cstamas:
> They are stateless beans, so they could be moved to the `@Singleton` scope. The question is: should they ?

In general, components should be stateless and immutable. By having them in singleton scope, in some circumstances you can preserve some memory and/or GC pressure, especially if instances are used across multiple components (is enlisted several times in object graph) or, (this is more like webapp scenario) are injected per request or so. 

Maven codebase in general applies this clearly and consistently, as IMHO the intent is clear when you see singleton annotation on a component... So my preference is yes, if a component is meant to be singleton, then they should be made singletons (as later on, as codebase evolve, this can lead to issues, IMHO be explicit and follow Maven "style" re components)

cstamas (approved):

PR #835, opened 2022-10-19
[MNG-6609] expose packaging via property "project.packaging"

kwin:
This can be used for packaging specific profiles with the property activator.

       pulling in other changes.

IT PR in https://github.com/apache/maven-integration-testing/pull/206

### DIFF
--- a/maven-model-builder/src/main/java/org/apache/maven/model/profile/ProfileActivationContext.java
+++ b/maven-model-builder/src/main/java/org/apache/maven/model/profile/ProfileActivationContext.java
@@ -30,6 +30,12 @@ import java.util.Map;
  */
 public interface ProfileActivationContext
 {
+    /**
+     * Key of the property containing the project's packaging.
+     * Available in {@link #getUserProperties()}.
+     * @since 4.0 
+     */
+    String PROPERTY_NAME_PACKAGING = "project.packaging";
 
     /**
      * Gets the identifiers of those profiles that should be activated by explicit demand.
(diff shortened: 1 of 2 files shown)


gnodet:
That really looks like a hack to avoid defining a new activation type to me.  Shouldn't we extend the model to add a packaging based activation ?

kwin:
Agreed, it is kind of a workaround, but given that I think it is consensus to stick with POM 4.0 for Maven 4 this is the only near term solution I could come up with.

maven-model-builder/src/main/java/org/apache/maven/model/profile/ProfileActivationContext.java
```diff
@@ -30,6 +30,12 @@
  */
 public interface ProfileActivationContext
 {
+    /**
+     * Key of the property containing the project's packaging.
+     * Available in {@link #getUserProperties()}.
+     * @since 4.0 
+     */
+    String PROPERTY_NAME_PACKAGING = "project.packaging";
```
gnodet, line 38: Shouldn't that be `packaging` to align with https://github.com/apache/maven/blob/2a9f39336cec1d8e52d30cc48503d51ed8672536/maven-model-builder/src/main/java/org/apache/maven/model/path/ProfileActivationFilePathInterpolator.java#L75-L82
I think the same reasoning apply, we don't support all `project.*` properties...

kwin: Good point, will change that

gnodet (approved):

jira-importer:
Resolve #8606

PR #838, opened 2022-10-20
Fix project versions not updated during release

nielsbasjes:
For an unknown reason the versions in the various modules were not updated correctly during the previous new snapshot version. 

This only fixes the current versions. It does not address the root cause at this point.

----

### DIFF
--- a/maven-model/pom.xml
+++ b/maven-model/pom.xml
@@ -37,16 +37,22 @@ under the License.
     <dependency>
       <groupId>org.apache.maven</groupId>
       <artifactId>maven-api-model</artifactId>
-      <version>4.0.0-alpha-3-SNAPSHOT</version>
+      <version>${project.version}</version>
     </dependency>
     <dependency>
       <groupId>org.apache.maven</groupId>
       <artifactId>maven-xml-impl</artifactId>
-      <version>4.0.0-alpha-3-SNAPSHOT</version>
+      <version>${project.version}</version>
     </dependency>
   </dependencies>
 
   <build>
+    <testResources>
+      <testResource>
+        <directory>src/test/resources</directory>
+        <filtering>true</filtering>
+      </testResource>
+    </testResources>
     <plugins>
       <plugin>
         <groupId>org.apache.maven.plugins</groupId>
@@ -64,7 +70,7 @@ under the License.
                 <artifactItem>
                   <groupId>org.apache.maven</groupId>
                   <artifactId>maven-api-model</artifactId>
-                  <version>4.0.0-alpha-1-SNAPSHOT</version>
+                  <version>${project.version}</version>
                   <type>mdo</type>
                   <outputDirectory>target/mdo/</outputDirectory>
                   <destFileName>maven.mdo</destFileName>
--- a/maven-model/src/test/resources/xml/pom.xml
+++ b/maven-model/src/test/resources/xml/pom.xml
@@ -25,7 +25,7 @@ under the License.
   <parent>
     <groupId>org.apache.maven</groupId>
     <artifactId>maven</artifactId>
-    <version>4.0.0-alpha-1-SNAPSHOT</version>
+    <version>dummy-version</version>
   </parent>
 
   <artifactId>maven-model</artifactId>
@@ -37,12 +37,12 @@ under the License.
     <dependency>
       <groupId>org.apache.maven</groupId>
       <artifactId>maven-api-model</artifactId>
-      <version>4.0.0-alpha-1-SNAPSHOT</version>
+      <version>${project.version}</version>
     </dependency>
     <dependency>
       <groupId>org.apache.maven</groupId>
       <artifactId>maven-xml-impl</artifactId>
-      <version>4.0.0-alpha-1-SNAPSHOT</version>
+      <version>${project.version}</version>
     </dependency>
   </dependencies>
 
@@ -64,7 +64,7 @@ under the License.
                 <artifactItem>
                   <groupId>org.apache.maven</groupId>
                   <artifactId>maven-api-model</artifactId>
-                  <version>4.0.0-alpha-1-SNAPSHOT</version>
+                  <version>${project.version}</version>
                   <type>mdo</type>
                   <outputDirectory>target/mdo/</outputDirectory>
                   <destFileName>maven.mdo</destFileName>
@@ -77,7 +77,7 @@ under the License.
       <plugin>
         <groupId>org.apache.maven</groupId>
         <artifactId>modello-plugin-velocity</artifactId>
-        <version>4.0.0-alpha-1-SNAPSHOT</version>
+        <version>${project.version}</version>
         <executions>
           <execution>
             <id>velocity</id>
--- a/maven-settings/pom.xml
+++ b/maven-settings/pom.xml
@@ -37,12 +37,12 @@ under the License.
     <dependency>
       <groupId>org.apache.maven</groupId>
       <artifactId>maven-api-settings</artifactId>
-      <version>4.0.0-alpha-3-SNAPSHOT</version>
+      <version>${project.version}</version>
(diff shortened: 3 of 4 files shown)


gnodet:
It's not only that, the `versions:set` goal did not update it during the release either:
  https://github.com/apache/maven/blob/maven-4.0.0-alpha-2/maven-model/pom.xml#L64-L71

maven-model/pom.xml
```diff
@@ -64,7 +64,7 @@ under the License.
                 <artifactItem>
                   <groupId>org.apache.maven</groupId>
                   <artifactId>maven-api-model</artifactId>
-                  <version>4.0.0-alpha-1-SNAPSHOT</version>
+                  <version>4.0.0-alpha-3-SNAPSHOT</version>
```
gnodet, line 67: Given the version is not changed during the release, this will break again.  Would you mind changing to `${project.version}` instead ?

maven-model/src/test/resources/xml/pom.xml
```diff
@@ -25,7 +25,7 @@ under the License.
   <parent>
     <groupId>org.apache.maven</groupId>
     <artifactId>maven</artifactId>
-    <version>4.0.0-alpha-1-SNAPSHOT</version>
+    <version>4.0.0-alpha-3-SNAPSHOT</version>
```
gnodet, line 28: This file is a test pom which is simply loaded, the version number is not important.  For clarity, if we really want to change it, maybe use something like `dummy-version` instead of something that can be misinterpreted as a real version.

maven-settings/pom.xml
```diff
@@ -68,7 +68,7 @@ under the License.
                 <artifactItem>
                   <groupId>org.apache.maven</groupId>
                   <artifactId>maven-api-settings</artifactId>
-                  <version>4.0.0-alpha-1-SNAPSHOT</version>
+                  <version>4.0.0-alpha-3-SNAPSHOT</version>
```
gnodet, line 71: Use `${project.version}`.

maven-toolchain-model/pom.xml
```diff
@@ -67,7 +67,7 @@ under the License.
                                 <artifactItem>
                                     <groupId>org.apache.maven</groupId>
                                     <artifactId>maven-api-toolchain</artifactId>
-                                    <version>4.0.0-alpha-1-SNAPSHOT</version>
+                                    <version>4.0.0-alpha-3-SNAPSHOT</version>
```
gnodet, line 70: Use `${project.version}`.

gnodet (changes requested):

maven-model/pom.xml
nielsbasjes: How about I change it everywhere where possible to `${project.version}`?

maven-model/src/test/resources/xml/pom.xml
nielsbasjes: Yes.
The tricky part here is that in this file all versions were missed.
I'll have to check if something as resource filtering can handle this file.

maven-model/pom.xml
nielsbasjes: I've set it to ${project.version} in all files I touched.

maven-model/src/test/resources/xml/pom.xml
nielsbasjes: Resource filtering works.
Even though the version is `dummy-version` the places where it says `${project.version}` now actually becomes the actual version of the project for testing purposes.

maven-model/pom.xml
```diff
@@ -37,16 +37,22 @@ under the License.
...
     </dependency>
   </dependencies>
 
   <build>
+    <testResources>
+      <testResource>
+        <directory>src/test/resources</directory>
+        <filtering>true</filtering>
+      </testResource>
+    </testResources>
```
gnodet, line 55: There's no real need for filtering resources here.  The exact content of the `xml/pom.xml` is really not important, the test just needs a model with a few things in it, but the values are not used.

maven-model/src/test/resources/xml/pom.xml
```diff
@@ -37,12 +37,12 @@ under the License.
     <dependency>
       <groupId>org.apache.maven</groupId>
       <artifactId>maven-api-model</artifactId>
-      <version>4.0.0-alpha-1-SNAPSHOT</version>
+      <version>${project.version}</version>
```
gnodet, line 40: I'd rather have all the versions in this file set to `dummy-version` with no filtering... see comment above.

maven-model/pom.xml
```diff
@@ -37,16 +37,22 @@ under the License.
     <dependency>
       <groupId>org.apache.maven</groupId>
       <artifactId>maven-api-model</artifactId>
-      <version>4.0.0-alpha-3-SNAPSHOT</version>
+      <version>${project.version}</version>
```
gnodet, line 40: Actually, this should be better added to dependency management of the parent pom along with all other modules defined in the build.

maven-model/pom.xml
```diff
@@ -37,16 +37,22 @@ under the License.
...
     </dependency>
     <dependency>
       <groupId>org.apache.maven</groupId>
       <artifactId>maven-xml-impl</artifactId>
-      <version>4.0.0-alpha-3-SNAPSHOT</version>
+      <version>${project.version}</version>
```
gnodet, line 45: Dependency management...

maven-settings/pom.xml
```diff
@@ -37,12 +37,12 @@ under the License.
     <dependency>
       <groupId>org.apache.maven</groupId>
       <artifactId>maven-api-settings</artifactId>
-      <version>4.0.0-alpha-3-SNAPSHOT</version>
+      <version>${project.version}</version>
```
gnodet, line 40: Dependency management...

maven-settings/pom.xml
```diff
@@ -37,12 +37,12 @@ under the License.
...
     </dependency>
     <dependency>
       <groupId>org.apache.maven</groupId>
       <artifactId>maven-xml-impl</artifactId>
-      <version>4.0.0-alpha-3-SNAPSHOT</version>
+      <version>${project.version}</version>
```
gnodet, line 45: Dependency management...

maven-toolchain-model/pom.xml
```diff
@@ -36,12 +36,12 @@ under the License.
         <dependency>
             <groupId>org.apache.maven</groupId>
             <artifactId>maven-api-toolchain</artifactId>
-            <version>4.0.0-alpha-3-SNAPSHOT</version>
+            <version>${project.version}</version>
```
gnodet, line 39: Dependency management...

maven-toolchain-model/pom.xml
```diff
@@ -36,12 +36,12 @@ under the License.
...
         </dependency>
         <dependency>
             <groupId>org.apache.maven</groupId>
             <artifactId>maven-xml-impl</artifactId>
-            <version>4.0.0-alpha-3-SNAPSHOT</version>
+            <version>${project.version}</version>
```
gnodet, line 44: Dependency management...

gnodet (changes requested):

maven-model/pom.xml
nielsbasjes: Will do. I'll get on it this weekend .

gnodet:
@nielsbasjes do you mind if I do the changes locally, that's really a nasty problem and should be fixed asap imho.

nielsbasjes:
Sure, go for it and I'll close this one.

gnodet:
Fixed by https://github.com/apache/maven/commit/0d70f7295de75fcd00d69789a19d2866348f97bb for the most important part.  A cleanup using the BOM / dependency management needs to follow.

PR #843, opened 2022-10-21
[MNG-7602] Move back model version from 5.0 to 4.2 and reactivate modello site generation

gnodet:


api/maven-api-model/src/main/mdo/maven.mdo
```diff
@@ -80,7 +80,7 @@
 
         <field xml.transient="true">
           <name>pomFile</name>
-          <version>5.0.0+</version>
+          <version>4.2.0</version>
```
cstamas, line 83: was "5.0.0+" and now is only "4.2.0", did you miss "+"?

api/maven-api-model/src/main/mdo/maven.mdo
```diff
@@ -434,7 +434,7 @@
           </code>
         </codeSegment>
         <codeSegment>
-          <version>5.0.0+</version>
+          <version>4.2.0</version>
```
cstamas, line 437: Same here, missing "+". Intentional?

api/maven-api-model/src/main/mdo/maven.mdo
```diff
@@ -2598,7 +2618,7 @@
...
```
cstamas, line 2621: Same, the missing plus...

maven-model/pom.xml
```diff
@@ -79,18 +79,38 @@ under the License.
...
           <execution>
-            <id>velocity</id>
+            <id>velocity-v3</id>
             <phase>generate-sources</phase>
             <goals>
               <goal>velocity</goal>
             </goals>
             <configuration>
-              <version>4.0.0</version>
+              <version>4.1.0</version>
```
cstamas, line 88: 4.1.0?

cstamas: Ah, two passed, one doing 4.1.0 and one 4.2.0? (below)

gnodet: Yes, 4.1.0 is the version used to generate the v3 model in maven4.  Version 4.2.0 is used for the v4 model.

kwin: I would still propose we use 4.0.1 and 4.0.2 to be more in line with semantic versioning as IMHO the API does not change at all!

api/maven-api-model/src/main/mdo/maven.mdo
gnodet: Yes, I fixed those.

api/maven-api-model/src/main/mdo/maven.mdo
gnodet: Fixed

maven-model/pom.xml
gnodet: > I would still propose we use 4.0.1 and 4.0.2 to be more in line with semantic versioning as IMHO the API does not change at all!

I kinda disagree.  If you use the 4.2.0 model removes some additional custom java code which makes the model generated by 4.1.0 and 4.2.0 already incompatible.  So the xml schema is compatible, but not the generated model.  I would think this warrant at least a minor increment.

kwin: I can follow this argument, but this will also change the generated XSD namespace which is IMHO not acceptable (no longer compatible with Maven Central)

gnodet: I don't get the argument.  The name is currently `http://maven.apache.org/POM/4.0.0`, so even a micro change could change that target namespace.  The only way is to actually force modello to generate the 4.0.0 version. I don't really see how that would change anything.

gnodet: See how the 4.0.0 version is forced for the schema generation:
  https://github.com/apache/maven/blob/bcbdd4d2724c36a7312697d90191905a9c498065/maven-model/pom.xml#L134

kwin (approved):
Sorry for the back and forth, looks good now, and I am fine with the versions 4.1/4.2 respectively.

gnodet:
> Sorry for the back and forth, looks good now, and I am fine with the versions 4.1/4.2 respectively.

Np, thx for reviewing !

PR #844, opened 2022-10-21
Move the modello velocity plugin out of the API

gnodet:
...  so that it won't be shown in the API documentation

https://maven.apache.org/ref/4-LATEST/maven-api/index.html

May be superseded by https://github.com/codehaus-plexus/modello/pull/269


gnodet:
Prefer https://github.com/apache/maven/pull/848 and https://github.com/codehaus-plexus/modello/pull/269 instead.

PR #846, opened 2022-10-23
change goal type to StringBuilder

zhangzq7:


michael-o (changes requested):
Please comply with our formatting rules. Checkstyle will tell you.

maven-core/src/main/java/org/apache/maven/lifecycle/internal/MojoDescriptorCreator.java
```diff
@@ -225,10 +225,10 @@ else if ( numTokens == 3 )
             plugin = findPluginForPrefix( prefix, session );
         }
 
-        int executionIdx = goal.indexOf( '@' );
+        int executionIdx = goal.toString().indexOf( '@' );
```
gnodet, line 228: maybe `goal.indexOf( "@" )` to avoid the string conversion ?

maven-core/src/main/java/org/apache/maven/lifecycle/internal/MojoDescriptorCreator.java
```diff
@@ -225,10 +225,10 @@ else if ( numTokens == 3 )
...
         if ( executionIdx > 0 )
         {
-            goal = goal.substring( 0, executionIdx );
+            goal = new StringBuilder(goal.substring(0, executionIdx));
```
gnodet, line 231: That looks like the kind of operation we'd like to avoid : a conversion to String, and a conversion back to a StringBuilder.  Use `goal.setLength( executionIdx )` ?
Not sure of the overall benefit if we manipulate `StringBuilder` to avoid a few string manipulation.   Maybe the use of the `StringBuilder` could be limited to refactoring the following code:
```
             goal = tok.nextToken();

             // This won't be valid, but it constructs something easy to read in the error message
             while ( tok.hasMoreTokens() )
             {
                 goal += ":" + tok.nextToken();
             }
```

gnodet (changes requested):

PR #855, opened 2022-10-27
[MNG-7583] Allow concurrent access to the MavenPluginManager

gnodet:
JIRA issue: https://issues.apache.org/jira/browse/MNG-7583

This would allow a better integration with mvnd.

At the moment, `mvnd` has to rewrite the `MavenPluginManager` to work around this synchronisation bottleneck.
See https://github.com/apache/maven-mvnd/blob/master/daemon/src/main/java/org/mvndaemon/mvnd/plugin/CliMavenPluginManager.java

Having the additional `get` methods on the cache would allow `mvnd` to simply provide a different cache, without having to rewrite the `DefaultMavenPluginManager`.

This will ease mvnd maintenance and maven upgrades a lot !


cstamas (approved):

jira-importer:
Resolve #8878

PR #856, opened 2022-10-27
[MNG-7584] The properties-maven-plugin fails to load project properties

gnodet:
JIRA issue: https://issues.apache.org/jira/browse/MNG-7584

### DIFF
--- a/maven-model/src/main/java/org/apache/maven/model/WrapperProperties.java
+++ b/maven-model/src/main/java/org/apache/maven/model/WrapperProperties.java
@@ -19,12 +19,13 @@ package org.apache.maven.model;
  * under the License.
  */
 
-import java.io.IOException;
 import java.io.InputStream;
-import java.io.OutputStream;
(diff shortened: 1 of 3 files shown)


slawekjaranowski:
Do we have three the same or similar classes?

gnodet:
> Do we have three the same or similar classes?

We have 3 copies of the `WrapperProperties`, `WrapperList`.  Those are non public classes used in the v3 models which wraps the new immutable v4 models.  I made copies in order to not add dependencies to some _utility_ module.
We also have 4 copies of `ImmutableCollections` (those are non public classes used in the new v4 models).
I've created https://issues.apache.org/jira/browse/MNG-7585 to try to find a way to get rid of those duplicates.

slawekjaranowski (approved):

cstamas (approved):

jira-importer:
Resolve #9284

PR #858, opened 2022-10-28
[MNG-6487] Add Sonatype dependency-check plugin to Parent POM

AbdelHajou:
JIRA issue: https://issues.apache.org/jira/browse/MNG-6487

This plugin checks dependencies for CVE vulnerabilities using Sonatype's vulnerability database. The build will fail when CVSS scores of >7.0 (HIGH) are found in any of the sub-modules. As discussed in MPOM-210, the OSS plugin is chosen in favour of OWASP Dependency-Check because the latter reports a lot of false positives and produces noise.

Only compile-time dependencies are included, because these are risky for Maven users and should be resolved before releasing.


jira-importer:
Resolve #7384

PR #859, opened 2022-10-28
[3.9.x] [MNG-7473] Backport Maven BOM to 3.9.0

cstamas:
---

https://issues.apache.org/jira/browse/MNG-7473

### DIFF
--- a/maven-bom/pom.xml
+++ b/maven-bom/pom.xml
@@ -1,0 +1,148 @@
+<?xml version="1.0" encoding="UTF-8"?>
+
+<!--
+Licensed to the Apache Software Foundation (ASF) under one
+or more contributor license agreements.  See the NOTICE file
+distributed with this work for additional information
+regarding copyright ownership.  The ASF licenses this file
+to you under the Apache License, Version 2.0 (the
+"License"); you may not use this file except in compliance
+with the License.  You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing,
+software distributed under the License is distributed on an
+"AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+KIND, either express or implied.  See the License for the
+specific language governing permissions and limitations
+under the License.
+-->
+
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 https://maven.apache.org/xsd/maven-4.0.0.xsd">
+  <modelVersion>4.0.0</modelVersion>
+  <parent>
+    <groupId>org.apache.maven</groupId>
+    <artifactId>maven-parent</artifactId>
+    <version>37</version>
+    <relativePath />
+  </parent>
+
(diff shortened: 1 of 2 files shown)


maven-bom/pom.xml
```diff
@@ -0,0 +1,148 @@
...
+    <site>
+      <id>apache.website</id>
+      <url>scm:svn:https://svn.apache.org/repos/asf/maven/website/components/${maven.site.path}</url>
+    </site>
+  </distributionManagement>
+
+  <dependencyManagement>
+    <!-- Not included:
+         - apache-maven, as it delivers the binaries
+         - maven-compat, only used by apache-maven distributions to support older releases
```
gnodet, line 67: I'm not sure it makes sense to not include `maven-compat`.  This is still required at runtime to run maven, so any plugin integration test will require it afaik.

maven-bom/pom.xml
```diff
@@ -0,0 +1,148 @@
...
+    <downloadUrl>https://maven.apache.org/download.html</downloadUrl>
+    <site>
+      <id>apache.website</id>
+      <url>scm:svn:https://svn.apache.org/repos/asf/maven/website/components/${maven.site.path}</url>
+    </site>
+  </distributionManagement>
+
+  <dependencyManagement>
+    <!-- Not included:
+         - apache-maven, as it delivers the binaries
```
gnodet, line 66: Can't the distribution be used for ITs ?
  https://github.com/apache/maven-build-cache-extension/blob/master/pom.xml#L401-L409
Unfortunately, I don't think the `m-dependency-p` can leverage the info from the BOM, but I wonder if there are other use cases ?

gnodet (changes requested):
Not sure about leaving out `apache-maven` and `maven-compat`...

gnodet (approved):
This is in line with current master, so out of topic for this PR.

slawekjaranowski (approved):

kwin:
What about the comments from https://issues.apache.org/jira/browse/MNG-6806?focusedCommentId=17532971&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-17532971.
This BOM is not really useful for plugin developers in its current form as putting the right version for those artifacts is a no-brainer. Can we have one which manages all artifacts exposed by the Core Classloader from the Maven distribution?

cstamas:
@gnodet I also think leaving out maven-compat makes no sense (at least for 3.x line)
@kwin I completely agree, but was down voted for doing that by arguments "it is not BOM"...

In essence, I wanted to backport what master/4.x has, would merely add maven-compat and later we can see for the rest...

kwin:
Maybe we can add another artifact then and call it maven-core-classloader-dep-mgmt. I don’t care about naming here, I primarily care about better support for plugin developers.

michael-o (approved):

cstamas:
Am somewhat aligned with @kwin here, and am now on the other side: backporting the BOM from master to 3.9.x (as is) does not makes much sense (plus as said above, where is compat, where is maven etc), so am rather to close this out w/o merge. Will give 24h for anyone interested to respond (or just +1) this comment as sign for "I agree: close out this PR w/o merge".

cstamas:
Am backing out from backporting BOM, it makes no sense as is IMHO

jira-importer:
Resolve #9128

PR #862, opened 2022-10-30
[MNG-7588] generate reader and some beans automatically for plugin

kwin:
descriptor

WIP (i.e. Core module is currently broken)


kwin:
@gnodet Do we need to keep the old beans in {{org.apache.maven.plugin.descriptor}} for backwards compatibility or is that not considered public API?

maven-plugin-api/pom.xml
```diff
@@ -112,6 +117,53 @@ under the License.
...
+              <goal>velocity</goal>
+            </goals>
+            <configuration>
+              <version>1.1.0</version>
+              <models>
+                <model>src/main/mdo/plugin.mdo</model>
+              </models>
+              <templates>
+                <template>src/main/mdo/model.vm</template>
+                <template>src/main/mdo/reader.vm</template>
```
kwin, line 133: should we generate a writer here or rather separately in m-plugin-tools?

gnodet:
> @gnodet Do we need to keep the old beans in `org.apache.maven.plugin.descriptor` for backwards compatibility or is that not considered public API?

We need to be very cautious for some time and avoid breaking compatibility as much as possible.  The new api is still in alpha and we need to upgrade most plugins before even deprecating all the old parts.

maven-plugin-api/pom.xml
```diff
@@ -112,6 +117,53 @@ under the License.
...
+            </goals>
+            <configuration>
+              <excludeDefaultDirectories>true</excludeDefaultDirectories>
+              <filesets>
+                <fileset>
+                  <directory>target/generated-sources/modello/org/apache/maven/plugin/descriptor</directory>
+                  <!-- remove those classes which need to ba maintained manually -->
+                  <includes>
+                    <include>MojoDescriptor.java</include>
+                    <include>PluginDescriptor.java</include>
```
gnodet, line 161: A better idea imho would be to tweak the `src/mdo/model.vm` to add a list of classes to not generate instead of generating them and removing them.  A variable similar to `packageModelV4` should work.

gnodet (changes requested):
The generated classes are incompatible with the existing ones, so I fear this is a big change.  The `maven-plugin-plugin` will probably be broken by such a change.

That's why for the v4 api, I've rather created entirely new packages so that we can phase the old ones out while keeping compatibility.

Maybe I'm wrong, but I think the fact that the `MojoDescriptor` and `PluginDescriptor` inherits the plexus `ComponentSetDescriptor` is only used internally in maven. If that's actually the case, I would favor trying to build a brand new and clean object model (without plexus) and only create the plexus required classes from those generated classes when needed.

kwin:
Ok, so when generating descriptor beans in a dedicated package, where should this live? IMHO it doesn't make sense to expose under https://github.com/apache/maven/tree/master/api/maven-api-core. This should rather be a private package (only for consumption of Maven Core and maybe Maven Plugin Tools). I don't think it would be useful to add this to the documented v4 API of Maven. So I am rather thinking about putting the new/clean/immutable plugin descriptor  classes into `org.apache.maven.plugin.descriptor.v11`. WDYT?

gnodet:
> Ok, so when generating descriptor beans in a dedicated package, where should this live? IMHO it doesn't make sense to expose under https://github.com/apache/maven/tree/master/api/maven-api-core. This should rather be a private package (only for consumption of Maven Core and maybe Maven Plugin Tools). I don't think it would be useful to add this to the documented v4 API of Maven. So I am rather thinking about putting the new/clean/immutable plugin descriptor classes into `org.apache.maven.plugin.descriptor.v11`. WDYT?

Sounds good to me.

kwin:
I would like to migrate all usage of Maven Core from the old `descriptor` to the new `immutabledescriptor` package. But the question is again, do we consider interfaces like https://github.com/apache/maven/blob/master/maven-core/src/main/java/org/apache/maven/plugin/BuildPluginManager.java API, i.e. do we need to evolve it in a backwards-compatible way? If yes, should we introduce new interfaces relying on the immutable descriptor or just add new methods to the existing interfaces?

jira-importer:
Resolve #8933

slawekjaranowski:
@kwin do you want plan to work on it?

kwin:
Not in the near future I fear

slachiewicz:
Closing this PR as superseded — thanks @kwin for the work and the design discussion!

**Rationale (based on the review history and current Maven 4 state):**

- As @gnodet noted in his review, the generated descriptor beans/readers are **incompatible with the existing `MojoDescriptor`/`PluginDescriptor` classes** and would break `maven-plugin-plugin` consumers.
- Maven 4 has since taken a different, phased approach: entirely **new packages in the v4 API** (`api/maven-api-*`) that replace the old descriptor model while keeping compatibility — which addresses MNG-7588's goal without breaking existing plugins.
- The PR has been inactive since 2022, and the author confirmed in April 2026 that there are no plans to work on it in the near future.

The underlying idea remains valid, so **[MNG-7588](https://github.com/apache/maven/issues/8933) stays open** as the tracking issue — if descriptor codegen is ever revived, it should start fresh against the Maven 4 API packages.

Thanks again for the contribution — closing does not diminish the value of the exploration!

PR #866, opened 2022-11-09
[MNG-7596] Upgrade to plexus 3.5.0

gnodet:
JIRA issue: https://issues.apache.org/jira/browse/MNG-7596
Upgrade to plexus 3.5.0 and change the xml merge accordingly

Note that I had to hack a bit the order to keep the current ordering, see https://issues.apache.org/jira/browse/MNG-5474

### DIFF
--- a/maven-xml-impl/src/main/java/org/apache/maven/internal/xml/Xpp3Dom.java
+++ b/maven-xml-impl/src/main/java/org/apache/maven/internal/xml/Xpp3Dom.java
@@ -29,6 +29,7 @@ import java.util.List;
 import java.util.ListIterator;
 import java.util.Map;
 import java.util.Objects;
+import java.util.Optional;
 import java.util.Set;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
@@ -180,8 +181,6 @@ public class Xpp3Dom implements Serializable, Dom {
      *   </ol></li>
      * <li> If mergeSelf == true
      *   <ol type="A">
-     *   <li> if the dominant root node's value is empty, set it to the recessive root node's value</li>
-     *   <li> For each attribute in the recessive root node which is not set in the dominant root node, set it.</li>
      *   <li> Determine whether children from the recessive DOM will be merged or appended to the dominant DOM as
      *        siblings (flag=mergeChildren).
      *     <ol type="i">
@@ -222,15 +221,10 @@ public class Xpp3Dom implements Serializable, Dom {
 
         if (mergeSelf) {
 
-            String value = null;
-            Object location = null;
+            String value = dominant.getValue();
+            Object location = dominant.getInputLocation();
             Map<String, String> attrs = null;
             List<Dom> children = null;
-
-            if (isEmpty(dominant.getValue()) && !isEmpty(recessive.getValue())) {
-                value = recessive.getValue();
-                location = recessive.getInputLocation();
-            }
 
             for (Map.Entry<String, String> attr : recessive.getAttributes().entrySet()) {
                 String key = attr.getKey();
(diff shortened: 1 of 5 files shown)


gnodet:
@kwin would you mind having a look at this PR ?

kwin (approved):
The double effort on maintenance is really nasty. We should indeed deprecate the original Xpp3Dom and Xpp3DomBuilder then.

cstamas (approved):

maven-xml-impl/src/main/java/org/apache/maven/internal/xml/Xpp3Dom.java
```diff
@@ -180,8 +181,6 @@ public void writeToSerializer(String namespace, XmlSerializer serializer) throws
      *   </ol></li>
      * <li> If mergeSelf == true
      *   <ol type="A">
-     *   <li> if the dominant root node's value is empty, set it to the recessive root node's value</li>
-     *   <li> For each attribute in the recessive root node which is not set in the dominant root node, set it.</li>
```
guylabs, line 184: @gnodet Why has this behavior been removed? Is there a migration path that you can share?

The following example has worked before that change where we have a parent pom and a child pom and a configuration that we'd like to merge. Here the parent and child pom definitions:

```xml
<!--parent pom-->
  ...
<pluginManagement>
  <plugins>
    <plugin>
      <groupId>foo.bar</groupId>
      <artifactId>foo-bar-plugin</artifactId>
      <configuration>
        <plugins>
          <plugin>
            <artifactId>maven-compiler-plugin</artifactId>
            <bar>
              <value>foo</value>
            </bar>
          </plugin>
          <plugin>
            <artifactId>maven-surefire-plugin</artifactId>
            <foo>
              <properties>
                <property>
                  <name>prop1</name>
                  <value>value1</value>
                </property>
              </properties>
            </foo>
          </plugin>
        </plugins>
      </configuration>
    </plugin>
  </plugins>
</pluginManagement>
  ...
  <!--child pom-->
  ...
<pluginManagement>
<plugins>
  <plugin>
    <groupId>foo.bar</groupId>
    <artifactId>foo-bar-plugin</artifactId>
    <configuration>
      <plugins>
        <plugin>
          <artifactId>maven-compiler-plugin</artifactId>
        </plugin>
        <plugin>
          <artifactId>maven-surefire-plugin</artifactId>
          <foo>
            <properties combine.children="append">
              <property>
                <name>prop2</name>
                <value>value2</value>
              </property>
            </properties>
          </foo>
        </plugin>
      </plugins>
    </configuration>
  </plugin>
</plugins>
</pluginManagement>
  ...
```

And the expected effective child pom that is used after the merge:

```xml
  <!--expected effective child pom-->
  ...
<pluginManagement>
<plugins>
  <plugin>
    <groupId>foo.bar</groupId>
    <artifactId>foo-bar-plugin</artifactId>
    <configuration>
      <plugins>
        <plugin>
          <artifactId>maven-compiler-plugin</artifactId>
          <bar>
            <value>foo</value>
          </bar>
        </plugin>
        <plugin>
          <artifactId>maven-surefire-plugin</artifactId>
          <foo>
            <properties combine.children="append">
              <property>
                <name>prop1</name>
                <value>value1</value>
              </property>
              <property>
                <name>prop2</name>
                <value>value2</value>
              </property>
            </properties>
          </foo>
        </plugin>
      </plugins>
    </configuration>
  </plugin>
</plugins>
</pluginManagement>
  ...
```

The problem we see now after this change, is that it will be merged the following way:

```xml
<!--expected effective child pom after this change-->
  ...
<pluginManagement>
<plugins>
  <plugin>
    <groupId>foo.bar</groupId>
    <artifactId>foo-bar-plugin</artifactId>
    <configuration>
      <plugins>
        <plugin>
          <artifactId>maven-compiler-plugin</artifactId>
          <foo>
            <properties>
              <property>
                <name>prop1</name>
                <value>value1</value>
              </property>
            </properties>
          </foo>
        </plugin>
        <plugin>
          <artifactId>maven-surefire-plugin</artifactId>
          <foo>
            <properties combine.children="append">
              <property>
                <name>prop2</name>
                <value>value2</value>
              </property>
            </properties>
          </foo>
        </plugin>
      </plugins>
    </configuration>
  </plugin>
</plugins>
</pluginManagement>
  ...
```

The problems are the following:

1. On the `maven-compiler-plugin` configuration the `<bar>` node is not merged from the parent pom configuration but rather the one from the `maven-surefire-plugin` configuration from the parent pom.
2. The `maven-surefire-plugin` configuration not as expected as the children of the `<properties>` node are not appended as configured via the `combine.children="append"` property.

Is that changed something that stays for Maven 4 ? I assume that then a migration path should be given as there are projects that use the behavior before this change and it would result in different effective pom files, which will change how the build is configured.

I'm also open to chat about it on the ASF Slack if required or if you have further questions.

hboutemy: I don't know if that can help, but big efforts were made in the past on maven-shared-utils reimplementation, with its test suite: https://github.com/apache/maven-shared-utils/blob/master/src/test/java/org/apache/maven/shared/utils/xml/pull/Xpp3DomTest.java
perhaps that can be reused

gnodet: @guylabs I'm not really the author of the change, I simply back ported it because the code is duplicated somehow from plexus-utils.  The original change is https://github.com/codehaus-plexus/plexus-utils/commit/67ac243dbc6434c03a9b1582d94aef6bf2b5cf95

jira-importer:
Resolve #8862

PR #868, opened 2022-11-10
[MNG-7542] Remove misleading warning about multi-module root directory

AzezSalami:
JIRA issue: https://issues.apache.org/jira/browse/MNG-7542

### DIFF
--- a/maven-core/src/main/java/org/apache/maven/project/collector/MultiModuleCollectionStrategy.java
+++ b/maven-core/src/main/java/org/apache/maven/project/collector/MultiModuleCollectionStrategy.java
@@ -41,6 +41,7 @@ import java.util.ArrayList;
 import java.util.Collections;
 import java.util.List;
 import java.util.Objects;
+import java.util.Optional;
 import java.util.function.Predicate;
 
 /**
@@ -64,11 +65,23 @@ public class MultiModuleCollectionStrategy implements ProjectCollectionStrategy
     @Override
     public List<MavenProject> collectProjects( MavenExecutionRequest request ) throws ProjectBuildingException
     {
-        File moduleProjectPomFile = getMultiModuleProjectPomFile( request );
-        List<File> files = Collections.singletonList( moduleProjectPomFile.getAbsoluteFile() );
+        Optional<File> moduleProjectPomFile = getMultiModuleProjectPomFile( request );
+        File absoluteFile = moduleProjectPomFile.orElse( request.getPom() ).getAbsoluteFile();
+        List<File> files = Collections.singletonList( absoluteFile );
         try
         {
             List<MavenProject> projects = projectsSelector.selectProjects( files, request );
+            if ( moduleProjectPomFile.isPresent()
+                    && projects.stream()
+                    .filter( MavenProject::hasParent )
+                    .noneMatch( p -> p.getParent().getBasedir() == null ) )
+            {
+                LOGGER.info( "Maven detected that the requested POM file is part of a multi-module project, "
+                                + "but could not find a pom.xml file in the multi-module root directory '{}'.",
+                        request.getMultiModuleProjectDirectory() );
+                LOGGER.info( "The reactor is limited to all projects under: " + request.getPom().getParent() );
+            }
+
             boolean isRequestedProjectCollected = isRequestedProjectCollected( request, projects );
             if ( isRequestedProjectCollected )
             {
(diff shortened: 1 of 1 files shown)


mthmulders:
Note to self: commit message could include

> Move the misleading logger from determining the multi-module project root POM to collecting the projects inside that directory.

maven-core/src/main/java/org/apache/maven/project/collector/MultiModuleCollectionStrategy.java
```diff
@@ -103,25 +118,20 @@ public List<MavenProject> collectProjects( MavenExecutionRequest request ) throw
         }
     }
 
-    private File getMultiModuleProjectPomFile( MavenExecutionRequest request )
+    private Optional<File> getMultiModuleProjectPomFile( MavenExecutionRequest request ) // return Optional<File>
```
mthmulders, line 121:
```suggestion
    private Optional<File> getMultiModuleProjectPomFile( MavenExecutionRequest request )
```

maven-core/src/main/java/org/apache/maven/project/collector/MultiModuleCollectionStrategy.java
```diff
@@ -64,23 +65,36 @@ public MultiModuleCollectionStrategy( ModelLocator modelLocator, ProjectsSelecto
...
     {
-        File moduleProjectPomFile = getMultiModuleProjectPomFile( request );
-        List<File> files = Collections.singletonList( moduleProjectPomFile.getAbsoluteFile() );
+        Optional<File> moduleProjectPomFile = getMultiModuleProjectPomFile( request );
+        File absoluteFile = moduleProjectPomFile.orElse( request.getPom() ).getAbsoluteFile();
+        List<File> files = Collections.singletonList( absoluteFile );
         try
         {
             List<MavenProject> projects = projectsSelector.selectProjects( files, request );
+            if ( moduleProjectPomFile.isPresent()
```
mthmulders, line 74: Are you sure you need to check for `moduleProjectPomFile.isPresent()`? Because previously, the warning would be emitted if the expected POM file did _not_ exist - in which case we now return `Optional.empty()`.

maven-core/src/main/java/org/apache/maven/project/collector/MultiModuleCollectionStrategy.java
```diff
@@ -64,23 +65,36 @@ public MultiModuleCollectionStrategy( ModelLocator modelLocator, ProjectsSelecto
...
+                    && projects.stream()
```
mthmulders, line 75: For better understanding of the `if` statement, I think it would help if the second part would be refactored to a separate boolean variable with a meaningful name.

maven-core/src/main/java/org/apache/maven/project/collector/MultiModuleCollectionStrategy.java
```diff
@@ -64,23 +65,36 @@ public MultiModuleCollectionStrategy( ModelLocator modelLocator, ProjectsSelecto
...
                 return projects;
             }
             else
             {
-                LOGGER.debug( "Multi module project collection failed:{}"
-                        + "Detected a POM file next to a .mvn directory in a parent directory ({}). "
-                        + "Maven assumed that POM file to be the parent of the requested project ({}), but it turned "
-                        + "out that it was not. Another project collection strategy will be executed as result.",
-                        System.lineSeparator(), moduleProjectPomFile.getAbsolutePath(),
+                LOGGER.debug(
```
mthmulders, line 92: It seems to me that these lines are unchanged apart from some whitespace. If that is the case, could you please revert the lines that aren't changed?

maven-core/src/main/java/org/apache/maven/project/collector/MultiModuleCollectionStrategy.java
```diff
@@ -91,7 +105,8 @@ public List<MavenProject> collectProjects( MavenExecutionRequest request ) throw
 
             if ( fallThrough )
             {
-                LOGGER.debug( "Multi module project collection failed:{}"
+                LOGGER.debug(
```
mthmulders, line 108: It seems to me that these lines are unchanged apart from some whitespace. If that is the case, could you please revert the lines that aren't changed?

mthmulders (commented):
Thanks for your contribution, @AzezSalami! I've identified a few minor things and left two questions. Would you mind looking into those?

gnodet (changes requested):
An IT would be welcomed so that we can reproduce the problem and ensure it's fixed.
The project that the JIRA points to seems to not produce such a warning.  My understanding is that the warning is caused by a mismatch between the project and the detected `maven.multiModuleProjectDirectory` property.
I've been able to reproduce the problem by adding a `.mvn` directory in the root of the project mentioned in the JIRA, but this looks like a wrong setup to me, as the `.mvn` directory does not point to the top of the reactor.

PR #872, opened 2022-11-15
[MNG-7597] Fix infinite loop when iterating PropertiesAsMap

gnodet:
JIRA issue: https://issues.apache.org/jira/browse/MNG-7597

### DIFF
--- a/maven-core/src/main/java/org/apache/maven/internal/impl/PropertiesAsMap.java
+++ b/maven-core/src/main/java/org/apache/maven/internal/impl/PropertiesAsMap.java
@@ -67,30 +67,30 @@ class PropertiesAsMap extends AbstractMap<String, String>
                     @Override
                     public Entry<String, String> next()
                     {
-                        if ( next == null )
+                        if ( !hasNext() )
                         {
                             throw new NoSuchElementException();
                         }
-                        Entry<Object, Object> next = this.next;
-                        this.next = null;
+                        Entry<Object, Object> ret = next;
+                        next = null;
                         return new Entry<String, String>()
                         {
                             @Override
                             public String getKey()
                             {
-                                return (String) next.getKey();
+                                return (String) ret.getKey();
                             }
 
                             @Override
                             public String getValue()
                             {
-                                return (String) next.getValue();
+                                return (String) ret.getValue();
                             }
 
                             @Override
                             public String setValue( String value )
                             {
-                                return (String) next.setValue( value );
+                                return (String) ret.setValue( value );
                             }
                         };
                     }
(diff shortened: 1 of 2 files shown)


maven-core/src/main/java/org/apache/maven/internal/impl/PropertiesAsMap.java
```diff
@@ -71,6 +71,8 @@ public Entry<String, String> next()
                         {
                             throw new NoSuchElementException();
                         }
+                        Entry<Object, Object> next = this.next;
```
slawekjaranowski, line 74: please use another name for local variable ... it is not clear which `next` is used

maven-core/src/test/java/org/apache/maven/internal/impl/PropertiesAsMapTest.java
```diff
@@ -0,0 +1,38 @@
...
+
+public class PropertiesAsMapTest
+{
+
+    @Test
+    public void testPropertiesAsMap()
+    {
+        Properties props = new Properties();
+        props.setProperty( "foo", "bar" );
+        Map<String, String> map = new HashMap<>( new PropertiesAsMap( props ) );
```
slawekjaranowski, line 36: please add assertions.

maybe also test for confirm, that PropertiesAsMap contains only has String as key and values

maven-core/src/main/java/org/apache/maven/internal/impl/PropertiesAsMap.java
gnodet: Fixed

maven-core/src/main/java/org/apache/maven/internal/impl/PropertiesAsMap.java
```diff
@@ -71,6 +71,8 @@ public Entry<String, String> next()
...
+                        this.next = null;
```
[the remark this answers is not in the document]
gnodet: Fixed

maven-core/src/test/java/org/apache/maven/internal/impl/PropertiesAsMapTest.java
gnodet: Fixed

maven-core/src/main/java/org/apache/maven/internal/impl/PropertiesAsMap.java
```diff
@@ -67,28 +67,30 @@ public boolean hasNext()
                     @Override
                     public Entry<String, String> next()
                     {
-                        if ( next == null )
+                        if ( !hasNext() )
```
kwin, line 70: although functionally correct I still find it weird to change the state in `hasNext()`. Why not put the logic to look for the next entry in `next` and simplify `hasNext()` with a simple check if `next == null` Also I find it very confusing that `next = null` when the end of the iterator has not yet been reached

gnodet: Fixed

kwin (approved):
Thanks, looks much cleaner now.

jira-importer:
Resolve #8858

PR #874, opened 2022-11-16
[MNG-7598] Enforce binary backwards-compatibility

kwin:
For now only enabled in core and plugin-api
Currently breaks build due to incompatible changes compared to 3.8.6


kwin:
@gnodet I fixed most of the issues in Maven Core. Can you look at the others and either fix or ignore with a comment?

gnodet:
I think it needs to be enabled on not only maven-core.  It's missing stuff like https://github.com/apache/maven/pull/841/files#diff-a660bdaa2e580478c65a557fd7001c80e129c0e9d4766840b4072c3b72c1863f

kwin:
This is just a start for sure to allow us to incrementally merge. Let us first fix Core and Plugin API and then follow-up with subsequent PRs for the other modules.

kwin:
@gnodet Any objection to merge this (after fixing the conflicts)?

maven-core/src/main/java/org/apache/maven/settings/SettingsUtils.java
```diff
@@ -51,9 +37,9 @@ private SettingsUtils()
      * @param recessive
      * @param recessiveSourceLevel
      */
-    public static Settings merge( Settings dominant, Settings recessive, String recessiveSourceLevel )
+    public static void merge( Settings dominant, Settings recessive, String recessiveSourceLevel )
     {
-        return new MavenSettingsMerger().merge( dominant, recessive, recessiveSourceLevel );
+        SettingsUtilsV4.merge( dominant.getDelegate(), recessive.getDelegate(), recessiveSourceLevel );
```
slawekjaranowski, line 42: `dominant` and `recessive` can be null in v3.

caller expect that `dominant` will have a result of merge , so we need

```
dominant.delegate = SettingsUtilsV4.merge( ...
```

gnodet: The method now reads:
```
     public static void merge(Settings dominant, Settings recessive, String recessiveSourceLevel) {
         if (dominant != null && recessive != null) {
             dominant.delegate = SettingsUtilsV4.merge(dominant.getDelegate(), recessive.getDelegate());
         }
     }
```

As for the source level, I think it would be better to leverage the InputLocationTracker in the model, which is way more powerful, for both settings and toolchains.  And get rid of the TrackableBase in the v4 models.

gnodet (approved):
LGTM, we need to continue the work on other modules in other PRs.
@slawekjaranowski can you double check the `SettingsUtils` work with the `m-invoker-p` plugin ?

slawekjaranowski (approved):
Works with m-invoker-p 👍

jira-importer:
Resolve #8827

PR #875, opened 2022-11-18
Chained LRM

cstamas:
Not for production, but usable in testing environments.
Depends on https://github.com/apache/maven-resolver/pull/223


slawekjaranowski:
Suppressed by: #889

PR #849, opened 2022-10-24
[MNG-6609] Maven 3.9: profile activation by packaging

kwin:
Following this checklist to help us incorporate your
contribution quickly and easily:

 - [x] Make sure there is a [JIRA issue](https://issues.apache.org/jira/browse/MNG) filed
       for the change (usually before you start working on it).  Trivial changes like typos do not
       require a JIRA issue. Your pull request should address just this issue, without
       pulling in other changes.
 - [x] Each commit in the pull request should have a meaningful subject line and body.
 - [x] Format the pull request title like `[MNG-XXX] SUMMARY`, where you replace `MNG-XXX`
       and `SUMMARY` with the appropriate JIRA issue. Best practice is to use the JIRA issue
       title in the pull request title and in the first line of the commit message.
 - [x] Write a pull request description that is detailed enough to understand what the pull request does, how, and why.
 - [x] Run `mvn clean verify` to make sure basic checks pass. A more thorough check will
       be performed on your pull request automatically.
 - [x] You have run the [Core IT][core-its] successfully.

If your pull request is about ~20 lines of code you don't need to sign an
[Individual Contributor License Agreement](https://www.apache.org/licenses/icla.pdf) if you are unsure
please ask on the developers list.

To make clear that you license your contribution under
the [Apache License Version 2.0, January 2004](http://www.apache.org/licenses/LICENSE-2.0)
you have to acknowledge this by using the following check-box.

 - [ ] I hereby declare this contribution to be licenced under the [Apache License Version 2.0, January 2004](http://www.apache.org/licenses/LICENSE-2.0)

 - [x] In any other case, please file an [Apache Individual Contributor License Agreement](https://www.apache.org/licenses/icla.pdf).

[core-its]: https://maven.apache.org/core-its/core-it-suite/

IT enabled in https://github.com/apache/maven-integration-testing/pull/208

PR #883, opened 2022-11-28
[MNG-6609] Profile activation based on packaging

cstamas:
In short: do NOT set request.setRawModel as NOTHING sets it. The Maven4 vs Maven3 is different, in Maven 3 NOTHING calls request.setRawModel

Full explanation: as ModelBuildingRequest is REUSED, and nothing sets this value in Maven3, once you set it here (as in original PR https://github.com/apache/maven/pull/849 ) results in awkward situation in Maven3: it will not load any other model...

---

https://issues.apache.org/jira/browse/MNG-6609

### DIFF
--- a/maven-model-builder/src/main/java/org/apache/maven/model/building/DefaultModelBuilder.java
+++ b/maven-model-builder/src/main/java/org/apache/maven/model/building/DefaultModelBuilder.java
@@ -276,8 +276,15 @@ public class DefaultModelBuilder
 
         DefaultModelProblemCollector problems = new DefaultModelProblemCollector( result );
 
+        // read and validate raw model
+        Model inputModel = request.getRawModel();
+        if ( inputModel == null )
+        {
+            inputModel = readModel( request.getModelSource(), request.getPomFile(), request, problems );
+        }
+
         // profile activation
-        DefaultProfileActivationContext profileActivationContext = getProfileActivationContext( request );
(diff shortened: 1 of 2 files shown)


kwin (approved):
LGTM, any reason why you just made it a draft?

cstamas:
Jut wanted to quickly "publish" my findings about the cause, but around same time moved to something else :smile: feel free to reuse (no need to keep my user ID, after all this is your change set with my one line removal). Or, I can pick it up later.

cstamas:
@kwin so does the new IT work? you mentioned something it does not fail with your commit?

kwin:
I could no longer bring it to fail with my original approach from https://github.com/apache/maven/pull/849 (i.e. it always succeeds)

cstamas:
IT fixed https://github.com/apache/maven-integration-testing/pull/214#pullrequestreview-1200730814

So am to merge this PR.

delanym:
@cstamas where is this documented? I don't see any change to https://maven.apache.org/guides/introduction/introduction-to-profiles.html

cstamas:
@delanym you mean the original issue? (as this PR was "just" a fire extinguish/fix for original one). Seems we missed it, @kwin ? Created https://issues.apache.org/jira/browse/MSITE-923

kwin:
Already tracked in https://issues.apache.org/jira/browse/MNGSITE-502, will prepare a PR for that.

jira-importer:
Resolve #8606

PR #884, opened 2022-11-28
[MNG-7607] Add M4 Transport API

cstamas:
Something simple to use and would reuse all the auth/proxy etc data from Maven. Intentionally super-trivial API.

If something more "serious" needed, plugin should probably roll it's own solution.

---

https://issues.apache.org/jira/browse/MNG-7607

### DIFF
--- a/api/maven-api-core/src/main/java/org/apache/maven/api/services/Transport.java
+++ b/api/maven-api-core/src/main/java/org/apache/maven/api/services/Transport.java
@@ -25,6 +25,7 @@ import java.nio.file.Path;
 import org.apache.maven.api.RemoteRepository;
 import org.apache.maven.api.annotations.Consumer;
 import org.apache.maven.api.annotations.Experimental;
+import org.apache.maven.api.annotations.Nonnull;
 import org.apache.maven.api.annotations.Nullable;
 
 /**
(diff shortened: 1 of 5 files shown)


api/maven-api-core/src/main/java/org/apache/maven/api/services/Transport.java
```diff
@@ -0,0 +1,50 @@
...
+import org.apache.maven.api.annotations.Consumer;
+import org.apache.maven.api.annotations.Experimental;
+
+/**
+ * @since 4.0
+ */
+@Experimental
+@Consumer
+public interface Transport
+{
```
gnodet, line 38: I would still add helpers for transferring the data as `byte[]` and `String`.  Even the `java.nio.Files` api now as those !

api/maven-api-core/src/main/java/org/apache/maven/api/services/Transport.java
```diff
@@ -0,0 +1,91 @@
...
+     */
+    void putBytes(byte[] source, URI relativeTarget);
+
+    /**
+     * PUTs the source string to target URI. The target MUST BE relative from the
+     * {@link RemoteRepository#getUrl()} root.
+     *
+     * @throws RuntimeException If PUT fails for any reason.
+     */
+    void putString(String source, Charset charset, URI relativeTarget);
```
gnodet, line 90: May I ask for a default:
```
default void putString(String source, URI relativeTarget) {
    putString(source, StandardCharsets.UTF_8, relativeTarget);
}
```

api/maven-api-core/src/main/java/org/apache/maven/api/services/Transport.java
```diff
@@ -0,0 +1,114 @@
...
+    String getString(@Nonnull URI relativeSource, @Nonnull Charset charset);
+
+    /**
+     * GETs the source URI content as string using UTF8 charset. The source MUST BE relative from the 
+     * {@link RemoteRepository#getUrl()} root.
+     *
+     * @return the string if operation succeeded, {@code null} if source does not exist.
+     * @throws RuntimeException If failed (and not due source not exists).
+     */
+    default String getString(@Nonnull URI relativeSource) {
```
gnodet, line 77: `@Nullable` ?

slawekjaranowski:
Great idea, I see that we can cover one case when user want to download a custom resource from remote repository.

We also have another case when user want to download a custom resource from any remote url. 
And also want to use authorization configuration from settings/servers and preserve settings/proxy for connections.

cstamas:
One big remark I just noticed: m4 API uses Optional quite extensively, while I went with my own preference. Better to not mix things, and rework this with Optional?

gnodet (approved):

cstamas:
> We also have another case when user want to download a custom resource from any remote url. And also want to use authorization configuration from settings/servers and preserve settings/proxy for connections.

@slawekjaranowski  But user can create own new instance of RemoteRepository (using whatever values/url he wants), and use that. Or is that not feasible?

jira-importer:
Resolve #8845

PR #885, opened 2022-11-28
[MNG-7608] Make native resolver transport the m4 default.

cstamas:

Altering the meaning of "default": is now same as NATIVE, still leaving it in place for future, as "default" at some point may again become something different than "native".

Depends on https://github.com/apache/maven-integration-testing/pull/215

---


cstamas:
Default changed (MavenITmng7470ResolverTransportTest.testResolverTransportDefault), and other 3 ITs expect Wagon to be used without overriding default, out of which 2 uses ancient way of setting UA header (using plexus configuration)

cstamas:
Superseded by https://github.com/apache/maven/pull/892

PR #886, opened 2022-11-29
[EXPERIMENT] Jetty transport

cstamas:
Add Jetty transport from resolver and make it selectable.

To build this Java17 is must (and raises Maven as well to 11).

Depends on https://github.com/apache/maven-resolver/pull/225

To build this (some dependencies brings in new license):
```
mvn clean install -Dremoteresources.skip
```

You will end up with HUGE maven 4 snapshot that embed Jetty Client.

---

To use it: use this settings: https://gist.github.com/cstamas/492747108c6112cedf887a5394e52eb0

```
~/tmp/apache-maven-4.0.0-alpha-3-SNAPSHOT/bin/mvn -V -s settings.xml clean install -Drat.skip -Dmaven.test.skip -Dmaven.resolver.transport=jetty
```
(rat.skip needed to leave alone local repo within build, tests skip to make it shorter, not wait for UT runs)

Transport values:
* maven.resolver.transport=jetty
* maven.resolver.transport=native
* maven.resolver.transport=wagon

settings.xml will use $CWD/local as local repo, nuke it before runs (to force all being re-downloaded).

### DIFF
--- a/apache-maven/pom.xml
+++ b/apache-maven/pom.xml
@@ -89,6 +89,10 @@ under the License.
       <artifactId>maven-resolver-transport-wagon</artifactId>
     </dependency>
     <dependency>
+      <groupId>org.apache.maven.resolver</groupId>
+      <artifactId>maven-resolver-transport-jetty</artifactId>
+    </dependency>
+    <dependency>
       <groupId>org.apache.maven</groupId>
       <artifactId>maven-slf4j-provider</artifactId>
     </dependency>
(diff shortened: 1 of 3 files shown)


cstamas:
This PR experiment is done, resolver has now JDK (added to Maven) and Jetty transport. Simplest way to use Jetty is to just define it as extension in `.mvn/extensions.xml` and it will prevail as it has highest priority. Is not included in Maven distro by default.

jira-importer:
Resolve #8447

PR #892, opened 2022-12-02
[MNG-7608] Make native transport the default

cstamas:

Altering the meaning of "default": is now same as "auto", but still leaving it in place for future, as "default" at some point may again become something different than "native".

---


gnodet (approved):

cstamas:
This PR needs more

kwin (approved):

rmannibucau (approved):

olamy:
before this it was possible for users to use their own wagon impls.
How is it possible now?

cstamas:
Nothing of legacy is lost, resolver transport (since 3.0) was always like this:
```
maven +-> transport-wagon +-> wagon-http
      |                   +-> wagon-foo
      |                   +-> ...
      +-> transport-http
      +-> transport-jetty (or whatever)
```

All what happened here is the "transport-wagon" got it's original priority for resolver applied (is demoted). Hence, to use custom wagon, all you need is to set transport-wagon, and from there, everything is "as before". This change merely exposed (and reordered) the first level of "transort-*" nodes...

jira-importer:
Resolve #8844

PR #893, opened 2022-12-02
[MNG-7615] Multithreaded model builder

gnodet:
Same PR as #803 but reformatted.

### DIFF
--- a/maven-core/src/main/java/org/apache/maven/internal/impl/DefaultProjectBuilder.java
+++ b/maven-core/src/main/java/org/apache/maven/internal/impl/DefaultProjectBuilder.java
@@ -150,10 +150,7 @@ public class DefaultProjectBuilder implements ProjectBuilder {
                         public String getLocation() {
                             StringBuilder buffer = new StringBuilder(256);
 
-                            if (getSource().length() > 0) {
-                                if (buffer.length() > 0) {
-                                    buffer.append(", ");
-                                }
+                            if (!getSource().isEmpty()) {
                                 buffer.append(getSource());
                             }
 
@@ -205,13 +202,6 @@ public class DefaultProjectBuilder implements ProjectBuilder {
                                 public Node getRoot() {
                                     return session.getNode(r.getDependencyGraph());
                                 }
-
-                                //                                @Override
-                                //                                public List<ArtifactResolverResult>
-                                // getArtifactResults()
-                                //                                {
-                                //                                    return Collections.emptyList();
-                                //                                }
                             });
                 }
             };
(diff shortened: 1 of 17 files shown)


maven-model-builder/src/main/java/org/apache/maven/model/superpom/DefaultSuperPomProvider.java
```diff
@@ -52,8 +53,8 @@ public DefaultSuperPomProvider(ModelProcessor modelProcessor) {
 
     @Override
     public Model getSuperModel(String version) {
-        if (superModel == null) {
-            String resource = "/org/apache/maven/model/pom-" + version + ".xml";
+        return SUPER_MODELS.computeIfAbsent(version, v -> {
```
cstamas, line 56: Does this mean that for example maven 4.0.1 could return super POM of Maven 3.8.6 (ie. newer could have all super POMs of older Mavens?)

cstamas:
Tried to test drive it, seems it deadlocked (console stopped at line 12):
https://gist.github.com/cstamas/d54715ffc24d7513e735353d17d8f78b

maven-model-builder/src/main/java/org/apache/maven/model/superpom/DefaultSuperPomProvider.java
gnodet: It could, though the only place it's called is with `4.0.0` afaik, and that's since ages.  I found the code does not reflect the javadoc and that the `version` parameter is not used, so I fixed it.
If that's not intended, then we should remove the parameter completely.
The version was originally hardcoded, see https://github.com/apache/maven/commit/dd635ca490bae271442d7b69eafdf595a627cafd
The commit added a version parameter, but it's only used the first time and subsequent calls return the cached version.

As for having multiple superpoms, I think the problem is that the version has never been updated when the pom changed.  So not sure what the future will be for the supermom...

cstamas:
Deadlock seems consistent, Ctrl-C after 6 minutes previous build, re-run and again:
https://gist.github.com/cstamas/62818af78abb293088cbb9383a278285

cstamas:
Those two runs above were Java 11. Switched to Java 17, first run was OK, but then it deadlocked again:
https://gist.github.com/cstamas/c44475bc3231363bc159d1a1d52f27c0

gnodet:
So the PR works with 1 one thread here, but may deadlock irrespective on the number of threads.
The deadlock happen in the following scenario : 3 projects P1, P2, P3, P3 depends on P2, P2 depends on P1.  The three projects are submitted to the FJ pool, which is able to steal work, so that jobs may be executed without any particular order.
So P2 is being executed, which requires P1. P1 is being joined by the FJ, but it decides to execute P3. P3 is being executed in the same thread and then waits for P2 being finished which can't happen.

cstamas (approved):

cstamas:
Ran ITs of this PR + resolver 2.0.0 (both combined) and ITs are OK

jira-importer:
Resolve #8837

PR #894, opened 2022-12-02
[MNG-7616] Fix missing exported artifacts

gnodet:

       pulling in other changes.

### DIFF
--- a/maven-core/src/main/resources/META-INF/maven/extension.xml
+++ b/maven-core/src/main/resources/META-INF/maven/extension.xml
@@ -136,6 +136,16 @@ under the License.
   </exportedPackages>
 
   <exportedArtifacts>
+    <!-- maven 4 api -->
+    <exportedArtifact>org.apache.maven:maven-api-core</exportedArtifact>
+    <exportedArtifact>org.apache.maven:maven-api-meta</exportedArtifact>
+    <exportedArtifact>org.apache.maven:maven-api-model</exportedArtifact>
+    <exportedArtifact>org.apache.maven:maven-api-settings</exportedArtifact>
+    <exportedArtifact>org.apache.maven:maven-api-toolchain</exportedArtifact>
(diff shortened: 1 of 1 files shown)


gnodet:
@cstamas I think a few resolver artifacts are missing, but I'm not even sure they are part of the default class loader.  Could you double check ?

maven-core/src/main/resources/META-INF/maven/extension.xml
```diff
@@ -136,6 +136,16 @@ under the License.
...
   <exportedArtifacts>
+    <!-- maven 4 api -->
+    <exportedArtifact>org.apache.maven:maven-api-core</exportedArtifact>
+    <exportedArtifact>org.apache.maven:maven-api-meta</exportedArtifact>
+    <exportedArtifact>org.apache.maven:maven-api-model</exportedArtifact>
+    <exportedArtifact>org.apache.maven:maven-api-settings</exportedArtifact>
+    <exportedArtifact>org.apache.maven:maven-api-toolchain</exportedArtifact>
+    <exportedArtifact>org.apache.maven:maven-api-xml</exportedArtifact>
+    <exportedArtifact>org.apache.maven:maven-xml-impl</exportedArtifact>
+    <exportedArtifact>org.apache.maven:plexus-utils</exportedArtifact>
```
kwin, line 147: I don't think this works, as only a handful of classes are actually exposed. For that reason the old plexus-utils was also no listed here.

gnodet: Good point. I've also removed the `maven-xml-impl` for the same reason.

kwin (approved):

cstamas:
> @cstamas I think a few resolver artifacts are missing, but I'm not even sure they are part of the default class loader. Could you double check ?

Nope, all good, connector and transport is not part of public api, they were never published

cstamas (approved):

PR #895, opened 2022-12-02
[MNG-7614] Translate Plexus XML to proper resolver configuration

cstamas:
Now that Maven 3.9.x contains other transports than Wagon, the Plexus XML is not just to be "stuffed", but translation should happen to proper resolver configuration to make them available to all transports.

---

https://issues.apache.org/jira/browse/MNG-7614


maven-core/src/main/java/org/apache/maven/internal/aether/DefaultRepositorySystemSessionFactory.java
```diff
@@ -209,6 +211,76 @@ public DefaultRepositorySystemSession newRepositorySession(MavenExecutionRequest
 
                 XmlPlexusConfiguration config = new XmlPlexusConfiguration(dom);
                 configProps.put("aether.connector.wagon.config." + server.getId(), config);
+
+                // Translate to proper resolver configuration properties as well (as Plexus XML above is Wagon specific
+                // only)
+                // but support only configuration/httpConfiguration/all, not the per-method nonsense
```
kwin, line 217: `not the per-method nonsense` :-) maybe just mention that only method `all` is supported.

maven-core/src/main/java/org/apache/maven/internal/aether/DefaultRepositorySystemSessionFactory.java
```diff
@@ -209,6 +211,76 @@ public DefaultRepositorySystemSession newRepositorySession(MavenExecutionRequest
...
+                    }
+                }
+
+                PlexusConfiguration connectTimeoutXml = config.getChild("connectTimeout", false);
+                if (connectTimeoutXml != null) {
+                    connectTimeout = Integer.parseInt(connectTimeoutXml.getValue());
+                } else {
+                    // fallback configuration name
+                    PlexusConfiguration httpConfiguration = config.getChild("httpConfiguration", false);
+                    if (httpConfiguration != null) {
```
kwin, line 242: maybe instead of the nested if, just `config.getChild("httpConfiguration", true).getChild("all", true).getChild("connectionTimeout", false)`

cstamas: that would mean am potentially mutating the XML (that is already passed to Wagon as Plexus config) -- so nope.

gnodet (approved):

jira-importer:
Resolve #8838

PR #937, opened 2022-12-26
[MNG-7618][MNG-7624] Backport to Maven 3.9.x

cstamas:
Lack of these changes are making IT suite to fail with maven-3.9.x branch.

OTOH, the change is nearly trivial, while value is high, so let's backport them.

---

https://issues.apache.org/jira/browse/MNG-7618
https://issues.apache.org/jira/browse/MNG-7624


michael-o (approved):
Now I know the reason :-D

PR #941, opened 2022-12-27
[MNG-7653] Integrate onSessionClose

cstamas:
Depends on https://github.com/apache/maven-resolver/pull/232

---

https://issues.apache.org/jira/browse/MNG-7653


cstamas:
Locally Maven IT suite passes OK with this change (w/ locally built related PR)

PR #948, opened 2023-01-04
[MNG-7662] Allow Graphbuilder to use session scoped components

laeubi:

 - [x] Make sure there is a [JIRA issue](https://issues.apache.org/jira/browse/MNG-7662) filed
       pulling in other changes.

### DIFF
--- a/maven-core/src/main/java/org/apache/maven/DefaultMaven.java
+++ b/maven-core/src/main/java/org/apache/maven/DefaultMaven.java
@@ -88,9 +88,6 @@ public class DefaultMaven implements Maven {
     @Requirement
     private DefaultRepositorySystemSessionFactory repositorySessionFactory;
 
-    @Requirement(hint = GraphBuilder.HINT)
-    private GraphBuilder graphBuilder;
-
     @Override
     public MavenExecutionResult execute(MavenExecutionRequest request) {
         MavenExecutionResult result;
@@ -449,7 +446,14 @@ public class DefaultMaven implements Maven {
     }
 
     private Result<? extends ProjectDependencyGraph> buildGraph(MavenSession session, MavenExecutionResult result) {
-        Result<? extends ProjectDependencyGraph> graphResult = graphBuilder.build(session);
+        Result<? extends ProjectDependencyGraph> graphResult;
+        try {
+            graphResult =
+                    container.lookup(GraphBuilder.class, GraphBuilder.HINT).build(session);
+        } catch (ComponentLookupException e) {
+            logger.error("Internal error GraphBuilder can not be fetched!", e);
+            return Result.error();
+        }
         for (ModelProblem problem : graphResult.getProblems()) {
             if (problem.getSeverity() == ModelProblem.Severity.WARNING) {
                 logger.warn(problem.toString());


laeubi:
@michael-o can you probably take a look?

cstamas:
Um, this looks wrong to me.... I guess this is to have injected some components that are session scoped? But why not use `Provider<X>` for components that are session scoped instead, as that will throw anyway if invoked out of session....

laeubi:
> Um, this looks wrong to me.... I guess this is to have injected some components that are session scoped?

Yes as described in the JIRA ticket

> But why not use `Provider<X>` for components that are session scoped instead, as that will throw anyway if invoked out of session....

Mostly because consumers should not need to know how the component is implemented, beside this I never got `Provider<X>` working (Maven 3.8.x) the build always complains something can not be injected. But maybe DefaultMavenComponent should all be `Provider<X>` ones so this do not conflict with implementation details?

gnodet:
Would it be possible to have scopes behave in a usual CDI way ? I.e. you should be able to inject a `@SessionScoped` bean into a `@Singleton` and the DI framework should wrap it in a proxy and fail _at use time_ if the bean is used outside of the session, but not at injection time.

laeubi:
@gnodet if you have an idea how to rewrite `DefaultMaven` that way it would be fine I think, I just don't know if it is possible as `DefaultMaven` is "special" in the way how it setup the session/scopes as part of the execute .... I think actually all references in `DefaultMaven` should probably be lazy until first access but wanted to keep this focused to the minimum...

gnodet:
> > Um, this looks wrong to me.... I guess this is to have injected some components that are session scoped?
> 
> Yes as described in the JIRA ticket
> 
> > But why not use `Provider<X>` for components that are session scoped instead, as that will throw anyway if invoked out of session....
> 
> Mostly because consumers should not need to know how the component is implemented, beside this I never got `Provider<X>` working (Maven 3.8.x) the build always complains something can not be injected. But maybe DefaultMavenComponent should all be `Provider<X>` ones so this do not conflict with implementation details?
> 
> I tried for example:
> 
> ```
> @Requirement( hint = GraphBuilder.HINT )
> private Provider<GraphBuilder> graphBuilder;
> ```
> 
> but the build fails with obscure errors...

Fwiw, I've just tested the injection of `javax.inject.Provider<GraphBuilder>` on master and it seems to work well...

laeubi:
> Fwiw, I've just tested the injection of `javax.inject.Provider<GraphBuilder>` on master and it seems to work well...

Do you like to submit a PR for this? This seems to be the easiest solution I think for now.... If it also works for 3.9.x I would be fine for this to be packported instead of my suggestion.

gnodet:
Also, there is an existing workaround which I think I've used somewhere, but I can't recall where exactly.
The idea is to define two beans:
```
@Named
@Singleton
class MyLazyGraphBuilder implements GraphBuilder {
    @Inject
    private final PlexusContainer container;
    private GraphBuilder delegate() {
       return container.lookup(GraphBuilder.class, "myGraphBuilder");
    }
    Result<? extends ProjectDependencyGraph> build(MavenSession session) {
        return delegate().build(session);
    }
}

@Named("myGraphBuilder")
@SessionScoped
class MyGraphBuilder implements GraphBuilder {
    ...
}
```

I still think this should be done automatically, but not sure I'll have the time to dig much into sisu/guice...

It may be preferable to use the workaround which will work on any maven version imho.

laeubi:
> Also, there is an existing workaround which I think I've used somewhere, but I can't recall where exactly.

This seems simlar to my aproach here, jsut that I more liek to fix it in maven itself than in my beans :-)

> I still think this should be done automatically, but not sure I'll have the time to dig much into sisu/guice...

maybe but I must confess that lazy lookup here seems easier ...

> It may be preferable to use the workaround which will work on any maven version imho.

I already have some workarounds, bt my goal i to get it fixed instead of running into the same issues again afterwards.

gnodet:
> > > Also, there is an existing workaround which I think I've used somewhere, but I can't recall where exactly.
> > 
> > 
> > This seems simlar to my aproach here, jsut that I more liek to fix it in maven itself than in my beans :-)
> > > I still think this should be done automatically, but not sure I'll have the time to dig much into sisu/guice...
> > 
> > 
> > maybe but I must confess that lazy lookup here seems easier ...
> > > It may be preferable to use the workaround which will work on any maven version imho.
> > 
> > 
> > I already have some workarounds, bt my goal i to get it fixed instead of running into the same issues again afterwards.
> 
> I think the [`SessionScoped`](https://github.com/apache/maven/blob/master/maven-core/src/main/java/org/apache/maven/session/scope/internal/SessionScope.java) class (and execution scope) could be a good candidate to implement transparent proxying.

See #950 @laeubi

gnodet (changes requested):
This looks like a workaround, but there's already one which can be implemented by doing the indirection of components manually in a custom component.  I'd rather have a global solution as the one outlined in #950

laeubi:
@gnodet I would be fine with a better solution, will #950 be included in Maven 3.9.0?

gnodet:
> @gnodet I would be fine with a better solution, will #950 be included in Maven 3.9.0?

If someone has the time to investigate and finalise the PR, I assume it could, though I'd rather target it for 4.x.  I'm not planning to work on it in the short term though.

gnodet:
Superseded by #950

PR #950, opened 2023-01-10
[MNG-7662] Use proxies for session scoped beans

gnodet:
JIRA issue: https://issues.apache.org/jira/browse/MNG-7662

The `SessionScope` will now create proxies to wrap beans when there's a need to inject a bean while the session scope is not yet available. Such proxies require the use of a `Typed` annotation, either the `org.eclipse.sisu.Typed` or `javax.enterprise.inject.Typed` annotation, to be put on the bean that requires to be wrapped by a proxy.
```
    @Named
    static class MySingletonBean {
        @Inject
        BeanItf myBean;
    }

    interface BeanItf {
        Session getSession();
    }

    @SessionScoped
    @Typed
    static class MySessionScopedBean implements BeanItf {
        @Inject
        Session session;

        public Session getSession() {
            return session;
        }
    }
```


maven-core/src/main/java/org/apache/maven/session/scope/internal/InstanceBuilder.java
```diff
@@ -0,0 +1,139 @@
...
+            final Object[] args = new Object[ctor.getParameterCount()];
+            return callConstructor(ctor, errors, args);
+        } catch (final NoSuchMethodException e) {
+            errors.addMessage(new Message(e.getMessage(), e));
+            return null;
+        }
+    }
+
+    private <T> Constructor<T> getConstructorFor(Class<T> proxyClass) throws NoSuchMethodException {
+        final Collection<Constructor<?>> ctors = Arrays.stream(proxyClass.getConstructors())
```
laeubi, line 113: Not sure if it is required, but from my testing I have the impression that one either need an `@Inject` annotated constructor or a default contructor without arguments.

maven-core/pom.xml
```diff
@@ -144,6 +144,12 @@ under the License.
       <groupId>org.slf4j</groupId>
       <artifactId>slf4j-api</artifactId>
     </dependency>
+    <dependency>
```
rmannibucau, line 147: since we already rely on build tools maybe let's just generate the proxy class at build time too and avoid to rely more on asm than the stack already does

maven-core/src/main/java/org/apache/maven/session/scope/internal/InstanceBuilder.java
gnodet: I ended up using objenesis in the latest commit to create the instance so that the requirements on the original class are lessened.

maven-core/pom.xml
gnodet: The choice to use a proxy is determined at runtime, depending if a bean to fulfil a given injection point is available or not.  This would end up generating proxy classes for all injection points I think (unless we try to use some heuristic), maybe a bit too much.

gnodet: Fwiw, in the distribution generated by this PR, I've made sure we use a byte-buddy version which does not embed ASM, so that we use a single version of it.

rmannibucau: well while asm is toggable it is fine (in several plugins) but if it becomes no more an option it also means a new release to use a new java version.
I'd like to avoid it.
Alternatively we can limit these scopes to interface injections where we don't need asm at all but ultimately we know upfront the types, the session scoped classes only so it is very generable at build time (like arc or openwebbeans do for ex).

gnodet: ASM has always been included in maven distribution.  It was previously embedded in sisu and guice.  We recently changed to use a standalone asm version.  This was done to support JDK 21 ;-)
At least, now, we're in control and we don't have one or more repackaged/shaded version of ASM in the distribution.
So byte-buddy would be third consumer of ASM, not the first...

rmannibucau: Well, this is not really the same, here you build an user facing feature (I'm thinking to extensions more than mojo) so it has to support this transparent java upgrade.
For "us" we don't care since our bytecode will always be consistent as you explained.
This is why I think we should generate the proxies at build time in out tooling stack more than at runtime - and I don't think doing it only for user code would be good.

this is my reasoning

gnodet: I'm not following.  What are you worried about exactly ? ASM is already used, so it's just another usage.  Related to class loaders, this PR does not change much, it simply adds byte buddy and objenesis to the maven core classloader, but ASM is already present, so there's actually no change related to ASM.
In the simplier form, I would have used a JDK `Proxy`, but this looks too limited, so I went directly with bytebuddy, but the effect is the same.  It's transparent for the user.
I really don't get your worry as the classloader is nearly unchanged and proxies are only generated when actually needed (i.e. with the current plugins, it should never happen).

gnodet: > the new interceptor is a new proxy basically high level i'd like to ensure we don't depend on java version as much as we can cause it starts to hurt all the ecosystem and at maven there is no reason we do depend on asm so no new usage should be added and ultimately we should drop them all from maven 4.0.0 IMHO

Due to the way scoping works in guice, the interception is built by wrapping the injected bean, not by wrapping the injection point.  The key method is the following:
https://google.github.io/guice/api-docs/4.2/javadoc/index.html?com/google/inject/Scope.html
```
  public <T> Provider<T> scope(Key<T> key, Provider<T> unscoped);
```
So we do wrap the `Provider<T>`, but `T` is the actual bean class, because that happens irrespective of where the bean will be injected (the injection point).
So even if the bean is injected into a field which refers to an interface the bean implementation provider will be _scope_.

Basically, we can't use JDK proxies here.

I do understand your concerns about avoiding the use of ASM.  We need to investigate, but it's actually used by sisu to discover beans.  That sounds duplicate work as we usually generate the META-INF files so that discovery should not have to scan all classes.
That said, problems happen only when the following combination happens:
 * the JDK in use is not supported by the ASM version embedded
 * we scan a class file which targets a JDK newer that the one supported by ASM
 That's why I've extracted ASM out of guice / sisu, so that we (or even the user) can easily upgrade the asm version in the distribution and get rid of those problems.

rmannibucau: Yes exactly, I think it is a case which becomes not rare to ignore.
About the interface thing it is mainly a matter of registering the scoped bean with an interface instead of its class so means `@SessionScoped(targetInterface = MyApi.class) class MyApiImpl implements Api {}` or something like that - once again we can relax it for internal usage but not for external ones - and `T` becomes `MyApi` instead of taking the actual instance.
Could it be a compromise to try to not shout in users foot?

gnodet: > Yes exactly, I think it is a case which becomes not rare to ignore. 

My point is that I mitigated the problem by externalising ASM.  Whenever a new ASM release comes, we can easily bump it and release maven.  But not sure it actually works, see the last point below.

> About the interface thing it is mainly a matter of registering the scoped bean with an interface instead of its class so means `@SessionScoped(targetInterface = MyApi.class) class MyApiImpl implements Api {}` or something like that - once again we can relax it for internal usage but not for external ones - and `T` becomes `MyApi` instead of taking the actual instance. Could it be a compromise to try to not shout in users foot?

I've reworked the PR in this way.  The injection is limited to beans which have a `Typed` annotation (sisu or inject) and the proxies now use JDK proxies.

Related to sisu, I investigated a bit, and ASM is not used to actually discover classes, that's correctly done with the generated META-INF files, but it's used to discover the injection points without having to load the classes, see [sisu SpaceScanner](https://github.com/eclipse/sisu.inject/blob/master/org.eclipse.sisu.inject/src/main/java/org/eclipse/sisu/space/SpaceScanner.java). 

Unfortunately, the sisu code does need [some small migration](https://github.com/eclipse/sisu.inject/commit/a9752361662edac045f16129a49e1a8b625aa604) when upgrading to a new ASM version to support reading new code...

Another point I just found is that the `java.lang.reflect.Proxy`... [is using ASM](https://github.com/openjdk/jdk/blob/dd9eab15c832c20e65681c21c5f91df11f4cddf9/src/java.base/share/classes/java/lang/reflect/ProxyGenerator.java#L56) to create the proxies !  So not sure we can really do much here...  However, I assume the JDK uses an ASM version which support the JDK itself somehow.

rmannibucau: yes the JVM embed a relocated asm version which always works with their version, this is why I want to be able to guarantee your extension developped with jvm N+3 when maven was released with last jvm N works.

your change is very appreciated and ok for me, thanks!

rmannibucau (approved):

cstamas:
Why is `@Typed` required? It has a serious side effect in Sisu regarding how bean is bound...

gnodet:
> Why is `@Typed` required? It has a serious side effect in Sisu regarding how bean is bound...

Oh, yes.  That's actually the semantics that we need.  The goal is to make sure the proxied bean is exposed only with interfaces.  So we need `@Typed` in order to make sure it will be injected properly.  
I experimented with adding an interface list to `@SessionScoped`, but then, if the injection point was using the actual implementation class instead of an interface, the injection would fail without a proper explanation (just that setting the field did not work, but not with a ClassCastException or anything that could give a hint).

cstamas:
Ah, true, yes, proxies are ONLY about interfaces.... kk

cstamas (approved):

PR #951, opened 2023-01-11
Jenkinsfile: pickup results from ITs

psiroky:
* trying to make Jenkins pick-up the IT results - currently, if some of the ITs fail, Jenkins won't pick-up the results and the build would be "green"
* not 100% sure this is the fix though, I have no way of testing this locally so will rely on the PR builds
* missing IT results can be seen for example here: https://ci-maven.apache.org/job/Maven/job/maven-box/job/maven/job/master/190/testReport/

       pulling in other changes.
 - [] Run `mvn clean verify` to make sure basic checks pass. A more thorough check will
 - [] You have run the [Core IT][core-its] successfully.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -56,7 +56,7 @@ node(jenkinsEnv.nodeSelection(osNode)) {
                     sh "mvn clean ${MAVEN_GOAL} -B -U -e -fae -V -Dmaven.test.failure.ignore -PversionlessMavenDist -Dmaven.repo.local=${WORK_DIR}/.repository"
                 }
             } finally {
-                junit testResults: '**/target/surefire-reports/*.xml', allowEmptyResults: true                
+                junit testResults: '**/target/surefire-reports/*.xml,**/target/failsafe-reports/*.xml', allowEmptyResults: true
             }    
             dir ('apache-maven/target') {
                 stash includes: 'apache-maven-bin.zip', name: 'maven-dist'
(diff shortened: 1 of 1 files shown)


Jenkinsfile
```diff
@@ -112,7 +112,7 @@ for (String os in runITsOses) {
                                 }
                             }
                         } finally {
-                            //junit testResults: '**/core-it-suite/**/target/surefire-reports/*.xml', allowEmptyResults: true                                            
+                            junit testResults: '**/core-it-suite/**/target/surefire-reports/*.xml', allowEmptyResults: true
```
slawekjaranowski, line 115: What with standard path, like `**target/surefire-reports/*.xml` and for failsafe?

psiroky: I am not sure. I assumed these patterns were tested before and it the junit step was accidentally left commented out.

I think we could go with `**target/surefire-reports/*.xml`. It will pick-up some test results from `core-it-support`, but that may be desirable.

There are not `failsafe-reports` directories, but to be safe we could include those as well.

psiroky:
I updated the PR to also pickup `failsafe-reports`. I haven't really tried locally, but according to https://www.jenkins.io/doc/pipeline/steps/junit/ separating the patterns with comma should work.

psiroky:
I noticed there is also [Jenkinsfile.its](https://github.com/apache/maven/blob/master/Jenkinsfile.its) which seems unused? Should it be just removed, or am I missing the usage somewhere?

slawekjaranowski:
@psiroky Thanks for PR, I will try it on branch in maven repo ... PR from fork are not build by jenkins

slawekjaranowski:
Suppressed by #953

PR #961, opened 2023-01-17
[3.9.x][MNG-7608] Make native transport the default

cstamas:

Altering the meaning of "default": is now semantically same as "auto", but still leaving it in place for future, as "default" at some point may again become something different than "native". This change is exactly same as happened in Maven 4.x and transports are aligned now.

---


cstamas:
This change breaks one IT adjusted in https://github.com/apache/maven-integration-testing/pull/237

michael-o (approved):
No objection here and progress output looks just like before, but with the release notes there must a s pecial note that the transport has changed and people must know exactly about it.

PR #963, opened 2023-01-17
[MNG-7646] Do not parse all projects in the reactor when building a subtree

gnodet:
Follow up to MNG-7629 and #954, the idea for this PR is to not parse all the reactor.

The idea of this PR is to parse all projects from the reactor only if the _makeBehavior_ is set (i.e. `-am` and/or `-amd` option has been used), or if a project activation has been set (i.e. `-pl`, `-r` or `-rf` is used).  Note that parents are always resolved from the file system, even if not included in the reactor.

The ITs only require a single modification: https://github.com/apache/maven-integration-testing/pull/239

### DIFF
--- a/maven-core/src/main/java/org/apache/maven/DefaultMaven.java
+++ b/maven-core/src/main/java/org/apache/maven/DefaultMaven.java
@@ -242,6 +242,13 @@ public class DefaultMaven implements Maven {
             return addExceptionToResult(result, e);
         }
 
+        try {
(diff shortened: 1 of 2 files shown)


cstamas (approved):

mthmulders (approved):
Nice improvement!

I've tested by building (`mvn test`) a particular module of a larger project, while a different module of the same project had an XML error. The build did not fail, until I ran `mvn -am test`. That gave a failing build because the other module POM could not be parsed.

PR #975, opened 2023-01-31
rename Maven's plexus-utils to maven-plexus-utils

hboutemy:
https://issues.apache.org/jira/browse/MNG-7655


elharo (changes requested):
This is dangerous (for a very weak definition of danger). In particular publishing the same classes in an artifact with a different artifact ID causes thorny, hard-to-debug, hard-to-understand dependency conflicts and even more failures in java 9+. -1

PR #977, opened 2023-01-31
[MNG-7676] Fix checksum plugin configuration

cstamas:
Wrong goal is being used that does not honor classifiers. Also, have to reshuffle the config, move it inside execution otherwise ASF parent pom execution is NOT overridden.

---

https://issues.apache.org/jira/browse/MNG-7676


kwin (approved):

cstamas:
Once merged (squashed+merged) this will need to be forward ported to master as well, as it suffers from same issue.

gnodet (approved):

PR #982, opened 2023-02-05
[MNG-7541] Implement powershell mvn command

JurrianFahner:

       pulling in other changes.


JurrianFahner:
Core IT tests fails currently on
`MavenITmng6609ProfileActivationForPackagingTest.testitMojoExecution:42->AbstractMavenIntegrationTestCase.assertEquals:668->AbstractMavenIntegrationTestCase.assertEquals:673 expected: <pom> but was: <null>`

It doesn't seem to be related to the changes made in this pull request.

JurrianFahner:
After pulling the latest version of the core-its, it was running fine. So the core-its ran successfully.

slachiewicz (approved):

apache-maven/src/bin/mvn.ps1
```diff
@@ -0,0 +1,158 @@
+<#
+Licensed to the Apache Software Foundation (ASF) under one
+or more contributor license agreements.  See the NOTICE file
+distributed with this work for additional information
+regarding copyright ownership.  The ASF licenses this file
+to you under the Apache License, Version 2.0 (the
+"License"); you may not use this file except in compliance
+with the License.  You may obtain a copy of the License at
+
+   http://www.apache.org/licenses/LICENSE-2.0
```
yeikel, line 10: Can we make this `https`?

JurrianFahner: @yeikel Good point. But why would we do it only in this occurrence? At this moment in time the http link is used in 1,663 places (it redirects properly to https). What would you think about the idea to create a separate pull request to resolve this for the whole codebase? For instance for the `3.9.x` and `master` branch.

yeikel: That's fair. I don't see any issue with that

gnodet:
I wonder if it would make sense to remove the `.cmd` script instead of keeping both `.ps1` and `.cmd` scripts.

apache-maven/src/bin/mvn.ps1
michael-o: Don't waste your time. This won't happen.

yeikel:
> I wonder if it would make sense to remove the `.cmd` script instead of keeping both `.ps1` and `.cmd` scripts.

That will likely break things so maybe that should be targeted for v4

gnodet:
> > I wonder if it would make sense to remove the `.cmd` script instead of keeping both `.ps1` and `.cmd` scripts.
> 
> That will likely break things so maybe that should be targeted for v4

Yes, I had not seen this PR was targeting 3.9.x specifically.

michael-o:
> > > I wonder if it would make sense to remove the `.cmd` script instead of keeping both `.ps1` and `.cmd` scripts.
> > 
> > 
> > That will likely break things so maybe that should be targeted for v4
> 
> Yes, I had not seen this PR was targeting 3.9.x specifically.

I'd require this to be on master first and then backported...at most

JurrianFahner:
@michael-o Should we remove `.cmd` in https://github.com/apache/maven/pull/877?

michael-o:
> @michael-o Should we remove `.cmd` in #877?

No, since the PR is only about addtion of a PS-based solution. No more, no less.

JurrianFahner:
@michael-o @slachiewicz What do we need to do to merge this merge request?

JurrianFahner:
@michael-o Can I do something to make this pull request merged?

michael-o:
Not now, ATM. I need to consult also a colleague who's proficient in PowerShell.

JurrianFahner:
> Not now, ATM. I need to consult also a colleague who's proficient in PowerShell.

Is there some progress?

michael-o:
> > Not now, ATM. I need to consult also a colleague who's proficient in PowerShell.
> 
> Is there some progress?

I hope to pick this up when my round with reporting stack is complete for next milestone. I guess somewere end of month. Please also raise your voice on dev@ to drag more attraction to this topic.

JurrianFahner:
> > > Not now, ATM. I need to consult also a colleague who's proficient in PowerShell.
> > 
> > 
> > Is there some progress?
> 
> I hope to pick this up when my round with reporting stack is complete for next milestone. I guess somewere end of month. Please also raise your voice on dev@ to drag more attraction to this topic.

Is there some progress?

michael-o:
For the time being, I want work on this since I don't have the capacity nor decent PowerShell scripting skill set. My knowledge in PS humble.

JurrianFahner:
@michael-o I understand your reserves, certainly if your see your PS knowledge as humble (I don't see myself as an expert either). 
At the same time @slachiewicz has given an approval. What's needed in order to get this PR merged?

Please let me know what is needed, then I can specifically ask for input/help on the mailing list.

apache-maven/src/bin/mvnDebug.ps1
```diff
@@ -0,0 +1,42 @@
...
+#>
+
+# set title
+$Host.UI.RawUI.WindowTitle = $MyInvocation.MyCommand
+
+if (-not $env:MAVEN_DEBUG_ADDRESS ) {
+    $env:MAVEN_DEBUG_ADDRESS = "localhost:8000"
+}
+
+$env:MAVEN_DEBUG_OPTS = "-Xdebug -Xrunjdwp:transport=dt_socket,server=y,suspend=y,address=$env:MAVEN_DEBUG_ADDRESS"
```
Giovds, line 38: I'm not as familiar with `.cmd` files, but I believe the [current](https://github.com/apache/maven/blob/maven-3.9.x/apache-maven/src/bin/mvnDebug.cmd) `.cmd` just makes it port `8000` (no `localhost`).

apache-maven/src/bin/mvn.ps1
```diff
@@ -0,0 +1,158 @@
...
+    Write-Error -Message $pom_file_error -ErrorAction Stop
+  }
+}
+else {
+  # if file flag is not found, then the pom.xml is relative to the working dir 
+  # and the jvm.config can be found in the maven project base dir. 
+
+  $basedir = $WDIR
+
+  while (IsNotRoot($WDIR.Path)) {
```
Giovds, line 126: If we do not find `.mvn` directory it will use the root directory as `$basedir` and as `-Dmaven.multiModuleProjectDirectory`?

I'm not entirely sure what the [`.cmd`](https://github.com/apache/maven/blob/0c8b7dfc605ce906f6c73db0aeab1735759da890/apache-maven/src/bin/mvn.cmd#L153) does though, I'm not as familiar with `.cmd`.

apache-maven/src/bin/mvn.ps1
```diff
@@ -0,0 +1,158 @@
...
+    }
+    else {
+      break
+    }  
+  }
+}
+
+$MAVEN_OPTS = (RetrieveContentsJvmConfig $basedir)
+
+$LAUNCHER_JAR = Get-Item $MAVEN_HOME"\boot\plexus-classworlds*.jar"
```
Giovds, line 143: If there are multiple jars (as the `*` suggests), I think this will fail as the separator would be a space.
(Note the [`.cmd`](https://github.com/apache/maven/blob/0c8b7dfc605ce906f6c73db0aeab1735759da890/apache-maven/src/bin/mvn.cmd#L169C1-L170C1) has an extra `-` and does not filter on `.jar`)

Giovds (commented):
I would love some powershell support.

apache-maven/src/bin/mvnDebug.ps1
JurrianFahner: Yes, you are right. I made here the assumption that people would use java > 9. However earlier versions of java seems not to support this notation. 

Please bear in mind that the behavior changes with java > 9, it will bind to only localhost instead of all internet interfaces. 

I'll change it, to bring it line with the .cmd file.

apache-maven/src/bin/mvn.ps1
JurrianFahner: Nope, it would use the current working directory as basedir.

apache-maven/src/bin/mvn.ps1
JurrianFahner: No, there is one jar. This is to be sure it can find the classworlds jar (even when the version changes).

apache-maven/src/bin/mvn.ps1
```diff
@@ -0,0 +1,158 @@
...
+  }
+  $i += 1
+}
+
+function IsNotRoot {
+  param (
+    [String] $path
+  )
+
+  return -not $path.endsWith(":\")
```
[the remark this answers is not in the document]
JurrianFahner: I wouldn't recommend using maven on a share, it would be not the fastest experience. 😉

apache-maven/src/bin/mvn.ps1
JurrianFahner: No there is only one jar. The * is there if the version changes, the script will stil work.

apache-maven/src/bin/mvn.ps1
```diff
@@ -0,0 +1,158 @@
...
+
+if (-not $env:MAVEN_SKIP_RC) {
+  if (Test-Path -Path $env:PROGRAMDATA\mavenrc.ps1 -PathType Leaf) { &$env:PROGRAMDATA"\mavenrc.ps1" $args }
+  if (Test-Path -Path $env:USERPROFILE\mavenrc.ps1 -PathType Leaf) { &$env:USERPROFILE"\mavenrc.ps1" $args }
+}
+
+if (-not (Test-path $env:JAVA_HOME)) {
+  $JAVACMD = (get-command java).Source 
+  if (-not $JAVACMD) {
+    Write-Error -ErrorAction Stop -Message "The $env:JAVA_HOME has not been set, JAVA_HOME environment variable is not defined correctly, so Apache Maven cannot be started."
```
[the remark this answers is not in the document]
JurrianFahner: Just tried to find an answer from the powershell community which would might give a best practice. 

I don't have a strong opinion on this one.

apache-maven/src/bin/mvn.ps1
```diff
@@ -0,0 +1,158 @@
...
+# set title
+$Host.UI.RawUI.WindowTitle = (Get-Variable MyInvocation -Scope Script).Value.InvocationName
+
+if (-not $env:MAVEN_SKIP_RC) {
+  if (Test-Path -Path $env:PROGRAMDATA\mavenrc.ps1 -PathType Leaf) { &$env:PROGRAMDATA"\mavenrc.ps1" $args }
+  if (Test-Path -Path $env:USERPROFILE\mavenrc.ps1 -PathType Leaf) { &$env:USERPROFILE"\mavenrc.ps1" $args }
+}
+
+if (-not (Test-path $env:JAVA_HOME)) {
+  $JAVACMD = (get-command java).Source
```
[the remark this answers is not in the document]
JurrianFahner: I agree! We need to do a try-catch here.

apache-maven/src/bin/mvn.ps1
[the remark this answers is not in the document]
Giovds: Me neither, but you never know what you find in the wild 🙈

michael-o:
As a matter of fact, this implemenation should copy the approach from `mvn.cmd` from `master`, not from `maven-3.9.x`. It does not matter to add something new and then skim a bit later.

michael-o (changes requested):
As a matter of fact, this implemenation should copy the approach from mvn.cmd from master, not from maven-3.9.x. It does not matter to add something new and then skim a bit later. Then it would be back ported to 3.9.x as-is logically.

apache-maven/src/bin/mvn.ps1
```diff
@@ -0,0 +1,158 @@
...
+
+  JAVA_HOME       (Optional) Points to a Java installation.
+  MAVEN_ARGS      (Optional) Arguments passed to Maven before CLI arguments.
+  MAVEN_OPTS      (Optional) Java runtime options used when Maven is executed.
+  MAVEN_SKIP_RC   (Optional) Flag to disable loading of mavenrc files.
+-----------------------------------------------------------------------------
+#>
+
+# set title
+$Host.UI.RawUI.WindowTitle = (Get-Variable MyInvocation -Scope Script).Value.InvocationName
```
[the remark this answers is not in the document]
JurrianFahner: @Giovds It is only part of mvnDebug, it shouldn't be part of mvn.ps1 IMHO. I like your idea, but it shouldn't be part of this PR, but in a separate issue. 
For now I'll remove this from mvn.ps1.

apache-maven/src/bin/mvn.ps1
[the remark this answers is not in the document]
JurrianFahner: - Just dived into this topic. The difference between with `throw` and `write-error` is that `write-error` writes to the error stream and `throw` doesn't.
 
- Since PowerShell 7.2 the default `$errorView` is set to ConciseView. Your example would show something like this:

image

Based on the above I think I don't need to change the implementation.

apache-maven/src/bin/mvn.ps1
```diff
@@ -0,0 +1,158 @@
...
+
+if (-not $env:MAVEN_SKIP_RC) {
+  if (Test-Path -Path $env:PROGRAMDATA\mavenrc.ps1 -PathType Leaf) { &$env:PROGRAMDATA"\mavenrc.ps1" $args }
```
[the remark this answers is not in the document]
michael-o: I wouldn't do it. New script, new behavior.

apache-maven/src/bin/mvn.ps1
```diff
@@ -0,0 +1,158 @@
...
+KIND, either express or implied.  See the License for the
+specific language governing permissions and limitations
+under the License.
+
+-----------------------------------------------------------------------------
+Apache Maven Startup Script
+
+Environment Variable Prerequisites
+
+  JAVA_HOME       (Optional) Points to a Java installation.
```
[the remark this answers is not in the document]
JurrianFahner: The `JAVA_HOME` variable is set as optional, in a similar way as the cmd implmentation. the cmd implementation also resolves the java exe on path. 
I consider this out of scope of this PR.

apache-maven/src/bin/mvn.ps1
```diff
@@ -0,0 +1,158 @@
...
+# Fallback to current working directory if not found.
+
+$WDIR = Get-Location
+
+# Look for the --file switch and start the search for the .mvn directory from the specified
+# POM location, if supplied.
+
+$i = 0
+$file_flag_found = $false
+foreach ($arg in $args) {
```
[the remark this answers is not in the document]
JurrianFahner: The nice thing about the choice I made is that the command line parameters are consistent through all shells. With using named parameters, the parameters will have only one `-`. 

It would also require changing `CLIManager`, the changes would become bigger than I anticipated.

JurrianFahner:
> As a matter of fact, this implemenation should copy the approach from mvn.cmd from master, not from maven-3.9.x. It does not matter to add something new and then skim a bit later. Then it would be back ported to 3.9.x as-is logically.

Wouldn't it not be better to get [master](https://github.com/apache/maven/pull/878) reviewed first and then retrofit the approach in master to the 3.9.x branch? Please be aware that there will be two different approaches in the component.xml, which makes reviewing a little bit harder.

michael-o:
> > As a matter of fact, this implemenation should copy the approach from mvn.cmd from master, not from maven-3.9.x. It does not matter to add something new and then skim a bit later. Then it would be back ported to 3.9.x as-is logically.
> 
> Wouldn't it not be better to get [master](https://github.com/apache/maven/pull/878) reviewed first and then retrofit the approach in master to the 3.9.x branch? Please be aware that there will be two different approaches in the component.xml, which makes reviewing a little bit harder.

True, but the content of the script won't change, only its arrangement.

JurrianFahner:
> > > As a matter of fact, this implemenation should copy the approach from mvn.cmd from master, not from maven-3.9.x. It does not matter to add something new and then skim a bit later. Then it would be back ported to 3.9.x as-is logically.
> > 
> > 
> > Wouldn't it not be better to get [master](https://github.com/apache/maven/pull/878) reviewed first and then retrofit the approach in master to the 3.9.x branch? Please be aware that there will be two different approaches in the component.xml, which makes reviewing a little bit harder.
> 
> True, but the content of the script won't change, only its arrangement.

Ok, in that case we are aligned. I'll update the master branch with the improvements on this branch. When the master is merged, I'll retrofit the changes into this branch.

@Giovds Are you available for reviewing the master PR? I first need to apply the changes on this branch to the master (#878).

Giovds (approved):

jira-importer:
Resolve #8909

gnodet (commented):
## AI Review — PR #982: [MNG-7541] Implement powershell mvn command (3.9.x)

This PR adds `mvn.ps1` and `mvnDebug.ps1` PowerShell scripts for Maven 3.9.x. It has been stalled since September 2023 with outstanding `CHANGES_REQUESTED` from committer michael-o and unresolved inline comments from Giovds.

### Bugs

**1. Syntax error — unbalanced parentheses** — The `IsNotRoot` function has mismatched parentheses on both lines:
```powershell
$root_path_base_name = ((get-item $path).Root.BaseName   # 2 open, 1 close
$path_base_name = ((get-item $path).BaseName              # 2 open, 1 close
```
This is a **parse error** — the script will fail to load. Fix: use single opening paren.

**2. `Test-Path` on null `$env:JAVA_HOME`** — When `JAVA_HOME` is not set, `Test-Path $null` throws a `ParameterBindingValidationException` rather than returning `$false`. The fallback logic to find `java` on PATH won't work.

**3. `$MAVEN_ARGS` vs `$env:MAVEN_ARGS`** — Line 149 references `$MAVEN_ARGS` (local variable, never assigned) instead of `$env:MAVEN_ARGS`. The `MAVEN_ARGS` environment variable will be silently ignored.

**4. `$MAVEN_DEBUG_OPTS` scope mismatch** — `mvnDebug.ps1` sets `$env:MAVEN_DEBUG_OPTS` (environment variable) but `mvn.ps1` reads `$MAVEN_DEBUG_OPTS` (local variable). Debug options are silently dropped when using `mvnDebug.ps1`.

### Other Issues

- **Path construction**: Uses string concatenation (`$env:JAVA_HOME + "\bin\java.exe"`) instead of `Join-Path`. Fragile with paths containing spaces.
- **Wildcard classpath**: `Get-Item $MAVEN_HOME"\boot\plexus-classworlds*.jar"` may return an array if multiple jars exist.
- **Debug address**: Comment says `localhost:8000` but code defaults to `"8000"` (listens on all interfaces on Java 9+).
- **Missing exit code propagation**: No `exit $LASTEXITCODE` after Java process invocation.

### Process Note

Committer michael-o requested that PR #878 (targeting master) be merged first and this PR be a backport. Both PRs share the same `IsNotRoot` syntax error. Several inline review comments from Giovds also remain unresolved.

---

PR #987, opened 2023-02-08
Patterns normalization

tbw777:
More shortest and simple Patterns.


tbw777:
i will create pr in 3.9.x

PR #991, opened 2023-02-08
Backport 3.8: [MNG-7686]

tbw777:
[MNG-7686] Speed up by replacing non-pattern #replaceAll() with #replace() or precompiled patterns


jira-importer:
Resolve #8767

PR #998, opened 2023-02-13
Using Set<>() for contains()

tbw777:


maven-core/src/main/java/org/apache/maven/ReactorReader.java
```diff
@@ -29,15 +29,7 @@
...
 import java.nio.file.StandardCopyOption;
-import java.util.ArrayDeque;
-import java.util.Arrays;
-import java.util.Collection;
-import java.util.Collections;
-import java.util.Deque;
-import java.util.HashMap;
-import java.util.List;
-import java.util.Map;
-import java.util.Objects;
```
michael-o, line 40: No, please undo this.

maven-core/src/main/java/org/apache/maven/ReactorReader.java
```diff
@@ -29,15 +29,7 @@
...
+import java.util.*;
```
michael-o, line 32: No, please undo this.

maven-core/src/main/java/org/apache/maven/ReactorReader.java
```diff
@@ -69,8 +61,12 @@ class ReactorReader implements MavenWorkspaceReader {
...
     public static final String PROJECT_LOCAL_REPO = "project-local-repo";
 
-    private static final Collection<String> COMPILE_PHASE_TYPES =
-            Arrays.asList("jar", "ejb-client", "war", "rar", "ejb3", "par", "sar", "wsr", "har", "app-client");
+    private static final Set<String> COMPILE_PHASE_TYPES;
+
+    static {
+        COMPILE_PHASE_TYPES = new HashSet<>(
+                Arrays.asList("jar", "ejb-client", "war", "rar", "ejb3", "par", "sar", "wsr", "har", "app-client"));
+    }
```
michael-o, line 69: Where is the benefit?

michael-o (changes requested):

maven-core/src/main/java/org/apache/maven/ReactorReader.java
tbw777: contains is faster on Set then List

michael-o: True, but does the type has to change to set from collection for this?

tbw777: https://gist.github.com/tbw777/716907b110df7eac1792de7148f8eb32

gnodet: I think what @michael-o meant is that we can keep the `Collection<String>` instead of changing the type to `Set<String>`:
```
private static final Collection<String> COMPILE_PHASE_TYPES =
             new HashSet<>(Arrays.asList("jar", "ejb-client", "war", "rar", "ejb3", "par", "sar", "wsr", "har", "app-client"));
```

michael-o: Correct.

maven-core/src/main/java/org/apache/maven/ReactorReader.java
```diff
@@ -69,8 +70,12 @@ class ReactorReader implements MavenWorkspaceReader {
 
     public static final String PROJECT_LOCAL_REPO = "project-local-repo";
 
-    private static final Collection<String> COMPILE_PHASE_TYPES =
-            Arrays.asList("jar", "ejb-client", "war", "rar", "ejb3", "par", "sar", "wsr", "har", "app-client");
+    private static final Collection<String> COMPILE_PHASE_TYPES;
+
```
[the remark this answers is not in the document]
gnodet: I really wish we could leverage jdk 11 and go for `Set.of(xxx)` instead.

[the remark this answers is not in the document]
tbw777: updated

gnodet (approved):

PR #1003, opened 2023-02-20
[EXPERIMENT] Halve logging

cstamas:
Of the resolver (downloading vs downloaded), just report after
the fact it was downloaded (or failed to download), no need
for double lines for same resource/dependency.

This experiment includes https://github.com/apache/maven/pull/989
but does not affects it (just to create "closer" Maven to
upcoming release).

As one could expect, IT suite does fail with this change
as some ITs do assert against the build logs (but the build
itself did pass OK). Failing ITs:

[ERROR] Failures:
[ERROR]   MavenITmng6240PluginExtensionAetherProvider.testPluginExtensionDependingOnMavenAetherProvider:83->AbstractMavenIntegrationTestCase.assertEquals:668->AbstractMavenIntegrationTestCase.assertEquals:673 expected: <2> but was: <0>
[INFO]
[ERROR] Tests run: 885, Failures: 1, Errors: 0, Skipped: 80

As to be expected, the IT asserts there is log line without
progress (so DownloadING one, the one that is removed).

### DIFF
--- a/maven-core/src/main/java/org/apache/maven/internal/aether/DefaultRepositorySystemSessionFactory.java
+++ b/maven-core/src/main/java/org/apache/maven/internal/aether/DefaultRepositorySystemSessionFactory.java
@@ -58,6 +58,10 @@ import org.eclipse.aether.repository.RepositoryPolicy;
 import org.eclipse.aether.repository.WorkspaceReader;
 import org.eclipse.aether.resolution.ResolutionErrorPolicy;
 import org.eclipse.aether.spi.localrepo.LocalRepositoryManagerFactory;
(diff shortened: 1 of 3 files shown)


michael-o (changes requested):
I wouldn't do it that way. We have several listeners. They need to be exposed with a name and new ones can be introduced from CP. Then the user can pick. With that we can deprecate `--no-transfer-progress` to the symbolic name.

cstamas:
I completely agree, this wrapping I did in this PoC is ugy, and serves the purpose only to see "how would it look"... This PR is def not to be reused, especially not merged.

elharo:
I like it. :-)

PR #1007, opened 2023-02-21
[MNG-7702] Update to Maven Deploy Plugin 3.1.0

cstamas:
And have it aligned with install plugin. These two now provides all the fixes and new functionalities like working deployAtEnd etc.

---

https://issues.apache.org/jira/browse/MNG-7702


michael-o (changes requested):
Commit summary shall contain "Update to Maven Deploy Plugin XXX"

Then it will properly appear in release notes, "latest" is a relative term

michael-o (approved):

PR #1009, opened 2023-02-21
[MNG-7706] Deprecate 'localRepository' parameter expression

cstamas:
This PR deprecates the 'localRepository' mojo parameter expression, and Core will emit warning if used by any Mojo.

---

https://issues.apache.org/jira/browse/MNG-7706

PR #1012, opened 2023-02-23
[MNG-7706] Deprecate 'localRepository' parameter expression (#1009)

cstamas:
This PR deprecates the 'localRepository' mojo parameter expression, and Core will emit warning if used by any Mojo.

Forward port of https://github.com/apache/maven/pull/1009

---

https://issues.apache.org/jira/browse/MNG-7706

### DIFF
--- a/maven-artifact/src/main/java/org/apache/maven/artifact/repository/ArtifactRepository.java
+++ b/maven-artifact/src/main/java/org/apache/maven/artifact/repository/ArtifactRepository.java
@@ -28,7 +28,10 @@ import org.apache.maven.repository.Proxy;
 /**
  * Abstraction of an artifact repository. Artifact repositories can be remote, local, or even build reactor or
  * IDE workspace.
+ *
+ * @deprecated Avoid use of this type, if you need access to local repository use repository system classes instead.
  */
+@Deprecated
 public interface ArtifactRepository {
     String pathOf(Artifact artifact);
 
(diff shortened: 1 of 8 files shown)


rmannibucau (approved):

paulwwisl:
Does this has to do with warning for test phase, when actually there are no tests at all?

mvn version 3.9.1

```
[INFO] --- surefire:2.22.2:test (default-test) @ ..... ---
[WARNING] Parameter 'localRepository' is deprecated core expression; Avoid use of ArtifactRepository type. If you need access to local repository, switch to '${repositorySystemSession}' expression and get LRM from it instead.
[INFO] No tests to run.
```

paulwwisl:
solved with latest 3.0.0 https://maven.apache.org/surefire/maven-surefire-plugin/

e.g. with Spring Boot

```xml
<maven-surefire-plugin.version>3.0.0</maven-surefire-plugin.version>
```

basil:
How should https://github.com/cloudbees/maven-license-plugin/blob/b58872cc1f4062ec01ca903d9a5546cc3b85a748/src/main/java/com/cloudbees/maven/license/ProcessMojo.java#L179 be adapted? The deprecation notice says to switch to `repositorySystemSession.getLocalRepository()`, but that does not compile:

> java: incompatible types: `org.eclipse.aether.repository.LocalRepository` cannot be converted to `org.apache.maven.artifact.repository.ArtifactRepository`

cstamas:
1. drop `localRepository`, remove `buildingRequest.setLocalRepository(localRepository);`
2. instead grab RepositorySystemSession   "${repositorySystemSession}"
3. on ProjectBuildingRequest setSession from 2 and remove dropped setter from 1
4. test it

basil:
> on `ProjectBuildingRequest` `setSession` from 2

The `setSession` method does not exist, but a `setRepositorySession` exists. But why should I call that when `session.getProjectBuildingRequest()` already calls it a few lines above?

cstamas:
Yup, setRepositorySession... if you are sure it is there, then not needed.

cstamas:
Just for future: please use ML for asks like this one :)
https://maven.apache.org/mailing-lists.html

basil:
This migration was poorly documented. What you asked me to do had nothing to do with the instructions in the deprecation warning. I think it was appropriate to complain about that in this PR.

PR #1016, opened 2023-02-26
remove unused code

elharo:
and we remove reference to class we want to delete

### DIFF
--- a/maven-xml-impl/src/main/java/org/codehaus/plexus/util/xml/Xpp3Dom.java
+++ b/maven-xml-impl/src/main/java/org/codehaus/plexus/util/xml/Xpp3Dom.java
@@ -27,7 +27,6 @@ import java.util.Map;
 
 import org.apache.maven.api.xml.XmlNode;
 import org.apache.maven.internal.xml.XmlNodeImpl;
-import org.codehaus.plexus.util.StringUtils;
 import org.codehaus.plexus.util.xml.pull.XmlSerializer;
 
 /**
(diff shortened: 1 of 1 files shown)


gnodet (approved):
Looks good.  However, the goal is to get rid of the `org.codehaus.plexus` classes in this artifact, see https://github.com/codehaus-plexus/plexus-utils/pull/236 and the [discussion on `dev@`](https://lists.apache.org/thread/tv1gt9lhjxoxh0cnrh84f12c895yf2n3)

elharo:
Yes, I've been looking at that. There are several somewhat parallel tracks here. The work you're doing on XML is the biggest part of it. I'd also like to get rid of a lot of the duplicate utility code that's been copied and pasted into Plexus utils, maven-shared-utils, apache commons, and other places over the years. These days a lot of it can ve found in the JDK. Ideally I'd like tp reach a point where org.codehaus is nowhere to be found in our code. One can dream. :-)

gnodet:
> Yes, I've been looking at that. There are several somewhat parallel tracks here. The work you're doing on XML is the biggest part of it. I'd also like to get rid of a lot of the duplicate utility code that's been copied and pasted into Plexus utils, maven-shared-utils, apache commons, and other places over the years. These days a lot of it can ve found in the JDK. Ideally I'd like tp reach a point where org.codehaus is nowhere to be found in our code. One can dream. :-)

I would begin by removing any dependency on `maven-shared-utils` because I think not many projects have migrated from plexus-utils.  So the path will be much easier in this way.

elharo:
It all moves to JDK classes when possible, and Apache commons when not so there's no intermediate step. This is true whether it's maven-shared-utils or plexus utils. I have already deprecated a lot of the functionality in maven-shared-utils so I hope that pokes a few projects to update themselves without me having to do the work piecemeal.

PR #1018, opened 2023-02-27
[maven-3.9.x] [MNG-7713] Drop legacy-local-repository option

cstamas:
There is really no need for it, and there is a resolver option if really must (but really should not, local repo is and should be considered as transient).

---

### DIFF
--- a/maven-embedder/src/main/java/org/apache/maven/cli/CLIManager.java
+++ b/maven-embedder/src/main/java/org/apache/maven/cli/CLIManager.java
@@ -101,8 +101,6 @@ public class CLIManager {
 
     public static final String THREADS = "T";
 
-    public static final String LEGACY_LOCAL_REPOSITORY = "llr";
-
     public static final String BUILDER = "b";
 
     public static final String NO_TRANSFER_PROGRESS = "ntp";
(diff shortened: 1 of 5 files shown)


maven-embedder/src/main/java/org/apache/maven/cli/CLIManager.java
```diff
@@ -256,8 +256,7 @@ public CLIManager() {
                 .build());
         options.addOption(Option.builder(LEGACY_LOCAL_REPOSITORY)
                 .longOpt("legacy-local-repository")
-                .desc(
-                        "Use Maven 2 Legacy Local Repository behaviour, ie no use of _remote.repositories. Can also be activated by using -Dmaven.legacyLocalRepo=true")
+                .desc("Ineffective, only kept for backward compatibility.")
```
slawekjaranowski, line 259: If such option will have  no effect - why we want to preserve it ... 
If user use it - it will not works as user expect - so we will have silently different behavior that user think

For me will be better to drop at all -  it will more clear for user and do not make confusing

cstamas: Agreed, i was drawn by other left defunct options...

slawekjaranowski (approved):

jira-importer:
Resolve #9268

PR #1019, opened 2023-03-01
[MNG-7728] Fix mutable api in PluginContainer

gnodet:
JIRA issue: https://issues.apache.org/jira/browse/MNG-7728

The PluginContainer from the v4 api uses the same additional code block than the v3 api and thus offers a `getPluginsAsMap()` method along with a `flushPluginMap()` which makes the whole class non immutable.
Given additional constructor arguments are not supported by the generator, using a lazy computed field should be enough to ensure that the content of the class never changes, even though the field itself can change, the value returned by `getPluginsAsMap()` cannot.

### DIFF
--- a/api/maven-api-model/src/main/mdo/maven.mdo
+++ b/api/maven-api-model/src/main/mdo/maven.mdo
@@ -624,7 +624,7 @@
       </fields>
       <codeSegments>
         <codeSegment>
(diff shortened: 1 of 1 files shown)


cstamas (approved):

jira-importer:
Resolve #9253

PR #1021, opened 2023-03-01
[MNG-7713] Make Maven fail if option present

cstamas:
As with previous PR (simple removal) the `-llr` got interpreted as `-l lr`, it logged all output to `lr` file. This would maean that use of `-llr` would still sneakily 'work' and probably cause surprise down the road to users.

Returned the option, and expicitly checking for it's presence to be able to fail with meaningful message.

---

### DIFF
--- a/maven-embedder/src/main/java/org/apache/maven/cli/CLIManager.java
+++ b/maven-embedder/src/main/java/org/apache/maven/cli/CLIManager.java
@@ -283,7 +283,7 @@ public class CLIManager {
         // Adding this back to make Maven fail if used
         options.addOption(Option.builder("llr")
                 .longOpt("legacy-local-repository")
-                .desc("Ineffective, use of this option will make Maven invocation fail.")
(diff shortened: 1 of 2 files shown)


maven-embedder/src/main/java/org/apache/maven/cli/CLIManager.java
```diff
@@ -280,6 +280,12 @@ public CLIManager() {
                 .desc("Ineffective, only kept for backward compatibility")
                 .build());
 
+        // Adding this back to make Maven fail if used
```
slawekjaranowski, line 283: Please add info about conflict with `-l lr` - why we need it

cstamas: We need it not only to avoid conflicting `-l lr` but also to emit sane error message (so we really want to parse it from CLI but fail if present). So, I started a new "section" here (below the "ineffective section") for those that should fail.

slawekjaranowski (approved):

PR #1022, opened 2023-03-01
[MNG-7713] Make Maven fail if option present (#1021)

cstamas:
As with previous PR (simple removal) the `-llr` got interpreted as `-l lr`, it logged all output to `lr` file. This would maean that use of `-llr` would still sneakily 'work' and probably cause surprise down the road to users.

Returned the option, and expicitly checking for it's presence to be able to fail with meaningful message.

---


gnodet (approved):

jira-importer:
Resolve #9268

PR #1030, opened 2023-03-02
[MNG-7717] Maven warns wrongly about deprecated parameter

cstamas:
The implementation for MNG-7706 is wrong: it checks parameter **name** where it should check **type** and **defaultValue** (expression) instead.

---

https://issues.apache.org/jira/browse/MNG-7717


slawekjaranowski (approved):

jira-importer:
Resolve #9264

PR #1040, opened 2023-03-07
[MNG-7720] Wrong build order of forked projects

cstamas:
Forward port of fix.

---

https://issues.apache.org/jira/browse/MNG-7720

### DIFF
--- a/maven-core/src/main/java/org/apache/maven/lifecycle/internal/LifecycleDependencyResolver.java
+++ b/maven-core/src/main/java/org/apache/maven/lifecycle/internal/LifecycleDependencyResolver.java
@@ -103,7 +103,10 @@ public class LifecycleDependencyResolver {
     private static Stream<MavenProject> getProjectAndSubModules(MavenProject project) {
         return Stream.concat(
                 Stream.of(project),
-                project.getCollectedProjects().stream().flatMap(LifecycleDependencyResolver::getProjectAndSubModules));
+                project.getCollectedProjects() == null
+                        ? Stream.empty()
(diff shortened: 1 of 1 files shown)


maven-core/src/main/java/org/apache/maven/lifecycle/internal/LifecycleDependencyResolver.java
```diff
@@ -90,7 +90,11 @@ public LifecycleDependencyResolver(
 
     public static List<MavenProject> getProjects(MavenProject project, MavenSession session, boolean aggregator) {
         if (aggregator && project.getCollectedProjects() != null) {
```
slawekjaranowski, line 92: Here we have check `project.getCollectedProjects() != null` 
but in method `getProjectAndSubModules` we list project `recursively` and we don't check if subproject has `getCollectedProjects() != null` - it is ok? line 102/106

cstamas: I think it can be only on "top level" null (ie. some goal invocation not project being built), and once top level is not null, no child will have it null either... this is just a wild guess.

cstamas: It seems this is really "just implementation details" (that when top level project collectedProjects is not null, then there is no project instance that would have collectedProjects null). So added nullcheck just to be on safe side.

slawekjaranowski (approved):

PR #1041, opened 2023-03-07
[MNG-7724] Don't log warnings when runtime is not broken (slf4j integrations)

rmannibucau:

       pulling in other changes.

The issue with current code is that using a SLF4J binding which is not known by maven will lead to WARNINGs.
There are several builds which intend to run without warning or consider warnings as error and therefore fail - not using maven CLI options but other mecanism - either log analyzis on CI, the old way with appenders/handlers or stdout/stderr check if the handlers support the redirection properly.
This behavior defeats a bit using a logger abstraction since the support is pretty limited.
The idea is to keep the message - we can refine it later if relevant - but log it at info level to not make the build failling when there is no issue loading the impl maven should use and keep warning when a loading failed (we can evaluate the move to error later too for this case).

### DIFF
--- a/maven-embedder/pom.xml
+++ b/maven-embedder/pom.xml
@@ -141,6 +141,12 @@ under the License.
       <optional>true</optional>
     </dependency>
     <dependency>
+      <groupId>org.slf4j</groupId>
+      <artifactId>slf4j-jdk14</artifactId>
+      <version>${slf4jVersion}</version>
+      <optional>true</optional>
+    </dependency>
+    <dependency>
       <groupId>ch.qos.logback</groupId>
       <artifactId>logback-classic</artifactId>
       <optional>true</optional>
--- a/maven-embedder/src/main/java/org/apache/maven/cli/logging/impl/JDKConfiguration.java
+++ b/maven-embedder/src/main/java/org/apache/maven/cli/logging/impl/JDKConfiguration.java
@@ -1,0 +1,52 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.maven.cli.logging.impl;
+
+import java.util.logging.Logger;
+
+import org.apache.maven.cli.logging.BaseSlf4jConfiguration;
+
+/**
+ * Configuration for slf4j-jdk14.
+ */
+public class JDKConfiguration extends BaseSlf4jConfiguration {
+    @Override
+    public void setRootLoggerLevel(Level level) {
+        java.util.logging.Level value;
+        switch (level) {
+            case DEBUG:
+                value = java.util.logging.Level.FINEST;
+                break;
+
+            case INFO:
+                value = java.util.logging.Level.INFO;
+                break;
+
+            default:
+                value = java.util.logging.Level.SEVERE;
+                break;
+        }
+        Logger.getLogger("").setLevel(value);
+    }
+
+    @Override
+    public void activate() {
+        // no op
+    }
+}
--- a/maven-embedder/src/main/java/org/apache/maven/cli/logging/impl/NoopConfiguration.java
+++ b/maven-embedder/src/main/java/org/apache/maven/cli/logging/impl/NoopConfiguration.java
@@ -1,0 +1,39 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.maven.cli.logging.impl;
+
+import org.apache.maven.cli.logging.BaseSlf4jConfiguration;
+
+/**
+ * Configuration available for users setting a custom binding and not using --fail-on-severity to avoid warnings.
+ * This is basically a no-op implementation not issuing any warning.
+ * To use it add to maven classpath (m2.conf) a folder/jar with a META-INF/maven/slf4j-configuration.properties
+ * containing {@code loggerClass = org.apache.maven.cli.logging.impl.NoopConfiguration}.
+ */
+public class NoopConfiguration extends BaseSlf4jConfiguration {
+    @Override
+    public void setRootLoggerLevel(Level level) {
+        // no-op
+    }
+
+    @Override
+    public void activate() {
+        // no-op
+    }
+}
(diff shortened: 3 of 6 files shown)


maven-embedder/src/test/java/org/apache/maven/cli/logging/Slf4jConfigurationFactoryTest.java
```diff
@@ -0,0 +1,107 @@
...
+            Files.deleteIfExists(fakeProvider);
+        }
+    }
+
+    private BiConsumer<Logger, String> findLogHandler(final UnsupportedSlf4jBindingConfiguration configuration) {
+        try {
+            final Field log = configuration.getClass().getDeclaredField("log");
+            log.setAccessible(true);
+            return (BiConsumer<Logger, String>) log.get(configuration);
+        } catch (final Exception e) {
```
elharo, line 82: 1, No need to mark this final
2. Make exception specific if possible

maven-embedder/src/main/java/org/apache/maven/cli/logging/Slf4jConfigurationFactory.java
```diff
@@ -65,8 +67,14 @@ public static Slf4jConfiguration getConfiguration(ILoggerFactory loggerFactory)
             }
         } catch (IOException | ClassNotFoundException | IllegalAccessException | InstantiationException e) {
             e.printStackTrace();
+            hadException = true;
         }
 
-        return new UnsupportedSlf4jBindingConfiguration(slf4jBinding, supported);
+        return new UnsupportedSlf4jBindingConfiguration(
+                slf4jBinding,
+                supported,
```
elharo, line 75: warning seems the right level error. If someone is treating this as an error, that's their choice and problem, not ours.

elharo (changes requested):

maven-embedder/src/test/java/org/apache/maven/cli/logging/Slf4jConfigurationFactoryTest.java
rmannibucau: Well this is to make it more readable, having 3 exception there does not help anything since it is really a "pass or fail" thing and handling is the same so for tests this looks a better compromise IMHO

maven-embedder/src/main/java/org/apache/maven/cli/logging/Slf4jConfigurationFactory.java
rmannibucau: Not really, maven is wrong assuming it knows the entire ecosystem, it is perfectly valid to use a slf4j binding maven does not know - or if you are right it means we should drop slf4j from maven because we don't use the abstraction at the end.
If it is right then we must not issue a warning (an error) when it is an expected case  - the use case triggering this change.
Guess it was more a default implementation than an intent to say a custom binding is intended.

Assuming any custom binding will bring a custom maven integration is also a high assumption for the little usage we do of that integration (a single CLI toggle which is rarely used in practise) so hope we don't go that path which would be the alternative.

maven-embedder/src/test/java/org/apache/maven/cli/logging/Slf4jConfigurationFactoryTest.java
elharo: Strongly disagree. Catching Exception here risks pulling in runtime exceptions we're not actually handling.

rmannibucau: we do handle them all -> fail, all good

maven-embedder/src/main/java/org/apache/maven/cli/logging/Slf4jConfigurationFactory.java
elharo: If this is valid, we shouldn't log at all. If it isn't, log the warning. An INFO level log isn't right here.

rmannibucau: I rewrote the PR to not play with levels, please check the new proposal to solve this issue

maven-embedder/pom.xml
```diff
@@ -140,6 +140,12 @@ under the License.
       <artifactId>slf4j-simple</artifactId>
       <optional>true</optional>
     </dependency>
+    <dependency>
+      <groupId>org.slf4j</groupId>
+      <artifactId>slf4j-jdk14</artifactId>
```
elharo, line 145: Does this introduce JDK 9+ byte code into the classpath? That would be a deal breaker for now.

maven-embedder/src/main/java/org/apache/maven/cli/logging/impl/JDKConfiguration.java
```diff
@@ -0,0 +1,52 @@
...
+package org.apache.maven.cli.logging.impl;
+
+import java.util.logging.Logger;
+
+import org.apache.maven.cli.logging.BaseSlf4jConfiguration;
+
+/**
+ * Configuration for slf4j-jdk14.
+ */
+public class JDKConfiguration extends BaseSlf4jConfiguration {
```
elharo, line 28: This is additional API I'd rather not carry around. I need to understand the root issue better, but it still seems to me either warning is correct, in which case we should do nothing, or we should remove the log message completely. I don't like half measures and work arounds for misconfigured builds.

maven-embedder/src/main/java/org/apache/maven/cli/logging/impl/NoopConfiguration.java
```diff
@@ -0,0 +1,39 @@
...
+
+import org.apache.maven.cli.logging.BaseSlf4jConfiguration;
+
+/**
+ * Configuration available for users setting a custom binding and not using --fail-on-severity to avoid warnings.
+ * This is basically a no-op implementation not issuing any warning.
+ * To use it add to maven classpath (m2.conf) a folder/jar with a META-INF/maven/slf4j-configuration.properties
+ * containing {@code loggerClass = org.apache.maven.cli.logging.impl.NoopConfiguration}.
+ */
+public class NoopConfiguration extends BaseSlf4jConfiguration {
```
elharo, line 29: I thought the point of log frameworks was that this sort of thing could be done in client projects without requiring support from projects doing the logging.

elharo (changes requested):

maven-embedder/pom.xml
rmannibucau: 14 = 1.4, code is from 2006 ;)

maven-embedder/src/main/java/org/apache/maven/cli/logging/impl/JDKConfiguration.java
rmannibucau: well, this one is easy, maven integrates with logback, log4j2, etc so it integrated with slf4j bindings, jdk14 is just a standard one so we must integrate with, no real ambiguity.
now if I didn't get it and you mean maven shouldn't integrate with slf4j binding - I can agree with that at some point - then I'd be happy to drop logback, log4j2 and other integrations too.

maven-embedder/src/main/java/org/apache/maven/cli/logging/impl/NoopConfiguration.java
rmannibucau: Fully agree but not how maven integrated the fail on severity flag, just trying to not drop the flag - shouldn't have been introduced since it is 100% a log framework feature for me - and get it running properly without warnings which actually just mention the exact case you setup (not supported by maven slf4j binding).

There are actually two cases assuming we keep this kind of integration:

* maven supports the binding but the loading fails (missing dep - missing logback-classic when adding logback-core for ex or the opposite) -> we should issue a warning
* maven does not know the binding -> guess we can discuss info (user intended to use a custom binding) or warning (user didnt intend to use the binding added to mvn classpath due to an extension transitive dep) but having this noop impl enables maven to issue a warning and the user to drop it when he knows it is not relevant

maven-embedder/src/main/java/org/apache/maven/cli/logging/impl/JDKConfiguration.java
elharo: OK, as long as this means JDK 1.4 and not JDK 14.0 no problem here. Time marches on...

elharo (commented):
I think I see how this is working now, though I'm still unclear on why this belongs in Maven core. Aren't there j.u.logging and Noop loggers already available that the user can install if they need one?

rmannibucau:
@elharo jul is jdk14 one (the PR adds its support, was not there not sure why), noop is kind of built-in but defeats the fact to log a warning on a random logger (but this is another ticket/issue I guess). The root cause is really the fact maven adds an integration layer with each slf4j implementation so at the end can't reliably rely on the slf4j abstraction.
For maven - as of today - slf4j binding == slf4j-$impl + maven-slf4j-$impl (in embedder module), this is what can need rework the future cause the relationship of slf4j binding should be sufficient - fully agree with your statement but not current state AFAIK.

elharo:
Why does Maven add that layer? Can it simply remove it instead? If I had a greenfield, I'd just use java.util.logging and not allow any other log libraries within ten dependency chains of core, but if wishes were horses...

rmannibucau:
Fully agree JUL is way saner for products like maven but this is a long debate.
If we remove it we drop the `--fail-on-severity` flag so not sure we can in 4.x.

gnodet (approved):

gnodet:
@rmannibucau can this be closed ? [MNG-7724](http://issues.apache.org/jira/browse/MNG-7724) has been closed with https://github.com/apache/maven/pull/1051

jira-importer:
Resolve #9257

PR #1049, opened 2023-03-10
[MNG-7730] remove unused MavenXmlException class

elharo:


gnodet (approved):
No problem, it was supposed to be used at some point, but it's not...

PR #1050, opened 2023-03-10
[MNG-7729] remove unused code

elharo:
Now that I look at it, this class probably doesn't belong in this repo at all. It should be repackaged or removed.


gnodet:
Yes, that's the plan, as discussed on the mailing lists and on [plexus-utils](https://github.com/codehaus-plexus/plexus-utils/pull/236).
This PR is not necessary, as this class will be moved to plexus-utils(-xml) 4.x and consumed from maven.

elharo:
If it's going to be moved, it should be moved without these methods.

gnodet:
I'm closing this one, as the location is now irrelevant, the code has changed and those constants are needed for compatibility and are actually used....

PR #1051, opened 2023-03-11
[MNG-7724] don't warn about unsupported SLF4J configs; no one cares

elharo:
This is a more direct approach to avoiding the warnings. Just take them out. They're not actionable and no one cares about this.


rmannibucau (approved):

slachiewicz:
what if plugin use SLF4J 2 (not supported by Maven) - then we will just hide issue

rmannibucau:
@slachiewicz not differently than today, keep in mind it was already not working since it was using the broken logger setup to log, only actual useful logging was the printstacktrace on stderr (without logger) when the integration was not working which can't happen here since this init is done before the init so tempted to say we support your case dropping the code ;).

PR #1053, opened 2023-03-11
[MNG-7726] Maven 3.9.0 resolves properties incorrectly

cstamas:
There was a typo: the stream contains entries, so e.getKey and e.getValue is needed.

---

https://issues.apache.org/jira/browse/MNG-7726


cstamas:
Tested the issue provided reproducer, all works now as expected.

slawekjaranowski (approved):

jira-importer:
Resolve #9254

PR #1059, opened 2023-03-16
[MNG-5659] Provide project-specific settings located in ${session.rootdir}/.mvn/settings.xml by default

gnodet:
This add support for project specific settings using:
  * a default location of `${session.rootDirectory}/.mvn/settings.xml`
  * an alternate location specified by `-ps` or `--project-settings` and resolved against the current dir if provided with a relative path

Note that project settings have some limitations:
  * no `localRepository`
  * no `interactiveMode`
  * no `offline`
  * no `proxies`
  * no `server[x].username`, `server[x].password`, `server[x].privateKey`, `server[x].passphrase`, `server[x].filePermissions`, `server[x].directoryPermissions`
  * no `usePluginRegistry`

Simple IT provided with https://github.com/apache/maven-integration-testing/pull/265

===
Related discussions to be considered (but note that this PR does not depend on the following):

* [MNG-6762](https://issues.apache.org/jira/browse/MNG-6762) suggests to also modify other file resolution, but there was a strong opposition to modify the existing behaviour.  This PR is coherent with the current code where relative files are resolved against the current directory.

### DIFF
--- a/api/maven-api-core/src/main/java/org/apache/maven/api/services/SettingsBuilder.java
+++ b/api/maven-api-core/src/main/java/org/apache/maven/api/services/SettingsBuilder.java
@@ -48,7 +48,7 @@ public interface SettingsBuilder extends Service {
     @Nonnull
     default SettingsBuilderResult build(
             @Nonnull Session session, @Nonnull Source globalSettingsSource, @Nonnull Source userSettingsSource) {
-        return build(SettingsBuilderRequest.build(session, globalSettingsSource, userSettingsSource));
+        return build(session, globalSettingsSource, null, userSettingsSource);
     }
 
     /**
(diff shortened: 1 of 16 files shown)


rmannibucau:
Can `${maven.multiModuleProjectDirectory}/.mvn/settings.xml` become `${maven.multiModuleProjectDirectory}/.mvn/project.settings.xml` or something like that since `.mvn/settings.xml` is sometimes used by users so something more unlikely would be great.

cstamas (approved):
While I like this, I'd generalize the approach (maybe no now) in similar way like "chained LRM" works. Example:

`-s` remains as today "user settings override"
And "settings tail" is s1, s2, s3.... global setting (that also can be overridden from CLI). And this list is ordered from "most dominant" to "most recessive", meaning, that even settings builder could be modified to make a for-loop to build effective settings in this order:
"user settings", s1, s2, s3.... "global settings"

jira-importer:
Resolve #6464

PR #1061, opened 2023-03-16
[MNG-7038] Introduce public property to point to a root directory of (multi-module) project

gnodet:
JIRA issue: [MNG-7038] Introduce public property to point to a root directory of (multi-module) project

This PR introduces three properties:
  * `project.rootDirectory`: _the project's directory or parent directory containing a `.mvn` subdirectory or a `pom.xml` flagged with the `root="true"` attribute_. If no such directory can be found, accessing the rootDirectory property will throw an `IllegalStateException`.

  * `session.topDirectory` : _the directory of the topmost project being built, usually the current directory or the directory pointed at by the `-f`/`--file` command line argument_. The `topDirectory` is similar to the `executionRootDirectory` property available on the session, but renamed to make it coherent with the new `rootDirectory` and to avoid using _root_ in its name.  The `topDirectory` property is computed by the CLI as the directory pointed at by the `-f`/`--file` command line argument, or the current directory if there's no such argument. 

  * `session.rootDirectory` : _the rootDirectory for the topDirectory project_.

The `topDirectory` and `rootDirectory` properties are made available on the `MavenSession` / `Session` and deprecate the `executionRootDirectory` and `multiModuleProjectDirectory` properties.  The `rootDirectory` should never change for a given project and is thus made available for profile activation and model interpolation (without the `project.` prefix, similar to `basedir`).   The goal is also to make the `rootDirectory` property also available during [command line arguments interpolation](https://github.com/apache/maven/pull/1062).

A `root` boolean attribute is also added to the model to indicate that the project is the root project. This attribute is only supported if the _buildconsumer_ feature is active and removed before the pom is installed or deployed.  It can be used as an alternative mechanism to the `.mvn` directory.

### DIFF
--- a/api/maven-api-core/src/main/java/org/apache/maven/api/Project.java
+++ b/api/maven-api-core/src/main/java/org/apache/maven/api/Project.java
@@ -86,7 +86,44 @@ public interface Project {
         return getModel().getId();
     }
 
+    /**
+     * @deprecated use {@link #isTopProject()} instead
+     */
+    @Deprecated
     boolean isExecutionRoot();
+
+    /**
+     * Returns a boolean indicating if the project is the top level project for
+     * this reactor build.  The top level project may be different from the
+     * {@code rootDirectory}, especially if a subtree of the project is being
+     * built, either because Maven has been launched in a subdirectory or using
+     * a {@code -f} option.
+     *
+     * @return {@code true} if the project is the top level project for this build
+     */
+    boolean isTopProject();
+
+    /**
+     * Returns a boolean indicating if the project is a root project,
+     * meaning that the {@link #getRootDirectory()} and {@link #getBasedir()}
+     * points to the same directory, and that either {@link Model#isRoot()}
+     * is {@code true} or that {@code basedir} contains a {@code .mvn} child
+     * directory.
+     *
+     * @return {@code true} if the project is the root project
+     * @see Model#isRoot()
+     */
+    boolean isRootProject();
+
+    /**
+     * Gets the root directory of the project, which is the parent directory
+     * containing the {@code .mvn} directory or flagged with {@code root="true"}.
+     *
+     * @throws IllegalStateException if the root directory could not be found
+     * @see Session#getRootDirectory()
+     */
+    @Nonnull
+    Path getRootDirectory();
 
     @Nonnull
     Optional<Project> getParent();
--- a/maven-core/src/main/java/org/apache/maven/execution/DefaultMavenExecutionRequest.java
+++ b/maven-core/src/main/java/org/apache/maven/execution/DefaultMavenExecutionRequest.java
@@ -19,6 +19,7 @@
 package org.apache.maven.execution;
 
 import java.io.File;
+import java.nio.file.Path;
 import java.util.ArrayList;
 import java.util.Date;
 import java.util.HashMap;
@@ -30,6 +31,7 @@ import java.util.Properties;
 import org.apache.maven.artifact.repository.ArtifactRepository;
 import org.apache.maven.eventspy.internal.EventSpyDispatcher;
 import org.apache.maven.model.Profile;
+import org.apache.maven.model.root.RootLocator;
 import org.apache.maven.project.DefaultProjectBuildingRequest;
 import org.apache.maven.project.ProjectBuildingRequest;
 import org.apache.maven.properties.internal.SystemProperties;
@@ -101,6 +103,10 @@ public class DefaultMavenExecutionRequest implements MavenExecutionRequest {
     private File multiModuleProjectDirectory;
 
     private File basedir;
+
+    private Path rootDirectory;
+
+    private Path topDirectory;
 
     private List<String> goals;
 
@@ -1051,14 +1057,41 @@ public class DefaultMavenExecutionRequest implements MavenExecutionRequest {
         return this;
     }
 
+    @Deprecated
     @Override
     public void setMultiModuleProjectDirectory(File directory) {
         this.multiModuleProjectDirectory = directory;
     }
 
+    @Deprecated
     @Override
     public File getMultiModuleProjectDirectory() {
         return multiModuleProjectDirectory;
+    }
+
+    @Override
+    public Path getRootDirectory() {
+        if (rootDirectory == null) {
+            throw new IllegalStateException(RootLocator.UNABLE_TO_FIND_ROOT_PROJECT_MESSAGE);
+        }
+        return rootDirectory;
+    }
+
+    @Override
+    public MavenExecutionRequest setRootDirectory(Path rootDirectory) {
+        this.rootDirectory = rootDirectory;
+        return this;
+    }
+
+    @Override
+    public Path getTopDirectory() {
+        return topDirectory;
+    }
+
+    @Override
+    public MavenExecutionRequest setTopDirectory(Path topDirectory) {
+        this.topDirectory = topDirectory;
+        return this;
     }
 
     @Override
--- a/maven-core/src/main/java/org/apache/maven/execution/MavenExecutionRequest.java
+++ b/maven-core/src/main/java/org/apache/maven/execution/MavenExecutionRequest.java
@@ -19,6 +19,7 @@
 package org.apache.maven.execution;
 
 import java.io.File;
+import java.nio.file.Path;
 import java.util.Date;
 import java.util.List;
 import java.util.Map;
@@ -91,8 +92,17 @@ public interface MavenExecutionRequest {
     // ----------------------------------------------------------------------
 
     // Base directory
+
+    /**
+     * @deprecated use {@link #setTopDirectory(Path)} instead
+     */
+    @Deprecated
     MavenExecutionRequest setBaseDirectory(File basedir);
 
+    /**
+     * @deprecated use {@link #getTopDirectory()} instead
+     */
+    @Deprecated
     String getBaseDirectory();
 
     // Timing (remove this)
@@ -494,13 +504,50 @@ public interface MavenExecutionRequest {
 
     /**
      * @since 3.3.0
+     * @deprecated use {@link #setRootDirectory(Path)} instead
      */
+    @Deprecated
     void setMultiModuleProjectDirectory(File file);
 
     /**
      * @since 3.3.0
+     * @deprecated use {@link #getRootDirectory()} instead
      */
+    @Deprecated
     File getMultiModuleProjectDirectory();
+
+    /**
+     * Sets the top directory of the project.
+     *
+     * @since 4.0.0
+     */
+    MavenExecutionRequest setTopDirectory(Path topDirectory);
+
+    /**
+     * Gets the directory of the topmost project being built, usually the current directory or the
+     * directory pointed at by the {@code -f/--file} command line argument.
+     *
+     * @since 4.0.0
+     */
+    Path getTopDirectory();
+
+    /**
+     * Sets the root directory of the project.
+     *
+     * @since 4.0.0
+     */
+    MavenExecutionRequest setRootDirectory(Path rootDirectory);
+
+    /**
+     * Gets the root directory of the top project, which is the parent directory containing the {@code .mvn}
+     * directory or a {@code pom.xml} file with the {@code root="true"} attribute.
+     * If there's no such directory, an {@code IllegalStateException} will be thrown.
+     *
+     * @throws IllegalStateException if the root directory could not be found
+     * @see #getTopDirectory()
+     * @since 4.0.0
+     */
+    Path getRootDirectory();
 
     /**
      * @since 3.3.0
--- a/maven-core/src/main/java/org/apache/maven/execution/MavenSession.java
+++ b/maven-core/src/main/java/org/apache/maven/execution/MavenSession.java
@@ -19,6 +19,7 @@
 package org.apache.maven.execution;
 
 import java.io.File;
+import java.nio.file.Path;
 import java.util.Arrays;
 import java.util.Date;
 import java.util.List;
@@ -141,8 +142,28 @@ public class MavenSession implements Cloneable {
         return projects;
     }
 
+    /**
+     * @deprecated use {@link #getTopDirectory()} ()}
+     */
+    @Deprecated
     public String getExecutionRootDirectory() {
         return request.getBaseDirectory();
+    }
+
+    /**
+     * @see MavenExecutionRequest#getTopDirectory()
+     * @since 4.0.0
+     */
+    public Path getTopDirectory() {
+        return request.getTopDirectory();
+    }
+
+    /**
+     * @see MavenExecutionRequest#getRootDirectory()
+     * @since 4.0.0
+     */
+    public Path getRootDirectory() {
+        return request.getRootDirectory();
     }
 
     public MavenExecutionRequest getRequest() {
--- a/maven-core/src/main/java/org/apache/maven/project/MavenProject.java
+++ b/maven-core/src/main/java/org/apache/maven/project/MavenProject.java
@@ -21,6 +21,7 @@ package org.apache.maven.project;
 import java.io.File;
 import java.io.IOException;
 import java.io.Writer;
+import java.nio.file.Path;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.HashMap;
@@ -64,6 +65,7 @@ import org.apache.maven.model.Repository;
 import org.apache.maven.model.Resource;
 import org.apache.maven.model.Scm;
 import org.apache.maven.model.io.xpp3.MavenXpp3Writer;
+import org.apache.maven.model.root.RootLocator;
 import org.apache.maven.project.artifact.InvalidDependencyVersionException;
 import org.apache.maven.project.artifact.MavenMetadataSource;
 import org.codehaus.plexus.classworlds.realm.ClassRealm;
@@ -104,6 +106,8 @@ public class MavenProject implements Cloneable {
     private File file;
 
     private File basedir;
+
+    private Path rootDirectory;
 
     private Set<Artifact> resolvedArtifacts;
 
@@ -1678,5 +1682,21 @@ public class MavenProject implements Cloneable {
     @Deprecated
     public void setProjectBuildingRequest(ProjectBuildingRequest projectBuildingRequest) {
         this.projectBuilderConfiguration = projectBuildingRequest;
+    }
+
+    /**
+     * @since 4.0.0
+     * @return the rootDirectory for this project
+     * @throws IllegalStateException if the rootDirectory cannot be found
+     */
+    public Path getRootDirectory() {
+        if (rootDirectory == null) {
+            throw new IllegalStateException(RootLocator.UNABLE_TO_FIND_ROOT_PROJECT_MESSAGE);
+        }
+        return rootDirectory;
+    }
+
+    public void setRootDirectory(Path rootDirectory) {
+        this.rootDirectory = rootDirectory;
     }
 }
--- a/maven-embedder/src/main/java/org/apache/maven/cli/CLIReportingUtils.java
+++ b/maven-embedder/src/main/java/org/apache/maven/cli/CLIReportingUtils.java
@@ -135,6 +135,13 @@ public final class CLIReportingUtils {
     }
 
     public static void showError(Logger logger, String message, Throwable e, boolean showStackTrace) {
+        if (logger == null) {
+            System.err.println(message);
+            if (showStackTrace && e != null) {
+                e.printStackTrace(System.err);
+            }
+            return;
+        }
         if (showStackTrace) {
             logger.error(message, e);
         } else {
--- a/maven-embedder/src/main/java/org/apache/maven/cli/MavenCli.java
+++ b/maven-embedder/src/main/java/org/apache/maven/cli/MavenCli.java
@@ -29,6 +29,8 @@ import java.io.InputStream;
 import java.io.PrintStream;
 import java.nio.charset.Charset;
 import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.Paths;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.HashSet;
@@ -37,6 +39,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.Map.Entry;
 import java.util.Properties;
+import java.util.ServiceLoader;
 import java.util.Set;
 import java.util.function.Consumer;
 import java.util.regex.Matcher;
@@ -87,6 +90,7 @@ import org.apache.maven.lifecycle.LifecycleExecutionException;
 import org.apache.maven.logwrapper.LogLevelRecorder;
 import org.apache.maven.logwrapper.MavenSlf4jWrapperFactory;
 import org.apache.maven.model.building.ModelProcessor;
+import org.apache.maven.model.root.RootLocator;
 import org.apache.maven.project.MavenProject;
 import org.apache.maven.properties.internal.EnvironmentUtils;
 import org.apache.maven.properties.internal.SystemProperties;
@@ -319,6 +323,54 @@ public class MavenCli {
                 cliRequest.multiModuleProjectDirectory = basedir.getAbsoluteFile();
             }
         }
+
+        // We need to locate the top level project which may be pointed at using
+        // the -f/--file option.  However, the command line isn't parsed yet, so
+        // we need to iterate through the args to find it and act upon it.
+        Path topDirectory = Paths.get(cliRequest.workingDirectory);
+        boolean isAltFile = false;
+        for (String arg : cliRequest.args) {
+            if (isAltFile) {
+                // this is the argument following -f/--file
+                Path path = Paths.get(arg);
+                if (Files.isDirectory(path)) {
+                    topDirectory = path;
+                } else if (Files.isRegularFile(topDirectory)) {
+                    topDirectory = path.getParent();
+                    if (!Files.isDirectory(topDirectory)) {
+                        System.err.println("Directory " + topDirectory
+                                + " extracted from the -f/--file command-line argument " + arg + " does not exist");
+                        throw new ExitException(1);
+                    }
+                } else {
+                    System.err.println(
+                            "POM file " + arg + " specified with the -f/--file command line argument does not exist");
+                    throw new ExitException(1);
+                }
+                break;
+            } else {
+                // Check if this is the -f/--file option
+                isAltFile = arg.equals(String.valueOf(CLIManager.ALTERNATE_POM_FILE)) || arg.equals("file");
+            }
+        }
+        try {
+            topDirectory = topDirectory.toAbsolutePath().toRealPath();
+        } catch (IOException e) {
+            System.err.println("Error computing real path from " + topDirectory);
+            throw new ExitException(1);
+        }
+        cliRequest.topDirectory = topDirectory;
+        // We're very early in the process and we don't have the container set up yet,
+        // so we rely on the JDK services to eventually lookup a custom RootLocator.
+        // This is used to compute {@code session.rootDirectory} but all {@code project.rootDirectory}
+        // properties will be compute through the RootLocator found in the container.
+        RootLocator rootLocator =
+                ServiceLoader.load(RootLocator.class).iterator().next();
+        Path rootDirectory = rootLocator.findRoot(topDirectory);
+        if (rootDirectory == null) {
+            System.err.println(RootLocator.UNABLE_TO_FIND_ROOT_PROJECT_MESSAGE);
+        }
+        cliRequest.rootDirectory = rootDirectory;
 
         //
         // Make sure the Maven home directory is an absolute path to save us from confusion with say drive-relative
@@ -1185,6 +1237,8 @@ public class MavenCli {
         request.setSystemProperties(cliRequest.systemProperties);
         request.setUserProperties(cliRequest.userProperties);
         request.setMultiModuleProjectDirectory(cliRequest.multiModuleProjectDirectory);
+        request.setRootDirectory(cliRequest.rootDirectory);
+        request.setTopDirectory(cliRequest.topDirectory);
         request.setPom(determinePom(commandLine, workingDirectory, baseDirectory));
         request.setTransferListener(determineTransferListener(quiet, verbose, commandLine, request));
         request.setExecutionListener(determineExecutionListener());
--- a/maven-model-builder/src/main/java/org/apache/maven/model/interpolation/AbstractStringBasedModelInterpolator.java
+++ b/maven-model-builder/src/main/java/org/apache/maven/model/interpolation/AbstractStringBasedModelInterpolator.java
@@ -21,6 +21,7 @@ package org.apache.maven.model.interpolation;
 import javax.inject.Inject;
 
 import java.io.File;
+import java.nio.file.Path;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Collections;
@@ -33,6 +34,7 @@ import org.apache.maven.model.building.ModelBuildingRequest;
 import org.apache.maven.model.building.ModelProblemCollector;
 import org.apache.maven.model.path.PathTranslator;
 import org.apache.maven.model.path.UrlNormalizer;
+import org.apache.maven.model.root.RootLocator;
 import org.codehaus.plexus.interpolation.AbstractValueSource;
 import org.codehaus.plexus.interpolation.InterpolationPostProcessor;
 import org.codehaus.plexus.interpolation.MapBasedValueSource;
@@ -74,10 +76,14 @@ public abstract class AbstractStringBasedModelInterpolator implements ModelInter
     private final PathTranslator pathTranslator;
     private final UrlNormalizer urlNormalizer;
 
+    private final RootLocator rootLocator;
+
     @Inject
-    public AbstractStringBasedModelInterpolator(PathTranslator pathTranslator, UrlNormalizer urlNormalizer) {
+    public AbstractStringBasedModelInterpolator(
+            PathTranslator pathTranslator, UrlNormalizer urlNormalizer, RootLocator rootLocator) {
         this.pathTranslator = pathTranslator;
         this.urlNormalizer = urlNormalizer;
+        this.rootLocator = rootLocator;
     }
 
     @Override
@@ -132,6 +138,20 @@ public abstract class AbstractStringBasedModelInterpolator implements ModelInter
             valueSources.add(baseUriValueSource);
             valueSources.add(new BuildTimestampValueSource(config.getBuildStartTime(), modelProperties));
         }
+
+        valueSources.add(new PrefixedValueSourceWrapper(
+                new AbstractValueSource(false) {
+                    @Override
+                    public Object getValue(String expression) {
+                        if ("rootDirectory".equals(expression)) {
+                            Path base = projectDir != null ? projectDir.toPath() : null;
+                            Path root = rootLocator.findMandatoryRoot(base);
+                            return root.toFile().getPath();
+                        }
+                        return null;
+                    }
+                },
+                PROJECT_PREFIXES));
 
         valueSources.add(projectPrefixValueSource);
 
--- a/maven-model-builder/src/main/java/org/apache/maven/model/path/ProfileActivationFilePathInterpolator.java
+++ b/maven-model-builder/src/main/java/org/apache/maven/model/path/ProfileActivationFilePathInterpolator.java
@@ -23,9 +23,11 @@ import javax.inject.Named;
 import javax.inject.Singleton;
 
 import java.io.File;
+import java.nio.file.Path;
 
 import org.apache.maven.api.model.ActivationFile;
 import org.apache.maven.model.profile.ProfileActivationContext;
+import org.apache.maven.model.root.RootLocator;
 import org.codehaus.plexus.interpolation.AbstractValueSource;
 import org.codehaus.plexus.interpolation.InterpolationException;
 import org.codehaus.plexus.interpolation.MapBasedValueSource;
@@ -42,9 +44,12 @@ public class ProfileActivationFilePathInterpolator {
 
     private final PathTranslator pathTranslator;
 
+    private final RootLocator rootLocator;
+
     @Inject
-    public ProfileActivationFilePathInterpolator(PathTranslator pathTranslator) {
+    public ProfileActivationFilePathInterpolator(PathTranslator pathTranslator, RootLocator rootLocator) {
         this.pathTranslator = pathTranslator;
+        this.rootLocator = rootLocator;
     }
 
     /**
@@ -78,6 +83,18 @@ public class ProfileActivationFilePathInterpolator {
         } else if (path.contains("${basedir}")) {
             return null;
         }
+
+        interpolator.addValueSource(new AbstractValueSource(false) {
+            @Override
+            public Object getValue(String expression) {
+                if ("rootDirectory".equals(expression)) {
+                    Path base = basedir != null ? basedir.toPath() : null;
+                    Path root = rootLocator.findMandatoryRoot(base);
+                    return root.toFile().getAbsolutePath();
+                }
+                return null;
+            }
+        });
 
         interpolator.addValueSource(new MapBasedValueSource(context.getProjectProperties()));
 
--- a/maven-model-builder/src/main/java/org/apache/maven/model/root/DefaultRootLocator.java
+++ b/maven-model-builder/src/main/java/org/apache/maven/model/root/DefaultRootLocator.java
@@ -1,0 +1,59 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.maven.model.root;
+
+import javax.inject.Named;
+
+import java.io.IOException;
+import java.io.InputStream;
+import java.nio.file.Files;
+import java.nio.file.Path;
+
+import org.codehaus.plexus.util.xml.pull.MXParser;
+import org.codehaus.plexus.util.xml.pull.XmlPullParserException;
+
+@Named
+public class DefaultRootLocator implements RootLocator {
+
+    public boolean isRootDirectory(Path dir) {
+        if (Files.isDirectory(dir.resolve(".mvn"))) {
+            return true;
+        }
+        // we're too early to use the modelProcessor ...
+        Path pom = dir.resolve("pom.xml");
+        try (InputStream is = Files.newInputStream(pom)) {
+            MXParser parser = new MXParser();
+            parser.setInput(is, null);
+            if (parser.nextTag() == MXParser.START_TAG && parser.getName().equals("project")) {
+                for (int i = 0; i < parser.getAttributeCount(); i++) {
+                    if ("root".equals(parser.getAttributeName(i))) {
+                        return Boolean.parseBoolean(parser.getAttributeValue(i));
+                    }
+                }
+            }
+        } catch (IOException | XmlPullParserException e) {
+            // The root locator can be used very early during the setup of Maven,
+            // even before the arguments from the command line are parsed.  Any exception
+            // that would happen here should cause the build to fail at a later stage
+            // (when actually parsing the POM) and will lead to a better exception being
+            // displayed to the user, so just bail out and return false.
+        }
+        return false;
+    }
+}
--- a/maven-model-builder/src/main/java/org/apache/maven/model/root/RootLocator.java
+++ b/maven-model-builder/src/main/java/org/apache/maven/model/root/RootLocator.java
@@ -1,0 +1,69 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.maven.model.root;
+
+import java.nio.file.Path;
+
+import org.apache.maven.api.annotations.Nonnull;
+import org.apache.maven.api.annotations.Nullable;
+
+/**
+ * Interface used to locate the root directory for a given project.
+ *
+ * The root locator is usually looked up from the plexus container.
+ * One notable exception is the computation of the early {@code session.rootDirectory}
+ * property which happens very early.  The implementation used in this case
+ * will be discovered using the JDK service mechanism.
+ *
+ * The default implementation will look for a {@code .mvn} child directory
+ * or a {@code pom.xml} containing the {@code root="true"} attribute.
+ *
+ * @see DefaultRootLocator
+ */
+public interface RootLocator {
+
+    String UNABLE_TO_FIND_ROOT_PROJECT_MESSAGE = "Unable to find the root directory. "
+            + "Create a .mvn directory in the root directory or add the root=\"true\""
+            + " attribute on the root project's model to identify it.";
+
+    @Nonnull
+    default Path findMandatoryRoot(Path basedir) {
+        Path rootDirectory = findRoot(basedir);
+        if (rootDirectory == null) {
+            throw new IllegalStateException(getNoRootMessage());
+        }
+        return rootDirectory;
+    }
+
+    @Nullable
+    default Path findRoot(Path basedir) {
+        Path rootDirectory = basedir;
+        while (rootDirectory != null && !isRootDirectory(rootDirectory)) {
+            rootDirectory = rootDirectory.getParent();
+        }
+        return rootDirectory;
+    }
+
+    @Nonnull
+    default String getNoRootMessage() {
+        return UNABLE_TO_FIND_ROOT_PROJECT_MESSAGE;
+    }
+
+    boolean isRootDirectory(Path dir);
+}
--- a/maven-model-builder/src/site/apt/index.apt
+++ b/maven-model-builder/src/site/apt/index.apt
@@ -41,7 +41,7 @@ Maven Model Builder
 
    ** profile activation: see {{{./apidocs/org/apache/maven/model/profile/activation/package-summary.html}available activators}}.
    Notice that model interpolation hasn't happened yet, then interpolation for file-based activation is limited to
-   <<<$\{basedir}>>> (since Maven 3), system properties and user properties
+   <<<$\{basedir}>>> (since Maven 3), <<<$\{rootDirectory}>>> (since Maven 4), system properties and user properties
 
    ** file model validation: <<<ModelValidator>>> ({{{./apidocs/org/apache/maven/model/validation/ModelValidator.html}javadoc}}),
    with its <<<DefaultModelValidator>>> implementation
@@ -51,7 +51,7 @@ Maven Model Builder
 
  * phase 2, with optional plugin processing
 
-   ** Build up a raw model by re-reading the file and enrich it based on information available in the reactor. Some features:
+   ** Build up a raw model by re-reading the file and enriching it based on information available in the reactor. Some features:
 
       *** Resolve version of versionless parents based on relativePath (including ci-friendly versions)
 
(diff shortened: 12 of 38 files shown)


maven-embedder/src/main/java/org/apache/maven/cli/CLIReportingUtils.java
```diff
@@ -136,6 +136,13 @@ static Properties getBuildProperties() {
     }
 
     public static void showError(Logger logger, String message, Throwable e, boolean showStackTrace) {
+        if (logger == null) {
```
michael-o, line 139: How can the logger be null here?

gnodet: If `showError` is called before the logger is initialised...

michael-o:
Do these three properties represent the same value?

gnodet:
> Do these three properties represent the same value?

I suppose you're talking about `topdir`, `project.topdir` and `session.topdir` ? They are aliases, but I don't think the `project.topdir` is a good idea, since it's not really a project attribute.

I'm still thinking about renaming the new property to `rootdir`, and introduce a `topdir` for the execution directory in a consistent way.  And deprecate `MavenSession.getExecutionRootDirectory()`, `MavenExecutionRequest.getBasedir()`, `MavenExecutionRequest.getMultiModuleProjectDirectory()` accordingly.  I'm open to any naming, but things are a bit inconsistent, so I'd rather fix it.

w6et:
> > Do these three properties represent the same value?
> 
> I suppose you're talking about `topdir`, `project.topdir` and `session.topdir` ? They are aliases, but I don't think the `project.topdir` is a good idea, since it's not really a project attribute.
> 
> I'm still thinking about renaming the new property to `rootdir`, and introduce a `topdir` for the execution directory in a consistent way. And deprecate `MavenSession.getExecutionRootDirectory()`, `MavenExecutionRequest.getBasedir()`, `MavenExecutionRequest.getMultiModuleProjectDirectory()` accordingly. I'm open to any naming, but things are a bit inconsistent, so I'd rather fix it.

just another idea，Maybe we can define build.json(like npm's package.json)，we can init topdir properties and other public properties in the build.json，then generate build-lock.json\yaml, every module can use the properties from build-lock.json, (BTW, maybe pom.xml->pom-lock.yaml，so  only use xxx-lock.yaml，so something don't need computed many times)
maybe  dependencies、plugins、repositories  ‘s properties can split to different properties file，so we can import & share it for different  maven project. or all in build.json like 
`{
  "name": "my app 001",
  "version": "0.0.1",
  "topdir": "xxxxx",
  "mvn.properties": {
	"java.version":"17",
	"maven.test.skip":"true"
  },
  "dependencies": {
    "compile": {
		"default":{
			"default is required":""
		},
		"spring":{
			"spring-amqp.version":"3.0.2"
		}
	}
  },
  "plugins": {

  },
  "repositories": {

  }
}
`

w6et:
> > > Do these three properties represent the same value?
> > 
> > 
> > I suppose you're talking about `topdir`, `project.topdir` and `session.topdir` ? They are aliases, but I don't think the `project.topdir` is a good idea, since it's not really a project attribute.
> > I'm still thinking about renaming the new property to `rootdir`, and introduce a `topdir` for the execution directory in a consistent way. And deprecate `MavenSession.getExecutionRootDirectory()`, `MavenExecutionRequest.getBasedir()`, `MavenExecutionRequest.getMultiModuleProjectDirectory()` accordingly. I'm open to any naming, but things are a bit inconsistent, so I'd rather fix it.
> 
> just another idea，Maybe we can define build.json(like npm's package.json)，we can init topdir properties and other public properties in the build.json，then generate build-lock.json\yaml, every module can use the properties from build-lock.json, (BTW, maybe pom.xml->pom-lock.yaml，so only use xxx-lock.yaml，so something don't need computed many times) maybe dependencies、plugins、repositories ‘s properties can split to different properties file，so we can import & share it for different maven project. or all in build.json like `{ "name": "my app 001", "version": "0.0.1", "topdir": "xxxxx", "mvn.properties": { "java.version":"17", "maven.test.skip":"true" }, "dependencies": { "compile": { "default":{ "default is required":"" }, "spring":{ "spring-amqp.version":"3.0.2" } } }, "plugins": {
> 
> }, "repositories": {
> 
> } } `

generated build-lock.json can contains system environment properties or something else configuration ，so we can sure the actual properties/configuration any which we used

michael-o:
Before we agree on new names, we need to agree what these properties should contain and the expectations/requirements from users.

gnodet:
> Before we agree on new names, we need to agree what these properties should contain and the expectations/requirements from users.

@michael-o  so I propose the following:

* `topdir` : _the directory of the top-most project being built_, this is usually the current directory, or the one pointed by the `-f`/`--file` option

* `rootdir` : _the parent directory containing a child `.mvn` directory_.  If there's no such directory, print a warning and use a `topdir`

This needs to be clarified when using `-pl`/`-am`/`-amd` options.

rmannibucau:
Think rootdir must be the first module of the hierarchy (child to parent order) without a parent in the same relativepath (not null) or parent dir and listing the child. Would be bad to require .mvn presence imho.

For me topdir is useless since we already have it somehow .

gnodet:
> Think rootdir must be the first module of the hierarchy (child to parent order) without a parent in the same relativepath (not null) or parent dir and listing the child. Would be bad to require .mvn presence imho.

The main use case for the `rootdir` is for build related files location, for example, locate files shared throughout the build.  This is currently a bit difficult to use.  I have not seen any real use case for the parent/child relationship.

> For me topdir is useless since we already have it somehow .

We do, it's called `executionRootDir`.

rmannibucau:
@gnodet sharing the use case, just adding that rootdir must not require .mvn dir presence as a marker so must use the pom hierarxhy and take the highest parent.

gnodet:
> @gnodet sharing the use case, just adding that rootdir must not require .mvn dir presence as a marker so must use the pom hierarchy and take the highest parent, was just detailling a bit the impl idea.

I think that would require another property,  I'm all for defining additional properties. But a property based on parent/child relationship would be made available after projects are built, so it would not be able to be used during interpolation imho.

rmannibucau:
@gnodet guess worse case we can do 2 passes since parent/child relationship does not support interpolation properly anyway by design. Point is that the .mvn requirement is high for a so often needed prop so 1. I'd live to not have it (technically it is cleaner too even if .mvn does not prevent to "stop bubbling") and 2. if kept like that it must be named mvn.mvn.rootdir and be made it obvious it must not be used in parent poms, boms and plugins probably (indeed i hope we avoid that since tech we dont need .mvn presence in 99% of the cases).

gnodet:
> @gnodet guess worse case we can do 2 passes since parent/child relationship does not support interpolation properly anyway by design. Point is that the .mvn requirement is high for a so often needed prop so 1. I'd live to not have it (technically it is cleaner too even if .mvn does not prevent to "stop bubbling") and 2. if kept like that it must be named mvn.mvn.rootdir and be made it obvious it must not be used in parent poms, boms and plugins probably (indeed i hope we avoid that since tech we dont need .mvn presence in 99% of the cases).

The goal is to make have a  public property that can be used, not to hide it.  We already have the hidden `multiModuleProjectDirectory` which is hidden.   What's the point of having it hidden ?

rmannibucau:
I'd be to make it public but also not controlled by scripts to always be accurate.

gnodet:
> I'd be to make it public but also not controlled by scripts to always be accurate.

I have no problem with that.    The point is to have a proper definition and thus computation for it.  Having a `.mvn` directory gives a clear and simple definition and computation. Not having it may lead to go outside the source tree (if the project is part of a bigger project) or unable to be detected (in which case the `multiModuleProjectDirectory` defaults to the current dir).

gnodet:
> I'd be to make it public but also not controlled by scripts to always be accurate.

Fwiw, that will mean that the computation needs to be duplicated somehow, so that the maven launch script can locate the `.mvn/jvm.config` file and read it before starting the JVM...

rmannibucau:
> Fwiw, that will mean that the computation needs to be duplicated somehow, so that the maven launch script can locate the .mvn/jvm.config file and read it before starting the JVM...

Well, we still have issues with `jvm.config` (spaces, quotes, EOL, ...) in terms of parsing so wonder if we shouldn't revisit it anyway. I had in mind a small native binary (either using graalvm without http and all the options making it fat or whatever is fast to launch) which could handle it to actually fork (unix)/trigger the real execution but, IMHO, it must belong to maven and not any enclosing tooling otherwise it is not part of maven and lost easily in any integration which must mimic our fragile script.

gnodet:
> > Fwiw, that will mean that the computation needs to be duplicated somehow, so that the maven launch script can locate the .mvn/jvm.config file and read it before starting the JVM...
> 
> Well, we still have issues with `jvm.config` (spaces, quotes, EOL, ...) in terms of parsing so wonder if we shouldn't revisit it anyway. I had in mind a small native binary (either using graalvm without http and all the options making it fat or whatever is fast to launch) which could handle it to actually fork (unix)/trigger the real execution but, IMHO, it must belong to maven and not any enclosing tooling otherwise it is not part of maven and lost easily in any integration which must mimic our fragile script.

That looks like `mvnd` but without keeping daemons alive...

rmannibucau:
> That looks like mvnd but without keeping daemons alive...

Not really since mvnd is a full mvn process whereas here it would just be a launcher responsible to handle the actual java command line to respect properly jvm.config. Agree the resolution can bring some resolver deps but shouldn't be as complete as mvnd.

Even if I agree with the overall statement, anything in between looks wrong to me since it means topdir will not be usable until you have a .mvn (which is not the most common thing), we'll keep broken properties for this very common need and we keep a broken jvm.config so from my window everything converges to get a launcher to maven if we want these features.

gnodet:
> > That looks like mvnd but without keeping daemons alive...
> 
> Not really since mvnd is a full mvn process whereas here it would just be a launcher responsible to handle the actual java command line to respect properly jvm.config. 

That's a wrong assumption.  Mvnd has a client which boots a JVM (the daemon) and which is compiled to native with graalvm.  The client does not load maven.

michael-o:
> @gnodet right but it has the network stack which makes the binary "fat" compared to a vanilla main compiled with graalvm but let's put graal aside and agree on the fact we need a light (not java sadly) launcher to run properly and portably maven. Then topdir feature comes into this module and is well defined compared to all alternatives we have which are all partial.

I have been thinking for years to write a C based launcher for Maven similar to Eclipse, but never found the time.

michael-o (changes requested):

maven-model-builder/src/site/apt/index.apt
```diff
@@ -192,6 +192,10 @@ Maven Model Builder
 *----+------+------+
 | <<<maven.home>>> | The path to the current Maven home. | <<<$\{maven.home\}>>> |
 *----+------+------+
+| <<<maven.rootdir>>> | the directory containing the root <<<pom.xml>>> file of a multi module project, in a single module project this is the same as <<<project.basedir>>> | <<<$\{maven.rootdir\}>>> |
+*----+------+------+
+| <<<maven.topdir>>> | the directory containing the top-level <<<pom.xml>>> in this reactor build | <<<$\{maven.topdir\}>>> |
+*----+------+------+
```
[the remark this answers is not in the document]
gnodet: Actually, `topdir` should not be part of any model interpolation, because it's mainly the current directory and should definitely not have any effect.  This will require an additional parameter to `ModelInterpolator.interpolate()`  or to add it to the `ModelBuildingRequest`. 

However, `rootdir` should be part of profile activation along `basedir` imho.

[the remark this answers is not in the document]
rmannibucau: ? `topdir` should be available in interpolation, it is its main usage to reference absolutely configurations otherwise not sure its interest for end users.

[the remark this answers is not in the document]
gnodet: > ? `topdir` should be available in interpolation, it is its main usage to reference absolutely configurations otherwise not sure its interest for end users.

No, that's would be the `rootdir`.  The `topdir` would be defined as the the top-most directory of executed projects in the reactor (basically, the current directory or the one pointed to when using `-f`/`--file`).

api/maven-api-core/src/main/java/org/apache/maven/api/Project.java
```diff
@@ -86,8 +86,34 @@ default String getId() {
...
     }
 
+    /**
+     * @deprecated use {@link #isTopdirProject()} instead
+     */
+    @Deprecated
     boolean isExecutionRoot();
 
+    /**
+     * Returns a boolean indicating if the project is the top leve project for
```
michael-o, line 96: level

api/maven-api-core/src/main/java/org/apache/maven/api/Project.java
```diff
@@ -86,8 +86,34 @@ default String getId() {
...
+     * this reactor build.  The top level project may be different from the
+     * {@code rootDirProject}, especially if a subtree of the project is being
+     * built, either because maven has been launched in a subdirectory or using
```
michael-o, line 99: Maven

maven-model-builder/src/site/apt/index.apt
```diff
@@ -190,6 +190,10 @@ Maven Model Builder
 *----+------+------+
 | <<<*>>> | model properties, such as project properties set in the pom | <<<$\{any.key\}>>> |
 *----+------+------+
+| <<<session.rootdir>>> | the directory containing the root <<<pom.xml>>> file of a multi module project, in a single module project this is the same as <<<project.basedir>>> | <<<$\{session.rootdir\}>>> |
+*----+------+------+
+| <<<session.topdir>>> | the directory containing the top-level <<<pom.xml>>> in this reactor build | <<<$\{session.topdir\}>>> |
+*----+------+------+
```
michael-o, line 196: Same concern as above.

michael-o (changes requested):

maven-core/src/main/java/org/apache/maven/execution/MavenExecutionRequest.java
```diff
@@ -494,14 +504,49 @@ public interface MavenExecutionRequest {
...
+     */
+    MavenExecutionRequest setRootdir(Path rootdir);
+
+    /**
+     * Gets the root dir of the project, which is the up-most directory of the reactor, usually containing
+     * the {@code .mvn} directory.
+     *
+     * @since 4.0.0
+     */
+    Path getRootdir();
```
cstamas, line 532: This getter may fail (lazily) if invoked by there is no `.mvn` directory present (the value cannot be calculated)?

maven-model-builder/src/site/apt/index.apt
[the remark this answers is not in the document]
rmannibucau: what's the point of topdir then? it does not have to match cwd nor the rootdir so sounds like a randomdir at usage

maven-core/src/main/java/org/apache/maven/execution/MavenExecutionRequest.java
gnodet: If there's no `.mvn` directory, the value currently defaults to `topdir`, i.e. the current directory or the one pointed to with the `-f`/`--file` option. 
@cstamas we could change the behaviour to: print a warning, default to `null`, throw an exception when accessed using a getter (i.e. from plugins, or when injected / interpolated).  Is that what you have in mind ?

maven-embedder/src/main/java/org/apache/maven/cli/MavenCli.java
```diff
@@ -321,6 +323,46 @@ void initialize(CliRequest cliRequest) throws ExitException {
             }
         }
 
+        Path topdir = Paths.get(cliRequest.workingDirectory);
+        boolean isAltFile = false;
+        for (String arg : cliRequest.args) {
+            if (isAltFile) {
+                Path path = Paths.get(arg);
+                if (Files.isDirectory(path)) {
```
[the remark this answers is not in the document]
gnodet: The `commandLine` variable is initialised in a following step.  Also, if we want to allow some interpolation in the command line arguments, we definitely need to compute this variable early.

maven-core/src/main/java/org/apache/maven/plugin/PluginParameterExpressionEvaluator.java
```diff
@@ -165,7 +165,11 @@ public Object evaluate(String expr, Class<?> type) throws ExpressionEvaluationEx
 
         MojoDescriptor mojoDescriptor = mojoExecution.getMojoDescriptor();
 
-        if ("localRepository".equals(expression)) {
+        if ("rootdir".equals(expression) || "session.rootdir".equals(expression)) {
+            value = session.getRootdir();
+        } else if ("topdir".equals(expression) || "session.topdir".equals(expression)) {
+            value = session.getTopdir();
```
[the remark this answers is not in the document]
gnodet: That's true, if we don't support the non prefixed expression, we can simply use the code which follows and which should handle `session.rootdir` and `session.topdir` perfectly.  Would you prefer that ?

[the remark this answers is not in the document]
gnodet: @michael-o  fwiw, the addition of `rootdir` and `topdir` was done to be in line with `basedir`, so maybe we should deprecate that one ?

maven-core/src/main/java/org/apache/maven/execution/DefaultMavenExecutionRequest.java
```diff
@@ -1051,16 +1057,44 @@ public MavenExecutionRequest setToolchains(Map<String, List<ToolchainModel>> too
...
+    @Deprecated
     @Override
     public File getMultiModuleProjectDirectory() {
         return multiModuleProjectDirectory;
     }
 
+    @Override
+    public Path getRootdir() {
+        if (rootdir == null) {
+            throw new IllegalArgumentException("Unable to find the root directory. "
```
michael-o, line 1075: This needs to be documented since it looks like a hard requirement now.

maven-core/src/main/java/org/apache/maven/execution/MavenExecutionRequest.java
```diff
@@ -494,14 +504,50 @@ public interface MavenExecutionRequest {
...
+     *
+     * @since 4.0.0
+     */
+    MavenExecutionRequest setRootdir(Path rootdir);
+
+    /**
+     * Gets the root directory of the project, which is the parent directory containing the {@code .mvn} directory.
+     * If there's no such file, an {@code IllegalArgumentException} will be thrown.
+     *
+     * @throws IllegalArgumentException if the root directory could not be found
```
michael-o, line 530: Isn't this `IllegalStateException`?

maven-model-builder/src/site/apt/index.apt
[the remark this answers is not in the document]
gnodet: `topdir` is not really new, it's mostly a rename of the `executionRootDir` property to make it more coherent with `basedir` and `rootdir`, else we end up with `root directory` and `execution root directory` for different things, and that will definitely lead to problems imho.

maven-embedder/src/main/java/org/apache/maven/cli/MavenCli.java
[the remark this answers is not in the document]
michael-o: Granted.

maven-model-builder/src/main/java/org/apache/maven/model/interpolation/AbstractStringBasedModelInterpolator.java
```diff
@@ -133,6 +134,21 @@ public Object getValue(String expression) {
...
 
+        valueSources.add(new AbstractValueSource(false) {
+            @Override
+            public Object getValue(String expression) {
+                if ("rootdir".equals(expression)) {
+                    Path path = config.getRootdir();
+                    if (path != null) {
+                        return path.toString();
+                    }
+                    throw new IllegalArgumentException("Unable to find the root directory. "
```
michael-o, line 145: ISE rather? The expression isn't invalid but the state within, no?

maven-model-builder/src/main/java/org/apache/maven/model/path/ProfileActivationFilePathInterpolator.java
```diff
@@ -79,6 +80,25 @@ public Object getValue(String expression) {
...
+                /*
+                 * We intentionally only support ${rootdir} and not ${session.rootdir} as the latter form
+                 * would suggest that other session.* expressions can be used which is beyond the design.
+                 */
+                if ("rootdir".equals(expression)) {
+                    Path path = context.getRootdir();
+                    if (path != null) {
+                        return path.toString();
+                    }
+                    throw new IllegalArgumentException("Unable to find the root directory. "
```
michael-o, line 95: Same here

maven-core/src/main/java/org/apache/maven/execution/DefaultMavenExecutionRequest.java
gnodet: The goal is to throw an exception when actually used.  So it's not a hard requirement unless you actually use it somehow in a plugin or in a model, in which case, yes, it becomes a hard requirement.

michael-o (changes requested):

maven-core/src/main/java/org/apache/maven/execution/DefaultMavenExecutionRequest.java
michael-o: But rather fail early than after 30 min of a build?

maven-core/src/main/java/org/apache/maven/plugin/PluginParameterExpressionEvaluator.java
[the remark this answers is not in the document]
michael-o: How does this then collide with `${project.basedir}`? The devs should be able to acces the current module project basedir. This should not change.

maven-core/src/main/java/org/apache/maven/execution/DefaultMavenExecutionRequest.java
gnodet: But we can't know early if someone uses the `${session.rootdir}`or `${rootdir}` in the model, profile activation or a plugin configuration.  If we fail early, this will mean that all projects will be required to add this `.mvn` direction.  Note that we already have a warning if this directory does not exist.

maven-model-builder/src/main/java/org/apache/maven/model/interpolation/AbstractStringBasedModelInterpolator.java
gnodet: Agreed, I'll fix them all.

maven-core/src/main/java/org/apache/maven/execution/DefaultMavenExecutionRequest.java
michael-o: Accepted

maven-core/src/main/java/org/apache/maven/plugin/PluginParameterExpressionEvaluator.java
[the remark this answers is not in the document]
michael-o: > That's true, if we don't support the non prefixed expression, we can simply use the code which follows and which should handle `session.rootdir` and `session.topdir` perfectly. Would you prefer that ?

Yes, strive for consistency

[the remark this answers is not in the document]
gnodet: > How does this then collide with `${project.basedir}`? The devs should be able to acces the current module project basedir. This should not change.

Yes, I meant to deprecate `${basedir}` and favor `${project.basedir}`.

rmannibucau:
I'm very confused by this PR, what I see:

1. we had 2 dir properties and they are deprecated in favor of 2 dir properties, can't we fix the previous ones?
2. the root dir is defined by the `.mvn` presence which is still not mainstream so will likely always fail instead of the top dir of linked project by the parent/child relationship defined by pom (which always works)

So overall think the impl should be fixed (2) and we should probably collapse most of the properties we have, in particular for v4 where it will just be a clean up and avoid the redefinition of new props.
The big pro of root dir is the ability to define absolutely a file config (style.xml is what i have in mind but there are multiple cases) so should absolutely be reliable for all projects, even without .mvn IMHO.

laeubi:
I only can add that I always have a `.mvn` dir in any setup, so "still not mainstream so will likely always fail " is probably depending on personal use-cases and experience.

Also I like it very much as it is both very basic, and easy to understand and **always** works and is global across the build, where any parent/child relation depends on.. well the parent-child relation, and I often build setups where there are not really parent/child but only an aggregator pom what includes other modules that probably have another parent so there is now way for a global property, eg. aggregator include `project1` that has `parentA` and it includes `project2` that has `parentB`, while mabye at the same reactor or even not, so what should this "parent/child defined property" point to?

rmannibucau:
> well the parent-child relation, and I often build setups where there are not really parent/child

as much as the `.mvn` thing actually, we are used to get it in the root dir but with fake parents (submodules for ex) it can be in the wrong dir so both strategies are as bad but relying on a recent addition and something not required by the project sounds very fragile upfront.

The parent child relationship is pretty straight forward: resolve all the graph while child is linked to a parent and relativepath != null and parent is not resolved from .m2/remote. Then take the higher folder.
It is as fragile as the .mvn but always works.
It is fine for me to use the .mvn as a "stop bubbling" marker if present but if none is found parent/child relationship should be used IMHO.

gnodet:
> I'm very confused by this PR, what I see:
> 
> 1. we had 2 dir properties and they are deprecated in favor of 2 dir properties, can't we fix the previous ones?

My option was to not break anything, it was easier to define a new set of properties and deprecate the old ones rather than trying to find all spots where things could be impacted.

> 2. the root dir is defined by the `.mvn` presence which is still not mainstream so will likely always fail instead of the top dir of linked project by the parent/child relationship defined by pom (which always works)

Well, the current definition is closer to `maven.multiModuleProjectDirectory` which has been used for years, even if not really public.  The idea to rely on parent/child relationship just does not work imho : first, it would have to be computed after all projects are parsed, which conflicts with the fact that one of the use case it during project interpolation, or even very early during the command line interpolation.  Although, there's absolutely no guarantee that it will point to the root level of the project, as I can point to lots of projects where this would not work. And the main use case imho it to locate files on the file system, so the parent/child relationship is completely unrelated and would not solve the issue.  It also have to never change for any build of the project, which would not be the case for any computation based on parent/child relationship.
 
> So overall think the impl should be fixed (2) and we should probably collapse most of the properties we have, in particular for v4 where it will just be a clean up and avoid the redefinition of new props. The big pro of root dir is the ability to define absolutely a file config (style.xml is what i have in mind but there are multiple cases) so should absolutely be reliable for all projects, even without .mvn IMHO.

It is at least very precisely defined and deterministic: you start from the `topdir` and go up until you find a `.mvn` tree.  If there's none, the property is not defined.  And it is fully compatible with all existing projects, that's the benefit of having new properties.

rmannibucau:
So this PR just duplicates the properties we have without solving any new feature but does it differently from before.
Git is different because the bubbling is cleaner whereas .mvn is an user dir pushed to the repos (compared to git which is not) and this is what breaks current impl as explained (agree it i as much as parent/child impl but as much both ways).
If we can't do better than what we have today, let's keep/enhance what we have instead of creating something very confusing for end user and not better in terms of reliability (which prop to use being said none is really what I want or has side effects).

>  If you aggregate projects using submodules in a bigger aggregator, the definition of the rootdir indicates it won't bubble up, and that's definitely what we want, else the project build may fail locating files that are pointed to. 

Hmm, you can make it run in both case depending how you define it (with .mvn in root but not children or with in children), so you still have broken cases.
If you aggregate you will get a single rootdir..so it is broken as soon as you get 2 subprojects - actually one since parent will not have its rootdir cleanly set.

Ideally we need a marker to set the root dir, we could use git as inspiration and force the presence of a `.mvn/.mvn_meta` which would be handled at "clone"/setup time somehow (do we fail if this file is missing, how do we ensure it is not committed and experience is smooth when a project is cloned?).

The parent/child resolution + filtering to ensure the resolution does not use .m2/remote is still way more reliable IMHO since it always works to get a deterministic rootdir. Only case it is broken is when the child does not reference the parent but then the intent of the user was to be able to run either from the parent (rootdir=parent) or child only (rootdir=child) so we always exactly match the intent whereas with .mvn we have several counter examples since the presence is optional.

gnodet:
> So this PR just duplicates the properties we have without solving any new feature but does it differently from before. Git is different because the bubbling is cleaner whereas .mvn is an user dir pushed to the repos (compared to git which is not) and this is what breaks current impl as explained (agree it i as much as parent/child impl but as much both ways). If we can't do better than what we have today, let's keep/enhance what we have instead of creating something very confusing for end user and not better in terms of reliability (which prop to use being said none is really what I want or has side effects).
> 
> > If you aggregate projects using submodules in a bigger aggregator, the definition of the rootdir indicates it won't bubble up, and that's definitely what we want, else the project build may fail locating files that are pointed to.
> 
> Hmm, you can make it run in both case depending how you define it (with .mvn in root but not children or with in children), so you still have broken cases. If you aggregate you will get a single rootdir..so it is broken as soon as you get 2 subprojects - actually one since parent will not have its rootdir cleanly set.
> 
> Ideally we need a marker to set the root dir, we could use git as inspiration and force the presence of a `.mvn/.mvn_meta` which would be handled at "clone"/setup time somehow (do we fail if this file is missing, how do we ensure it is not committed and experience is smooth when a project is cloned?).
> 
> The parent/child resolution + filtering to ensure the resolution does not use .m2/remote is still way more reliable IMHO since it always works to get a deterministic rootdir. Only case it is broken is when the child does not reference the parent but then the intent of the user was to be able to run either from the parent (rootdir=parent) or child only (rootdir=child) so we always exactly match the intent whereas with .mvn we have several counter examples since the presence is optional.

But the parent/child relationship has to be done later in the process, so can't be used for early interpolation.    If you can come up with a nice definition, we can define additional properties.  Although, those are already available from various plugins, so not sure we do actually need them, unless there is pressing needs, but I still don't grab the use case for those.  For example, they could not be used to locate checkstyle rules, headers files and such, as those have to be local to the project when aggregated in a bigger reactor. For those, the `.mvn` does work reliably.  So what's your use case ?

rmannibucau:
> But the parent/child relationship has to be done later in the process, so can't be used for early interpolation.

Interpolation does not really works there anyway so no, but yes it will require a 2 phases processing, guess it is not a big blocker as of today.
I'm not to add new properties, really just to get a single functional rootdir one (hopefully merged with multi one but if not at least reliable on any maven compatible projects).
My definition enables to use it for checkstyles, headers etc even without a .mvn (whereas it works in a minority of cases - statistically - if not done this way). It can also work when using an aggregator if the pom uses properties to define paths exactly as .mvn case which does not solve this issue since for a project with `root > {child1/.mvn>subchild1,child2/.mvn>subchild2}` it will randomly take child1 or child2 as rootdir.


Side note: using .mvn also has a nasty side effect, we couldnt use it in nested directories for module specific configuration anymore whereas it can come later if we support lock files, dependency routing or other metadata per module we don't want to put in the pom to keep it readable. It would be a natural location for me.

Maybe thinking out loud but do we attack the issue by the right side, for configuration location we want to be able to pin absolutely a file (often the root pom), shouldn't we enable it? Main issue is that `${project.xxxx}` is interpolated per pom so maybe we should enable a definition pinning, ie `${project.definition.basedir}` - name is bad but sharing the idea - would stay `/opt/foo` for `/opt/foo/pom.xml` but also its child `/opt/foo/child/pom.xml`? This looks pretty easy to handle in interpolation and drops this `rootdir` need in properties for user/mojo configuration. Can it be an outcome?

gnodet:
> > But the parent/child relationship has to be done later in the process, so can't be used for early interpolation.
> 
> Interpolation does not really works there anyway so no, but yes it will require a 2 phases processing, guess it is not a big blocker as of today. I'm not to add new properties, really just to get a single functional rootdir one (hopefully merged with multi one but if not at least reliable on any maven compatible projects). My definition enables to use it for checkstyles, headers etc even without a .mvn (whereas it works in a minority of cases - statistically - if not done this way). It can also work when using an aggregator if the pom uses properties to define paths exactly as .mvn case which does not solve this issue since for a project with `root > {child1/.mvn>subchild1,child2/.mvn>subchild2}` it will randomly take child1 or child2 as rootdir.
> 
> Side note: using .mvn also has a nasty side effect, we couldnt use it in nested directories for module specific configuration anymore whereas it can come later if we support lock files, dependency routing or other metadata per module we don't want to put in the pom to keep it readable. It would be a natural location for me.
> 
> Maybe thinking out loud but do we attack the issue by the right side, for configuration location we want to be able to pin absolutely a file (often the root pom), shouldn't we enable it? Main issue is that `${project.xxxx}` is interpolated per pom so maybe we should enable a definition pinning, ie `${project.definition.basedir}` - name is bad but sharing the idea - would stay `/opt/foo` for `/opt/foo/pom.xml` but also its child `/opt/foo/child/pom.xml`? This looks pretty easy to handle in interpolation and drops this `rootdir` need in properties for user/mojo configuration. Can it be an outcome?

Imho, we need this definition before parsing any project POM , else use cases such as https://github.com/apache/maven/pull/1059 or https://github.com/apache/maven/pull/1062 cannot be covered.  So I'm not opposed to a different way of pinning, but not from pom files.

rmannibucau:
> Imho, we need this definition before parsing any project POM , else use cases such as https://github.com/apache/maven/pull/1059 or https://github.com/apache/maven/pull/1062 cannot be covered. So I'm not opposed to a different way of pinning, but not from pom files.

For the second one it is not a technical issue AFAIK.
For first one it looks it is a bit more nasty indeed but `maven.multiModuleProjectDirectory` seems fully dedicated to this lifecycle case and looks like having the same issues so if it will somehow assume `.mvn` is there, I would prefer `.mvn/jvm.config` enabling a minimum interpolation (`$PWD` or alike) which can be set in `jvm.config` this way user can set all the system props needed there and avoid the ambiguity to introduce something magic with a lot of corner cases.

gnodet:
@michael-o @rmannibucau @cstamas  please check the recent addition of the `root="true"` model attribute as an alternative mechanism to locate the root directory.  Both are conceptually similar and the nearest  directory having either a `.mvn` sub-directory or a `pom.xml` flagged with this attribute will be used.

maven-core/src/main/java/org/apache/maven/execution/MavenExecutionRequest.java
```diff
@@ -494,14 +504,51 @@ public interface MavenExecutionRequest {
...
+
+    /**
+     * Gets the root directory of the project, which is the parent directory containing the {@code .mvn} directory
+     * or a {@code pom.xml} file with the {@code root="true"} attribute.
+     * If there's no such file, an {@code IllegalStateException} will be thrown.
+     *
+     * @throws IllegalStateException if the root directory could not be found
+     * @since 4.0.0
+     */
+    Path getRootdir() throws IllegalStateException;
```
michael-o, line 534: That's a runtime exception and redundant in the signature.

michael-o (commented):
I do not feel very comfortable to add runtime information into a static model. This does not feel right. Maybe we should just require `.mvn` to be present now, period?

maven-core/src/main/java/org/apache/maven/execution/MavenExecutionRequest.java
gnodet: It's not mandatory, but definitely not redundant. That's documentation ;-)

gnodet:
> I do not feel very comfortable to add runtime information into a static model. This does not feel right. Maybe we should just require `.mvn` to be present now, period?

I'll let you discuss with @rmannibucau about that, because that's precisely the very point he disagree with...

maven-core/src/main/java/org/apache/maven/execution/MavenExecutionRequest.java
michael-o: The documentation is already present in Javadoc. We don't declare REs anywhere, but Javadoc. It should be consistent throughout.

laeubi:
> > I do not feel very comfortable to add runtime information into a static model. This does not feel right. Maybe we should just require `.mvn` to be present now, period?
> 
> I'll let you discuss with @rmannibucau about that, because that's precisely the very point he disagree with...

My feeling is that hes the only one disagreeing here, so maybe we can not make everyone happy ;-)
Beside that i find "root" confusing if it is only used to define a path but has no other meaning.

Also as an excerpt from the discussion, I think that almost all use-cases would better to have a property that has the full path of the current **pom.xml file** (!).

So lets say I have a parent `pom.xml` and I use the property `${this.location}` it should always resolve to the path of the parent pom even if used in a child at any level. Thats simple and can be used independent of the project structure.

gnodet:
> > > I do not feel very comfortable to add runtime information into a static model. This does not feel right. Maybe we should just require `.mvn` to be present now, period?
> > 
> > 
> > I'll let you discuss with @rmannibucau about that, because that's precisely the very point he disagree with...
> 
> My feeling is that hes the only one disagreeing here, so maybe we can not make everyone happy ;-) Beside that i find "root" confusing if it is only used to define a path but has no other meaning.
> 
> Also as an excerpt from the discussion, I think that almost all use-cases would better to have a property that has the full path of the current **pom.xml file** (!).
> 
> So lets say I have a parent `pom.xml` and I use the property `${this.location}` it should always resolve to the path of the parent pom even if used in a child at any level. Thats simple and can be used independent of the project structure.

Some of the use cases involve arguments interpolation, so this has to be done _very early_ in the process.

Also, I disagree that this is _weird_ somehow.  I think it's actually mapping quite well with the property `isRootProject()` which has already been added to the `Project`, and actually nicely reflects the correct semantic and what we're looking for.

laeubi:
> Some of the use cases involve arguments interpolation, so this has to be done very early in the process.

That are probably just other use-cases that where requested! I just wanted remind of the original request, that only wanted to make `multiModuleDirectory` official to be **used inside a pom**, it was never asked to work on commandline arguments, and at the point where the **model(!)** is read the location is well known and there are no inconsistency and it is early enough for this use-case.

> Also, I disagree that this is weird somehow. I think it's actually mapping quite well with the property isRootProject() 

I think you mean `org.apache.maven.project.MavenProject.isExecutionRoot()`? Beside from Maven developers I assume no one ever has heard of it and I can't find the term in the current maven documentation anywhere. Beside that, its currently defined as the project that is the one where I executed maven.

rmannibucau:
> I do not feel very comfortable to add runtime information into a static model. This does not feel right. Maybe we should just require `.mvn` to be present now, period?

Strictly speaking it is as runtime as the gav of the pom so this part is fine IMHO.

Current proposal looks good and avoids the pitfall `.mvn` presence will imply so looks good to me.

gnodet:
> > Some of the use cases involve arguments interpolation, so this has to be done very early in the process.
> 
> That are probably just other use-cases that where requested! I just wanted remind of the original request, that only wanted to make `multiModuleDirectory` official to be **used inside a pom**, it was never asked to work on commandline arguments, and at the point where the **model(!)** is read the location is well known and there are no inconsistency and it is early enough for this use-case.

JIRA seems down for me, but I can point to MNG-5659 or https://github.com/apache/maven-mvnd/issues/813#issuecomment-1466916434

> > Also, I disagree that this is weird somehow. I think it's actually mapping quite well with the property isRootProject()
> 
> I think you mean `org.apache.maven.project.MavenProject.isExecutionRoot()`? Beside from Maven developers I assume no one ever has heard of it and I can't find the term in the current maven documentation anywhere. Beside that, its currently defined as the project that is the one where I executed maven.

I was referring to https://github.com/apache/maven/pull/1061/files#diff-5e32d3e141f6d87a19ff7f26d7ff4f2d427a8d98622e71ef24a1168772d23e13R117

maven-core/src/main/java/org/apache/maven/execution/MavenExecutionRequest.java
gnodet: Fixed

gnodet:
I think we should make the `rootdir` property available from the _project_ rather than from the _session_.  If we want to support über-aggregations, we should prepare for each project to have a different _rootdir_.
The current `session.rootdir` would only be the `rootdir` for the _top project_.
Consider the following:
```
a
 |- b
 |   \- c
 \- d
     \- e
```
where _b_ and _d_ are GitHub repositories and _a_ is a project aggregating _b_ and _d_.  In this case, _c_ rootdir should point to _b_ and _e_ rootdir should point to _d_.  If we don't do this, projects would be build differently because the rootdir would point to _a_, which could be really bad.

I'm going to add a `getRootdir()` on `Project` and make sure that this one is used when interpolating projects or injecting mojos.  This would not affect early usage of the `rootdir` property obviously.

The other possibility is to clearly document the limitation and that projects using `rootdir` may not be properly aggregated (which is the case currently for projects using the `maven.multiModuleProjectDirectory` or other mechanisms), but if that could be fixed as well...

rmannibucau:
@gnodet issue is that we likely dont want a "subroot" being a root (plugin/prop override with syst prop) with uberagg but guess we can put this case aside for now since interpolation on user props should make it smooth and these props would use root dir placeholder in its default so LGTM like that

michael-o (changes requested):

maven-core/src/main/java/org/apache/maven/project/MavenProject.java
```diff
@@ -1679,4 +1681,13 @@ public ProjectBuildingRequest getProjectBuildingRequest() {
...
     }
+
+    /**
+     * @since 4.0.0
+     * @return the rootdir for this project
+     * @throws IllegalArgumentException if the rootdir cannot be found
+     */
+    public Path getRootdir() {
+        return new DefaultRootLocator().locateMandatoryRoot(basedir != null ? basedir.toPath() : null);
+    }
```
gnodet, line 1692: I think that's wrong.  We should not be lazy here and the use of `new DefaultRootLocator()` is wrong.

maven-core/src/main/java/org/apache/maven/execution/MavenExecutionRequest.java
```diff
@@ -494,14 +504,49 @@ public interface MavenExecutionRequest {
...
+
+    /**
+     * Sets the top dir of the project.
+     *
+     * @since 4.0.0
+     */
+    MavenExecutionRequest setTopdir(Path topdir);
+
+    /**
+     * Gets the top dir of the project, which is the up-most directory of the reactor projects.
```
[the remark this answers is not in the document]
gnodet: I've removed the confusing part.

gnodet:
I'm done with the changes and added an IT.

michael-o:
I'll try to review this weekend.

michael-o:
I think as far as @elharo's comments are concerned, he's right. Names should be written out, i.e., `getRootDirectory()`, `getTopDirectory()`, etc.

gnodet:
> I think as far as @elharo's comments are concerned, he's right. Names should be written out, i.e., `getRootDirectory()`, `getTopDirectory()`, etc.

Should that imply that the properties become `project.rootDirectory`, `session.rootDirectory` and `session.topDirectory` ?  
I think people are used to `basedir` so I wanted to align with it and make things coherent. If so, I'd rename that one to `project.baseDirectory` too.

michael-o:
> > I think as far as @elharo's comments are concerned, he's right. Names should be written out, i.e., `getRootDirectory()`, `getTopDirectory()`, etc.
> 
> Should that imply that the properties become `project.rootDirectory`, `session.rootDirectory` and `session.topDirectory` ? I think people are used to `basedir` so I wanted to align with it and make things coherent. If so, I'd rename that one to `project.baseDirectory` too.

I wouldn't touch the properties, only Java code. But if the properties *map* to Java code, then this is a problem, of course. Is this the case?

michael-o:
> > > I think as far as @elharo's comments are concerned, he's right. Names should be written out, i.e., `getRootDirectory()`, `getTopDirectory()`, etc.
> > 
> > 
> > Should that imply that the properties become `project.rootDirectory`, `session.rootDirectory` and `session.topDirectory` ? I think people are used to `basedir` so I wanted to align with it and make things coherent. If so, I'd rename that one to `project.baseDirectory` too.
> 
> I wouldn't touch the properties, only Java code. But if the properties _map_ to Java code, then this is a problem, of course. Is this the case?

@elharo, please tell us what you think about this inconsistency.

maven-core/src/main/java/org/apache/maven/execution/MavenExecutionRequest.java
```diff
@@ -494,14 +504,51 @@ public interface MavenExecutionRequest {
...
 
     /**
      * @since 3.3.0
+     * @deprecated use {@link #getRootdir()} instead
      */
+    @Deprecated
     File getMultiModuleProjectDirectory();
 
+    /**
+     * Sets the top dir of the project.
```
elharo, line 520: "directory" and that's even if the name remains setTopdir

maven-core/src/main/java/org/apache/maven/execution/MavenExecutionRequest.java
```diff
@@ -494,14 +504,51 @@ public interface MavenExecutionRequest {
...
+    /**
+     * Gets the directory of the topmost project being built, usually the current directory or the
+     * directory pointed at by the {@code -f/--file} command line argument.
+     *
+     * @since 4.0.0
+     */
+    Path getTopdir();
+
+    /**
+     * Sets the root dir of the project.
```
elharo, line 535: directory

elharo:
> Should that imply that the properties become `project.rootDirectory`, `session.rootDirectory` and `session.topDirectory` ?
> I think people are used to `basedir` so I wanted to align with it and make things coherent. If so, I'd rename that one to `project.baseDirectory` too.

I do not think consistency is the most important virtue here. If these are new properties in 4.0, then by all means name the properties rootDirectory, topDirectory, etc. or perhaps some case variant of that like root_directory. I think basedir already exists so I wouldn't rename it. 

My general train of thought is to make the API and properties as clean and readable as possible, even if that introduces some inconsistencies with older API that does not follow best practices.

maven-model-builder/src/site/apt/index.apt
```diff
@@ -183,13 +183,17 @@ Maven Model Builder
...
 | <<<build.timestamp>>>\
 <<<maven.build.timestamp>>> | the UTC timestamp of build start, in <<<yyyy-MM-dd'T'HH:mm:ss'Z'>>> default format, which can be overridden with <<<maven.build.timestamp.format>>> POM property | <<<$\{maven.build.timestamp\}>>> |
 *----+------+------+
 | <<<*>>> | user properties, set from CLI with <<<-Dproperty=value>>> | <<<$\{skipTests\}>>> |
 *----+------+------+
 | <<<*>>> | model properties, such as project properties set in the pom | <<<$\{any.key\}>>> |
 *----+------+------+
+| <<<session.rootdir>>> | the directory containing the root <<<pom.xml>>> file, identified by the existence of a <<<.mvn>>> child directory | <<<$\{session.rootdir\}>>> |
+*----+------+------+
```
michael-o, line 196: Where is the documentation for the top directory?

michael-o (changes requested):
Rereading the discussions and looking into our source code, I believe that @elharo is right. Our code base from `MavenProject` consistently uses the term `directory`, everything is spelled out: directories, dependencies, repositories. Nothing abbreviated. Therefore, it seems logical that we should have `${rootDirectory}`, `${project.rootDirectory}`, `${session.rootDirectory}`, `${session.topDirectory}`

maven-model-builder/src/site/apt/index.apt
gnodet: `session.topdir` depends on the current directory so it should not be used during profile activation or model interpolation.  The `session.rootdir` property depends on `session.topdir` and thus should not be made available either.  Those two properties are available for profile injection, but this page is about the model building phase, which should be limited to `project.rootDirectory`.

api/maven-api-core/src/main/java/org/apache/maven/api/Project.java
```diff
@@ -86,8 +86,42 @@ default String getId() {
...
+    /**
+     * @deprecated use {@link #isTopProject()} instead
+     */
+    @Deprecated
     boolean isExecutionRoot();
 
+    /**
+     * Returns a boolean indicating if the project is the top level project for
+     * this reactor build.  The top level project may be different from the
+     * {@code rootDirProject}, especially if a subtree of the project is being
```
michael-o, line 98: missed this one

api/maven-api-core/src/main/java/org/apache/maven/api/Project.java
```diff
@@ -86,8 +86,42 @@ default String getId() {
...
+     * built, either because Maven has been launched in a subdirectory or using
+     * a {@code -f} option.
+     *
+     * @return {@code true} if the project is the top level project for this build
+     */
+    boolean isTopProject();
+
+    /**
+     * Returns a boolean indicating if the project is a root project,
+     * meaning that .
```
michael-o, line 108: incomplete

maven-core/src/main/java/org/apache/maven/execution/MavenExecutionRequest.java
michael-o: Yep

maven-core/src/main/java/org/apache/maven/execution/MavenExecutionRequest.java
michael-o: Yeo

maven-core/src/main/java/org/apache/maven/execution/MavenSession.java
```diff
@@ -141,10 +142,30 @@ public List<MavenProject> getProjects() {
         return projects;
     }
 
+    /**
+     * @deprecated use {@link #getTopdir()}
```
michael-o, line 146: missed this one

maven-core/src/main/java/org/apache/maven/execution/MavenSession.java
```diff
@@ -141,10 +142,30 @@ public List<MavenProject> getProjects() {
...
+    @Deprecated
     public String getExecutionRootDirectory() {
         return request.getBaseDirectory();
     }
 
+    /**
+     * @see MavenExecutionRequest#getTopDirectory()
+     * @since 4.0.0
+     */
+    public Path getTopdir() {
```
michael-o, line 157: missed this one

maven-embedder/src/main/java/org/apache/maven/cli/MavenCli.java
```diff
@@ -320,6 +324,49 @@ void initialize(CliRequest cliRequest) throws ExitException {
...
+                        throw new ExitException(1);
+                    }
+                } else {
+                    System.err.println(
+                            "POM file " + arg + " specified with the -f/--file command line argument does not exist");
+                    throw new ExitException(1);
+                }
+                break;
+            } else {
+                isAltFile = arg.equals(String.valueOf(CLIManager.ALTERNATE_POM_FILE)) || arg.equals("file");
```
michael-o, line 348: This looks pretty low level, can't we use high level CLI to get this value?

maven-model-builder/src/main/java/org/apache/maven/model/root/RootLocator.java
```diff
@@ -0,0 +1,69 @@
...
+        }
+        return rootDirectory;
+    }
+
+    @Nonnull
+    default String getNoRootMessage() {
+        return UNABLE_TO_FIND_ROOT_PROJECT_MESSAGE;
+    }
+
+    boolean isRootDir(Path dir);
```
michael-o, line 68: missed this one

maven-model-builder/src/site/apt/index.apt
```diff
@@ -41,7 +41,7 @@ Maven Model Builder
 
    ** profile activation: see {{{./apidocs/org/apache/maven/model/profile/activation/package-summary.html}available activators}}.
    Notice that model interpolation hasn't happened yet, then interpolation for file-based activation is limited to
-   <<<$\{basedir}>>> (since Maven 3), system properties and user properties
+   <<<$\{basedir}>>> (since Maven 3), <<<$\{rootDirectory}>>> (since Maven 4) system properties and user properties
```
michael-o, line 44: comma missing after "... Maven 4)"

michael-o (changes requested):
A few misses

maven-embedder/src/main/java/org/apache/maven/cli/MavenCli.java
gnodet: Well, the CLI isn't built yet, this is really one of the first things done.  Also the point is to be able to leverage those properties during argument interpolation (see https://github.com/apache/maven/pull/1062), so this really has to be done very early.

michael-o: Aha, then a comment should accompany that code,

michael-o (commented):
Looks much better now. Will try this with the IT you have written.

maven-model-builder/src/main/java/org/apache/maven/model/root/DefaultRootLocator.java
```diff
@@ -0,0 +1,53 @@
...
+            parser.setInput(is, null);
+            if (parser.nextTag() == MXParser.START_TAG && parser.getName().equals("project")) {
+                for (int i = 0; i < parser.getAttributeCount(); i++) {
+                    if ("root".equals(parser.getAttributeName(i))) {
+                        return Boolean.parseBoolean(parser.getAttributeValue(i));
+                    }
+                }
+            }
+        } catch (Exception e) {
+            // Ignore
```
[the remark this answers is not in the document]
gnodet: Not sure how to handle that.  If there's really `pom.xml` file which is not readable / parseable, the build should fail a bit later with a cleaner exception that we can display here.
That's why I decided to go this way.  The problem is that doing it in a cleaner way will involve differentiating what can be ignored and what errors should be displayed to the user, and how to display them, and whether this should fail the build or not.  Also, even the check is minimalist as we don't check the namespace.
What would you suggest ?

maven-embedder/src/main/java/org/apache/maven/cli/MavenCli.java
gnodet: Added some comments.

gnodet:
> Looks much better now. Will try this with the IT you have written.

I can also deprecate `basedir` and add a `baseDirectory`, but this may be better located in a subsequent PR.

maven-model-builder/src/main/java/org/apache/maven/model/root/DefaultRootLocator.java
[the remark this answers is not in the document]
gnodet: I can add a comment :-)

michael-o:
> > Looks much better now. Will try this with the IT you have written.
> 
> I can also deprecate `basedir` and add a `baseDirectory`, but this may be better located in a subsequent PR.

Let's move this to new PR where a separate ticket will make it visible.

maven-model-builder/src/main/java/org/apache/maven/model/root/DefaultRootLocator.java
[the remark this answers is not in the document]
michael-o: Yes, depicting that a later stage will handle this better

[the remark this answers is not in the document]
gnodet: Comment added.

maven-model-builder/src/main/java/org/apache/maven/model/root/DefaultRootLocator.java
```diff
@@ -0,0 +1,57 @@
...
+            MXParser parser = new MXParser();
+            parser.setInput(is, null);
+            if (parser.nextTag() == MXParser.START_TAG && parser.getName().equals("project")) {
+                for (int i = 0; i < parser.getAttributeCount(); i++) {
+                    if ("root".equals(parser.getAttributeName(i))) {
+                        return Boolean.parseBoolean(parser.getAttributeValue(i));
+                    }
+                }
+            }
+        } catch (Exception e) {
```
elharo, line 48: It should be possible to catch more specific exceptions here.

michael-o:
@gnodet Please don't forget to update the commit summary and JIRA summary since the property names have changed.

maven-model-builder/src/main/java/org/apache/maven/model/root/DefaultRootLocator.java
gnodet: > It should be possible to catch more specific exceptions here.

Fixed

michael-o (approved):
Looks good to me now except the summary and the commit description.

jira-importer:
Resolve #8000

PR #1063, opened 2023-03-17
[MNG-7720] Simplify code

gnodet:
Simplify code from MNG-7720 because the getCollectedProjects() already returns all submodules recursively, so no need to recurse again.


michael-o (approved):
Nice, I was right. If IT passes, I am fine with that. Needs backport to 3.9.x as well.

jira-importer:
Resolve #9262

PR #1070, opened 2023-03-24
Provide a maven-compat usage dumper

gnodet:
Small tool to dump all calls to maven-compat classes.


cstamas:
Made it into plain single-jar extension, no Maven code modification needed, thanks!

cstamas:
https://github.com/cstamas/mojosniffer for now, but IMHO we should move this into Maven (but creating new repo is tedious, maybe have some like "maven-extensions" for simple ones?)

PR #1071, opened 2023-03-24
[MNG-7751] Provide a way to inject XmlNode into plugins

gnodet:
JIRA issue [MNG-7751](https://issues.apache.org/jira/browse/MNG-7751)

This PR provides the ability to inject an `XmlNode` object into a plugin.  This can be currently done by injecting a `PlexusConfiguration`, but injecting an `XmlNode` ensures there's no dependency on the internal plexus configuration bits and limits the plugin exposure to the maven v4 api.


slachiewicz (commented):
Can we similar way incject Path?

gnodet:
> Can we similar way incject Path?

Yes, that's already possible for v4 plugins.

PR #1074, opened 2023-03-25
[MNG-7714] correct comment

elharo:
A minor detail while trying to resolve an underlying bug. The doc comment here does not accurately describe how the string 1.0alpha1 is actually parsed, and this is significant.


michael-o (approved):

slachiewicz:
please close jira

jira-importer:
Resolve #9267

PR #1075, opened 2023-03-26
[MNG-7750] Executions for plugins from an active profile are interpolated in the originalModel.

nielsbasjes:
Following this checklist to help us incorporate your
contribution quickly and easily:

 - [x] Make sure there is a [JIRA issue](https://issues.apache.org/jira/browse/MNG) filed
       for the change (usually before you start working on it).  Trivial changes like typos do not
       require a JIRA issue. Your pull request should address just this issue, without
       pulling in other changes.
 - [x] Each commit in the pull request should have a meaningful subject line and body.
 - [x] Format the pull request title like `[MNG-XXX] SUMMARY`,
       where you replace `MNG-XXX` and `SUMMARY` with the appropriate JIRA issue.
 - [x] Also format the first line of the commit message like `[MNG-XXX] SUMMARY`.
       Best practice is to use the JIRA issue title in both the pull request title and in the first line of the commit message.
 - [x] Write a pull request description that is detailed enough to understand what the pull request does, how, and why.
 - [x] Run `mvn clean verify` to make sure basic checks pass. A more thorough check will
       be performed on your pull request automatically.
 - [x] You have run the [Core IT][core-its] successfully.

If your pull request is about ~20 lines of code you don't need to sign an
[Individual Contributor License Agreement](https://www.apache.org/licenses/icla.pdf) if you are unsure
please ask on the developers list.

To make clear that you license your contribution under
the [Apache License Version 2.0, January 2004](http://www.apache.org/licenses/LICENSE-2.0)
you have to acknowledge this by using the following check-box.

 - [x] I hereby declare this contribution to be licenced under the [Apache License Version 2.0, January 2004](http://www.apache.org/licenses/LICENSE-2.0)

 - [x] In any other case, please file an [Apache Individual Contributor License Agreement](https://www.apache.org/licenses/icla.pdf).

[core-its]: https://maven.apache.org/core-its/core-it-suite/

This is ONLY the reproduction of the problem I found.

If you disable the `activeProfile` (in buildPom and remove the assert) then all checks pass.

If you enable this profile (provided code) then in several places the `${project.basedir}` has been interpolated and shows the path of the build and the assertions fail the build.

PR #1076, opened 2023-03-26
[MNG-7750] Ensure no unwanted interpolation in plugins from profiles.

nielsbasjes:
This is the port of https://github.com/apache/maven/pull/1075 to the master branch.

The good thing is that this is ONLY the tests because the maven 4 does not have the problem of MNG-7750.

A few small changes in the tests were needed to handle the internal differences between maven 3.x and maven 4.x.


nielsbasjes:
FYI: @rmannibucau @slachiewicz 
Just to be sure the maven 4.x does not have the same problem.

slachiewicz (approved):

PR #1083, opened 2023-04-11
Improve error message for deprecated core expressions

kwin:
Compare with discussion at https://the-asf.slack.com/archives/C7Q9JB404/p1681245267656379

### DIFF
--- a/maven-core/src/main/java/org/apache/maven/plugin/internal/DeprecatedCoreExpressionValidator.java
+++ b/maven-core/src/main/java/org/apache/maven/plugin/internal/DeprecatedCoreExpressionValidator.java
@@ -40,7 +40,7 @@ class DeprecatedCoreExpressionValidator extends AbstractMavenPluginParametersVal
     private static final HashMap<String, String> DEPRECATED_CORE_PARAMETERS;
 
     private static final String ARTIFACT_REPOSITORY_REASON =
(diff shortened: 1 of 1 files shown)


maven-core/src/main/java/org/apache/maven/plugin/internal/DeprecatedCoreExpressionValidator.java
```diff
@@ -51,7 +51,7 @@ class DeprecatedCoreExpressionValidator extends AbstractMavenPluginParametersVal
 
     @Override
     protected String getParameterLogReason(Parameter parameter) {
-        return "is deprecated core expression; " + DEPRECATED_CORE_PARAMETERS.get(parameter.getDefaultValue());
+        return "is leveraging a deprecated core expression: " + DEPRECATED_CORE_PARAMETERS.get(parameter.getDefaultValue());
```
kwin, line 54: Probably we should even log the used expression.

kwin:
I see this is partially fixed already in https://github.com/apache/maven/pull/1079/files#diff-c8dae894a4075bd2b8071c01ebefbf6c809009809929df69125b98fc3b3c6c17.

kwin:
Superseded by #1079

slawekjaranowski:
source branch deleted

PR #1085, opened 2023-04-13
[MNG-7763] Provide UTF-8 as a default value for project.build.sourceEncoding and project.reporting.outputEncoding

gnodet:
JIRA issue: https://issues.apache.org/jira/browse/MNG-7763
Discussion: https://lists.apache.org/thread/0mjosk1wp58fjvhz812htqfjot0n930g

### DIFF
--- a/maven-core/src/test/java/org/apache/maven/project/PomConstructionTest.java
+++ b/maven-core/src/test/java/org/apache/maven/project/PomConstructionTest.java
@@ -1219,7 +1219,7 @@ public class PomConstructionTest {
     @Test
     public void testPropertiesNoDuplication() throws Exception {
         PomTestWrapper pom = buildPom("properties-no-duplication/sub");
-        assertEquals(1, ((Properties) pom.getValue("properties")).size());
+        assertEquals(3, ((Properties) pom.getValue("properties")).size());
         assertEquals("child", pom.getValue("properties/pomProfile"));
     }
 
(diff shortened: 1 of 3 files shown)


rmannibucau (approved):

michael-o:
The outcome is then there is no way to remove this and it will also apply to ALL projects.

gnodet:
> The outcome is then there is no way to remove this and it will also apply to ALL projects.

Isn't that the goal ?
Also you can't remove the property, but you can override it, which is very often the case (else you actually have a warning).

rmannibucau:
while you can override it in your own project (which is the case AFAIK) it is fine and I don't see how null and not be deterministic can be better than forcing the value to the OS one so guess we are good like that without handling null in the code.

slachiewicz (approved):

kwin:
FTR: overwriting default values by empty string is not supported due to https://github.com/eclipse/sisu.plexus/issues/29

ge0ffrey:
For those that want to override it to the Maven 3 behaviour, maybe you can introduce a magic value "OS_DEFAULT" that gets translated to not setting the encoding deeper down the implementation?

michael-o:
> For those that want to override it to the Maven 3 behaviour, maybe you can introduce a magic value "OS_DEFAULT" that gets translated to not setting the encoding deeper down the implementation?
> 
> +1 to default to UTF-8

I'd prefer `system`. This is what I have done in other Maven components where empty was not allowed.

ge0ffrey:
Good point.

ge0ffrey:
For what's it worth, I ran an online poll around this topic a few days ago:

image

I might have revealed my bias a bit in how I posed the question :)

khmarbaise (approved):

gnodet:
> > For those that want to override it to the Maven 3 behaviour, maybe you can introduce a magic value "OS_DEFAULT" that gets translated to not setting the encoding deeper down the implementation?
> > +1 to default to UTF-8
> 
> I'd prefer `system`. This is what I have done in other Maven components where empty was not allowed.

I'd suggest that people that want the default encoding use `${file.encoding}` so that we don't have to do any hack in maven.

michael-o:
> > > For those that want to override it to the Maven 3 behaviour, maybe you can introduce a magic value "OS_DEFAULT" that gets translated to not setting the encoding deeper down the implementation?
> > > +1 to default to UTF-8
> > 
> > 
> > I'd prefer `system`. This is what I have done in other Maven components where empty was not allowed.
> 
> I'd suggest that people that want the default encoding use `${file.encoding}` so that we don't have to do any hack in maven.

You are totally right. Stupid me.

slawekjaranowski (approved):

PR #1087, opened 2023-04-13
[MNG-6934] Suggest plugin upgrade in case of failure

Giovds:
image
(Locally installed multiple versions of a dummy plugin, version 2.0.0-SNAPSHOT just throws an exception)

I'm not sure if this is the correct approach to this issue and I would like to hear your thoughts.

1. I'm currently checking `MojoExecutionException` as trigger to check if there are newer plugin versions available, but I'm unaware of alternatives to this. I've tried the `AbstractMojoExecutionException` to catch `MojoFailureException` as well, however this will also trigger the check for test failures for example which seems an unlikely case of wanting to upgrade the test plugin.

    [These docs](https://maven.apache.org/developers/mojo-api-specification.html#org-apache-maven-plugin-mojo) lead me to believe the `MojoExecutionException` is something the plugin can not recover from, which makes me think this is a case of perhaps upgrading your plugin. Or perhaps I should enrich via the `PluginManagerException` or [once any exception comes out of the mojo.execute()](https://github.com/apache/maven/blob/b7a3d857136ec89bdfe35aea05ed2746f09de162/maven-core/src/main/java/org/apache/maven/lifecycle/internal/MojoExecutor.java#L341). I'm not sure if these are correct assumptions.
    
    Perhaps any uncaught exceptions should be caught and rethrown while adding some metadata/new exception to identify it in the ExceptionHandler?
    
    Trying to filter a blacklist of plugins that should not to be checked in some in-memory list doesn't seem like a better approach.

2. It would be nice to fill the reference with some sort of link to the plugin or the version that is suggested. So that the user can see `-> [Help X]` and `[Help x] http://link-to-plugin-or-version`. I can't find one in the `PluginDescriptor` or similar classes through the `MavenPluginManager` and `PluginResolver` classes. Perhaps there is another way of retrieving this?
---

 - [x] Make sure there is a [MNG-6934](https://issues.apache.org/jira/browse/MNG-6934) filed
       pulling in other changes.


Giovds:
I don't think there are unnecessary connections as they are required to check if the plugin is using the latest version _if the Jira ticket is still something Maven wants to provide_. I don't think you can resolve this issue within Maven without outbound connections?

An alternative could be to check only the local repository, but I don't think this is a good idea as the user may or may not have higher versions of the plugin installed.

Or it could just log a suggestion like: "Please consider checking you are using the latest version of the 'xxx' plugin", however that kind of defeats the purpose of automatically checking for newer version.

Giovds:
Closing PR as discussed in ticket and here. We don't see a correct way to be absolutely sure an upgrade will result in resolving the issue, and you can not depend on the exceptions thrown by the plugins.

PR #1093, opened 2023-04-17
[MNG-7655] Switch to plexus-utils / plexus-xml 4

gnodet:
Following up various discussions, this PR removes the custom packaged `plexus-utils` and  switches to `plexus-utils` and `plexus-xml` 4.


jira-importer:
Resolve #8797

PR #1097, opened 2023-04-28
[MNG-7774] Maven config and command line interpolation

cstamas:
This is "mild introduction" of subset of new Maven4 features into Maven 3.9.x land, with single intent: be able to store resolver configuration along with checked in project sources in SCM.

Changes:
* interpolate properties and arguments
* introduce two new special properties "session.rootDirectory" and "session.topDirectory"
* these two properties may be undefined (the .mvn may not be found)
* these two properties if undefined, but referenced during interpolation, causes Maven to bail out (error out), and will prevent starting.
* some other minor cosmetic cleanup

---

https://issues.apache.org/jira/browse/MNG-7774

### DIFF
--- a/maven-embedder/src/main/java/org/apache/maven/cli/MavenCli.java
+++ b/maven-embedder/src/main/java/org/apache/maven/cli/MavenCli.java
@@ -579,7 +579,7 @@ public class MavenCli {
             System.err.println(message);
             throw new UnrecognizedOptionException(message);
         } catch (IllegalUseOfUndefinedProperty e) {
-            String message = "ERROR: Invalid use of undefined property: " + e.property;
+            String message = "ERROR: Illegal use of undefined property: " + e.property;
             System.err.println(message);
             if (cliRequest.multiModuleProjectDirectory == null) {
                 System.err.println();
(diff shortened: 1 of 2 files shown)


maven-embedder/src/main/java/org/apache/maven/cli/MavenCli.java
```diff
@@ -301,11 +307,17 @@ void initialize(CliRequest cliRequest) throws ExitException {
...
+            if (new File(basedir, DOT_MVN).isDirectory()) {
+                try {
+                    cliRequest.multiModuleProjectDirectory = basedir.getCanonicalFile();
+                } catch (IOException e) {
+                    cliRequest.multiModuleProjectDirectory = basedir.getAbsoluteFile();
+                }
+            }
+        } else {
+            if (!isMultiModuleProjectDirectoryAcceptable(cliRequest.multiModuleProjectDirectory)) {
+                cliRequest.multiModuleProjectDirectory = null;
```
slawekjaranowski, line 320: Will be ok that multiModuleProjectDirectory will be null.?
What other impact it can have?

cstamas: This PR corrects this: if not found .mvn folder, it will be null, yes, before it could have any contents. This makes it more consistent than before, when it could contain any path...

cstamas:
superseded by https://github.com/apache/maven/pull/1098

PR #1099, opened 2023-05-02
[MNG-7714] Fixed a scenario in version sorting where sp1 is less than final

CrazyHZM:

       pulling in other changes.

### DIFF
--- a/maven-artifact/src/main/java/org/apache/maven/artifact/versioning/ComparableVersion.java
+++ b/maven-artifact/src/main/java/org/apache/maven/artifact/versioning/ComparableVersion.java
@@ -130,6 +130,7 @@ public class ComparableVersion implements Comparable<ComparableVersion> {
                     return -1;
 
                 case STRING_ITEM:
+                    return 1;
                 case COMBINATION_ITEM:
                     return 1; // 1.1 > 1-sp
 
@@ -202,6 +203,7 @@ public class ComparableVersion implements Comparable<ComparableVersion> {
                     return -1;
 
                 case STRING_ITEM:
+                    return 1;
                 case COMBINATION_ITEM:
                     return 1; // 1.1 > 1-sp
 
@@ -273,6 +275,7 @@ public class ComparableVersion implements Comparable<ComparableVersion> {
                     return value.compareTo(((BigIntegerItem) item).value);
 
                 case STRING_ITEM:
+                    return 1;
                 case COMBINATION_ITEM:
                     return 1; // 1.1 > 1-sp
 
@@ -446,18 +449,18 @@ public class ComparableVersion implements Comparable<ComparableVersion> {
 
         Item digitValue;
 
-        CombinationItem(String v) {
+        CombinationItem(String value) {
             int index = 0;
-            for (int i = 0; i < v.length(); i++) {
-                char c = v.charAt(i);
+            for (int i = 0; i < value.length(); i++) {
+                char c = value.charAt(i);
                 if (Character.isDigit(c)) {
                     index = i;
                     break;
                 }
             }
 
-            stringValue = new StringItem(v.substring(0, index), true);
-            digitValue = parseItem(true, v.substring(index));
+            stringValue = new StringItem(value.substring(0, index), true);
+            digitValue = parseItem(true, value.substring(index));
         }
 
         @Override
@@ -590,6 +593,7 @@ public class ComparableVersion implements Comparable<ComparableVersion> {
                     return -1; // 1-1 < 1.0.x
 
                 case STRING_ITEM:
+                    return 1;
                 case COMBINATION_ITEM:
                     return 1; // 1-1 > 1-sp
 
--- a/maven-artifact/src/main/java/org/apache/maven/artifact/versioning/ComparableVersion.java
+++ b/maven-artifact/src/main/java/org/apache/maven/artifact/versioning/ComparableVersion.java
@@ -26,13 +26,14 @@ import java.util.Deque;
 import java.util.Iterator;
 import java.util.List;
 import java.util.Locale;
+import java.util.Objects;
 import java.util.Properties;
 
 /**
  * <p>
  * Generic implementation of version comparison.
  * </p>
- *
+ * <p>
  * Features:
  * <ul>
  * <li>mixing of '<code>-</code>' (hyphen) and '<code>.</code>' (dot) separators,</li>
@@ -58,9 +59,9 @@ import java.util.Properties;
  *   over {@code 1.0.0.X1}.</li>
  * </ul>
  *
- * @see <a href="https://maven.apache.org/pom.html#version-order-specification">"Versioning" in the POM reference</a>
  * @author <a href="mailto:kenney@apache.org">Kenney Westerhof</a>
  * @author <a href="mailto:hboutemy@apache.org">Hervé Boutemy</a>
+ * @see <a href="https://maven.apache.org/pom.html#version-order-specification">"Versioning" in the POM reference</a>
  */
 public class ComparableVersion implements Comparable<ComparableVersion> {
     private static final int MAX_INTITEM_LENGTH = 9;
@@ -79,6 +80,7 @@ public class ComparableVersion implements Comparable<ComparableVersion> {
         int BIGINTEGER_ITEM = 0;
         int STRING_ITEM = 1;
         int LIST_ITEM = 2;
+        int COMBINATION_ITEM = 5;
 
         int compareTo(Item item);
 
@@ -128,6 +130,7 @@ public class ComparableVersion implements Comparable<ComparableVersion> {
                     return -1;
 
                 case STRING_ITEM:
+                case COMBINATION_ITEM:
                     return 1; // 1.1 > 1-sp
 
                 case LIST_ITEM:
@@ -199,6 +202,7 @@ public class ComparableVersion implements Comparable<ComparableVersion> {
                     return -1;
 
                 case STRING_ITEM:
+                case COMBINATION_ITEM:
                     return 1; // 1.1 > 1-sp
 
                 case LIST_ITEM:
@@ -269,6 +273,7 @@ public class ComparableVersion implements Comparable<ComparableVersion> {
                     return value.compareTo(((BigIntegerItem) item).value);
 
                 case STRING_ITEM:
+                case COMBINATION_ITEM:
                     return 1; // 1.1 > 1-sp
 
                 case LIST_ITEM:
@@ -309,13 +314,11 @@ public class ComparableVersion implements Comparable<ComparableVersion> {
     private static class StringItem implements Item {
         private static final List<String> QUALIFIERS =
                 Arrays.asList("alpha", "beta", "milestone", "rc", "snapshot", "", "sp");
+        private static final List<String> RELEASE_QUALIFIERS = Arrays.asList("ga", "final", "release");
 
         private static final Properties ALIASES = new Properties();
 
         static {
-            ALIASES.put("ga", "");
-            ALIASES.put("final", "");
-            ALIASES.put("release", "");
             ALIASES.put("cr", "rc");
         }
 
@@ -353,15 +356,15 @@ public class ComparableVersion implements Comparable<ComparableVersion> {
 
         @Override
         public boolean isNull() {
-            return (comparableQualifier(value).compareTo(RELEASE_VERSION_INDEX) == 0);
+            return value == null || value.isEmpty();
         }
 
         /**
          * Returns a comparable value for a qualifier.
-         *
+         * <p>
          * This method takes into account the ordering of known qualifiers then unknown qualifiers with lexical
          * ordering.
-         *
+         * <p>
          * just returning an Integer with the index here is faster, but requires a lot of if/then/else to check for -1
          * or QUALIFIERS.size and then resort to lexical ordering. Most comparisons are decided by the first character,
          * so this is still fast. If more characters are needed then it requires a lexical sort anyway.
@@ -370,6 +373,10 @@ public class ComparableVersion implements Comparable<ComparableVersion> {
          * @return an equivalent value that can be used with lexical comparison
          */
         public static String comparableQualifier(String qualifier) {
+            if (RELEASE_QUALIFIERS.contains(qualifier)) {
+                return String.valueOf(QUALIFIERS.indexOf(""));
+            }
+
             int i = QUALIFIERS.indexOf(qualifier);
 
             return i == -1 ? (QUALIFIERS.size() + "-" + qualifier) : String.valueOf(i);
@@ -389,6 +396,13 @@ public class ComparableVersion implements Comparable<ComparableVersion> {
 
                 case STRING_ITEM:
                     return comparableQualifier(value).compareTo(comparableQualifier(((StringItem) item).value));
+
+                case COMBINATION_ITEM:
+                    int result = this.compareTo(((CombinationItem) item).getStringValue());
+                    if (result == 0) {
+                        return -1;
+                    }
+                    return result;
 
                 case LIST_ITEM:
                     return -1; // 1.any < 1-1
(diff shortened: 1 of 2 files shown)


maven-artifact/src/main/java/org/apache/maven/artifact/versioning/ComparableVersion.java
```diff
@@ -619,7 +748,13 @@ public final void parseVersion(String version) {
     }
 
     private static Item parseItem(boolean isDigit, String buf) {
-        if (isDigit) {
+        return parseItem(false, isDigit, buf);
+    }
+
+    private static Item parseItem(boolean isCombination, boolean isDigit, String buf) {
```
elharo, line 754: What is "buf"? It's not a StringBuffer. Is there a more descriptive name for this variable?

maven-artifact/src/test/java/org/apache/maven/artifact/versioning/ComparableVersionTest.java
```diff
@@ -147,10 +154,10 @@ void testVersionsEqual() {
         checkVersionsEqual("1.0.0x", "1-x");
 
         // aliases
-        checkVersionsEqual("1ga", "1");
-        checkVersionsEqual("1release", "1");
-        checkVersionsEqual("1final", "1");
-        checkVersionsEqual("1cr", "1rc");
+        checkVersionsEqualOrder("1ga", "1");
```
elharo, line 157: I'd be more comfortable if existing tests didn't change

maven-artifact/src/test/java/org/apache/maven/artifact/versioning/ComparableVersionTest.java
```diff
@@ -162,14 +169,14 @@ void testVersionsEqual() {
...
         checkVersionsEqual("1M", "1m");
-        checkVersionsEqual("1Ga", "1");
-        checkVersionsEqual("1GA", "1");
-        checkVersionsEqual("1RELEASE", "1");
-        checkVersionsEqual("1release", "1");
-        checkVersionsEqual("1RELeaSE", "1");
-        checkVersionsEqual("1Final", "1");
-        checkVersionsEqual("1FinaL", "1");
-        checkVersionsEqual("1FINAL", "1");
+        checkVersionsEqualOrder("1Ga", "1");
```
elharo, line 172: These really belong in a new test method. There's too much in this test method already.

maven-artifact/src/test/java/org/apache/maven/artifact/versioning/ComparableVersionTest.java
```diff
@@ -383,4 +390,15 @@ void testMng7644() {
...
         }
     }
+
+    @Test
+    public void testMng7714() {
+        String f = "1.0.final-redhat";
+        String sp1 = "1.0-sp1-redhat";
+        String sp2 = "1.0-sp-1-redhat";
+        String sp3 = "1.0-sp.1-redhat";
+        checkVersionsOrder(f, sp1);
```
elharo, line 400: This is unclear since there's nothing in the method call to say which is expected to be greater. If these can be tested directly, it would be easier to follow. That is, show the comparisons between f and sp1, etc.

maven-artifact/src/main/java/org/apache/maven/artifact/versioning/ComparableVersion.java
```diff
@@ -199,6 +202,7 @@ public int compareTo(Item item) {
                     return -1;
 
                 case STRING_ITEM:
+                case COMBINATION_ITEM:
```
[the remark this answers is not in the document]
CrazyHZM: What does `every case` mean, it's just a comparison of `LONG_ITEM` and `COMBINATION_ITEM`.

maven-artifact/src/main/java/org/apache/maven/artifact/versioning/ComparableVersion.java
```diff
@@ -422,6 +436,106 @@ public String toString() {
...
+
+        StringItem stringValue;
+
+        Item digitValue;
+
+        CombinationItem(String v) {
+            int index = 0;
+            for (int i = 0; i < v.length(); i++) {
+                char c = v.charAt(i);
+                if (Character.isDigit(c)) {
```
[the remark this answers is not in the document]
CrazyHZM: It is part of the CombinationItem, as well as part of the version, and should not be discarded.

maven-artifact/src/main/java/org/apache/maven/artifact/versioning/ComparableVersion.java
CrazyHZM: This is the previous name, I think try not to change this name in this change, it will cause trouble to code review.

maven-artifact/src/test/java/org/apache/maven/artifact/versioning/ComparableVersionTest.java
CrazyHZM: Because there was a previous case where `final` was replaced with an empty string and the hash comparison was different, I don't think this change will affect anything.

maven-artifact/src/test/java/org/apache/maven/artifact/versioning/ComparableVersionTest.java
CrazyHZM: I think it's possible to restandardize and redefine the approach to these tests.

maven-artifact/src/main/java/org/apache/maven/artifact/versioning/ComparableVersion.java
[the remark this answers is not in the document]
elharo: STRING_ITEM should have its own return.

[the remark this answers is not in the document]
CrazyHZM: Does that mean separate returns? Like
```
                case STRING_ITEM:
                    return 1;
                case COMBINATION_ITEM:
                    return 1; 
```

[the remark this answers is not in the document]
elharo: yes.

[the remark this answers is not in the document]
CrazyHZM: Done.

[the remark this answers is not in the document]
elharo: I don't see the change. Perhaps you need to push the branch.

maven-artifact/src/main/java/org/apache/maven/artifact/versioning/ComparableVersion.java
```diff
@@ -422,6 +436,106 @@ public String toString() {
...
+     * Represents a combination in the version item list.
+     * It is usually a combination of a string and a number, with the string first and the number second
+     */
+    private static class CombinationItem implements Item {
+
+        StringItem stringValue;
+
+        Item digitValue;
+
+        CombinationItem(String v) {
```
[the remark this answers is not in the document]
elharo: or version?

maven-artifact/src/main/java/org/apache/maven/artifact/versioning/ComparableVersion.java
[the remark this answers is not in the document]
elharo: Please elaborate with reference to the docs. I don't think non-ASCII digits are included in maven's version comparison algorithm. Though now that I look at it I don't think we say that:

https://maven.apache.org/pom.html

We should probably bring this up on the dev mailing list to see if there are any unexpected consequences here of non-ASCII digits.

maven-artifact/src/test/java/org/apache/maven/artifact/versioning/ComparableVersionTest.java
elharo: still easier to follow if you add new tests without changing the existing test methods.

elharo (changes requested):

maven-artifact/src/main/java/org/apache/maven/artifact/versioning/ComparableVersion.java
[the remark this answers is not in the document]
CrazyHZM: I looked at the documentation and it doesn't seem to conflict with what the documentation says
. This is just for combinations, like sp.1, sp-1, and so on.

maven-artifact/src/test/java/org/apache/maven/artifact/versioning/ComparableVersionTest.java
CrazyHZM: I did not think of a particularly good way to ensure that the original test does not change, because the difference between checkVersionsEqualOrder and checkVersionsEqual is that the hash is no longer the same, let `1ga` be `1ga` itself, I think this is a good improvement. What do you think?

elharo: Does checkVersionsEqual("1ga", "1") now fail? This is important.

If it does, then I need to dig into this much more carefully to understand what's going on.

If it doesn't, then the code shouldn't change here. It should be line by line identical.

New cases not previously tested should be in completely new test methods.

related: Existing utility methods should not change. New ones can (but probably shouldn't) be added. 

Assume the existing tests are correct unless you're deliberately changing behavior, and then call that out. However, do not assume the existing tests are well written examples of good testing practice that should be imitated. They aren't. The existing code makes a number of novice mistakes (testing more than one case in a single method; unclear method names, hiding test behavior in utility methods) that should not be copied in new code. There is no benefit to being consistent with bad practices.

maven-artifact/src/test/java/org/apache/maven/artifact/versioning/ComparableVersionTest.java
```diff
@@ -101,6 +101,13 @@ private void checkVersionsEqual(String v1, String v2) {
         assertEquals(c2, c1, "expected " + v2 + ".equals( " + v1 + " )");
     }
 
+    private void checkVersionsEqualOrder(String v1, String v2) {
```
elharo, line 104: The name of this method does not make it obvious what it does. It probably isn;t needed at all. Inlining it would likely be easier to read.

maven-artifact/src/test/java/org/apache/maven/artifact/versioning/ComparableVersionTest.java
```diff
@@ -101,6 +101,13 @@ private void checkVersionsEqual(String v1, String v2) {
...
+        Comparable c1 = newComparable(v1);
+        Comparable c2 = newComparable(v2);
```
elharo, line 106: Inlining newComparable is probably also a good idea. Reading it now, I don't know what this method does.

maven-artifact/src/test/java/org/apache/maven/artifact/versioning/ComparableVersionTest.java
```diff
@@ -147,10 +154,10 @@ void testVersionsEqual() {
         checkVersionsEqual("1.0.0x", "1-x");
 
         // aliases
-        checkVersionsEqual("1ga", "1");
-        checkVersionsEqual("1release", "1");
-        checkVersionsEqual("1final", "1");
-        checkVersionsEqual("1cr", "1rc");
+        checkVersionsEqualOrder("1ga", "1");
+        checkVersionsEqualOrder("1release", "1");
+        checkVersionsEqualOrder("1final", "1");
```
elharo, line 159: What you seem to have done is remove existing tests and replace them with new ones. Instead, leave the existing tests in place and unchanged. Then. add new tests in new methods.

We need to be really sure that existing behavior doesn't change unless we want it to.

maven-artifact/src/test/java/org/apache/maven/artifact/versioning/ComparableVersionTest.java
CrazyHZM: This code change will cause checkVersionsEqual("1ga", "1") to fail, and I think you can take a closer look at the specific changes.
The difference between `checkVersionsEqualOrder` and the `checkVersionsEqual` method is that `checkVersionsEqualOrder` only compares the order of versions, `checkVersionsEqual` compares Comparable HashCodes.
As for why I changed this test, I think I have explained above, because the previous logic was to directly change the contents of `1ga` to `1`, but now I will let `1ga` still stand for `1ga`, which will cause the hashcode in `checkVersionsEqual` to fail. This requires a new `checkVersionsEqualOrder` method to test that the order between `1ga` and `1` is consistent.

maven-artifact/src/test/java/org/apache/maven/artifact/versioning/ComparableVersionTest.java
CrazyHZM: The `checkVersionsEqualOrder` method is only used to check whether two versions are in the same order. Is there anything unreadable about this? Do you think newComparable affects the semantics of this method?

elharo: Yes, what's unreadable is that I don't know which is greater and which is lesser. Or if it's doing something else then I am being confused by the name. Perhaps a clearer name would help, but it'ds also OK to just inline it.

maven-artifact/src/test/java/org/apache/maven/artifact/versioning/ComparableVersionTest.java
elharo: only new tests please. Do not change existing tests unless the behavior has changed.

maven-artifact/src/test/java/org/apache/maven/artifact/versioning/ComparableVersionTest.java
CrazyHZM: @elharo 
I have modified it, please review again.

maven-artifact/src/main/java/org/apache/maven/artifact/versioning/ComparableVersion.java
```diff
@@ -422,6 +439,106 @@ public String toString() {
         }
     }
 
+    /**
+     * Represents a combination in the version item list.
+     * It is usually a combination of a string and a number, with the string first and the number second
+     */
+    private static class CombinationItem implements Item {
+
+        StringItem stringValue;
```
elharo, line 448: This confused since I initially thought these were different representations of the same value. How about stringPart and digitPart?

maven-artifact/src/main/java/org/apache/maven/artifact/versioning/ComparableVersion.java
```diff
@@ -688,7 +828,7 @@ public int hashCode() {
      * </pre>
      *
      * @param args the version strings to parse and compare. You can pass arbitrary number of version strings and always
-     * two adjacent will be compared
+     *             two adjacent will be compared
```
elharo, line 831: nit: period at end

maven-artifact/src/test/java/org/apache/maven/artifact/versioning/ComparableVersionTest.java
elharo: I see what's happening here now, but only because I've been over this a few times. Consider renaming this method to checkVersionsHaveSameOrder or checkVersionsAreOrderedTheSame

maven-artifact/src/test/java/org/apache/maven/artifact/versioning/ComparableVersionTest.java
```diff
@@ -101,6 +102,21 @@ private void checkVersionsEqual(String v1, String v2) {
...
     }
 
+    private void checkVersionsEqualOrder(String v1, String v2) {
+        ComparableVersion c1 = new ComparableVersion(v1);
+        ComparableVersion c2 = new ComparableVersion(v2);
+        assertEquals(0, c1.compareTo(c2), "expected " + v1 + " == " + v2);
+        assertEquals(0, c2.compareTo(c1), "expected " + v2 + " == " + v1);
+    }
+
+    private void checkVersionObjectInequality(String v1, String v2) {
```
elharo, line 112: checkVersionsAreNotEqual

maven-artifact/src/test/java/org/apache/maven/artifact/versioning/ComparableVersionTest.java
```diff
@@ -162,21 +173,39 @@ void testVersionsEqual() {
...
-        checkVersionsEqual("1FINAL", "1");
         checkVersionsEqual("1Cr", "1Rc");
         checkVersionsEqual("1cR", "1rC");
         checkVersionsEqual("1m3", "1Milestone3");
         checkVersionsEqual("1m3", "1MileStone3");
         checkVersionsEqual("1m3", "1MILESTONE3");
     }
 
+    @Test
+    void testVersionsEqualOrderAndObjectInequality() {
```
elharo, line 184: testVersionsHaveSameOrderButAreNotEqual

maven-artifact/src/test/java/org/apache/maven/artifact/versioning/ComparableVersionTest.java
```diff
@@ -162,21 +173,39 @@ void testVersionsEqual() {
...
+        checkVersionsEqualOrder("1ga", "1");
+        checkVersionObjectInequality("1ga", "1");
```
elharo, line 186: Is inequality part of the API for ComparableVersion or is it an implementation detail? If the former it should be added to the class JavaDoc. If the latter, it shouldn't be tested.

maven-artifact/src/test/java/org/apache/maven/artifact/versioning/ComparableVersionTest.java
```diff
@@ -383,4 +412,15 @@ void testMng7644() {
             checkVersionsEqual("2.0." + x, "2.0.0." + x); // previously ordered, now equals
         }
     }
+
+    @Test
+    public void testMng7714() {
+        String f = "1.0.final-redhat";
```
elharo, line 418: Make these variables ComparableVersions, not strings, and then inline checkVersionsOrder in this method so it's easier to understand in one place what is being tested. `checkVersionsOrder(f, sp1)` is unclear because it does not indicate whether f is expected to be less than, equal to, or greater than sp1.

elharo (commented):
This looks good. I do think there's still some work to make clearer what the intention of the tests is. Possibly we also need to expand the Javadoc on ComparableVersion to indicate this class is no longer consistent with equals.

maven-artifact/src/main/java/org/apache/maven/artifact/versioning/ComparableVersion.java
CrazyHZM: That's a good idea. I changed it.

maven-artifact/src/test/java/org/apache/maven/artifact/versioning/ComparableVersionTest.java
CrazyHZM: It was a special behavior before the change, but now that it's not a special behavior, I've removed the test case.

maven-artifact/src/test/java/org/apache/maven/artifact/versioning/ComparableVersionTest.java
CrazyHZM: Done.

maven-artifact/src/main/java/org/apache/maven/artifact/versioning/ComparableVersion.java
```diff
@@ -353,15 +359,15 @@ public int getType() {
...
 
         /**
          * Returns a comparable value for a qualifier.
-         *
+         * <p>
          * This method takes into account the ordering of known qualifiers then unknown qualifiers with lexical
          * ordering.
-         *
+         * <p>
          * just returning an Integer with the index here is faster, but requires a lot of if/then/else to check for -1
```
elharo, line 371: From here to the end is implementation detail. It's a good comment but it isn't needed in API doc. Move this part to a regular comment.

maven-artifact/src/main/java/org/apache/maven/artifact/versioning/ComparableVersion.java
```diff
@@ -422,6 +439,106 @@ public String toString() {
         }
     }
 
+    /**
+     * Represents a combination in the version item list.
+     * It is usually a combination of a string and a number, with the string first and the number second
```
elharo, line 444: nit: period at end

elharo (approved):
some minor optional comments on javadoc, but algorithm and test wise I think this is ready. Thanks!

maven-artifact/src/main/java/org/apache/maven/artifact/versioning/ComparableVersion.java
CrazyHZM: Done.

PR #1100, opened 2023-05-02
[MNG-7038] Fixes for topDirectory

cstamas:
Changes:
* resolve file argument against workingDirectory explicitly
* the check Files.isRegularFile should happen against path

### DIFF
--- a/maven-embedder/src/main/java/org/apache/maven/cli/MavenCli.java
+++ b/maven-embedder/src/main/java/org/apache/maven/cli/MavenCli.java
@@ -332,10 +332,10 @@ public class MavenCli {
         for (String arg : cliRequest.args) {
             if (isAltFile) {
                 // this is the argument following -f/--file
-                Path path = Paths.get(arg);
+                Path path = topDirectory.resolve(arg);
                 if (Files.isDirectory(path)) {
                     topDirectory = path;
-                } else if (Files.isRegularFile(topDirectory)) {
+                } else if (Files.isRegularFile(path)) {
                     topDirectory = path.getParent();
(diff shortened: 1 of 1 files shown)


maven-embedder/src/main/java/org/apache/maven/cli/MavenCli.java
```diff
@@ -332,10 +332,10 @@ void initialize(CliRequest cliRequest) throws ExitException {
         for (String arg : cliRequest.args) {
             if (isAltFile) {
                 // this is the argument following -f/--file
-                Path path = Paths.get(arg);
+                Path path = topDirectory.resolve(arg);
```
gnodet, line 335: That's really weird.  That would make file argument order dependent as the `topDirectory` is modified inside the loop.

gnodet: Even if this was not the case, the goal of the loop is to compute the `topDirectory`.
However, introducing a `workingDirectory` variable and using it to resolve the `path` could make sense.

cstamas: I still don't understand why would this loop enter 2 or more times? (the user specifies several -f parameters?)

cstamas: but even then, there is a break (or throws) so first entrance will either break or throw

gnodet: Missed the `break` ...

gnodet (approved):

jira-importer:
Resolve #8000

PR #1101, opened 2023-05-05
[MNG-7769] Update Resolver to 1.9.10

cstamas:
From 1.9.8

---

https://issues.apache.org/jira/browse/MNG-7769


jira-importer:
Resolve #9218

PR #1103, opened 2023-05-05
[MNG-7778] - Include suppressed exceptions when logging failures

marcphilipp:
Instead of only including causes, suppressed exceptions are now included
as well. A similar indentation strategy as in
`Throwable.printStackTrace` is used.

---

https://issues.apache.org/jira/browse/MNG-7778


cstamas:
You need to reformat: pls `mvn spotless:apply` and commit

cstamas (approved):

rmannibucau (approved):

jira-importer:
Resolve #9209

PR #1104, opened 2023-05-09
[MNG-6829] Replace any StringUtils#isEmpty(String) and #isNotEmpty(String)

timtebeek:
Part of an ongoing effort to phase out StringUtils.
https://issues.apache.org/jira/browse/MNG-6829

Use this link to re-run the recipe: https://public.moderne.io/recipes/org.openrewrite.java.migrate.apache.commons.lang.IsNotEmptyToJdk?organizationId=QXBhY2hlIE1hdmVu

Review requested of @elharo 

---

       pulling in other changes.

### DIFF
--- a/maven-core/src/main/java/org/apache/maven/lifecycle/internal/DefaultMojoExecutionConfigurator.java
+++ b/maven-core/src/main/java/org/apache/maven/lifecycle/internal/DefaultMojoExecutionConfigurator.java
@@ -37,7 +37,6 @@ import org.apache.maven.plugin.descriptor.Parameter;
 import org.apache.maven.project.MavenProject;
 import org.apache.maven.shared.utils.logging.MessageBuilder;
 import org.apache.maven.shared.utils.logging.MessageUtils;
-import org.codehaus.plexus.util.StringUtils;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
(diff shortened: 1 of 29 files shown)


elharo (approved):
I'm surprised none of these change the imports

timtebeek:
Some of these remove imports though right? But only when there's no other methods being called. There's quite a few `StringUtils.replace` that I'd like to tackle next, which should help remove further imports.

I'd not reconsidered replacing imports yet (going from one string utils implementation to another), as it seemed removing was preferred over replacement. Is that still the case?

We can discuss what to tackle next and how to approach things going forward, as there were other concerns as well such as matching up per project Jira issues. Perhaps we can schedule a brief call or chat at a conference to discuss?

elharo:
I didn't notice any import changes. Maybe I just missed them. I'd expect that in some if not all of these files these would be the only methods used from StringUtils so I expected to see some import removals.

maven-core/src/main/java/org/apache/maven/lifecycle/internal/DefaultMojoExecutionConfigurator.java
```diff
@@ -37,7 +37,6 @@
 import org.apache.maven.project.MavenProject;
 import org.apache.maven.shared.utils.logging.MessageBuilder;
 import org.apache.maven.shared.utils.logging.MessageUtils;
-import org.codehaus.plexus.util.StringUtils;
```
elharo, line 40: OK, looks like it is sometimes removing an import. That's good.

elharo (approved):

jira-importer:
Resolve #8458

PR #1105, opened 2023-05-10
[MNG-7740] Remove old temporary consumer*pom files from buildDir

Giovds:
This PR removes consumer*pom files from previous runs when the new consumer pom file is created. This way the build directory is not flooded with lots of `consumer*pom` files after a couple of runs.

---

 - [x] Make sure there is a [MNG-7740](https://issues.apache.org/jira/browse/MNG-7740) filed
       pulling in other changes.

### DIFF
--- a/maven-core/src/main/java/org/apache/maven/internal/transformation/ConsumerPomArtifactTransformer.java
+++ b/maven-core/src/main/java/org/apache/maven/internal/transformation/ConsumerPomArtifactTransformer.java
@@ -29,7 +29,10 @@ import java.nio.file.Paths;
 import java.nio.file.StandardCopyOption;
 import java.util.ArrayList;
 import java.util.Collection;
+import java.util.List;
 import java.util.function.BiConsumer;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
 
 import org.apache.maven.feature.Features;
 import org.apache.maven.model.building.DefaultBuildPomXMLFilterFactory;
@@ -77,10 +80,31 @@ public final class ConsumerPomArtifactTransformer {
                 Files.createDirectories(buildDir);
                 generatedFile = Files.createTempFile(buildDir, CONSUMER_POM_CLASSIFIER, "pom");
             }
+
+            removeOldConsumerPomFiles(generatedFile);
+
             project.addAttachedArtifact(new ConsumerPomArtifact(project, generatedFile, session));
         } else if (project.getModel().isRoot()) {
             throw new IllegalStateException(
                     "The use of the root attribute on the model requires the buildconsumer feature to be active");
+        }
+    }
+
+    private void removeOldConsumerPomFiles(Path generatedFile) throws IOException {
+        List<Path> oldConsumerPomFiles;
+        String newestFileName = generatedFile.getFileName().toString();
+        try (Stream<Path> stream = Files.walk(generatedFile.getParent(), 1)) {
+            oldConsumerPomFiles = stream.filter(path -> {
+                        String fileName = path.getFileName().toString();
+                        return !fileName.equals(newestFileName)
+                                && fileName.startsWith(CONSUMER_POM_CLASSIFIER)
+                                && fileName.endsWith("pom");
+                    })
+                    .collect(Collectors.toList());
+        }
+
+        for (Path path : oldConsumerPomFiles) {
+            Files.delete(path);
         }
     }
 
--- a/maven-core/src/main/java/org/apache/maven/internal/transformation/ConsumerPomArtifactTransformer.java
+++ b/maven-core/src/main/java/org/apache/maven/internal/transformation/ConsumerPomArtifactTransformer.java
@@ -78,10 +78,9 @@ public final class ConsumerPomArtifactTransformer {
             } else {
                 Path buildDir = Paths.get(buildDirectory);
                 Files.createDirectories(buildDir);
+                removeOldConsumerPomFiles(buildDir);
                 generatedFile = Files.createTempFile(buildDir, CONSUMER_POM_CLASSIFIER, "pom");
             }
-
-            removeOldConsumerPomFiles(generatedFile);
 
             project.addAttachedArtifact(new ConsumerPomArtifact(project, generatedFile, session));
         } else if (project.getModel().isRoot()) {
@@ -90,15 +89,12 @@ public final class ConsumerPomArtifactTransformer {
         }
     }
 
-    private void removeOldConsumerPomFiles(Path generatedFile) throws IOException {
+    private void removeOldConsumerPomFiles(Path buildDir) throws IOException {
         List<Path> oldConsumerPomFiles;
-        String newestFileName = generatedFile.getFileName().toString();
-        try (Stream<Path> stream = Files.walk(generatedFile.getParent(), 1)) {
+        try (Stream<Path> stream = Files.walk(buildDir, 1)) {
             oldConsumerPomFiles = stream.filter(path -> {
                         String fileName = path.getFileName().toString();
-                        return !fileName.equals(newestFileName)
-                                && fileName.startsWith(CONSUMER_POM_CLASSIFIER)
-                                && fileName.endsWith("pom");
+                        return fileName.startsWith(CONSUMER_POM_CLASSIFIER) && fileName.endsWith("pom");
                     })
                     .collect(Collectors.toList());
         }


maven-core/src/main/java/org/apache/maven/internal/transformation/ConsumerPomArtifactTransformer.java
```diff
@@ -77,13 +80,34 @@ public void injectTransformedArtifacts(MavenProject project, RepositorySystemSes
                 Files.createDirectories(buildDir);
                 generatedFile = Files.createTempFile(buildDir, CONSUMER_POM_CLASSIFIER, "pom");
```
michael-o, line 81: The suffix should be `.pom`, look at the Javadoc of this method.
It is *not* the extension.

maven-core/src/main/java/org/apache/maven/internal/transformation/ConsumerPomArtifactTransformer.java
```diff
@@ -77,13 +80,34 @@ public void injectTransformedArtifacts(MavenProject project, RepositorySystemSes
...
 
+    private void removeOldConsumerPomFiles(Path generatedFile) throws IOException {
+        List<Path> oldConsumerPomFiles;
+        String newestFileName = generatedFile.getFileName().toString();
+        try (Stream<Path> stream = Files.walk(generatedFile.getParent(), 1)) {
+            oldConsumerPomFiles = stream.filter(path -> {
+                        String fileName = path.getFileName().toString();
+                        return !fileName.equals(newestFileName)
+                                && fileName.startsWith(CONSUMER_POM_CLASSIFIER)
+                                && fileName.endsWith("pom");
```
michael-o, line 101: Ditto.

michael-o (changes requested):

maven-core/src/main/java/org/apache/maven/internal/transformation/ConsumerPomArtifactTransformer.java
Giovds: This was added in [MNG-7622](https://issues.apache.org/jira/browse/MNG-7622) at [this line](https://github.com/apache/maven/pull/907/files#diff-b0e0e02760045bbda4d9839297d968b0218de307f1c4a35a3262d8cb406e0ea3R70). The newly created files currently are extensionless and suffixed `pom` (see screenshot 1). 

If I suffix the newly created files with `.pom` it will be seen as a `POM` extension in my file explorer (screenshot 2).
Would this not be a different ticket to change that behaviour? I'm happy to change it here if not, but I'll have to check if it impacts other places.

[1]
image
[2]
image
image

maven-core/src/main/java/org/apache/maven/internal/transformation/ConsumerPomArtifactTransformer.java
```diff
@@ -77,13 +80,34 @@ public void injectTransformedArtifacts(MavenProject project, RepositorySystemSes
...
+
             project.addAttachedArtifact(new ConsumerPomArtifact(project, generatedFile, session));
         } else if (project.getModel().isRoot()) {
             throw new IllegalStateException(
                     "The use of the root attribute on the model requires the buildconsumer feature to be active");
         }
     }
 
+    private void removeOldConsumerPomFiles(Path generatedFile) throws IOException {
+        List<Path> oldConsumerPomFiles;
```
mthmulders, line 94: Why does the code collect all the files in a list if it could as well remove them in the terminal operation of the `j.u.s.Stream`?

Giovds: I had a couple of reasons for it. The first reason for me is readability (as I'll show down below). 

The second is that I don't think its a good practice to wrap the checked exception into a runtime exception (also introducing more code or a wrapper 'delete' method). Especially since `injectTransformedArtifacts()` already throws a checked exception which is probably accounted for somewhere at some point.

As far as I know when you rewrite the stream to something like `stream.filter(path -> /*filter*/ }).forEach(Files::delete);` you either have to wrap the checked exception thrown by `Files.delete(path)` into a runtime exception:
```java
    private void delete( Path path ) {
        try { Files.delete( path ); }
        catch ( IOException e ) { throw new RuntimeException("", e); }
    }
    
    /* obfuscated*/
    
    stream.filter(path -> /*filter*/ })
           .forEach(this::delete);
```
Or handle it:
```java
stream.filter(path -> /*filter*/ })
           .forEach(path -> {
                try { Files.delete(path); }
                catch (IOException ex) { /*handle the exception*/ }
           });
```

mthmulders: That makes perfect sense to me. Thanks for commenting on this!

maven-core/src/main/java/org/apache/maven/internal/transformation/ConsumerPomArtifactTransformer.java
```diff
@@ -77,13 +80,34 @@ public void injectTransformedArtifacts(MavenProject project, RepositorySystemSes
                 Files.createDirectories(buildDir);
                 generatedFile = Files.createTempFile(buildDir, CONSUMER_POM_CLASSIFIER, "pom");
             }
+
+            removeOldConsumerPomFiles(generatedFile);
```
gnodet, line 84: Can we move the call to `removeOldConsumerPomFiles` before generating the new file and remove them all instead of having to pass the newly created file as an argument ?

gnodet: Also, I think this removing *all* matching files from the temporary directory is problematic.  If we have two concurrent builds running, one of the files will be lost.  This may be done only when `buildDirectory != null`.

gnodet (changes requested):
I think it's easier to remove all files, then generate the new one...
And it can only be done for the `buildDirectory`, not the OS temp dir.

maven-core/src/main/java/org/apache/maven/internal/transformation/ConsumerPomArtifactTransformer.java
Giovds: > I think it's easier to remove all files, then generate the new one...

Fair point. It makes it simpler and clearer, nice catch!

> Also, I think this removing _all_ matching files from the temporary directory is problematic. If we have two concurrent builds running, one of the files will be lost. This may be done only when `buildDirectory != null`.

It would be a problem. I don't think we can safely delete any files in that case without introducing too much complexity, so the behaviour would be to just let the files live in the temp dir and _maybe_ get cleaned up by the OS. Some do and some don't.

I didn't know the temp dir would/could be used as build directory (I just looked and the files were indeed there). Out of curiosity; when would this be the case?

gnodet: > > Also, I think this removing _all_ matching files from the temporary directory is problematic. If we have two concurrent builds running, one of the files will be lost. This may be done only when `buildDirectory != null`.
> 
> It would be a problem. I don't think we can safely delete any files in that case without introducing too much complexity, so the behaviour would be to just let the files live in the temp dir and _maybe_ get cleaned up by the OS. Some do and some don't.

There may be a better alternative.  I wonder if we could just add a `@PreDestroy` method that would keep a list of generated files (be it in the temp directory or in the build directory) and delete them when the bean is destroy...

> I didn't know the temp dir would/could be used as build directory (I just looked and the files were indeed there). Out of curiosity; when would this be the case?

Actually, I did not know either, I just noticed it while reviewing this PR.

gnodet: @Giovds have a look at https://github.com/gnodet/maven/commit/11009e9f6921657e31384bd9838d5ab1edb1aacc, could you double check that it works for you ?

Giovds: > @Giovds have a look at https://github.com/gnodet/maven/commit/11009e9f6921657e31384bd9838d5ab1edb1aacc, could you double check that it works for you ?

Cool, TIL! Just verified it and it does work for me. The files are cleaned up. 

Do note that the final result means there will be no consumer*pom file in the build dir whatsoever. I don't know if it was kept for a reason. Perhaps the file is/will be (re)used for build-consumer features beyond a single run? If it's not and it's a temp file, why not always use the temp dir e.g.?

\* Without to much research: * Seems like (in this class at least) the `CONSUMER_POM_CLASSIFIER` is private so not used by other classes and the build dir is never queried for the file, only in-memory artifact lists?

gnodet: > > @Giovds have a look at [gnodet@11009e9](https://github.com/gnodet/maven/commit/11009e9f6921657e31384bd9838d5ab1edb1aacc), could you double check that it works for you ?
> 
> Cool, TIL! Just verified it and it does work for me. The files are cleaned up.
> 
> Do note that the final result means there will be no consumer*pom file in the build dir whatsoever. I don't know if it was kept for a reason. Perhaps the file is/will be (re)used for build-consumer features beyond a single run? If it's not and it's a temp file, why not always use the temp dir e.g.?
> 
> * Without to much research: * Seems like (in this class at least) the `CONSUMER_POM_CLASSIFIER` is private so not used by other classes and the build dir is never queried for the file, only in-memory artifact lists?

Yes, I think it was written to disk because it was needed as a file and there was no easy way to delete it at the end of the process, but they're not supposed to outlive the session.

gnodet:
Closing this PR in favour of #1117

PR #1108, opened 2023-05-15
[MNG-7780] DefaultArtifact.equals throws NullPointerException if o.version is null

CrazyHZM:

       pulling in other changes.

### DIFF
--- a/maven-artifact/src/main/java/org/apache/maven/artifact/DefaultArtifact.java
+++ b/maven-artifact/src/main/java/org/apache/maven/artifact/DefaultArtifact.java
@@ -19,11 +19,7 @@
 package org.apache.maven.artifact;
 
 import java.io.File;
-import java.util.Collection;
-import java.util.Collections;
-import java.util.HashMap;
-import java.util.List;
-import java.util.Map;
+import java.util.*;
 
 import org.apache.maven.artifact.handler.ArtifactHandler;
 import org.apache.maven.artifact.metadata.ArtifactMetadata;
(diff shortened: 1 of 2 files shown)


slachiewicz:
Please add also test case

CrazyHZM:
> Please add also test case

Done.

maven-artifact/src/main/java/org/apache/maven/artifact/DefaultArtifact.java
```diff
@@ -314,7 +314,9 @@ public boolean equals(Object o) {
             return false;
         } else if (!a.getArtifactId().equals(artifactId)) {
             return false;
-        } else if (!a.getVersion().equals(version)) {
+        } else if (a.getVersion() != null && !a.getVersion().equals(version)) {
```
gnodet, line 317: How is that different from `!a.getGroupId().equals(groupId)` or `!a.getArtifactId().equals(artifactId)` ?

gnodet: So `validateIdentity` ensures some fields are non null or non empty.
This is not the case for `version` because the test is `(version == null) && (versionRange == null)`

maven-artifact/src/main/java/org/apache/maven/artifact/DefaultArtifact.java
```diff
@@ -314,7 +314,9 @@ public boolean equals(Object o) {
...
+            return false;
+        } else if (a.getVersion() == null && version != null) {
```
gnodet, line 319: I would suggest rewriting the whole logic in a single statement:
```
    return Objects.equals(a.getGroupId(), this.groupId)
        && Objects.equals(a.getArtifactId(), this.artifactId)
        && Objects.equals(a.getVersion(), this.version)
        && Objects.equals(a.getType(), this.type)
        && Objects.equals(a.getClassifier(), this.classifier);
```

gnodet (changes requested):
Please simplify the logic which has not been updated since JDK 1.7

maven-artifact/src/main/java/org/apache/maven/artifact/DefaultArtifact.java
CrazyHZM: Agreed. I have revised this part.

CrazyHZM:
> Please simplify the logic which has not been updated since JDK 1.7

@gnodet Done.

gnodet (approved):

PR #1112, opened 2023-05-19
[MNG-7786] Fix plugin validation misleading message

cstamas:
Reword the validation warning and add new check for real plugin dependencies in wrong scopes (do not rely on build-time derived descriptor, but on real data instead).

---

https://issues.apache.org/jira/browse/MNG-7786


rmannibucau (approved):

jira-importer:
Resolve #9201

PR #1114, opened 2023-05-19
[MNG-7788] Plugin Validation Report logged before summary

cstamas:
This change uses different technique to hook into "session end", and thus is invoked before ExecutionEventLogger is.

---

https://issues.apache.org/jira/browse/MNG-7788


cstamas:
Example of "before" and "after" outputs: https://gist.github.com/cstamas/7c9bddbf4e8d96cdd9b73670d878e997

PR #1115, opened 2023-05-19
[MNG-7789] Dependency validation rules used wrong data

cstamas:
Refactored and now dropped the unused pluginDescriptor/dependencies validation. Now we have "plugin dependency" (using POM) components to perform dependency validation.

Also, two "inlined" validation from plugin dependencies resolver factored out to components, along with 3 existing moved to newly added (they were wrongly validating descriptor dependencies while intent was to validate "real" dependencies) validators.

---

https://issues.apache.org/jira/browse/MNG-7789


cstamas:
This resolves my personal problem as well in really natural way: "plugin dependencies" are validated per-plugin level, while "plugin descriptor dependencies" were validated (and reported) per Mojo level, while it is plugin POM (so applicable to all Mojos within plugin) that may contains dependency problems, not standalone Mojos...

gnodet (commented):
I have a bit of a hard time understanding the purpose of the generated dependencies in the plugin descriptor.  What are they used for and how do they differ from the ones from the pom ?

maven-core/src/main/java/org/apache/maven/plugin/internal/DefaultPluginValidationManager.java
```diff
@@ -46,6 +42,13 @@
 @Singleton
 @Named
 public final class DefaultPluginValidationManager extends AbstractEventSpy implements PluginValidationManager {
+    /**
+     * The collection of "G:A" combinations that do NOT belong to Maven Core, hence, should be excluded from
+     * "expected in provided scope" type of checks.
+     */
+    static final Collection<String> EXPECTED_PROVIDED_SCOPE_EXCLUSIONS_GA =
+            Collections.unmodifiableCollection(Arrays.asList(
+                    "org.apache.maven:maven-archiver", "org.apache.maven:maven-jxr", "org.apache.maven:plexus-utils"));
```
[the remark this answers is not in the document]
cstamas: You tell me 😄 https://repo.maven.apache.org/maven2/org/apache/maven/plexus-utils/

cstamas:
Today, descriptor/dependencies are not used anywhere (maybe in some UTs) in maven, the POM drives plugin dependencies. Hence the realization and this PR (it moves existing code to check POM originating deps).
The descriptor/deps is at (plugin) build time derived from POM, and as we saw, is affected by possible mpp bugs like that one in 3.6.2

maven-core/src/main/java/org/apache/maven/plugin/internal/DefaultPluginValidationManager.java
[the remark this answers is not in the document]
gnodet: Yes, and this very artifact will not exist anymore in 4.0.0-m6, and that's why I asked the question.  It does not exist in 3.9.x either.  But afaik, you added a check on this artifact on https://github.com/apache/maven/commit/53b64732378fe27a91cb66dad5693d6d0c71628a#diff-95336d0fc4e39e625e90137247157a9d73ed8094dfb4e3431b589a9d94c3dedeR38
Do you recall for which use case you added this check ?

gnodet:
> Today, descriptor/dependencies are not used anywhere (maybe in some UTs) in maven, the POM drives plugin dependencies. Hence the realization and this PR (it moves existing code to check POM originating deps).
> The descriptor/deps is at (plugin) build time derived from POM, and as we saw, is affected by possible mpp bugs like that one in 3.6.2

Then, why checking those and add validation warnings ? Wouldn't it make sense to actually deprecate them, not generate them anymore with the next m-p-p plugins ?

cstamas:
This PR does not add any check to it, it actually moves them off from it. One check remains, thay yes, may be killed off actually.

gnodet:
> This PR does not add any check to it, it actually moves them off from it. One check remains, thay yes, may be killed off actually.

Can we then delete the AbstractMavenPluginDescriptorDependenciesValidator and MavenArtifactsMavenPluginDescriptorDependenciesValidator classes which are new ?
This would remove all usage of `PluginDescriptor#getDependencies()` from maven (apart from the `clone` and UTs).

cstamas:
Will do that tomorrow, agreed.

slawekjaranowski:
Plugin pom/descriptor dependencies can something else that is resolved by dependency collection and what finally is used to construct plugin class realm
like: https://gist.github.com/slawekjaranowski/c65612fca4094ec8a2d09c6db3b8f78a

maven-core/src/main/java/org/apache/maven/plugin/internal/DefaultPluginValidationManager.java
```diff
@@ -22,11 +22,7 @@
 import javax.inject.Singleton;
 
 import java.io.File;
-import java.util.Arrays;
-import java.util.LinkedHashMap;
-import java.util.LinkedHashSet;
-import java.util.Locale;
-import java.util.Map;
+import java.util.*;
```
hgschmie, line 25: *cringe*

maven-core/src/main/java/org/apache/maven/plugin/internal/Maven2DependenciesValidator.java
```diff
@@ -45,19 +46,22 @@ class Maven2DependenciesValidator extends AbstractMavenPluginDependenciesValidat
...
+                .map(Artifact::getVersion)
                 .filter(v -> v.startsWith("2."))
                 .collect(Collectors.toSet());
 
         if (!maven2Versions.isEmpty()) {
             pluginValidationManager.reportPluginValidationIssue(
-                    mavenSession,
-                    mojoDescriptor,
-                    "Plugin is a Maven 2.x plugin, which will be not supported in Maven 4.x");
+                    session, pluginArtifact, "Plugin is a Maven 2.x plugin, which will be not supported in Maven 4.x");
```
hgschmie, line 64: this says "will not be supported"

maven-core/src/main/java/org/apache/maven/plugin/internal/Maven3CompatDependenciesValidator.java
```diff
@@ -0,0 +1,61 @@
...
+            Artifact pluginArtifact,
+            ArtifactDescriptorResult artifactDescriptorResult) {
+        for (org.eclipse.aether.graph.Dependency dependency : artifactDescriptorResult.getDependencies()) {
+            if ("org.apache.maven".equals(dependency.getArtifact().getGroupId())
+                    && "maven-compat".equals(dependency.getArtifact().getArtifactId())
+                    && !JavaScopes.TEST.equals(dependency.getScope())) {
+                pluginValidationManager.reportPluginValidationIssue(
+                        session,
+                        pluginArtifact,
+                        "Plugin depends on the deprecated Maven 2.x compatibility layer, which may not be supported in Maven 4.x");
```
hgschmie, line 57: this says "may not be supported". It implies that there is chance that it may still be supported. This should be the same wording as above.

maven-core/src/main/java/org/apache/maven/plugin/internal/MavenArtifactsMavenPluginDescriptorDependenciesValidator.java
```diff
@@ -0,0 +1,74 @@
...
+                        d.getGroupId() + ":" + d.getArtifactId()))
+                .filter(d -> d.getVersion().startsWith("3."))
+                .map(d -> d.getGroupId() + ":" + d.getArtifactId() + ":" + d.getVersion())
+                .collect(Collectors.toSet());
+
+        if (!mavenArtifacts.isEmpty()) {
+            pluginValidationManager.reportPluginValidationIssue(
+                    mavenSession,
+                    mojoDescriptor,
+                    "Plugin descriptor should not contain these Maven artifacts: " + mavenArtifacts);
```
hgschmie, line 71: there should be an explanation *why* the descriptor should not contain them (and what a plugin author can do, e.g. "update the version of the maven-plugin-plugin to at least ...").

maven-core/src/main/java/org/apache/maven/plugin/internal/MavenScopeDependenciesValidator.java
```diff
@@ -0,0 +1,70 @@
...
+                        a.getGroupId() + ":" + a.getArtifactId()))
+                .filter(a -> a.getVersion().startsWith("3."))
+                .map(a -> a.getGroupId() + ":" + a.getArtifactId() + ":" + a.getVersion())
+                .collect(Collectors.toSet());
+
+        if (!mavenArtifacts.isEmpty()) {
+            pluginValidationManager.reportPluginValidationIssue(
+                    session,
+                    pluginArtifact,
+                    "Plugin should declare these Maven artifacts in `provided` scope: " + mavenArtifacts);
```
hgschmie, line 67: same as above. Adding the line "if the plugin already declares them in `provided` scope, update the plugin-plugin to at least version x.y.z (see <link to maven bug here)` goes a *long* way to acceptance of bug reports.

maven-core/src/main/java/org/apache/maven/plugin/internal/PlexusContainerDefaultDependenciesValidator.java
```diff
@@ -40,14 +41,17 @@ class PlexusContainerDefaultDependenciesValidator extends AbstractMavenPluginDep
...
+            Artifact pluginArtifact,
+            ArtifactDescriptorResult artifactDescriptorResult) {
+        boolean pcdPresent = artifactDescriptorResult.getDependencies().stream()
+                .filter(d -> "org.codehaus.plexus".equals(d.getArtifact().getGroupId()))
+                .anyMatch(d -> "plexus-container-default".equals(d.getArtifact().getArtifactId()));
 
         if (pcdPresent) {
             pluginValidationManager.reportPluginValidationIssue(
-                    mavenSession, mojoDescriptor, "Plugin depends on plexus-container-default, which is EOL");
+                    session, pluginArtifact, "Plugin depends on plexus-container-default, which is EOL");
```
hgschmie, line 54: this looks like it could be a more generic "check for end of life" component checker if there other stuff that is EOLed, too.

maven-core/src/main/java/org/apache/maven/plugin/internal/MavenArtifactsMavenPluginDescriptorDependenciesValidator.java
cstamas: This class (that uses plugin descriptor / dependencies) is about to be removed.

maven-core/src/main/java/org/apache/maven/plugin/internal/PlexusContainerDefaultDependenciesValidator.java
cstamas: For that, we'd need some "EOL marker" or some metadata?

Historically, the was p-c-d, that was 14 years ago stopped being developed (essentially)  got drop-in replacement in form of "plexus-shim" (that is sisu plexus). Furthermore, JSR330 migration started around same time, so today, there is really no reason to depend on p-c-d artifact (nor plexus shim, except in test scope, as PlexusTestCase is in there still, but also advisable to move off from that Junir3 class).

cstamas:
Pushed two commits:
* removed descriptor dependencies validator (3 classes)
* reworded/aligned emitted messages

maven-core/src/main/java/org/apache/maven/plugin/internal/MavenScopeDependenciesValidator.java
cstamas: While I agree, I just went with "upgrade to latest", as I cannot wholeheartedly recommend 3.6.3 (that would fix your case), nor any other "historical" version, as if you look at release notes, each release do contain a bug fix, that this or that way just improves things.

cstamas:
So this PR now contains all the changes I would like to have in here (as these needs to be ported to master as well).

Once this merged, then we should create another PR that _splits_ (by concern bearer) the emitted messages and/or reports them different or maybe even at different channels. That PR will also need to be ported to master as well, I guess.

maven-core/src/main/java/org/apache/maven/plugin/internal/DefaultPluginDependenciesResolver.java
```diff
@@ -109,34 +110,8 @@ public Artifact resolve(Plugin plugin, List<RemoteRepository> repositories, Repo
...
-                        .collect(Collectors.toSet());
-
-                if (!mavenArtifacts.isEmpty()) {
-                    pluginValidationManager.reportPluginValidationIssue(
-                            session,
-                            pluginArtifact,
-                            "Plugin should declare these Maven artifacts in `provided` scope: " + mavenArtifacts);
-                }
+            for (MavenPluginDependenciesValidator dependenciesValidator : dependenciesValidators) {
+                dependenciesValidator.validate(session, pluginArtifact, result);
```
slawekjaranowski, line 114: I think that dependency list returned by `repoSystem.readArtifactDescriptor` is different than `repoSystem.collectDependencies`

But for building plugin classloder we use output of `resolveInternal` method ... so maybe we should check dependencies returned by `repoSystem.collectDependencies`

cstamas: readArtifactDescriptor == POM
resolveInternal = collection and resolving of POM (that above) augmented with things like WagonExcluder, ExclusionsDependencyFilter (excludes core and extension exported artifacts).

So, first is kinda "raw" while second is "processed", and due exclusions, 2nd will have core stuff excluded always,

slawekjaranowski: Not every artifacts is excluded, pleas look: https://gist.github.com/slawekjaranowski/c65612fca4094ec8a2d09c6db3b8f78a

like this:

```
[DEBUG] Populating class realm plugin>org.apache.maven.plugins:maven-war-plugin:3.4.0-SNAPSHOT
...
[DEBUG]   Included: org.sonatype.sisu:sisu-inject-bean:jar:1.4.2
[DEBUG]   Included: org.sonatype.sisu:sisu-guice:jar:noaop:2.1.7
[DEBUG]   Included: org.codehaus.plexus:plexus-component-annotations:jar:1.5.5
[DEBUG]   Included: org.sonatype.plexus:plexus-sec-dispatcher:jar:1.3
[DEBUG]   Included: org.sonatype.plexus:plexus-cipher:jar:1.4
...
```

slawekjaranowski: I know that every core artifacts are stripped ... 
but we check if plugin has such on dependency list - so we have two different source of checked and used dependencies 

In my example we have:
```
[DEBUG]    org.apache.maven.shared:maven-mapping:jar:3.0.0:compile
[DEBUG]       org.apache.maven:maven-core:jar:3.0:compile
```
but validator do not detect that `maven-core` is in compile scope on dependencies

cstamas: @slawekjaranowski unsure what dependency plugin uses to display tree, but (to me) it seems it lacks information. Here is the 'verbose" tree that current resolver emits (in verbose mode): Don't be mislead by war version, as there is java snippet as well, I manipulate war POM BEFORE collect (basically making maven-core dep provided in war 3.3.2.

My point is, that "enhanced" DependencyGraphDumper (since 1.9.8) prints out extra info (more than dep plugin). 

This is the "verbose" (and dirty) tree as there was no graph transformation applied). The fact that line has `(xxx)` at ends, means that node **is eliminated**, but is left for verbosity.

So we may want to look into here, but IMHO there is no "two sources" and the dep-tree seems to lack information...

slawekjaranowski: @cstamas - ok, let's leave as is - we can improve later if needed

slawekjaranowski (approved):

maven-core/src/main/java/org/apache/maven/plugin/internal/PlexusContainerDefaultDependenciesValidator.java
hgschmie: I meant "we maintain a list of EOLed artifacts here and report them if found in the tree", not "look up one specifically". I guess we can rewrite this if a second "this is EOL" artifact shows up.

jira-importer:
Resolve #9198

PR #1121, opened 2023-05-24
[3.9.x][MNG-7587] Upgrade to sisu 0.9.0.M2

gnodet:
JIRA issue: https://issues.apache.org/jira/browse/MNG-7587

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -143,6 +143,7 @@ under the License.
     <cipherVersion>2.0</cipherVersion>
     <jxpathVersion>1.3</jxpathVersion>
     <resolverVersion>1.9.10</resolverVersion>
+    <sisuVersion>0.9.0.M2</sisuVersion>
     <slf4jVersion>1.7.36</slf4jVersion>
     <xmlunitVersion>2.9.1</xmlunitVersion>
     <powermockVersion>2.0.9</powermockVersion>


slachiewicz (approved):

cstamas:
What about:
* using sisu "no_asm" classifier
* using guice "classes" classifier
* declaring asm (whatever version) as maven dep?

kwin:
> What about:

> using sisu "no_asm" classifier
> using guice "classes" classifier
> declaring asm (whatever version, for start 9.5) as maven dep?

I would defer to add this additional complexity until there is a new version of ASM released which is not integrated in Sisu.

gnodet:
> > What about:
> 
> > using sisu "no_asm" classifier
> > using guice "classes" classifier
> > declaring asm (whatever version, for start 9.5) as maven dep?
> 
> I would defer to add this additional complexity until there is a new version of ASM released which is not (yet) integrated in Sisu.

We can't defer.  We use juice 5.1.0 which embeds asm 9.2, so we need to use version which does not embed asm.

cstamas (approved):
For 3.9.x "simple upgrade" is the way to go, and let's to the "asm trick" on master

jira-importer:
Resolve #8936

PR #1124, opened 2023-05-24
[MNG-7097] Plugin Dependency Resolution

cstamas:
Idea: maven plugin dependency resolution could do it smarter: currently, it _resolves_ all dependencies of a plugin, only to leave them (exclude from classpath) if they are present in Maven Core and are exported. So, why not filter ahead instead?

Plugin resolution filter, part one: The Filter

## Simple "tests" for numbers

CMD: `mvn clean install -Dmaven.repo.local=local -Drat.skip` (and always emptying local repo naturally)

### Building maven-3.9.x

The local repo size and file count contained.

w/ 3.9.2:
Size: 147MB
Files: 3301

w/ 3.9.3 + this PR:
Size: 131MB
Files: 2506

### Building maven-3.8.x

w/ 3.9.2:
Size: 142MB
Files: 4500

w/ 3.9.3 + this PR:
Size: 117MB
Files: 3362

Conclusion: the "saving" (in transported byte count and file count, downloaded from remote) is bigger as "older" the plugins are (3.9.x uses parent 39, while 3.8.x parent 36). Essentially, putting maven bits to "provided" scope results in same result as this PR makes it. The difference is that this filter works for ALL, even old, non-provided plugins as well.

---

https://issues.apache.org/jira/browse/MNG-7097

### DIFF
--- a/maven-core/src/main/java/org/apache/maven/plugin/internal/DefaultPluginDependenciesResolver.java
+++ b/maven-core/src/main/java/org/apache/maven/plugin/internal/DefaultPluginDependenciesResolver.java
@@ -19,12 +19,14 @@
 package org.apache.maven.plugin.internal;
 
 import java.util.Collection;
+import java.util.HashSet;
 import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
 
 import org.apache.maven.RepositoryUtils;
+import org.apache.maven.extension.internal.CoreExportsProvider;
 import org.apache.maven.model.Dependency;
 import org.apache.maven.model.Plugin;
 import org.apache.maven.plugin.PluginResolutionException;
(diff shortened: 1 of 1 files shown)


maven-core/src/main/java/org/apache/maven/plugin/internal/DefaultPluginDependenciesResolver.java
```diff
@@ -300,4 +311,40 @@ public boolean visitLeave(DependencyNode node) {
...
+
+    private static class CoreDependencySelector implements DependencySelector {
+        private final HashSet<String> coreArtifacts;
+
+        private CoreDependencySelector(final Collection<String> coreArtifacts) {
+            this.coreArtifacts = new HashSet<>(coreArtifacts);
+        }
+
+        @Override
+        public boolean selectDependency(final org.eclipse.aether.graph.Dependency dependency) {
```
gnodet, line 323: I'm not a big fan of the `final` keyword on arguments and local variable... They are actually only useful on fields.

gnodet (approved):

cstamas:
As expected, with "part one" IT results:
```
Error:  Tests run: 889, Failures: 2, Errors: 5, Skipped: 80
```

Failed ITs:
* MavenITmng4274PluginRealmArtifactsTest.testit
* MavenITmng5783PluginDependencyFiltering.testSLF4j

Errored ITs:
* MavenITmng3536AppendedAbsolutePathsTest
* MavenITmng3693PomFileBasedirChangeTest.testitMNG3693
* MavenITmng4360WebDavSupportTest.testitJackrabbitBasedImpl
* MavenITmng4360WebDavSupportTest.testitSlideBasedImpl:
* MavenITmng7045DropUselessAndOutdatedCdiApiTest.testShouldNotLeakCdiApi:

gnodet:
I think the IT bootstrap needs some adjustments as some artifacts are missing.

PR #1128, opened 2023-05-25
[3.9.x][MNG-7795] IllegalArgumentException: 'other' has different root durin…

adangel:
…ing plugin validation

Checks the paths before relativizing them.

https://issues.apache.org/jira/browse/MNG-7795

       pulling in other changes.

### DIFF
--- a/maven-core/src/main/java/org/apache/maven/plugin/internal/DefaultPluginValidationManager.java
+++ b/maven-core/src/main/java/org/apache/maven/plugin/internal/DefaultPluginValidationManager.java
@@ -22,6 +22,7 @@ import javax.inject.Named;
 import javax.inject.Singleton;
 
 import java.io.File;
+import java.nio.file.Path;
 import java.util.*;
 import java.util.concurrent.ConcurrentHashMap;
 
@@ -222,16 +223,14 @@ public final class DefaultPluginValidationManager extends AbstractEventSpy imple
                 if (location.contains("://")) {
                     stringBuilder.append(" (").append(location).append(")");
                 } else {
-                    File rootBasedir = mavenSession.getTopLevelProject().getBasedir();
-                    File locationFile = new File(location);
-                    if (location.startsWith(rootBasedir.getPath())) {
-                        stringBuilder
-                                .append(" (")
-                                .append(rootBasedir.toPath().relativize(locationFile.toPath()))
-                                .append(")");
-                    } else {
-                        stringBuilder.append(" (").append(location).append(")");
+                    Path rootBasedir =
+                            mavenSession.getTopLevelProject().getBasedir().toPath();
+                    Path locationPath =
+                            new File(location).toPath().toAbsolutePath().normalize();
+                    if (locationPath.startsWith(rootBasedir)) {
+                        locationPath = rootBasedir.relativize(locationPath);
                     }
+                    stringBuilder.append(" (").append(locationPath).append(")");
                 }
             }
             stringBuilder.append(" @ line ").append(inputLocation.getLineNumber());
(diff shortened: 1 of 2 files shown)


maven-core/src/main/java/org/apache/maven/plugin/internal/DefaultPluginValidationManager.java
```diff
@@ -247,7 +247,11 @@ private String pluginOccurrence(MavenSession mavenSession) {
         File currentPom = prj.getFile();
         if (currentPom != null) {
             File rootBasedir = mavenSession.getTopLevelProject().getBasedir();
-            result += " (" + rootBasedir.toPath().relativize(currentPom.toPath()) + ")";
+            if (currentPom.getPath().startsWith(rootBasedir.getPath())) {
```
gnodet, line 250: I wonder if the path should be absoluted / normalized.  Also, this should be done against the new `Session#getRootDirectory()` imho:
```
            Path root = mavenSession.getRootDirectory();
            Path current = currentPom.toPath().toAbsolutePath().normalize();
            if (current.startsWith(root)) {
                current = root.relativize(current);
            }
            result += " (" + current + ")";
```

gnodet (changes requested):

maven-core/src/main/java/org/apache/maven/plugin/internal/DefaultPluginValidationManager.java
gnodet: Note that `rootDirectory` always refer to the root of the project, while `topDirectory` refer to the current path (those may be different when building a subset of a project when using `cd xxx && mvn ...` or `mvn -f xxx ...`

adangel: `rootDirectory` seems to be available for maven 4 only...

gnodet: On 3.9.x, you could use `session.getRequest().getMultiModuleProjectDirectory()`.

@cstamas what's your take on that ?  would you rather keep poms relative to the top directory ?

adangel: I've refactored the parts using your ideas (determine the relative path first before adding to the result string). The commits should probably be squashed before merged.

I've changed it also for `pluginDeclaration()`, to have consistent behavior.

Open question is now: Should I use `session.getRequest().getMultiModuleProjectDirectory().toPath()` instead of `mavenSession.getTopLevelProject().getBasedir().toPath()` for the root directory (for both pluginDeclaration() and pluginOccurrence()?

gnodet: I think we need to be consistent with
https://github.com/apache/maven/blob/e9d570889ffce29c3f92494510499ae0e32a02cf/maven-embedder/src/main/java/org/apache/maven/cli/event/ExecutionEventLogger.java#L299-L300
So better continue using the top directory.  However, does the ExecutionEventLogger suffer from the same problem ?

adangel: > And can we make the local variable name consistent, i.e. name it `topBasedir` instead of `rootBasedir` ?

Done.

I've also updated ExecutionEventLogger. It seems, this was the only other place, where Path::relativize is used.

cstamas (approved):

gnodet (approved):

PR #1132, opened 2023-05-28
[MNG-7797] Return BRIEF mode

hgschmie:
Make BRIEF a legal value again.

---

https://issues.apache.org/jira/browse/MNG-7797


rmannibucau:
Can be beneficial to comment it is to not break existing users otherwise code looks wrong. Except that +1 to get it asap to restore it.

cstamas (approved):

PR #1134, opened 2023-06-01
[MNG-7698] Support comments in maven.config

gnodet:
JIRA issue: https://issues.apache.org/jira/projects/MNG/issues/MNG-7698


cstamas (approved):

cstamas:
Just FTR: _these very same_ "comments" (empty lines and "starts with #") allows resolver as well in various files. So this is good change IMHO.

slawekjaranowski (approved):

PR #1137, opened 2023-06-01
[MNG-7604] Removal of pom.* interpolation makes some older plugins defunct

gnodet:
This reverts commit b2a21f12f8a0598dd4a286177e340bda0148e9ba.

JIRA issue: https://issues.apache.org/jira/browse/MNG-7604

### DIFF
--- a/maven-model-builder/src/main/java/org/apache/maven/model/building/ModelBuildingRequest.java
+++ b/maven-model-builder/src/main/java/org/apache/maven/model/building/ModelBuildingRequest.java
@@ -53,7 +53,7 @@ public interface ModelBuildingRequest {
     int VALIDATION_LEVEL_MAVEN_3_0 = 30;
 
     /**
-     * Denotes validation as performed by Maven 3.1. This validation level is meant for new projects.
+     * Denotes validation as performed by Maven 3.1. This validation level is meant for existing projects.
      */
     int VALIDATION_LEVEL_MAVEN_3_1 = 31;
 
--- a/maven-model-builder/src/main/java/org/apache/maven/model/building/ModelBuildingRequest.java
+++ b/maven-model-builder/src/main/java/org/apache/maven/model/building/ModelBuildingRequest.java
@@ -58,9 +58,14 @@ public interface ModelBuildingRequest {
     int VALIDATION_LEVEL_MAVEN_3_1 = 31;
 
     /**
+     * Denotes validation as performed by Maven 3.1. This validation level is meant for new projects.
+     */
+    int VALIDATION_LEVEL_MAVEN_4_0 = 40;
+
+    /**
      * Denotes strict validation as recommended by the current Maven version.
      */
-    int VALIDATION_LEVEL_STRICT = VALIDATION_LEVEL_MAVEN_3_1;
+    int VALIDATION_LEVEL_STRICT = VALIDATION_LEVEL_MAVEN_4_0;
 
     /**
      * Gets the file model to build (with profile activation).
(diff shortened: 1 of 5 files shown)


maven-model-builder/src/main/java/org/apache/maven/model/building/ModelBuildingRequest.java
```diff
@@ -57,10 +57,15 @@ public interface ModelBuildingRequest {
      */
     int VALIDATION_LEVEL_MAVEN_3_1 = 31;
 
+    /**
+     * Denotes validation as performed by Maven 3.1. This validation level is meant for new projects.
```
mthmulders, line 61:
```suggestion
     * Denotes validation as performed by Maven 4.0. This validation level is meant for new projects.
```

mthmulders: Also, for `VALIDATION_LEVEL_MAVEN_3_1` I think the Javadoc should now read "This validation level is meant for existing projects."

mthmulders (changes requested):
Thanks for creating this one. I had no idea how to solve it, but this seems like a nice approach.

This text in the PR body caught me off guard

> This reverts commit https://github.com/apache/maven/commit/b2a21f12f8a0598dd4a286177e340bda0148e9ba.

If I understand correctly, this PR doesn't _revert_ b2a21f12, but refines or enhances it by _only_ performing the validation if it needs to.

Out of curiosity: how does Maven determine which validation level to apply?

gnodet:
> Thanks for creating this one. I had no idea how to solve it, but this seems like a nice approach.
> 
> This text in the PR body caught me off guard
> 
> > This reverts commit [b2a21f1](https://github.com/apache/maven/commit/b2a21f12f8a0598dd4a286177e340bda0148e9ba).
> 
> If I understand correctly, this PR doesn't _revert_ [b2a21f1](https://github.com/apache/maven/commit/b2a21f12f8a0598dd4a286177e340bda0148e9ba), but refines or enhances it by _only_ performing the validation if it needs to.

Right, my original idea was to do a two step thing ...

 
> Out of curiosity: how does Maven determine which validation level to apply?

The default value is `VALIDATION_LEVEL_STRICT`.  When reading poms which are external to the workspace, the code usually set the validation level on the `ModelBuildingRequest` to `VALIDATION_LEVEL_MINIMAL`, as done [here](https://github.com/apache/maven/blob/8150c62749a637a98fbb9489fb8162034956220b/maven-resolver-provider/src/main/java/org/apache/maven/repository/internal/DefaultArtifactDescriptorReader.java#L196).  I don't think other values are actually used.

mthmulders:
> > Out of curiosity: how does Maven determine which validation level to apply?
> 
> The default value is `VALIDATION_LEVEL_STRICT`. When reading poms which are external to the workspace, the code usually set the validation level on the `ModelBuildingRequest` to `VALIDATION_LEVEL_MINIMAL`, as done [here](https://github.com/apache/maven/blob/8150c62749a637a98fbb9489fb8162034956220b/maven-resolver-provider/src/main/java/org/apache/maven/repository/internal/DefaultArtifactDescriptorReader.java#L196). I don't think other values are actually used.

Thanks for clarifying!

cstamas (approved):
This change affects only master (and the preceding "removal" of interpolation as well)?

gnodet:
> This change affects only master (and the preceding "removal" of interpolation as well)?

3.9 is not affected afaik, https://issues.apache.org/jira/browse/MNG-7244 was only targeted at 4.x

jira-importer:
Resolve #8848

PR #1138, opened 2023-06-01
[MNG-7548] Kill off "legacy" repository metadata support

gnodet:
JIRA issue: https://issues.apache.org/jira/projects/MNG/issues/MNG-7548


cstamas (approved):
LGTM, but as issue mentions, we have other places to look into as well

gnodet:
> LGTM, but as issue mentions, we have other places to look into as well

@cstamas What are you referring it ? I read the issue again but I don't see...

cstamas:
Ah, I meant "we have many other places **in ecosystem**" like that invoker plugin etc. So we need to search'n'destroy all those places.

jira-importer:
Resolve #8902

PR #1139, opened 2023-06-01
[MNG-7059][MNG-4645] Move Central repo definition out of Maven's core so it can be more easily changed

gnodet:
JIRA issue: https://issues.apache.org/jira/browse/MNG-4645
IT PR: https://github.com/apache/maven-integration-testing/pull/266
Supersedes: https://github.com/apache/maven/pull/419

### DIFF
--- a/apache-maven/src/assembly/maven/conf/settings.xml
+++ b/apache-maven/src/assembly/maven/conf/settings.xml
@@ -43,9 +43,9 @@ under the License.
  | values (values used when the setting is not specified) are provided.
  |
  |-->
-<settings xmlns="http://maven.apache.org/SETTINGS/1.2.0"
+<settings xmlns="http://maven.apache.org/SETTINGS/1.3.0"
           xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
-          xsi:schemaLocation="http://maven.apache.org/SETTINGS/1.2.0 https://maven.apache.org/xsd/settings-1.2.0.xsd">
+          xsi:schemaLocation="http://maven.apache.org/SETTINGS/1.3.0 https://maven.apache.org/xsd/settings-1.3.0.xsd">
   <!-- localRepository
    | The path to the local repository maven will use to store artifacts.
    |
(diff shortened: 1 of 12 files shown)


rmannibucau:
Do we intend to break builds overriding settings? Guess injection logic should be kept until we force an implicit settings.xml or do not define any implicit packaging/lifecycle, no?

gnodet:
> Do we intend to break builds overriding settings? Guess injection logic should be kept until we force an implicit settings.xml or do not define any implicit packaging/lifecycle, no?

That's a really good question.  I've added back the injection of maven central in case the settings read have no repositories defined.  The most usual use case will be if the global settings have been overridden by the command line.

gnodet:
Kinda weird that the only IT failing is because it's now using the newly accepted syntax, but we're testing a failure: https://github.com/apache/maven-integration-testing/blob/941353757073754cedd91f49c3d1d9c216adda0d/core-it-suite/src/test/resources/mng-3748/settings.xml ...

apache-maven/src/assembly/maven/conf/settings.xml
```diff
@@ -166,6 +166,38 @@ under the License.
...
+  <pluginRepositories>
+    <pluginRepository>
+      <id>central</id>
+      <name>Central Repository</name>
+      <url>https://repo.maven.apache.org/maven2</url>
+      <snapshots>
+        <enabled>false</enabled>
+      </snapshots>
+      <releases>
+        <updatePolicy>never</updatePolicy>
```
cstamas, line 195: why never? why the diff between central repo and central plugin repo?

cstamas:
Does this means maven4 settings.xml will not be parsable by Maven3? So projects will have to "take a stance", and be "maven4 only" or 'maven3 only"?

apache-maven/src/assembly/maven/conf/settings.xml
gnodet: That's the current value, unchanged. 
See https://github.com/apache/maven/pull/1142/commits/04268f9dd323dd31630a55004ee4ca10e412c360 for the follow-up commit which changes the value.

gnodet:
> Does this means maven4 settings.xml will not be parsable by Maven3? So projects will have to "take a stance", and be "maven4 only" or 'maven3 only"?

That's the `settings.xml` which will come with the maven 4 distribution and located in `[MVN_HOME]/conf/settings.xml`, so I don't think this is an issue.  If people want to modify their own user settings and leverage new features, then yes, but as a workaround, it's always possible to use a profile activated by default instead.

cstamas (approved):

rmannibucau (approved):

PR #1142, opened 2023-06-02
[MNG-6112] Make central update policy coherent

gnodet:
- [MNG-6112] Align artifacts and plugins policies using the default value
This PR is based on #1139 , so please review only the last commit https://github.com/apache/maven/pull/1142/commits/04268f9dd323dd31630a55004ee4ca10e412c360

### DIFF
--- a/apache-maven/src/assembly/maven/conf/settings.xml
+++ b/apache-maven/src/assembly/maven/conf/settings.xml
@@ -43,9 +43,9 @@ under the License.
  | values (values used when the setting is not specified) are provided.
  |
  |-->
-<settings xmlns="http://maven.apache.org/SETTINGS/1.2.0"
+<settings xmlns="http://maven.apache.org/SETTINGS/1.3.0"
           xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
-          xsi:schemaLocation="http://maven.apache.org/SETTINGS/1.2.0 https://maven.apache.org/xsd/settings-1.2.0.xsd">
+          xsi:schemaLocation="http://maven.apache.org/SETTINGS/1.3.0 https://maven.apache.org/xsd/settings-1.3.0.xsd">
   <!-- localRepository
    | The path to the local repository maven will use to store artifacts.
    |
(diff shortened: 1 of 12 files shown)


apache-maven/src/assembly/maven/conf/settings.xml
```diff
@@ -257,9 +286,11 @@ under the License.
   <!-- activeProfiles
    | List of profiles that are active for all builds.
    |
+  -->
   <activeProfiles>
-    <activeProfile>alwaysActiveProfile</activeProfile>
-    <activeProfile>anotherAlwaysActiveProfile</activeProfile>
+    <activeProfile>maven:core:central-repo</activeProfile>
```
slawekjaranowski, line 291: Where is used?

gnodet: I need to revert that part, that comes from the original commit.

gnodet: So this PR is based on #1139, please check only commit https://github.com/apache/maven/pull/1142/commits/1d7045dcf631926d3a6515bc8527d5e3241cc482

rmannibucau (approved):

maven-settings-builder/src/main/java/org/apache/maven/settings/building/DefaultSettingsBuilder.java
```diff
@@ -104,6 +106,26 @@ public SettingsBuildingResult build(SettingsBuildingRequest request) throws Sett
...
+            central.setId("central");
+            central.setName("Central Repository");
+            central.setUrl("https://repo.maven.apache.org/maven2");
+            RepositoryPolicy disabledPolicy = new RepositoryPolicy();
+            disabledPolicy.setEnabled(false);
+            central.setSnapshots(disabledPolicy);
+            userSettings.getRepositories().add(central);
+            central = central.clone();
+            RepositoryPolicy updateNeverPolicy = new RepositoryPolicy();
+            disabledPolicy.setUpdatePolicy("never");
```
slawekjaranowski, line 124: here we override update policy - but in settings.xml we don't such it

PR #1143, opened 2023-06-02
[MNG-7801] Update the model doc to reflect the real default policy

gnodet:


gnodet:
Actually, it seems the default is `daily`. The reason for the mix-up is that the default value in the resolver is `never`, but maven does never pass `null` to the resolver, as indicated by https://github.com/apache/maven/blob/8150c62749a637a98fbb9489fb8162034956220b/maven-core/src/main/java/org/apache/maven/internal/impl/DefaultRepositoryFactory.java#L68, where `daily` is used by default.

jira-importer:
Resolve #9193

PR #1144, opened 2023-06-02
[MNG-7802] Fix behaviour of the maven update policy

gnodet:


cstamas:
Looks good, but:
* the CLI `-U` should be changed (not snapshot update but all update, I mean change the help text)
* as IT failures show, IF `-U` present, do what happened before (stick ALWAYS to session), if not present, do partially what in PR

gnodet:
Will update the PR when https://github.com/apache/maven-resolver/pull/297 is released.

cstamas:
To be done in resolver 2.0.0, please once out upgrade to this instead https://issues.apache.org/jira/browse/MRESOLVER-377

PR #1147, opened 2023-06-05
[MNG-7804] Ability to order plugin executions using the priority xml element

gnodet:
JIRA issue: https://issues.apache.org/jira/browse/MNG-7804
IT PR: https://github.com/apache/maven-integration-testing/pull/268

Note that this introduce a new element in a new 4.1.0 model.

### DIFF
--- a/api/maven-api-model/src/main/mdo/maven.mdo
+++ b/api/maven-api-model/src/main/mdo/maven.mdo
@@ -2634,7 +2634,7 @@
         </field>
         <field xml.transient="true">
           <name>priority</name>
-          <version>4.0.0+</version>
+          <version>4.0.0</version>
           <type>int</type>
           <description>
             <![CDATA[
@@ -2643,6 +2643,15 @@
             it is not part of the public API. In particular, this property can be changed or deleted without prior
             notice.
             ]]>
+          </description>
+        </field>
+        <field>
+          <name>priority</name>
+          <version>4.1.0+</version>
+          <type>int</type>
+          <description>The priority of this execution compared to other executions which are bound to the same phase.
+            Executions derived from the default lifecycle have a negative priority by default so that they are executed
+            before any custom plugin executions.
           </description>
         </field>
         <field>
--- a/maven-model/pom.xml
+++ b/maven-model/pom.xml
@@ -90,9 +90,15 @@ under the License.
             <id>modello-site-docs</id>
             <goals>
               <goal>xdoc</goal>
-              <goal>xsd</goal>
             </goals>
             <phase>pre-site</phase>
+          </execution>
+          <execution>
+            <id>modello-schema</id>
+            <goals>
+              <goal>xsd</goal>
+            </goals>
+            <phase>generate-resources</phase>
           </execution>
           <execution>
             <id>model-v3</id>
--- a/maven-model/src/test/java/org/apache/maven/model/v4/MavenModelVersionTest.java
+++ b/maven-model/src/test/java/org/apache/maven/model/v4/MavenModelVersionTest.java
@@ -19,8 +19,12 @@
 package org.apache.maven.model.v4;
 
 import java.io.InputStream;
+import java.util.Collections;
 
+import org.apache.maven.api.model.Build;
 import org.apache.maven.api.model.Model;
+import org.apache.maven.api.model.Plugin;
+import org.apache.maven.api.model.PluginExecution;
 import org.junit.jupiter.api.BeforeAll;
 import org.junit.jupiter.api.Test;
 
@@ -57,6 +61,15 @@ class MavenModelVersionTest {
     @Test
     void testV4ModelPreserveModelVersion() {
         Model m = model.withPreserveModelVersion(true);
+        assertEquals("4.1.0", new MavenModelVersion().getModelVersion(m));
+    }
+
+    @Test
+    void testV4V4ModelPriority() {
+        Model m = model.withBuild(Build.newInstance()
+                .withPlugins(Collections.singleton(Plugin.newInstance()
+                        .withExecutions(Collections.singleton(
+                                PluginExecution.newInstance().withPriority(5))))));
         assertEquals("4.1.0", new MavenModelVersion().getModelVersion(m));
     }
 }
--- a/src/mdo/model-version.vm
+++ b/src/mdo/model-version.vm
@@ -138,7 +138,7 @@ public class ${className} {
         #end
         #foreach ( $field in $classToFields.get( $class ) )
             #if ( $field.isManyMultiplicity() )
-                #if ( $classToFields.containsKey( $model.getClass( $field.type ) ) )
+                #if ( $classToFields.containsKey( $field.toClass ) )
             $pfx ${var}.get${Helper.capitalise($field.name)}().stream().anyMatch(this::is_${v}) // ${class.name} : ${field.name}
                 #else
             $pfx !${var}.get${Helper.capitalise($field.name)}().isEmpty() // ${class.name} : ${field.name}


api/maven-api-model/src/main/mdo/maven.mdo
```diff
@@ -2646,6 +2646,15 @@
             ]]>
           </description>
         </field>
+        <field>
+          <name>priority</name>
+          <version>4.2.0+</version>
```
hboutemy, line 2651: IIUC, this is a new XML element, then not compatible with previous pom.xml schema
is this change for Maven 4 only? and with proper removing in consumer POM?

did you try to define the field as XML attribute to avoid the issue (and be able to add the feature to Maven 3.9)?

hboutemy (changes requested):
I'd appreciate to have a review on #1145, that is necessary both if this PR is merged in the future or not

api/maven-api-model/src/main/mdo/maven.mdo
```diff
@@ -2646,6 +2646,15 @@
...
+          <type>int</type>
+          <description>The priority of this execution compared to other executions which are bound to the same phase.
+            Executions derived from the default lifecycle have a negative priority by default so that they are executed
+            before any custom plugin executions.
```
hboutemy, line 2655: nice description
idea: IIRC, priority is set to -1 for lifecycle bindings, where default priority is 0
while introducing this priority, using -1000 instead of -1 would permit users to insert between lifecycle binding and default 0 (and would start to clarify why priority vales is not ideal, but just an easy opportunistic feature to use only when necessary)

note for self: I also need to check that when bindings have been defined with default priority, we can in a child pom just override the priority by just putting execution id

api/maven-api-model/src/main/mdo/maven.mdo
gnodet: Not at all.

This element can not be removed if present in a POM, as this will change the semantics of the POM.  The build/consumer POM is all about inferred (or removal of unused) data.  If we were to upload a project which defines such a new element/attribute but use it with a maven version that does not know about it, the build results could be different and that would be bad.  So I don't think we should upload a POM stripped of this element.

So I think a new POM is needed for that one. 4.1.0 is kinda used internally for the api v4 migration, so I went with 4.2.0.  This field is a bit specific as it does already exist in the 4.0.0 model, but it is flagged as `xml.transient=true`, which means it is not read from the xml, but there's a field + getter/setter already generated.

On the compatibility side, I think this defines a new 4.2.0 POM.  Note that the way the xml reader is generated, it has two modes: the first mode when reading projects being built where a strict validation is used, i.e. any unknown element will be rejected and cause a parsing failure.  When using this mode, i.e. when building a project with a 4.2.0 `modelVersion`, maven 3.x will reject it.  However, if uploaded on maven central and used as a dependency, the lenient parsing mode is used, where any unknown element or attribute is simply ignored (see the `checkUnknownElement` and `checkUnknownAttribute` of the `MavenXpp3Reader`).  That will mean that projects with a modelVersion of 4.2.0 will be able to be consumed by older maven 3.x tools / projects.

hboutemy: Maven is not the only one using these pom.xml, but many tools in the wild: that's why we always avoided adding XML elements, but only added XML attributes when really necessary = the only accepted changes (in addition to the "import" value for dependencyManagement scope
see attributes in https://maven.apache.org/ref/3.9.2/maven-model/maven.html

on build vs consumer, we'll need to again define what is consumer, but to me, this strict convention on available elements in published POMs will remain forever

gnodet: > Maven is not the only one using these pom.xml, but many tools in the wild: that's why we always avoided adding XML elements, but only added XML attributes when really necessary = the only accepted changes (in addition to the "import" value for dependencyManagement scope see attributes in https://maven.apache.org/ref/3.9.2/maven-model/maven.html
> 
> on build vs consumer, we'll need to again define what is consumer, but to me, this strict convention on available elements in published POMs will remain forever

What tools to you have in mind ? If our own parser can be quite lenient, I don't see why other tools would not be able to do the same.  We can't just decide to never update the model just because _other tools_ are using it.

gnodet: And what's the difference between adding an attribute and adding an element ? In our own parser, both are treated the same way (either discarded or rejected, depending if it's in strict/lenient mode).

api/maven-api-model/src/main/mdo/maven.mdo
gnodet: The priority for lifecycle plugin executions is negative, but not -1, see https://github.com/apache/maven/blob/master/maven-core/src/main/java/org/apache/maven/lifecycle/internal/DefaultLifecyclePluginAnalyzer.java#L191

api/maven-api-model/src/main/mdo/maven.mdo
hboutemy: it is the convention that has been defined from the beginning for the whole consumer ecosystem, and the reason build bom vs consumer pom has been introduced https://cwiki.apache.org/confluence/display/MAVEN/Build+vs+Consumer+POM

we can test options and go back to the M dev ML to have feedback from others: this change is not only about our internal code

gnodet: > it is the convention that has been defined from the beginning for the whole consumer ecosystem, and the reason build bom vs consumer pom has been introduced https://cwiki.apache.org/confluence/display/MAVEN/Build+vs+Consumer+POM
> 
> we can test options and go back to the M dev ML to have feedback from others: this change is not only about our internal code

I think the initial assumption and driver for the build/consumer feature, i.e. that "changing the POM version and publishing artifacts on Maven Central with this new model would break consumers using either older Maven versions or other build tools" was done with a wrong assumption.  I'll double check with more tests, but afaik, consumers using old versions of maven won't be broken when _using_ the artifacts as a dependency (they would be broken if using such a project as a parent though).  So yes, this definitely would mean that if you use a 4.2.0 modelVersion (though you should not be forced to unless you actually need something in it) imposes to use a recent maven when building a project that inherit it.

I'm not sure about other tools.

Anyway, I'll send a mail to dev@ to discuss that.

hboutemy, line 2651: thinking at it, if we keep it as a XML element, it should be version 5.0.0, not 4.2.0
and associated to Maven 5
= https://cwiki.apache.org/confluence/display/MAVEN/Maven+Core+Roadmap

gnodet: Could you clarify on the ML what you consider breaking and non breaking ?  I don't think we're on the same page, so this clearly needs some clarification.

gnodet:
A few questions have been asked by @ctubbsii [on the IT PR](https://github.com/apache/maven-integration-testing/pull/268#pullrequestreview-1478538677):

> - Does this require the POM/4.2.0 model? Which versions of Maven can use that POM version?

This does indeed require POM 4.1.0 that comes with Maven 4.

> - How does this work with priorities defined in a parent POM or profile?

The rule has not changed (as this `priority` element did exist but was hidden): it's done in the [DefaultLifecycleBindingsInjector](https://github.com/apache/maven/blob/0b3246381d12e69deef63888d0db274fb49c6697/maven-core/src/main/java/org/apache/maven/model/plugin/DefaultLifecycleBindingsInjector.java#L168-L171) class.

- What is the numerical type of the priority? Is it an integer, decimal, long, non-negative, strictly positive?

The type is `int` as defined [in the model](https://github.com/apache/maven/pull/1147/files#diff-ccc5b1903a864adb4a01d6072c9c2ff937c71fb06bea6e5ec3df6097f0d01199R2651).  The doc also hints to the fact that the value can be negative.

- What if two executions have the same priority? Is that an error, or is it arbitrarily ordered?

Arbitrarily (un-)ordered within a given priority.

gnodet:
@hboutemy would you mind another look please ?

cstamas (approved):

jira-importer:
Resolve #9186

PR #1150, opened 2023-06-08
[MNG-7794][MNG-7809] Deprecate RepositorySystem and make maven-compat optional

gnodet:
- Move maven-compat to jsr330
- Remove more ties to RepositorySystem
- Move more stuff to maven-compat
- Move PluginManager and LocalRepositoryNotAccessibleException to maven-compat
- Deprecate RepositorySystem [MNG-7794](https://issues.apache.org/jira/browse/MNG-7794)

This PR effectively cut the runtime dependency from maven-core to maven-compat.
When all plugins are upgrade to their latest version (which will come when releasing 
the parent POMs and upgrading to those), the maven distribution can build maven
after the removal of the maven-compat library from the distribution.


gnodet:
We could also move the `ArtifactFactory` and its implementation to maven-compat, but this would require getting rid of the [`MavenProject#createArtifacts`](https://github.com/apache/maven/blob/5367c0119ec8a88d1941449647f4315166a27f8b/maven-core/src/main/java/org/apache/maven/project/MavenProject.java#L1313-L1318) method has been deprecated 9 years ago afaik.

cstamas (approved):

PR #1172, opened 2023-06-16
[MNG-7816] Update parent to 40, align maven requirements with those from plugins

cstamas:
Update parent to 40. Also align plugin requirements with those from parent.

This lowers plugin validation "external" warnings from 12 to 3 (site, plexus-component-metadata and rat).

---

https://issues.apache.org/jira/browse/MNG-7816


slawekjaranowski (approved):

jira-importer:
Resolve #9103

PR #1179, opened 2023-06-22
Remove references to deprecated ReaderFactory in non generated code

gnodet:
- Prefix the key name for the TransformerContext
- Remove references to deprecated ReaderFactory in non generated code

       pulling in other changes.

### DIFF
--- a/maven-compat/src/main/java/org/apache/maven/artifact/repository/metadata/AbstractRepositoryMetadata.java
+++ b/maven-compat/src/main/java/org/apache/maven/artifact/repository/metadata/AbstractRepositoryMetadata.java
@@ -29,8 +29,8 @@ import org.apache.maven.artifact.repository.ArtifactRepository;
 import org.apache.maven.artifact.repository.ArtifactRepositoryPolicy;
 import org.apache.maven.artifact.repository.metadata.io.xpp3.MetadataXpp3Reader;
 import org.apache.maven.artifact.repository.metadata.io.xpp3.MetadataXpp3Writer;
-import org.codehaus.plexus.util.ReaderFactory;
-import org.codehaus.plexus.util.WriterFactory;
+import org.codehaus.plexus.util.xml.XmlStreamReader;
+import org.codehaus.plexus.util.xml.XmlStreamWriter;
 import org.codehaus.plexus.util.xml.pull.XmlPullParserException;
 
 /**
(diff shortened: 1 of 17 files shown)


maven-compat/src/main/java/org/apache/maven/artifact/repository/metadata/AbstractRepositoryMetadata.java
```diff
@@ -29,8 +29,8 @@
 import org.apache.maven.artifact.repository.ArtifactRepositoryPolicy;
 import org.apache.maven.artifact.repository.metadata.io.xpp3.MetadataXpp3Reader;
 import org.apache.maven.artifact.repository.metadata.io.xpp3.MetadataXpp3Writer;
-import org.codehaus.plexus.util.ReaderFactory;
-import org.codehaus.plexus.util.WriterFactory;
+import org.codehaus.plexus.util.xml.XmlStreamReader;
```
elharo, line 32: org.apache.commons.io.input.XmlStreamReader is preferred here

elharo (changes requested):

maven-compat/src/main/java/org/apache/maven/artifact/repository/metadata/AbstractRepositoryMetadata.java
gnodet: I don't want to introduce a new dependency for one class.  This one is in plexus-xml which won't go away any time soon.  We may be able to get rid of plexus-utils, but not plexus-xml which contains classes that are part of the maven 3 api.

elharo: It;s not just about going away. I have more expectation that the version in apache commons will be maintained and correct than the one in plexus.

gnodet: Actually, I think the usage of those classes is just wrong and they already have caused issues with tycho.  So I'll modify this PR to simply use plain InputStream / OutputStream.

elharo: best of all possible worlds. +1

PR #1182, opened 2023-06-23
[MNG-7823] Improve plugin validation level parsing

cstamas:
Changes:
* always parse it at session start
* hence, will WARN if needed there as well, and will warn once
* do not re-parse and possibly warn always, reuse parsed enum

---

https://issues.apache.org/jira/browse/MNG-7823

### DIFF
--- a/maven-core/src/main/java/org/apache/maven/plugin/internal/DefaultPluginValidationManager.java
+++ b/maven-core/src/main/java/org/apache/maven/plugin/internal/DefaultPluginValidationManager.java
@@ -86,8 +86,7 @@ public final class DefaultPluginValidationManager extends AbstractEventSpy imple
             if (executionEvent.getType() == ExecutionEvent.Type.SessionStarted) {
                 RepositorySystemSession repositorySystemSession =
                         executionEvent.getSession().getRepositorySession();
-                ValidationReportLevel level = parseValidationReportLevel(repositorySystemSession);
-                repositorySystemSession.getData().set(MAVEN_PLUGIN_VALIDATION_KEY, level);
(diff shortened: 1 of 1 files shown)


maven-core/src/main/java/org/apache/maven/plugin/internal/DefaultPluginValidationManager.java
```diff
@@ -83,13 +83,23 @@ private enum ValidationReportLevel {
...
+                repositorySystemSession.getData().set(MAVEN_PLUGIN_VALIDATION_KEY, level);
+            } else if (executionEvent.getType() == ExecutionEvent.Type.SessionEnded) {
                 reportSessionCollectedValidationIssues(executionEvent.getSession());
             }
         }
     }
 
     private ValidationReportLevel validationReportLevel(RepositorySystemSession session) {
+        return (ValidationReportLevel) session.getData()
+                .computeIfAbsent(MAVEN_PLUGIN_VALIDATION_KEY, () -> parseValidationReportLevel(session));
```
slawekjaranowski, line 98: Can be empty here?
We use the same key and assume that will contains correct type.
If code from onEvent will not execute we can have wrong type.

cstamas: no, this is session.data while string is in session.configProperties, two distinct maps

cstamas: changed to not reuse keys as this may cause misunderstanding

slawekjaranowski (approved):

slawekjaranowski (approved):

PR #1187, opened 2023-06-26
[MNG-7827] Ensure maven 4 Log API is the primary and documented API for mojos

rmannibucau:

       pulling in other changes.

### DIFF
--- a/maven-plugin-api/pom.xml
+++ b/maven-plugin-api/pom.xml
@@ -52,6 +52,20 @@ under the License.
       <groupId>org.codehaus.plexus</groupId>
       <artifactId>plexus-classworlds</artifactId>
     </dependency>
+
+    <dependency>
+      <groupId>${project.groupId}</groupId>
+      <artifactId>maven-api-core</artifactId>
+      <version>${project.version}</version>
+      <scope>provided</scope>
+    </dependency>
+
+    <dependency>
+      <groupId>org.slf4j</groupId>
+      <artifactId>slf4j-jdk14</artifactId>
+      <version>${slf4jVersion}</version>
+      <scope>test</scope>
+    </dependency>
   </dependencies>
 
   <build>
--- a/maven-plugin-api/src/main/java/org/apache/maven/plugin/AbstractMojo.java
+++ b/maven-plugin-api/src/main/java/org/apache/maven/plugin/AbstractMojo.java
@@ -18,10 +18,17 @@
  */
 package org.apache.maven.plugin;
 
+import java.lang.reflect.Constructor;
+import java.lang.reflect.InvocationTargetException;
+import java.lang.reflect.Method;
+import java.lang.reflect.Proxy;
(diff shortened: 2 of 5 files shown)


gnodet:
@rmannibucau not sure what the point of this PR is.  I would not force 3.x plugins to use the maven 4 api, I think it's better to keep it for 4.x plugins.

rmannibucau:
@gnodet there are two goals:

1. Drop the irrelevant documentation we never agreed upon to use SLF4J (api) for plugins
2. Ensure the default is the actual maven logging system and not stdout/stderr directly for other cases if it is present

v4 API is more a way to do the bridge easily and with less internal efforts (since it is hidden anyway)

gnodet:
> 2. Ensure the default is the actual maven logging system and not stdout/stderr directly for other cases if it is present

I don't understand the actual use case.  All mojos called by maven are [injected with a proper log](https://github.com/apache/maven/blob/e08c95a6735a82c7abc25e2dc67ea4773ed9d05d/maven-core/src/main/java/org/apache/maven/plugin/internal/DefaultMavenPluginManager.java#L557-L560) already.  The `SystemStreamLog` is really a fallback.  In which case would this new code actually be used ?

rmannibucau:
@gnodet good for me too, I wondered about this point too and if true we should delete SystemStreamLog and just fail instead probably. If useful it means we must forward the fallback to the actual expected logging system behind `Log` and now it is v4 mecanism thanks new `Log` API, this was the intent behind this PR but happy to just drop it too cause I share your reasoning.

cstamas:
> 1. Drop the irrelevant documentation we never agreed upon to use SLF4J (api) for plugins

Then your memory is lossy :smile: 
See https://lists.apache.org/thread/bh9vr9fwt2cqf3ltg88jm1fh5y0d57b0

Please, stop with this "change the logging..." madness.

rmannibucau:
@cstamas this is only one of the 4-5 explaining why we must not leak our default impl and where we didnt find an agreement in terms of api leak yes.
Note we dont want to change the logging, just keep the one we had and fix the bug introduced in the javadoc which doesnt reflects discussion outcomes.
Technically, there are multiple cons leaking slf4j but community wise a vote wouldnt pass but in force so both are bad.
Last is that maven didnt migrate its api to slf4j in plugin - never, we miss several steps to say it is done so current absusive comment is even wrong - so let's fix it, will impact nobody as of today and not mislead mojo dev.

gnodet:
Closing as duplicate of #793

slawekjaranowski:
source branch deleted

jira-importer:
Resolve #9092

slawekjaranowski:
@rmannibucau how it corresponding to 3.10.0 version?

rmannibucau:
@slawekjaranowski hmm no more sure but guess plan was to backport but was not done cause lack of time/holidays

PR #1192, opened 2023-06-30
Small changes

gnodet:


api/maven-api-core/src/main/java/org/apache/maven/api/Artifact.java
```diff
@@ -33,7 +33,7 @@ public interface Artifact {
 
     /**
      * Returns a unique identifier for this artifact.
-     * The identifier is composed of groupId, artifactId, version, classifier, extension.
+     * The identifier is composed of groupId, artifactId, extension, classifier, version.
```
elharo, line 36: and version

maven-core/src/main/java/org/apache/maven/plugin/internal/DefaultMavenPluginManager.java
```diff
@@ -229,7 +228,7 @@ private PluginDescriptor extractPluginDescriptor(Artifact pluginArtifact, Plugin
                 File pluginXml = new File(pluginFile, getPluginDescriptorLocation());
 
                 if (pluginXml.isFile()) {
-                    try (InputStream is = new BufferedInputStream(new FileInputStream(pluginXml))) {
+                    try (InputStream is = Files.newInputStream(pluginXml.toPath())) {
```
elharo, line 231: This is not buffered. Does that matter?

gnodet: No, the woodstox parser will always use an internal buffer.  I think it's best to hand it the raw stream rather than trying to make any assumption on what would be beneficial for it.

api/maven-api-core/src/main/java/org/apache/maven/api/Artifact.java
```diff
@@ -33,7 +33,7 @@ public interface Artifact {
 
     /**
      * Returns a unique identifier for this artifact.
-     * The identifier is composed of groupId, artifactId, version, classifier, extension.
+     * The identifier is composed of groupId, artifactId, extension, classifier and version.
```
elharo, line 36: Oxford comma: classifier, and version

api/maven-api-model/pom.xml
```diff
@@ -76,6 +76,26 @@ under the License.
           </excludes>
         </configuration>
       </plugin>
+      <plugin>
```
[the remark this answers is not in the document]
gnodet: It's not needed yet, but I'm writing an IT for https://github.com/apache/maven/pull/1160/ which will use the model to generate a hocon-based pom parser.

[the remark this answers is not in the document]
elharo: Probably belongs in that PR then.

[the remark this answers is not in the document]
gnodet: Ok, I'll remove from that one.

elharo (approved):

PR #1193, opened 2023-07-02
Improve API doc for ArtifactHandler

elharo:
Noticed this was a unclear and out of date

### DIFF
--- a/maven-artifact/src/main/java/org/apache/maven/artifact/handler/ArtifactHandler.java
+++ b/maven-artifact/src/main/java/org/apache/maven/artifact/handler/ArtifactHandler.java
@@ -19,10 +19,10 @@
 package org.apache.maven.artifact.handler;
 
 /**
- * An artifact handler provides metadata for an artifact derived from the dependency element that references the artifact:<ul>
- * <li>extension and classifier, to be able to download the file,</li>
+ * An artifact handler contains metadata derived from the dependency element that references the artifact:<ul>
+ * <li>extension and classifier, to be able to download the file</li>
  * <li>information on how to use the artifact: whether to add it to the classpath, or to take into account its
- * dependencies.</li>
+ * dependencies</li>
  * </ul>
  *
  * @author <a href="mailto:jason@maven.org">Jason van Zyl</a>
@@ -32,7 +32,7 @@ public interface ArtifactHandler {
     String ROLE = ArtifactHandler.class.getName();
 
     /**
-     * Returns the file name extension used for dependencies of that type;
+     * Returns the file name extension of the artifact;
      * e.g. "jar", "pom", "xml", etc.
      *
      * @return the file extension
--- a/maven-artifact/src/main/java/org/apache/maven/artifact/handler/ArtifactHandler.java
+++ b/maven-artifact/src/main/java/org/apache/maven/artifact/handler/ArtifactHandler.java
@@ -19,7 +19,7 @@
 package org.apache.maven.artifact.handler;
 
 /**
- * An artifact handler defines for a dependency type, defined as Plexus role:<ul>
+ * An artifact handler provides metadata for an artifact derived from the dependency element that references the artifact:<ul>
  * <li>extension and classifier, to be able to download the file,</li>
  * <li>information on how to use the artifact: whether to add it to the classpath, or to take into account its
  * dependencies.</li>
@@ -32,7 +32,8 @@ public interface ArtifactHandler {
     String ROLE = ArtifactHandler.class.getName();
 
     /**
-     * Get the file extension associated to the file represented by the dependency type.
+     * Returns the file name extension used for dependencies of that type;
+     * e.g. "jar", "pom", "xml", etc.
      *
      * @return the file extension
      */
@@ -41,7 +42,7 @@ public interface ArtifactHandler {
     String getDirectory();
 
     /**
-     * Get the classifier associated to the dependency type.
+     * Returns the classifier of the dependency.
      *
      * @return the classifier
      */
--- a/maven-artifact/src/main/java/org/apache/maven/artifact/handler/ArtifactHandler.java
+++ b/maven-artifact/src/main/java/org/apache/maven/artifact/handler/ArtifactHandler.java
@@ -19,10 +19,10 @@
 package org.apache.maven.artifact.handler;
 
 /**
- * An artifact handler contains metadata derived from the dependency element that references the artifact:<ul>
- * <li>extension and classifier, to be able to download the file</li>
- * <li>information on how to use the artifact: whether to add it to the classpath, or to take into account its
- * dependencies</li>
+ * An artifact handler contains information explaining how an artifact plugs into the Maven build:<ul>
+ * <li>Information needed to find the artifact file in a repository including extension and classifier</li>
+ * <li>Information on how to use the artifact as a dependency: whether to add it to the classpath, whether to load its
+ * dependencies transitively</li>
  * </ul>
  *
  * @author <a href="mailto:jason@maven.org">Jason van Zyl</a>
@@ -42,7 +42,7 @@ public interface ArtifactHandler {
     String getDirectory();
 
     /**
-     * Returns the classifier of the dependency.
+     * Returns the default classifier used if a different one is not set in pom.xml.
      *
      * @return the classifier
      */


maven-artifact/src/main/java/org/apache/maven/artifact/handler/ArtifactHandler.java
```diff
@@ -41,7 +42,7 @@ public interface ArtifactHandler {
     String getDirectory();
 
     /**
-     * Get the classifier associated to the dependency type.
+     * Returns the classifier of the dependency.
```
hboutemy, line 45: while we are at clarifying, it is the **default** classifier: user may override the value in his pom.xml dependency (by defining a classifier)

hboutemy (changes requested):
on packaging field, i think this is the most confusing parameter that lead to misunderstandings for year: packaging field is the "probable" packaging that created the artifact (if there is any probable one, as any packaging may create a source or javadoc artifact): I'm sure this API is unused; because I don't see how it can be reasonably used

on directory, I don't know at all what the intent was

everything I wrote in the documentation is based on years of thoughts trying to understand the initial intent of the code, that was written before i was involved...

thanks for helping digging into it

maven-artifact/src/main/java/org/apache/maven/artifact/handler/ArtifactHandler.java
elharo: so this is the classifier from the artifact's own pom.xml?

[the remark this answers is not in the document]
elharo: I'm not sure there is such a thing as a "dependency type" and I'd prefer to avoid that term completely. If it's confusing us, it will certainly confuse end users. I think there are three things in play here:

1. The binary artifact served from the repository, such as a jar file, a signature file, or a pom.xml file 

2. The pom.xml for the artifact. For a jar file this is the pom.xml from which the jar was built, though I'm not sure this is meaningful for anything but the regular jar containing .class files. I'm not sure if it makes sense for javadoc or source jars or non-jars. This is the pom found in the same directory  as the jar.

3. The pom.xml that imports the artifact using a dependency element. 

This dependency can declare a different packaging and other properties of the artifact. 

So does ArtifactHandler report number 3 or number 2 if both are present? Or does the answer to this question depend on which property we're looking at?

maven-artifact/src/main/java/org/apache/maven/artifact/handler/ArtifactHandler.java
```diff
@@ -19,7 +19,7 @@
 package org.apache.maven.artifact.handler;
 
 /**
- * An artifact handler defines for a dependency type, defined as Plexus role:<ul>
+ * An artifact handler provides metadata for an artifact derived from the dependency element that references the artifact:<ul>
  * <li>extension and classifier, to be able to download the file,</li>
```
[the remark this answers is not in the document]
elharo: Does this always provide the info from the artifact's own pom.xml?

I'm a little confused because if the classifier is used to download the file as seen in the first bullet point then it has to come from the dependency element that loads the artifact, not the artifact's own pom.xml. Maybe the  first bullet point is wrong?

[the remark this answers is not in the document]
gnodet: See below, the artefacts own pom.xml is not used at all here afaik.

maven-artifact/src/main/java/org/apache/maven/artifact/handler/ArtifactHandler.java
gnodet: The `ArtifactHandler` is unrelated to the artifact's own pom.xml.
The `ArtifactHandler` is used to determine a few informations about a dependency.  It is retrieved based on the [_dependency type_](https://maven.apache.org/ref/4-LATEST/apidocs/org/apache/maven/api/model/Dependency.html#getType()).  But the [`Dependency`](https://maven.apache.org/ref/4-LATEST/apidocs/org/apache/maven/api/model/Dependency.html) can define a classifier, which takes precedence over the default one provided by the artifact handler.

maven-artifact/src/main/java/org/apache/maven/artifact/handler/ArtifactHandler.java
```diff
@@ -19,10 +19,10 @@
 package org.apache.maven.artifact.handler;
 
 /**
- * An artifact handler defines for a dependency type, defined as Plexus role:<ul>
- * <li>extension and classifier, to be able to download the file,</li>
+ * An artifact handler contains metadata derived from the dependency element that references the artifact:<ul>
```
gnodet, line 22: I don't think that's true.
The metadata is not `derived from the dependency element`.  The set of artifact handlers is reduced and provides metadata related to the `dependency type`.

maven-artifact/src/main/java/org/apache/maven/artifact/handler/ArtifactHandler.java
```diff
@@ -19,10 +19,10 @@
...
+ * <li>extension and classifier, to be able to download the file</li>
```
gnodet, line 23: `extension and classifier` -> `défaut extension and classifier`

maven-artifact/src/main/java/org/apache/maven/artifact/handler/ArtifactHandler.java
```diff
@@ -19,10 +19,10 @@
...
  * <li>information on how to use the artifact: whether to add it to the classpath, or to take into account its
- * dependencies.</li>
+ * dependencies</li>
```
gnodet, line 25: Should we keep the `.` and add `,` at the end of the previous `<li>`.

maven-artifact/src/main/java/org/apache/maven/artifact/handler/ArtifactHandler.java
```diff
@@ -32,7 +32,8 @@ public interface ArtifactHandler {
     String ROLE = ArtifactHandler.class.getName();
 
     /**
-     * Get the file extension associated to the file represented by the dependency type.
+     * Returns the file name extension of the artifact;
```
gnodet, line 35: `the file name extension` -> `the default file name extension`

maven-artifact/src/main/java/org/apache/maven/artifact/handler/ArtifactHandler.java
gnodet: Agreed, `classifier` -> `default classifier`

gnodet (changes requested):
We should make it more explicit that the artifact handler provides default values.

maven-artifact/src/main/java/org/apache/maven/artifact/handler/ArtifactHandler.java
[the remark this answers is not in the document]
hboutemy: on extension, there is no way to override it: then it's directly the extension
that's one reason why most people are confused by the (lack of) difference between type and extension, while there is a difference between the default classifier and the effective classifier

maven-artifact/src/main/java/org/apache/maven/artifact/handler/ArtifactHandler.java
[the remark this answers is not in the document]
hboutemy: in fact, if Artifact in Maven core did not mix features from Artifact and Dependency (as Maven Resolver documented and solved 10 years later), this ArtifactHandler should have been named DependencyHandler
Because it's what that class does: define how to handle a dependency
IMHO, saying "artifact/dependency handler" in the javadoc could help understand reality

maven-artifact/src/main/java/org/apache/maven/artifact/handler/ArtifactHandler.java
hboutemy: finally, it's more `extension and default classifier`

maven-artifact/src/main/java/org/apache/maven/artifact/handler/ArtifactHandler.java
hboutemy: no, not that one :)

maven-artifact/src/main/java/org/apache/maven/artifact/handler/ArtifactHandler.java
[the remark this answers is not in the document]
gnodet: Note that the v4 api provides [`Type`](https://github.com/apache/maven/blob/master/api/maven-api-core/src/main/java/org/apache/maven/api/Type.java) as an equivalent to v3 `ArtifactHandler`, directly accessible using [`Dependency#getType()`](https://github.com/apache/maven/blob/master/api/maven-api-core/src/main/java/org/apache/maven/api/Dependency.java#L25-L31).  Maybe the api could be improved at the same time (replace _artifact type_ with _dependency type_ as a first step for example).

elharo (commented):
PTAL

gnodet (approved):

PR #1194, opened 2023-07-03
[MNG-6401] Support interpolation of the proxy port in settings.xml

gnodet:
JIRA: https://issues.apache.org/jira/browse/MNG-6401
IT PR: https://github.com/apache/maven-integration-testing/pull/275

### DIFF
--- a/api/maven-api-settings/src/main/mdo/settings.mdo
+++ b/api/maven-api-settings/src/main/mdo/settings.mdo
@@ -468,6 +468,7 @@
             <![CDATA[
             Whether this proxy configuration is the active one. Note: While the type of this field
             is {@code String} for technical reasons, the semantic type is actually {@code boolean}.
+            @see #isActive()
             ]]>
           </description>
           <type>String</type>
@@ -508,8 +509,9 @@
           <version>1.0.0+</version>
           <description>
             <![CDATA[
-            The proxy port. Note: While the type of this field is {@code String<} for technical
+            The proxy port. Note: While the type of this field is {@code String} for technical
             reasons, the semantic type is actually {@code int}.
+            @see #getPort()
             ]]>
           </description>
           <type>String</type>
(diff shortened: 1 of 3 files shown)


api/maven-api-settings/src/main/mdo/settings.mdo
```diff
@@ -459,17 +459,19 @@
...
       <fields>
-        <field>
-          <name>active</name>
+        <field xml.tagName="active">
+          <name>activeString</name>
           <version>1.0.0+</version>
           <required>false</required>
           <defaultValue>true</defaultValue>
           <description>
             <![CDATA[
```
elharo, line 468: CDATA not needed here

api/maven-api-settings/src/main/mdo/settings.mdo
```diff
@@ -502,15 +504,17 @@
           </description>
           <type>String</type>
         </field>
-        <field>
-          <name>port</name>
+        <field xml.tagName="port">
+          <name>portString</name>
           <version>1.0.0+</version>
           <description>
             <![CDATA[
```
elharo, line 511: CDATA not needed here

api/maven-api-settings/src/main/mdo/settings.mdo
```diff
@@ -535,6 +539,62 @@
           <type>String</type>
         </field>
       </fields>
+      <codeSegments>
+        <codeSegment>
+          <version>1.0.0/1.3.0</version>
+          <code>
+            <![CDATA[
```
elharo, line 546: CDATA not needed here

api/maven-api-settings/src/main/mdo/settings.mdo
```diff
@@ -535,6 +539,62 @@
...
+          <version>2.0.0+</version>
+          <code>
+            <![CDATA[
+    /**
+     * Indicates if this proxy is active or not.
+     * To allow interpolation of this field, this method lazily parses
+     * the {@link #getActiveString()} value as a boolean and defaults to {@code true}
+     * if not set.
+     *
+     * @return a boolean indicating whether this proxy is active or not
```
elharo, line 575: delete "or not"

api/maven-api-settings/src/main/mdo/settings.mdo
```diff
@@ -535,6 +539,62 @@
...
+    public void setActive(boolean active) {
+        setActiveString(String.valueOf(active));
+    }
+
+    public int getPort() {
+        return (getPortString() != null) ? Integer.parseInt(getPortString()) : 8080;
+    }
+
+    public void setPort(int port) {
+        setPortString(String.valueOf(port));
```
elharo, line 560: This allows any port value, including negative numbers and other ints that aren't legal port values. An IllegalArgumentException might helpful.


[IllegalArgumentException](https://docs.oracle.com/javase/8/docs/api/java/lang/IllegalArgumentException.html) - if the port parameter is outside the specified range of valid port values, which is between 0 and 65535, inclusive.

api/maven-api-settings/src/main/mdo/settings.mdo
```diff
@@ -535,6 +539,62 @@
...
+    }
+
+    /**
+     * Returns the port to use for this proxy.
+     * To allow interpolation of this field, this method lazily parses
+     * the {@link #getPortString()} value as an integer and defaults to {@code 8080}
+     * if not set.
+     *
+     * @return an integer indicating the port to use for this proxy
+     * @throws NumberFormatException if the port string is not a valid integer
```
elharo, line 588: This should happen when the port string is set, not later when the getter method is called

elharo (changes requested):

api/maven-api-settings/src/main/mdo/settings.mdo
gnodet: We don't do any validation in the model afaik, it is done later using custom components such as `SettingsValidator` or `ModelValidator`.  Although the goal of those methods is to provide compatibility with the ones that are currently automatically generated which look like:
```
    /**
     * Get the proxy port.
     * 
     * @return int
     */
    public int getPort()
    {
        return this.port;
    } //-- int getPort()
```
and
```
    /**
     * Set the proxy port.
     * 
     * @param port a port object.
     */
    public void setPort( int port )
    {
        this.port = port;
    } //-- void setPort( int )
```

api/maven-api-settings/src/main/mdo/settings.mdo
gnodet: All descriptions in this file are enclosed in CDATA blocks, so while I don't mind removing those, I think it belongs in a separate PR.  I suppose the idea was to add CDATA blocks to all human readable texts and java code.

api/maven-api-settings/src/main/mdo/settings.mdo
gnodet: This is usually not set manually, but only when parsing the xml and I've only added the exception because it is actually thrown.  Also, the whole idea is to allow interpolation, so I can't restrict `setPortString` to valid integers, else, the whole PR is pointless.

api/maven-api-settings/src/main/mdo/settings.mdo
elharo: It's syntax sugar, unnecessary here though and just excess code. It can be fixed one by one, or not, as you like. There's no need to change everything at once.

api/maven-api-settings/src/main/mdo/settings.mdo
elharo: Interesting. Why don't we? In general, the ealrier one can detect an error such as a negative port the better.

api/maven-api-settings/src/main/mdo/settings.mdo
elharo: NumberFormatException is an IllegalArgumentException. It is very weird for a method with no arguments to throw an IllegalArgumentException and very unexpected for a getter to throw any exception at all.

Where/when does interpolation happen? Can you throw the exception at that point and store an int for later use here?

gnodet: Fwiw, the port can be negative, in which case a default value will be used, depending on the protocol.  At least, that what is done in the resolver for the http transport.  So I'll abstain from any semantic validation, but I'll add some code to ensure that it can be parsed as an integer.

cstamas (approved):

jira-importer:
Resolve #7440

PR #1197, opened 2023-07-05
[MNG-7836] Support alternative syntaxes for POMs

gnodet:
JIRA issue: https://issues.apache.org/jira/browse/MNG-7836
IT PR: https://github.com/apache/maven-integration-testing/pull/276

### DIFF
--- a/api/maven-api-spi/src/main/java/org/apache/maven/api/spi/ModelParser.java
+++ b/api/maven-api-spi/src/main/java/org/apache/maven/api/spi/ModelParser.java
@@ -1,0 +1,71 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.maven.api.spi;
+
+import java.nio.file.Path;
+import java.util.Map;
+import java.util.Optional;
+
+import org.apache.maven.api.annotations.Experimental;
+import org.apache.maven.api.annotations.Nonnull;
+import org.apache.maven.api.annotations.Nullable;
+import org.apache.maven.api.model.Model;
+import org.apache.maven.api.services.Source;
+
+/**
+ * The {@code ModelParser} interface is used to locate and read {@link Model}s from the file system.
+ * This allows plugging in additional syntaxes for the main model read by Maven when building a project.
+ */
+@Experimental
+public interface ModelParser {
+
+    /**
+     * Locates the pom in the given directory.
+     *
+     * @param dir the directory to locate the pom for, never {@code null}
+     * @return a {@code Source} pointing to the located pom or an empty {@code Optional} if none was found by this parser
+     */
+    @Nonnull
+    Optional<Source> locate(@Nonnull Path dir);
+
+    /**
+     * Parse the model obtained previously by a previous call to {@link #locate(Path)}.
+     *
+     * @param source the source to parse, never {@code null}
+     * @param options possible parsing options, may be {@code null}
+     * @return the parsed {@link Model}, never {@code null}
+     * @throws ModelParserException if the model cannot be parsed
+     */
+    @Nonnull
+    Model parse(@Nonnull Source source, @Nullable Map<String, ?> options) throws ModelParserException;
+
+    /**
+     * Locate and parse the model in the specified directory.
+     * This is equivalent to {@code locate(dir).map(s -> parse(s, options))}.
+     *
+     * @param dir the directory to locate the pom for, never {@code null}
+     * @param options possible parsing options, may be {@code null}
+     * @return an optional parsed {@link Model} or {@code null} if none could be found
+     * @throws ModelParserException if the located model cannot be parsed
+     */
+    default Optional<Model> locateAndParse(@Nonnull Path dir, @Nullable Map<String, ?> options)
+            throws ModelParserException {
+        return locate(dir).map(s -> parse(s, options));
+    }
+}
--- a/api/maven-api-spi/src/main/java/org/apache/maven/api/spi/ModelParserException.java
+++ b/api/maven-api-spi/src/main/java/org/apache/maven/api/spi/ModelParserException.java
@@ -1,0 +1,66 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.maven.api.spi;
+
+import org.apache.maven.api.annotations.Experimental;
+import org.apache.maven.api.services.MavenException;
+
+@Experimental
+public class ModelParserException extends MavenException {
+
+    /**
+     * The one-based index of the line containing the error.
+     */
+    private final int lineNumber;
+
+    /**
+     * The one-based index of the column containing the error.
+     */
+    private final int columnNumber;
+
+    public ModelParserException() {
+        this(null, null);
+    }
+
+    public ModelParserException(String message) {
+        this(message, null);
+    }
+
+    public ModelParserException(String message, Throwable cause) {
+        this(message, -1, -1, cause);
+    }
+
+    public ModelParserException(String message, int lineNumber, int columnNumber, Throwable cause) {
+        super(message, cause);
+        this.lineNumber = lineNumber;
+        this.columnNumber = columnNumber;
+    }
+
+    public ModelParserException(Throwable cause) {
+        this(null, cause);
+    }
+
+    public int getLineNumber() {
+        return lineNumber;
+    }
+
+    public int getColumnNumber() {
+        return columnNumber;
+    }
+}
--- a/maven-model-builder/src/main/java/org/apache/maven/model/building/DefaultModelProcessor.java
+++ b/maven-model-builder/src/main/java/org/apache/maven/model/building/DefaultModelProcessor.java
@@ -27,10 +27,18 @@ import java.io.IOException;
 import java.io.InputStream;
 import java.io.Reader;
 import java.nio.file.Path;
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.List;
 import java.util.Map;
 import java.util.Objects;
+import java.util.Optional;
 
+import org.apache.maven.api.model.Model;
+import org.apache.maven.api.spi.ModelParser;
+import org.apache.maven.api.spi.ModelParserException;
 import org.apache.maven.building.Source;
+import org.apache.maven.model.io.ModelParseException;
 import org.apache.maven.model.io.ModelReader;
 import org.apache.maven.model.locator.ModelLocator;
 import org.eclipse.sisu.Typed;
@@ -65,11 +73,14 @@ import org.eclipse.sisu.Typed;
 @Typed(ModelProcessor.class)
 public class DefaultModelProcessor implements ModelProcessor {
 
+    private final Collection<ModelParser> modelParsers;
     private final ModelLocator modelLocator;
     private final ModelReader modelReader;
 
     @Inject
-    public DefaultModelProcessor(ModelLocator modelLocator, ModelReader modelReader) {
+    public DefaultModelProcessor(
+            Collection<ModelParser> modelParsers, ModelLocator modelLocator, ModelReader modelReader) {
+        this.modelParsers = modelParsers;
         this.modelLocator = modelLocator;
         this.modelReader = modelReader;
     }
@@ -81,7 +92,15 @@ public class DefaultModelProcessor implements ModelProcessor {
 
     public Path locatePom(Path projectDirectory) {
         // Note that the ModelProcessor#locatePom never returns null
-        Path pom = modelLocator.locatePom(projectDirectory.toFile()).toPath();
+        // while the ModelParser#locatePom needs to return an existing path!
+        Path pom = modelParsers.stream()
+                .map(m -> m.locate(projectDirectory)
+                        .map(org.apache.maven.api.services.Source::getPath)
+                        .orElse(null))
+                .filter(Objects::nonNull)
+                .findFirst()
+                .orElseGet(
+                        () -> modelLocator.locatePom(projectDirectory.toFile()).toPath());
         if (!pom.equals(projectDirectory) && !pom.getParent().equals(projectDirectory)) {
             throw new IllegalArgumentException("The POM found does not belong to the given directory: " + pom);
         }
@@ -95,8 +114,15 @@ public class DefaultModelProcessor implements ModelProcessor {
 
     public Path locateExistingPom(Path projectDirectory) {
         // Note that the ModelProcessor#locatePom never returns null
-        File f = modelLocator.locateExistingPom(projectDirectory.toFile());
-        Path pom = f != null ? f.toPath() : null;
+        // while the ModelParser#locatePom needs to return an existing path!
+        Path pom = modelParsers.stream()
+                .map(m -> m.locate(projectDirectory).map(s -> s.getPath()).orElse(null))
+                .filter(Objects::nonNull)
+                .findFirst()
+                .orElseGet(() -> {
+                    File f = modelLocator.locateExistingPom(projectDirectory.toFile());
+                    return f != null ? f.toPath() : null;
+                });
         if (pom != null && !pom.equals(projectDirectory) && !pom.getParent().equals(projectDirectory)) {
             throw new IllegalArgumentException("The POM found does not belong to the given directory: " + pom);
         }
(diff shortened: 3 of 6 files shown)


api/maven-api-core/src/main/java/org/apache/maven/api/services/Source.java
```diff
@@ -20,17 +20,35 @@
...
 import java.io.InputStream;
+import java.nio.file.Path;
 
 import org.apache.maven.api.annotations.Experimental;
+import org.apache.maven.api.annotations.Nonnull;
+import org.apache.maven.api.annotations.Nullable;
 
 /**
- * The source for a project's XML model.
+ * The source for parsing an object.
```
elharo, line 30: Is this a Java object being parsed somehow? If not, some synonym would be better

api/maven-api-core/src/main/java/org/apache/maven/api/services/Source.java
```diff
@@ -20,17 +20,35 @@
...
  *
  * @since 4.0.0
  */
 @Experimental
 public interface Source {
+
+    /**
+     * Provides access the file to be parsed, if this source is backed by a file.
```
elharo, line 38: access to

or perhaps just "Returns the path of the file being parsed"

api/maven-api-core/src/main/java/org/apache/maven/api/services/Source.java
```diff
@@ -20,17 +20,35 @@
...
+    Path getPath();
+
+    /**
+     * Creates a readable stream.
+     */
+    @Nonnull
     InputStream getInputStream() throws IOException;
 
+    /**
+     * Retrieves the location for this source, useful to display to the user.
```
elharo, line 50: How does this differ from the path? What is a location if not  a path?

api/maven-api-spi/src/main/java/org/apache/maven/api/spi/ModelParser.java
```diff
@@ -0,0 +1,55 @@
...
+
+import org.apache.maven.api.annotations.Experimental;
+import org.apache.maven.api.annotations.Nonnull;
+import org.apache.maven.api.annotations.Nullable;
+import org.apache.maven.api.model.Model;
+import org.apache.maven.api.services.Source;
+
+/**
+ * The {@code ModelParser} interface is used to locate and read {@link Model}s from the file system.
+ * This allows plugging in additional syntaxes for the main model read by maven when building a project.
```
elharo, line 32: Maven

api/maven-api-spi/src/main/java/org/apache/maven/api/spi/ModelParser.java
```diff
@@ -0,0 +1,55 @@
...
+ */
+@Experimental
+public interface ModelParser {
+
+    /**
+     * Locates the pom in the given directory
```
elharo, line 38: period

api/maven-api-spi/src/main/java/org/apache/maven/api/spi/ModelParser.java
```diff
@@ -0,0 +1,55 @@
...
+     * @param dir the directory to locate the pom for, never {@code null}
+     * @return the located pom or <code>null</code> if none was found by this parser
+     */
+    @Nullable
+    Path locatePom(@Nonnull Path dir);
+
+    /**
+     * Parse the model.
+     *
+     * @param source the source to use, never {@code null}
```
elharo, line 49: use --> parse

api/maven-api-spi/src/main/java/org/apache/maven/api/spi/ModelParserException.java
```diff
@@ -0,0 +1,66 @@
...
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.maven.api.spi;
+
+import org.apache.maven.api.annotations.Experimental;
+import org.apache.maven.api.services.MavenException;
+
+@Experimental
+public class ModelParserException extends MavenException {
```
elharo, line 25: This should be a checked exception since it is thrown when accepting input from outside the program

maven-core/src/main/java/org/apache/maven/internal/transformation/ConsumerPomArtifactTransformer.java
```diff
@@ -165,31 +178,50 @@ private ConsumerPomArtifact(MavenProject mavenProject, Path target, RepositorySy
...
-                    .get(reader, pomFile);
-            return XmlUtils.writeDocument(reader);
+    void transform(Path pomFile, Path destPomFile, TransformerContext context) throws IOException, XMLStreamException {
+        ModelParser parser = modelParsers.values().stream()
+                .filter(p -> pomFile.equals(p.locatePom(pomFile.getParent())))
+                .findAny()
+                .orElse(null);
+        if (parser != null) {
+            try (OutputStream output = Files.newOutputStream(destPomFile)) {
+                Model model = parser.parse(new FileSource(pomFile), null);
```
elharo, line 205: This line can throw an unhandled exception.

maven-core/src/main/java/org/apache/maven/internal/transformation/ConsumerPomArtifactTransformer.java
```diff
@@ -165,31 +178,50 @@ private ConsumerPomArtifact(MavenProject mavenProject, Path target, RepositorySy
...
+                // transform
+                model = model.withRoot(false).withModules(null);
+                if (model.getParent() != null) {
+                    model = model.withParent(model.getParent().withRelativePath(null));
+                }
+                MavenStaxWriter w = new MavenStaxWriter();
+                w.write(output, model);
```
elharo, line 212: One time use objects such MavenStaxWriter.write should be a static utility method.

maven-core/src/main/java/org/apache/maven/internal/transformation/ConsumerPomArtifactTransformer.java
```diff
@@ -165,31 +178,50 @@ private ConsumerPomArtifact(MavenProject mavenProject, Path target, RepositorySy
...
-            };
-        }
+    BiConsumer<Path, Path> transformer(RepositorySystemSession session) {
+        TransformerContext context = (TransformerContext) session.getData().get(TransformerContext.KEY);
+        return (src, dest) -> {
+            try {
+                Files.createDirectories(dest.getParent());
+                transform(src, dest, context);
+            } catch (XMLStreamException | IOException e) {
+                throw new RuntimeException(e);
```
elharo, line 190: Unless it can be proved these exceptions cannot be thrown, converting to a runtime exception is not OK. If this means not using a lambda here, that's fine.

maven-model-builder/src/main/java/org/apache/maven/model/building/DefaultModelProcessor.java
```diff
@@ -63,32 +75,110 @@
...
 
     @Override
     public File locatePom(File projectDirectory) {
-        return locator.locatePom(projectDirectory);
+        return locatePom(projectDirectory.toPath()).toFile();
+    }
+
+    public Path locatePom(Path projectDirectory) {
+        // Note that the ModelProcessor#locatePom never returns null
+        // while the ModelParser#locatePom needs to return an existing path !
```
elharo, line 97: no space before !

maven-model-builder/src/main/java/org/apache/maven/model/building/DefaultModelProcessor.java
```diff
@@ -63,32 +75,110 @@
...
+        Path pom = modelParsers.stream()
+                .map(m -> m.locatePom(projectDirectory))
+                .filter(Objects::nonNull)
+                .filter(Files::exists)
+                .findFirst()
+                .orElseGet(
+                        () -> modelLocator.locatePom(projectDirectory.toFile()).toPath());
+        if (!pom.getParent().equals(projectDirectory)) {
+            throw new IllegalArgumentException("The POM found does not belong to the given directory: " + pom);
```
elharo, line 106: This should be a checked exception

maven-model-builder/src/main/java/org/apache/maven/model/building/DefaultModelProcessor.java
```diff
@@ -63,32 +75,110 @@
...
 
     @Override
-    public Model read(Reader input, Map<String, ?> options) throws IOException {
-        return reader.read(input, options);
+    public org.apache.maven.model.Model read(InputStream input, Map<String, ?> options) throws IOException {
+        Objects.requireNonNull(input, "input cannot be null");
+        try (InputStream in = input) {
+            org.apache.maven.api.model.Model model = read(null, in, null, options);
+            return new org.apache.maven.model.Model(model);
+        } catch (ModelParserException e) {
```
elharo, line 171: I had to read this twice. ModelParserException and ModelParseException are confusingly similarly named since they differ by a single letter.

maven-model-builder/src/main/java/org/apache/maven/model/building/FileSource.java
```diff
@@ -0,0 +1,47 @@
...
+        return pom;
+    }
+
+    @Override
+    public InputStream getInputStream() throws IOException {
+        return Files.newInputStream(pom);
+    }
+
+    @Override
+    public String getLocation() {
```
elharo, line 44: Now I see the location is just the string form of the path. I'm not sure that's worth carrying around an extra method for. I suggest deleting getLocation.

elharo (dismissed):

maven-model-builder/src/main/java/org/apache/maven/model/building/DefaultModelProcessor.java
gnodet: I agree, but that's always the problem to bridge between the v3 and v4 apis.

maven-model-builder/src/main/java/org/apache/maven/model/building/FileSource.java
gnodet: No, we may have other implementations such as:
 * https://github.com/apache/maven/blob/de19cfcd2bc8e774818d87472e8e64dc37c0b93d/maven-builder-support/src/main/java/org/apache/maven/building/StringSource.java
 * https://github.com/apache/maven/blob/de19cfcd2bc8e774818d87472e8e64dc37c0b93d/maven-builder-support/src/main/java/org/apache/maven/building/UrlSource.java

api/maven-api-core/src/main/java/org/apache/maven/api/services/Source.java
gnodet: Yeah, the wording is not really good.  The idea is that the _thing_ pointed to by the `Source` will be be parsed into.  
I'll fix them to be closer to those from [`org.apache.maven.building.Source`](https://github.com/apache/maven/blob/de19cfcd2bc8e774818d87472e8e64dc37c0b93d/maven-builder-support/src/main/java/org/apache/maven/building/Source.java)

api/maven-api-core/src/main/java/org/apache/maven/api/services/Source.java
gnodet: The location could be a URL.  The point is that `getPath()` returns a path, but may be `null` if the source is created with an `InputStream`.

api/maven-api-spi/src/main/java/org/apache/maven/api/spi/ModelParserException.java
gnodet: We can discuss that, but this is a general problem in the api, as all exceptions inherit from `MavenException` which is `RuntimeException`.

maven-core/src/main/java/org/apache/maven/internal/transformation/ConsumerPomArtifactTransformer.java
gnodet: Yes, that's exactly why I prefer unchecked exceptions.
The way to handle that, if `ModelParserException` would be a checked exception, is to add another exception type to the [catch clause](https://github.com/gnodet/maven/blob/MNG-7836-Support-alternative-syntaxes-for-POMs/maven-core/src/main/java/org/apache/maven/internal/transformation/ConsumerPomArtifactTransformer.java#L189) so that it can be rethrown as a `RuntimeException`.   The catch/rethrow with a different checked exception is a very common pattern, which imho does not make sense.

maven-core/src/main/java/org/apache/maven/internal/transformation/ConsumerPomArtifactTransformer.java
gnodet: Ouch, no, it has options (file comment, string formatter), which are not used here.

maven-core/src/main/java/org/apache/maven/internal/transformation/ConsumerPomArtifactTransformer.java
gnodet: So you want me to define an interface with `throw Exception` just to avoid using unchecked exception ?
Worse, define a new interface + exception type, both would only be used in this context, just to wrap another checked exception for the sole purpose of not using unchecked exceptions ?
Again, that's a more general discussion we can have, but I don't think it's the right place.  Feel free to start a discussion thread on exception handling in the v4 api on the mailing list.

maven-model-builder/src/main/java/org/apache/maven/model/building/DefaultModelProcessor.java
gnodet: Yes, French bias. Always an unbreakable space before double punctation signs in French.

api/maven-api-core/src/main/java/org/apache/maven/api/services/Source.java
gnodet: No, because the Source is not limited to a file.

maven-core/src/main/java/org/apache/maven/internal/transformation/ConsumerPomArtifactTransformer.java
gnodet: This discussion does not belong to this PR, as this is a design decision in the v4 api and can't be changed in this smaller PR.

api/maven-api-core/src/main/java/org/apache/maven/api/services/Source.java
```diff
@@ -20,17 +20,54 @@
...
+    @Nullable
+    Path getPath();
+
+    /**
+     * Gets a byte stream to the source contents. Closing the returned stream is the responsibility of the caller.
+     *
+     * @return A byte stream to the source contents, never {@code null}.
+     * @throws IOException in case of IO issue
+     */
+    @Nonnull
```
laeubi, line 62: i would call this method `openStream()` to make it more clear that this always creates a new stream and is not a simple getter.

api/maven-api-spi/src/main/java/org/apache/maven/api/spi/ModelParser.java
```diff
@@ -0,0 +1,71 @@
...
+    /**
+     * Parse the model.
+     *
+     * @param source the source to parse, never {@code null}
+     * @param options possible parsing options, may be {@code null}
+     * @return the parsed {@link Model}, never {@code null}
+     * @throws ModelParserException if the model cannot be parsed
+     */
+    @Nonnull
+    Model parse(@Nonnull Source source, @Nullable Map<String, ?> options) throws ModelParserException;
```
laeubi, line 56: Will the source be the same as returned from locate (what would be good? the javadoc is unclear in this regards.

api/maven-api-spi/src/main/java/org/apache/maven/api/spi/ModelParser.java
```diff
@@ -0,0 +1,71 @@
...
+    /**
+     * Locate and parse the model in the specified directory.
+     * This is equivalent to {@code locate(dir).map(s -> parse(s, options))}.
+     *
+     * @param dir the directory to locate the pom for, never {@code null}
+     * @param options possible parsing options, may be {@code null}
+     * @return an optional parsed {@link Model} or {@code null} if none could be found
+     * @throws ModelParserException if the located model cannot be parsd
+     */
+    default Optional<Model> locateAndParse(@Nonnull Path dir, @Nullable Map<String, ?> options)
```
laeubi, line 67: maybe simply `getModel( ... )`  ?

gnodet: Having replaced `getInputStream` with `openStream` to avoid the _getter_ semantics, I'd rather keep it the way it is.  Plus, having `locate`, `parse` and `locateAndParse` makes it quite easy to understand the underlying semantics.

api/maven-api-spi/src/main/java/org/apache/maven/api/spi/ModelParser.java
gnodet: I think so.  I made that more clear in the javadoc.

api/maven-api-spi/src/main/java/org/apache/maven/api/spi/ModelParserException.java
elharo: I'm tempted to say that either this shouldn't extend MavenException then, or that this PR should wait ion resolving that issue. It's a pretty serious API design problem. This is a new exception in 4.0, right?

maven-model-builder/src/main/java/org/apache/maven/model/building/DefaultModelProcessor.java
```diff
@@ -81,7 +92,15 @@ public File locatePom(File projectDirectory) {
...
+        Path pom = modelParsers.stream()
+                .map(m -> m.locate(projectDirectory)
+                        .map(org.apache.maven.api.services.Source::getPath)
+                        .orElse(null))
+                .filter(Objects::nonNull)
+                .findFirst()
+                .orElseGet(
+                        () -> modelLocator.locatePom(projectDirectory.toFile()).toPath());
         if (!pom.equals(projectDirectory) && !pom.getParent().equals(projectDirectory)) {
             throw new IllegalArgumentException("The POM found does not belong to the given directory: " + pom);
```
elharo, line 105: I'm not sure this is the right exception here. This isn't necessarily a problem with he argument itself. It's more akin to a checked IOException, again arising from conditions external to the program itself.

maven-model-builder/src/main/java/org/apache/maven/model/building/DefaultModelProcessor.java
```diff
@@ -95,8 +114,15 @@ public File locateExistingPom(File projectDirectory) {
...
         // Note that the ModelProcessor#locatePom never returns null
-        File f = modelLocator.locateExistingPom(projectDirectory.toFile());
-        Path pom = f != null ? f.toPath() : null;
+        // while the ModelParser#locatePom needs to return an existing path!
+        Path pom = modelParsers.stream()
+                .map(m -> m.locate(projectDirectory).map(s -> s.getPath()).orElse(null))
+                .filter(Objects::nonNull)
+                .findFirst()
+                .orElseGet(() -> {
+                    File f = modelLocator.locateExistingPom(projectDirectory.toFile());
```
elharo, line 123: probably don't need to convert to a File here, just use the Path

elharo (commented):
I'll move this to a comment instead of a request changes, but I still think there are pretty serious issues between checked and runtime exceptions here, and those need to be resolved ASAP, and certainly before the 4.0 API is locked down and ships.

maven-model-builder/src/main/java/org/apache/maven/model/building/DefaultModelProcessor.java
gnodet: `ModelLocator` is part of the Maven 3 public api and uses the `File` instead of `Path`.
The Maven 4 api uses `Path` and not `File`, but I don't think that's a good idea to start modifying the Maven 3 api in incompatible ways, or even to duplicate all the methods to leverage the nio API at this point.
So I'm not really sure what can be done here.

gnodet:
> I'll move this to a comment instead of a request changes, but I still think there are pretty serious issues between checked and runtime exceptions here, and those need to be resolved ASAP, and certainly before the 4.0 API is locked down and ships.

I don't think there are issues at this point, as this was a concious design decision.  A change can be discussed, feel free to start a discussion and make your point.

maven-model-builder/src/main/java/org/apache/maven/model/building/DefaultModelProcessor.java
gnodet: I strongly disagree.  This is a health check on the response. Both `ModelLocator#locatePom` and `ModelParser#locate` are supposed to return a file in the provided directory or the directory itself.  This is quite explicit in their respective javadoc.  So it's a programming error and can not be recovered.

api/maven-api-spi/src/main/java/org/apache/maven/api/spi/ModelParserException.java
gnodet: All exceptions in the Maven 4 API extends `MavenException`.

maven-model-builder/src/main/java/org/apache/maven/model/building/DefaultModelProcessor.java
laeubi: @elharo I agree with @gnodet that checked exceptions should be used if calling code can do something about it (not necessarily "recover") and **should** do something, programming errors are usually nothing code can act uppon and will only result in try-catch-forget ...

cstamas (approved):

jira-importer:
Resolve #9083

PR #1198, opened 2023-07-05
Javadoc fixes

gnodet:


api/maven-api-core/src/main/java/org/apache/maven/api/Type.java
```diff
@@ -22,7 +22,7 @@
 import org.apache.maven.api.annotations.Immutable;
 
 /**
- * An artifact's{@code Type} represents a known kind of artifacts.
+ * A dependency's {@code Type} represents a known kind of dependencies.
```
elharo, line 25: dependency (singular)

This comment gives an example, which is good, but it still doesn't fully explain what a dependency type *is*. see https://issues.apache.org/jira/browse/MNGSITE-521

api/maven-api-core/src/main/java/org/apache/maven/api/Type.java
```diff
@@ -22,7 +22,7 @@
...
  * Such types are often associated to an extension and possibly
```
elharo, line 26: associated with

Is this often or always?

api/maven-api-core/src/main/java/org/apache/maven/api/Dependency.java
```diff
@@ -23,9 +23,9 @@
 public interface Dependency extends Artifact {
```
elharo, line 23: Realizing now this inheritance hierarchy is wonky. This suggests a Dependency *is* an artifact, which isn't true. Nothing to do about that in this PR of course, but I wanted to call it out while I'm thinking about it.

elharo (approved):

api/maven-api-core/src/main/java/org/apache/maven/api/Dependency.java
gnodet: I think this is true in this context.  This is unrelated to the model's dependency list.
Here, the `Dependency` is the resolved `Artifact` that come from a `DependencyCoordinate`.  This is more related to the dependency collection and artifact resolution.  The `DependencyCoordinate` is what maps closely to the model's `Dependency`.

api/maven-api-core/src/main/java/org/apache/maven/api/Type.java
gnodet: Always.

api/maven-api-core/src/main/java/org/apache/maven/api/Type.java
gnodet: I've reworded again.

api/maven-api-core/src/main/java/org/apache/maven/api/Type.java
```diff
@@ -20,14 +20,19 @@
...
- * It is also used to determine if a given dependency should be
- * included in the classpath or if its transitive dependencies should.
+ * A dependency's {@code Type} is uniquely identified by a {@code String},
+ * and semantically represents a known <i>kind</i> of dependency.
+ * <p>
+ * It provides information about the file type (or extension) of the associated artifact,
+ * its default classifier, and how the artifact will be used in the build when creating
+ * classpaths.
+ * <p>
+ * For example, the type {@code java-source} has a {@code jar} extension and a
```
elharo, line 33: classifier, the artifact
-->
classifier. The artifact

api/maven-api-core/src/main/java/org/apache/maven/api/Type.java
```diff
@@ -43,27 +48,43 @@ public interface Type {
...
+     * Get the file extension of artifacts of this type.
      *
      * @return the file extension
      */
     String getExtension();
 
     /**
-     * Get the classifier associated to the dependency type.
+     * Get the default classifier associated to the dependency type.
+     * The default classifier can be overriden when specifying
```
elharo, line 67: overridden

api/maven-api-core/src/main/java/org/apache/maven/api/Type.java
```diff
@@ -43,27 +48,43 @@ public interface Type {
...
      *
-     * @return the classifier
+     * @return the default classifier
      */
     String getClassifier();
 
-    boolean isIncludesDependencies();
-
+    /**
+     * Specifies if the artifact contains java code and should be
```
elharo, line 75: Java byte code?

api/maven-api-core/src/main/java/org/apache/maven/api/Type.java
```diff
@@ -43,27 +48,43 @@ public interface Type {
...
+     * added to the classpath.
+     *
+     * @return if the artifact should be added to the classpath
+     */
     boolean isAddedToClasspath();
+
+    /**
+     * Specifies if the artifacts already embeds its own dependencies.
```
elharo, line 83: artifacts --> artifact

api/maven-api-core/src/main/java/org/apache/maven/api/Type.java
```diff
@@ -43,27 +48,43 @@ public interface Type {
...
+     * This is the case for JEE packages or similar artifacts such as
+     * WARs, EARs, etc...
```
elharo, line 85: etc. (no ellipsis)

elharo (commented):
Looking much clearer, thanks. Just some nits

PR #1199, opened 2023-07-07
[MNG-7838] Fix usage of older packaged artifacts from project local repository

gnodet:
JIRA issue: https://issues.apache.org/jira/browse/MNG-7838
IT PR: https://github.com/apache/maven-integration-testing/pull/278

I'm not sure yet it solves the original reported problem though.

### DIFF
--- a/maven-core/src/main/java/org/apache/maven/ReactorReader.java
+++ b/maven-core/src/main/java/org/apache/maven/ReactorReader.java
@@ -101,9 +101,9 @@ class ReactorReader implements MavenWorkspaceReader {
         MavenProject project = getProject(artifact);
 
         if (project != null) {
-            File file = findArtifact(project, artifact);
+            File file = findArtifact(project, artifact, true);
             if (file == null && project != project.getExecutionProject()) {
-                file = findArtifact(project.getExecutionProject(), artifact);
+                file = findArtifact(project.getExecutionProject(), artifact, true);
             }
             return file;
         }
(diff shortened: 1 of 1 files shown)


gnodet:
A similar problem can easily be reproduced in maven master:
```
> cd maven-settings
> mvn install
```

```
[INFO] --- japicmp:0.17.2:cmp (default) @ maven-settings ---
[WARNING] File 'maven-settings/target/classes/org/apache/maven/settings/WrapperList.class' is more recent than the packaged artifact for 'maven-settings', please run a full `mvn package` build
[WARNING] File 'maven-settings/target/classes/org/apache/maven/settings/WrapperList.class' is more recent than the packaged artifact for 'maven-settings', please run a full `mvn package` build
Listening for transport dt_socket at address: 8000
[INFO] Written file '/Users/gnodet/work/git/maven/maven-settings/target/japicmp/japicmp.diff'.
[INFO] Written file '/Users/gnodet/work/git/maven/maven-settings/target/japicmp/japicmp.xml'.
[INFO] Written file '/Users/gnodet/work/git/maven/maven-settings/target/japicmp/japicmp.html'.
```

In this case, the `japicmp:cmp` goal is executed after the `jar` goal, but before the `install` goal.  And the artifacts are copied into the project local repo after the project is fully built. With this PR, the project's artifacts take precedence over the ones from the project local repo, so it actually fixes the above problem.

gnodet:
@rmannibucau it would be nice if you could build this branch and see if this reproduces your actual problem.  A small reproducer project would be even better !

rmannibucau:
@gnodet does not look helping much (and didnt manage to get it functional), guess it is cause it does not take the highest "lastModified" value from [local-repo, m2/repo] - last one being the aether resolver indeed.

jira-importer:
Resolve #9080

PR #1200, opened 2023-07-13
[MNG-7843] Allow glob patterns in dependency exclusions

gnodet:
Excluding a bunch of dependencies is quite verbose. A simple glob pattern would really help.

### DIFF
--- a/maven-core/src/main/java/org/apache/maven/artifact/resolver/filter/ExclusionArtifactFilter.java
+++ b/maven-core/src/main/java/org/apache/maven/artifact/resolver/filter/ExclusionArtifactFilter.java
@@ -18,9 +18,12 @@
  */
 package org.apache.maven.artifact.resolver.filter;
 
+import java.lang.reflect.Proxy;
+import java.nio.file.FileSystems;
+import java.nio.file.Path;
+import java.nio.file.PathMatcher;
 import java.util.List;
 import java.util.function.Predicate;
-import java.util.regex.Pattern;
 import java.util.stream.Collectors;
 
 import org.apache.maven.artifact.Artifact;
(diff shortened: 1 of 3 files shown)


maven-core/src/main/java/org/apache/maven/artifact/resolver/filter/ExclusionArtifactFilter.java
```diff
@@ -28,37 +30,120 @@
...
-                .or(groupIdIsWildcard.and(sameArtifactId(artifact)))
-                .or(artifactIdIsWildcard.and(sameGroupId(artifact)))
-                .or(sameGroupId(artifact).and(sameArtifactId(artifact)));
-    }
-
-    @Override
-    public boolean include(Artifact artifact) {
-        return !exclusions.stream().anyMatch(exclude(artifact));
+    /**
+     * Converts a standard POSIX Shell globbing pattern into a regular expression
```
elharo, line 58: This is unnecessary. There are already globstar patterns in the JDK. See  PathMatcher.getMatcher("glob")

And there's glob support in maven-shared-utils that predates this being added to the JDK.

elharo (changes requested):

elharo (commented):
In case, that wasn't clear the idea of globstars for dependency exclusion is useful. We just don't need to write more glob code to support this

maven-core/src/main/java/org/apache/maven/artifact/resolver/filter/ExclusionArtifactFilter.java
gnodet: `PathMatcher` is not really usable in this context, unless we want to implement a `Path` class just for that, but it looks worse to me. And I thought we wanted to avoid `maven-shared-utils` which is not a dependency in maven atm...

maven-core/src/test/java/org/apache/maven/artifact/resolver/filter/ExclusionArtifactFilterTest.java
```diff
@@ -140,4 +145,15 @@ void testMultipleExclusionsExcludeGroupIdWildcard() {
 
         assertThat(filter.include(artifact), is(false));
     }
+
+    @Test
+    void testExcludeWithGlob() {
+        Exclusion exclusion = new Exclusion();
+        exclusion.setGroupId("*");
```
elharo, line 152: needs tests for ** as well here, or maybe you're not using globstar? Docs on the format are needed here.

maven-core/src/main/java/org/apache/maven/artifact/resolver/filter/ExclusionArtifactFilter.java
elharo: I agree we shouldn't add a dependency on maven-shared-utils, but I do think PathMatcher is the way to go. This code is very complex, and if we can push concerns about its correctness into the JDK that's a big win, even if we have to do a but of hacking to adapt the string to paths.

elharo (changes requested):

cstamas (approved):

gnodet:
> In case, that wasn't clear the idea of globstars for dependency exclusion is useful. We just don't need to write more glob code to support this

@elharo are you fine with the changes that leverage the JDK glob matchers ?

jira-importer:
Resolve #9074

PR #1203, opened 2023-07-17
Fix codestyle in model code snippets

gnodet:


api/maven-api-model/src/main/mdo/maven.mdo
```diff
@@ -2503,27 +2501,23 @@
...
+
     /**
      * @return a Map of executions field with {@code PluginExecution#getId()} as key
      * @see PluginExecution#getId()
      */
-    public java.util.Map<String, PluginExecution> getExecutionsAsMap()
-    {
-        if ( executionMap == null )
-        {
+    public java.util.Map<String, PluginExecution> getExecutionsAsMap() {
```
elharo, line 2512: Why java.util.Map instead of just Map?

api/maven-api-model/src/main/mdo/maven.mdo
```diff
@@ -2534,34 +2528,29 @@
...
      */
-    public String getKey()
-    {
-        return constructKey( getGroupId(), getArtifactId() );
+    public String getKey() {
+        return constructKey(getGroupId(), getArtifactId());
     }
 
     /**
      * @param groupId The group ID of the plugin in the repository
```
elharo, line 2549: The --> the per Javadoc style guide

api/maven-api-model/src/main/mdo/maven.mdo
gnodet: Using classes without the full package may lead to compilation errors if the generator does not add the related imports.  Removing the package would tie the code block to the way the generator works.  For example, it can be done for the `java.util.Map` but not for the `java.util.Iterator` below because that one is not imported.

elharo: I would expect the code generator to figure out what imports it needs and do the right thing. If it doesn't, I guess that's another reason not to use a code generator.

api/maven-api-model/src/main/mdo/maven.mdo
gnodet: Fixed

api/maven-api-model/src/main/mdo/maven.mdo
gnodet: Fixed

cstamas (approved):

PR #1204, opened 2023-07-17
Trivial fixes to the generated stax readers

gnodet:


elharo (approved):
If you like. Personally I don't think single use static constants add much if anything compared to just using the string literals.

PR #1205, opened 2023-07-18
[MNG-5862] Support XML entities and XInclude

gnodet:
- [MNG-5862] XML entities support
- [MNG-5862] Support XInclude
- [IT PR](https://github.com/apache/maven-integration-testing/pull/279)

This PR is built on top of #1245 which can be integrated separately.  This feature could be opt-in or even extracted as a separate extension (hence the PR to give the needed support).

Note that the tests do ensure that any POM installed/deployed is a standalone one, i.e. the xinclude / entities are inlined during the consumer pom transformation step.


maven-model-transform/src/main/java/org/apache/maven/model/transform/stax/XmlUtils.java
```diff
@@ -124,6 +122,38 @@ public static void copy(XMLStreamReader reader, XMLStreamWriter writer, boolean
...
                     return;
+                case XMLStreamConstants.ENTITY_REFERENCE:
+                    if (!((Boolean) reader.getProperty(XMLInputFactory.IS_REPLACING_ENTITY_REFERENCES))) {
+                        writer.writeEntityRef(reader.getLocalName());
+                    }
+                    break;
+                case XMLStreamConstants.ATTRIBUTE:
+                    throw new IllegalStateException();
+                case XMLStreamConstants.DTD:
+                    /*
```
elharo, line 133: delete or uncomment

gnodet:
> I think this PR only supports external entities, not XInclude?

Both, but the code for XInclude support is in third-party library for now: https://github.com/gnodet/stax-xinclude
Also, the IT uses xinclude at https://github.com/apache/maven-integration-testing/pull/279/files#diff-f29a4cbe7d6f8b4fb5f7534e23bfe9622d437605cf1df65821d94389d2efe5e6R29


> There might be some security implications here. There certainly will be people who claim there are.

Those new options are conditioned by reading a local file with strict mode (which means not when loading a POM from a dependency, when only for projects being built).   If needed, we could use a custom resolver and restrict to loading external files for entities / xinclude from the file system, but I'd like to avoid any restriction if they are not needed.  The files are loaded from locally built POM files, so the user should be in control of those files.

> This likely needs doc changes as well.

Yes, I'll try to find where...

gnodet:
> > There might be some security implications here. There certainly will be people who claim there are.
> 
> Those new options are conditioned by reading a local file with strict mode (which means not when loading a POM from a dependency, i.e. only for projects being built). If needed, we could use a custom resolver and restrict to loading entities / xinclude from the file system, or even the project tree, but I'd like to avoid any restriction if they are not needed. As the files are loaded from locally built POM files, the user should be in control of those files.

I've modified the xinclude support so that it reuses the XMLResolver.  This allows using a single implementation to load external stuff.  I've thus modified it to reject any non relative URI, which means the code can only access files under the user's control, so that should be fine from a security perspective.

norrisjeremy:
Is this wise?
XML parsers are littered with a long history of security vulnerabilities around loading crap remotely.
Introducing that into a build system seems scary to me.

maven-model-builder/src/main/java/org/apache/maven/model/io/DefaultModelReader.java
```diff
@@ -104,16 +113,53 @@ private TransformerContext getTransformerContext(Map<String, ?> options) {
...
+                factory.setProperty(XMLInputFactory.IS_REPLACING_ENTITY_REFERENCES, true);
+                factory.setProperty(XMLInputFactory.IS_SUPPORTING_EXTERNAL_ENTITIES, true);
+                factory.setProperty(WstxInputProperties.P_ENTITY_RESOLVER, resolver);
+                parser = new XIncludeStreamReader(
+                        factory,
+                        outputFactory,
+                        pomFile.toUri().toString(),
+                        factory.createXMLStreamReader(new StaxPathInputSource(pomFile, input)));
+            } else {
+                parser = factory.createXMLStreamReader(new StreamSource(input));
```
norrisjeremy, line 156: Should there be a `factory.setProperty(XMLInputFactory.IS_REPLACING_ENTITY_REFERENCES, false);` call here?

gnodet:
> Is this wise?
> XML parsers are littered with a long history of security vulnerabilities around loading crap remotely.
> Introducing that into a build system seems scary to me.

I tried to mitigate the risks by only loading from rejecting any absolute URI.  Also, no file with such entities / xinclude import should end up in maven central, those are translated when installed / uploaded.  So this should only happen at build time, for the pom.xml that are parts of the build.  The security problems are then kinda in the hand of the developper I would think.

norrisjeremy:
> I tried to mitigate the risks by only loading from rejecting any absolute URI. Also, no file with such entities / xinclude import should end up in maven central, those are translated when installed / uploaded. So this should only happen at build time, for the pom.xml that are parts of the build. The security problems are then kinda in the hand of the developper I would think.
> 
> That's said, I'd like to find a consensus, and if it's seen as too risky, we could go with only mixins. That's the whole point of the discussion I started on dev.

Would you consider adding a CLI option that a user can specify to Maven to explicitly tell it not to support remote entities / xinclude stuff (and possibly even default the option to having remote entities / xinclude off)?
That could help allay concerns, in that users would have to explicitly opt-in (or have an ability to explicitly opt-out) to enabling this new potentially risky behavior.

maven-model-builder/src/main/java/org/apache/maven/model/io/DefaultModelReader.java
```diff
@@ -104,16 +113,53 @@ private TransformerContext getTransformerContext(Map<String, ?> options) {
...
+                    if (sysUri.isAbsolute()) {
+                        throw new XMLStreamException("systemID must be a relative URL: " + systemID);
+                    }
+                    try {
+                        URL url = URI.create(baseURI).resolve(sysUri).toURL();
+                        return new StreamSource(url.openStream(), url.toExternalForm());
+                    } catch (IOException e) {
+                        throw new XMLStreamException("Unable to create Source for " + systemID, e);
+                    }
+                };
```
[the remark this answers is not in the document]
gnodet: Fixed

elharo (dismissed):
Since the library you wrote uses the org.apache.maven package, it needs to move into this repo (package protected?) or into another package or something before maven core can depend on it.

gnodet:
> > I tried to mitigate the risks by only loading from rejecting any absolute URI. Also, no file with such entities / xinclude import should end up in maven central, those are translated when installed / uploaded. So this should only happen at build time, for the pom.xml that are parts of the build. The security problems are then kinda in the hand of the developper I would think.
> > That's said, I'd like to find a consensus, and if it's seen as too risky, we could go with only mixins. That's the whole point of the discussion I started on dev.
> 
> Would you consider adding a CLI option that a user can specify to Maven to explicitly tell it not to support remote entities / xinclude stuff (and possibly even default the option to having remote entities / xinclude off)? That could help allay concerns, in that users would have to explicitly opt-in (or have an ability to explicitly opt-out) to enabling this new potentially risky behavior.

I've added a way to opt-out using `-Dmaven.experimental.xinclude=false` ...

norrisjeremy:
> > Would you consider adding a CLI option that a user can specify to Maven to explicitly tell it not to support remote entities / xinclude stuff (and possibly even default the option to having remote entities / xinclude off)? That could help allay concerns, in that users would have to explicitly opt-in (or have an ability to explicitly opt-out) to enabling this new potentially risky behavior.
> 
> I've added a way to opt-out using `-Dmaven.experimental.xinclude=false` ...

Thank you!

maven-model-builder/src/main/java/org/apache/maven/model/io/DefaultModelReader.java
gnodet: The factory is new and `false` is always the default, it should not be necessary.

gnodet:
@elharo can I go ahead with this PR ?

api/maven-api-core/src/main/java/org/apache/maven/api/feature/Features.java
```diff
@@ -0,0 +1,72 @@
...
+ *
+ * @since 4.0.0
+ */
+public final class Features {
+
+    public static final String BUILDCONSUMER = "maven.experimental.buildconsumer";
+
+    public static final String XINCLUDE = "maven.experimental.xinclude";
+
+    public static final String TRUE = "true";
```
elharo, line 39: Inline this. The constant adds nothing.

api/maven-api-core/src/main/java/org/apache/maven/api/feature/Features.java
```diff
@@ -0,0 +1,72 @@
...
+
+    private Features() {}
+
+    @Nonnull
+    public static boolean buildConsumer(@Nonnull Properties userProperties) {
+        return TRUE.equals(Objects.requireNonNull(userProperties).getOrDefault(BUILDCONSUMER, TRUE));
```
elharo, line 45: In fact, reading this I realize the constant actively obscures the type, which is obvious with a string literal.

api/maven-api-core/src/main/java/org/apache/maven/api/feature/Features.java
```diff
@@ -0,0 +1,72 @@
...
+    }
+
+    @Nonnull
+    public static boolean buildConsumer(@Nonnull Map<String, String> userProperties) {
+        return TRUE.equals(Objects.requireNonNull(userProperties).getOrDefault(BUILDCONSUMER, TRUE));
```
elharo, line 50: If there's a known default value. maybe null is acceptable? That is,

if (userProperties == null) {
  return true;
} else{ 
...

Just a thought

maven-core/src/test/java/org/apache/maven/project/ProjectBuilderTest.java
```diff
@@ -234,7 +234,7 @@ void testReadInvalidPom() throws Exception {
 
         // single project build entry point
         Exception ex = assertThrows(Exception.class, () -> projectBuilder.build(pomFile, configuration));
-        assertThat(ex.getMessage(), containsString("Received non-all-whitespace CHARACTERS or CDATA event"));
+        assertThat(ex.getMessage(), containsString("expected START_TAG or END_TAG not CHARACTERS"));
```
elharo, line 237: END_TAG, not
(add comma)

maven-model-transform/src/main/java/org/apache/maven/model/transform/stax/BufferingParser.java
```diff
@@ -476,13 +510,20 @@ protected Event bufferEvent() throws XMLStreamException {
                 event.name = pp.getLocalName();
                 event.namespace = pp.getNamespaceURI();
                 event.prefix = pp.getPrefix();
-                // event.text = pp.getText();
                 break;
             case CHARACTERS:
```
elharo, line 514: case fallthrough is risky for code maintenance; better to break explictly for each case

maven-model-transform/src/main/java/org/apache/maven/model/transform/stax/XmlUtils.java
```diff
@@ -124,6 +122,22 @@ public static void copy(XMLStreamReader reader, XMLStreamWriter writer, boolean
...
+                    }
+                    break;
+                case XMLStreamConstants.ATTRIBUTE:
+                    throw new IllegalStateException();
+                case XMLStreamConstants.DTD:
+                    break;
+                case XMLStreamConstants.CDATA:
+                    writer.writeCData(normalize(reader.getText()));
+                    break;
+                case XMLStreamConstants.NAMESPACE:
```
elharo, line 137: avoid fallthrough

maven-model-transform/src/test/java/org/apache/maven/model/transform/ConsumerPomXMLFilterTest.java
```diff
@@ -226,4 +235,40 @@ void lexicalHandler() throws Exception {
...
+        factory.setProperty(WstxInputProperties.P_TREAT_CHAR_REFS_AS_ENTS, false);
+        factory.setProperty(WstxInputProperties.P_UNDECLARED_ENTITY_RESOLVER, resolver);
+        factory.setProperty(WstxInputProperties.P_ENTITY_RESOLVER, resolver);
+        XMLStreamReader parser = factory.createXMLStreamReader(new StringReader(input));
+        XMLStreamReader filter = getFilter(parser);
+        ByteArrayOutputStream baos = new ByteArrayOutputStream();
+        XmlUtils.writeDocument(filter, baos);
+        String actual = baos.toString();
+
+        System.out.println(actual);
```
elharo, line 271: No println in tests please. Only failing tests should generate output so failures are easier to spot and debug

maven-stax-xinclude/src/main/java/org/apache/maven/stax/xinclude/DOMXMLElementEvaluator.java
```diff
@@ -0,0 +1,101 @@
...
+ * This class is based upon a class of the same name in the Apache Woden.
+ */
+public class DOMXMLElementEvaluator extends XMLElementEvaluator<Element> {
+
+    private DTDSubset dtd;
+
+    /**
+     * Constructs a new DOMXMLElementEvaluator to evaluate a XPointer on a DOM Element.
+     *
+     * @param xpointer an XPointer to evaluate.
```
elharo, line 45: per Oracle guidelines, no period when not a complete sentence

maven-stax-xinclude/src/main/java/org/apache/maven/stax/xinclude/FragmentReader.java
```diff
@@ -0,0 +1,70 @@
...
+ * under the License.
+ */
+package org.apache.maven.stax.xinclude;
+
+import javax.xml.stream.XMLStreamException;
+import javax.xml.stream.XMLStreamReader;
+
+public class FragmentReader extends StreamReaderDelegate {
+
+    XMLStreamReader delegate;
```
elharo, line 26: private

maven-stax-xinclude/src/main/java/org/apache/maven/stax/xpointer/ElementPointerPart.java
```diff
@@ -0,0 +1,245 @@
...
+import java.util.Collections;
+import java.util.List;
+
+/**
+ * ElementPointerPart is a class which represents the element() scheme for the XPointer Framework.
+ * The specification is defined at <a href="http://www.w3.org/TR/xptr-element/">http://www.w3.org/TR/xptr-element/</a>
+ * <p>
+ * This class is immutable.
+ * <p>
+ * This class is based upon a class of the same name in the Apache Woden.
```
elharo, line 31: the Apache --> Apache

maven-stax-xinclude/src/main/java/org/apache/maven/stax/xpointer/ElementPointerPart.java
```diff
@@ -0,0 +1,245 @@
...
+ */
+public class ElementPointerPart implements PointerPart {
```
elharo, line 33: This PR exposes a lot of new public API that's going to be hard and expensive to support. Is there anyway to make less of it public? E.g. by using a single package?

maven-stax-xinclude/src/main/java/org/apache/maven/stax/xpointer/InvalidXPointerException.java
```diff
@@ -0,0 +1,143 @@
...
+     *
+     * @param message  a String message of error.
+     * @param fragment a String fragment of the cause.
+     */
+    public InvalidXPointerException(String message, String fragment) {
+        this(message, fragment, null, null, null);
+    }
+
+    /**
+     * Constructs a InvalidXPointerException with a message and fragment properties.
```
elharo, line 43: a InvalidXPointerException --> an InvalidXPointerException

maven-stax-xinclude/src/main/java/org/apache/maven/stax/xpointer/XMLElementEvaluator.java
```diff
@@ -0,0 +1,201 @@
...
+import java.util.ArrayDeque;
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.Deque;
+import java.util.Iterator;
+import java.util.List;
+import java.util.stream.Collectors;
+
+/**
+ * This class Evaluates an XPointer on a XMLElement, using the XPointer model.
```
elharo, line 30: evaluates (lower case)

src/mdo/reader-stax.vm
```diff
@@ -70,6 +73,11 @@ public class ${className} {
      */
     private static final String XSI_NAMESPACE = "http://www.w3.org/2001/XMLSchema-instance";
 
+    /**
+     * XML namespace
+     */
+    private static final String XML_NAMESPACE = "http://www.w3.org/XML/1998/namespace";
```
elharo, line 79: reinventing javax.xml.XMLConstants

elharo (commented):
I'm still not convinced this is a good idea, but if it's going in, there are still some ciode issues to be addressed.

gnodet:
> I'm still not convinced this is a good idea, but if it's going in, there are still some ciode issues to be addressed.

Well, I tried to kick [a discussion on dev@](https://lists.apache.org/thread/7sz91y04zb79lqkvnw77bxf70f9rny3b) but not much interest there...  The only concerns were the same that have been raised here around security, and while I agree it could have been a problem, I think those concerns have been addressed in the PR.
So I assume lazy consensus...

The idea of xinclude/xml entities is slightly redundant with pom mixins, so I would have expected that this would have been raised...  This discussion was started mid-august, definitely not the best time to have people involved.

norrisjeremy:
FYI, just from a simple user's perspective, it seems peculiar that "experimental" (`maven.experimental.xxx`) features are being enabled by default.

Typically I would expect a feature/option that a software developer considers "experimental" to be something that a user has to opt-in to enable, not opt-out to disable.

I'm not sure if that means these new features should be disabled by default, or if the options should be renamed to better indicate they aren't considered "experimental"?

maven-stax-xinclude/src/main/java/org/apache/maven/stax/xpointer/ElementPointerPart.java
gnodet: FTR, I do really question the fact that a class being public is supposed to be supported when consumed by other projects.   Especially with Maven 4, the API will be defined by everything in under the `org.apache.maven.api` and anything not under this package will not be part of the API.

Anyway, I'll refactor into a single package with less public classes...

gnodet:
> FYI, just from a simple user's perspective, it seems peculiar that "experimental" (`maven.experimental.xxx`) features are being enabled by default.
> 
> Typically I would expect a feature/option that a software developer considers "experimental" to be something that a user has to opt-in to enable, not opt-out to disable.
> 
> I'm not sure if that means these new features should be disabled by default, or if the options should be renamed to better indicate they aren't considered "experimental"?

I agree, I think those "experimental" flags will need to be removed or the features disabled by default before GA.  But you're right, experimental + enabled by default looks borderline.

maven-core/src/test/java/org/apache/maven/project/ProjectBuilderTest.java
gnodet: This is a test, so I can't change the message here.  I'd be happy to raise a follow-up as this message is sent by the generated xml parsers.

pdeneve:
@gnodet Thanks for you work. Any plans on resolving the requested change?

maven-model-builder/src/main/java/org/apache/maven/model/io/LocalXmlResolver.java
```diff
@@ -0,0 +1,85 @@
...
+        if (!base.startsWith(rootDirectory)) {
+            return null;
+        }
+        Path sys = Paths.get(sysUri.getSchemeSpecificPart()).normalize();
+        if (sys.isAbsolute()) {
+            throw new XMLStreamException("systemID must be a relative path: " + systemID);
+        }
+        Path res = base.resolveSibling(sys).normalize();
+        if (!res.startsWith(rootDirectory)) {
+            throw new XMLStreamException("systemID cannot refer to outside rootDirectory: " + systemID);
```
elharo, line 77: refer to --> refer to a path

maven-model-builder/src/main/java/org/apache/maven/model/io/ModelReader.java
```diff
@@ -45,6 +45,12 @@ public interface ModelReader {
      */
     String INPUT_SOURCE = "org.apache.maven.model.io.inputSource";
 
+    /**
+     * Name of the property used to store a boolean {@code true} if XInclude supports
```
elharo, line 49: supports --> support

maven-stax-xinclude/src/main/java/org/apache/maven/stax/xinclude/DOMXMLElementEvaluator.java
```diff
@@ -0,0 +1,98 @@
...
+                }
+            }
+        }
+        return false;
+    }
+
+    /**
+     * Evaluates the XPointer on the root Element and returns the resulting Element or null.
+     *
+     * @return an Element from the resultant evaluation of the root Element or null if evaluation fails
```
elharo, line 82: delete resultant

gnodet:
One possibility would be to create an extension to support this feature.  This should be possible as the entities/xinclude are only processed at build time and the consumer pom is flattened.

maven-stax-xinclude/src/main/java/org/apache/maven/stax/xinclude/ElementPointerPart.java
```diff
@@ -0,0 +1,245 @@
...
+     * @throws IllegalArgumentException if the schemeData has invalid scheme syntax.
+     */
+    static ElementPointerPart parseFromString(final String schemeData) throws InvalidXPointerException {
+        List<Integer> childSequence;
+        String elementID = null;
+        int startChar;
+        int endChar;
+
+        // Find an NCName if it exists?
+        startChar = schemeData.indexOf("/");
```
elharo, line 174: declare here

maven-stax-xinclude/src/main/java/org/apache/maven/stax/xinclude/InvalidXPointerException.java
```diff
@@ -0,0 +1,143 @@
...
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.maven.stax.xinclude;
+
+/**
+ * This class represents Exceptions that can happen during parsing an XPointer Expression.
```
elharo, line 22: during --> while

maven-stax-xinclude/src/main/java/org/apache/maven/stax/xinclude/NCName.java
```diff
@@ -0,0 +1,176 @@
...
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.maven.stax.xinclude;
+
+/**
+ * This class represents the data type NCName use for XML non-colonized names.
+ */
+@SuppressWarnings("checkstyle:MagicNumber")
+class NCName {
```
elharo, line 25: This is a lot to reinvent and maintain for one function. Maybe we need this to avoid extra dependencies, but I'm sure this already exists in Xerces, XML Commons, or any of several other projects.

elharo (commented):
Mostly nits.

More generally, rereading this I'm struck at how much work this is. You've provided a nearly complete XInclude implementation. The maintenance burden on this code is likely to be high. It feels like this belongs in the parser, or perhaps a separate library on top of the parser, and all Maven should do when reading a pom is flip the flag to turn on XInclude support.
Maven's XML parsing is already quite non-standard and avoids the normal parsers like Xerces or the JDK's because of decisions made 20 years ago when the easier path was not so obvious. This is unfortunately baked into a lot of the code and even the public APIs, so I'm not sure how much of that can be repaired now. 

Nonetheless, I do wonder if this code belongs here instead of in the parser.

api/maven-api-core/src/main/java/org/apache/maven/api/feature/Features.java
gnodet: `userProperties` is now `Nullable`

maven-model-builder/src/main/java/org/apache/maven/model/io/DefaultModelReader.java
```diff
@@ -100,14 +104,39 @@ private Path getRootDirectory(Map<String, ?> options) {
...
+            }
+
+            XMLStreamReader parser;
+            // We only support xml entities and xinclude when reading a file in strict mode
+            if (pomFile != null && strict && getXInclude(options)) {
+                parser = XInclude.xinclude(xmlSource, new LocalXmlResolver(rootDirectory));
+            } else {
+                XMLInputFactory factory = new com.ctc.wstx.stax.WstxInputFactory();
+                factory.setProperty(XMLInputFactory.IS_REPLACING_ENTITY_REFERENCES, false);
+                factory.setProperty(XMLInputFactory.IS_SUPPORTING_EXTERNAL_ENTITIES, false);
```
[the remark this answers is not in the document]
gnodet: > Nothing you can do here, but this property should have been named SUPPORTS_EXTERNAL_ENTITIES

Agreed

maven-model-builder/src/main/java/org/apache/maven/model/io/DefaultModelReader.java
```diff
@@ -147,4 +175,18 @@ private Model read(Reader reader, Path pomFile, Map<String, ?> options) throws I
             throw new IOException("Unable to transform pom", e);
         }
     }
+
+    private static class StaxPathInputSource extends Stax2FileSource {
```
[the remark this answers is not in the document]
gnodet: I removed that class as it was not really needed.

maven-stax-xinclude/src/main/java/org/apache/maven/stax/xinclude/NCName.java
gnodet: I reused the woodstox methods.

api/maven-api-core/src/main/java/org/apache/maven/api/feature/Features.java
```diff
@@ -62,6 +64,18 @@ public static boolean buildConsumer(@Nullable Session session) {
         return buildConsumer(session != null ? session.getUserProperties() : null);
     }
 
+    public static boolean xinclude(@Nullable Properties userProperties) {
+        return doGet(userProperties, XINCLUDE, true);
```
norrisjeremy, line 68: I would still advocate for this being off by default (opt-in) instead of on by default (opt-out) since XML parsers have   a long history of security vulnerabilities surrounding this stuff.

gnodet: Done

gnodet:
> Mostly nits.

They should be all fixed now.

> More generally, rereading this I'm struck at how much work this is. You've provided a nearly complete XInclude implementation. The maintenance burden on this code is likely to be high. It feels like this belongs in the parser, or perhaps a separate library on top of the parser, and all Maven should do when reading a pom is flip the flag to turn on XInclude support. Maven's XML parsing is already quite non-standard and avoids the normal parsers like Xerces or the JDK's because of decisions made 20 years ago when the easier path was not so obvious. This is unfortunately baked into a lot of the code and even the public APIs, so I'm not sure how much of that can be repaired now.
> 
> Nonetheless, I do wonder if this code belongs here instead of in the parser.

We switched to a standard StaX parser in 4.x.  The xml parser is now Woodstox, which is a fully conformant parser.  Unfortunately, I don't know any StaX parser that provides support for XInclude, so I ended up writing one, borrowing the XPointer implementation from Apache Woden.  I'd rather keep it separated from the parser so that we can eventually switch to Aalto which is slightly faster than Woodstox, but there were a few problems when I tried initially.

As for the location, I originally put it in an external repository (https://github.com/gnodet/stax-xinclude), but you asked to move it back in this repository.  I agree, this library is independent of Maven and would be better outside.  I can maintain it as a personal project (and rename the package) if you think it's better, I don't really mind.

elharo:
A personal project won't really work, but is there any way this can go into Woodstox or the Apache XML project or something like that?

gnodet:
> A personal project won't really work, but is there any way this can go into Woodstox or the Apache XML project or something like that?

@cotowncoder ? @scantor ?

scantor:
> > A personal project won't really work, but is there any way this can go into Woodstox or the Apache XML project or something like that?
> 
> @cotowncoder ? @scantor ?

Umm. Maybe I was referenced by mistake, but I'm not sure of my relevance to this?

As a maintainer of software that uses Maven itself, I certainly have "opinions" but I think you all addressed the main one, which is to turn this off by default.

All I can say is, please don't ever enable this by default. Ever. I say that as somebody that's worked with XML for 25 years and has been fixing security bugs in the use of it for about 20. Maybe that's why you asked me?

delanym:
The stigma surrounding xinclude is really quite infuriating. Is there a feature more maligned than this? The disparity between its notoriety and its simplicity is something almost poetic.

The fact is pom files already compose in at least 3 other ways - so if there's some "security principle" at play its already broken. It could be argued the whole purpose of pom files is to compose.
Whether processing xinclude introduces risks has nothing to do with XML per se (a standard designed to be as open and dynamic as possible so no wonder there's so many horror stories) and everything to do with this implementation. You'd have to be half-mad not to use this one anyway.

I'm also not sure that xincludes will really work in practice, but I've yet to see why not.

scantor:
> The stigma surrounding xinclude is really quite infuriating. Is there a feature more maligned than this?

Yes, entities, deservedly so. That was actually more my concern. I have no background on what this issue is about or why I was asked about it, but apologies for not clarifying which of the two I was expressing the opinion about. I know very little about XInclude as it's very poorly supported in general, it came too late, so I've generally ignored it.

My *opinion* is that it's not sensible for Maven to even consider including a one-off implementation of any XML standard addendum. But that's a decision for its maintainers.

cowtowncoder:
I remember thinking about XInclude at some point wrt Woodstox, but it seemed rather complicated to support for a streaming parser. So likely it'd instead need to be some processor on top of generic Stax (or SAX interface), in which case Woodstox as well as any other compliant implementation would work.

I don't think I personally have time to work on such thing, although I can see how it'd be useful.

gnodet:
This should become a separate Maven extension.

gnodet:
An extension will be available at 
  https://github.com/apache/maven-xinclude-extension

pdeneve:
Cool, will check it out!

jira-importer:
Resolve #7776

PR #1209, opened 2023-07-20
[MNG-5102] Add support for POM mixins

gnodet:
- [MNG-7810] Provide a SPI for the v4 api
- [MNG-7836] Support alternative syntaxes for POMs
- Add a bom packaging, trim down consumer pom, add build pom
- [MNG-5102] Add support for POM mixins


its/core-it-suite/src/test/resources/mng-5102-mixins/gav/project/pom.xml
```diff
@@ -0,0 +1,60 @@
...
+          </execution>
+        </executions>
+      </plugin>
+    </plugins>
+  </build>
+
+  <mixins>
+    <mixin>
+      <groupId>org.apache.maven.its.mng5102</groupId>
+      <artifactId>mixin-2</artifactId>
```
talios, line 56: Can this take a classifier? One of the early issues we identified with tiles - which is why we have a pom.xml and a tile.xml - is for publishing to a repo.

The pom.xml would have release plugin information, which tne tile/mixin wouldn't want.

I like what I'm seeing here, and it looks like it should support my own needs - will try find some time to pull/build this locally and give it a renewed test.

gnodet: It's now supported :-)

jimisola: Great news. I looked at the tests quickly and it seemed like the mixins there only contains properties.

What can be included (or not included) in the mixins?

gnodet: The mixins are merged in their declaration order as if they were parents, so there are no restrictions at this point. I haven't given much thought if there should be yet...

jimisola: @talios Our organization use your Maven Tiles (thank you!). Trying to think if there is something in Maven Tiles that this current implementation of Maven Mixins does not support. With Maven Tiles we have issues with profiles (the are active but Maven does not alwats acknowledge them) and properties (properties aren't propagated).

From what Guillaume wrote it seems as if mixins, currently not having any restrictions, might function even better that Maven Tiles.

jimisola: @gnodet Is it likely that mixins will make the next Maven 4 release candidate? Would really like to try it out with our setup.

jimisola: Not sure if it is of interest for @gnodet but here is our Maven Tiles setup.

We use tiles and composites where are used to group tiles.

Hence, we have a need for including mixins in mixins for us to be able to group mixins into composites.

Parent Tile

- maven-tiles-parent which is used to set the Maven Tiles version for all tiles (won't be needed with Maven Mixins)

Tiles: 
- common-tile
- java-17-tile
- java-21-tile
- lfv-development-environment-tile
- lint-tile
- maven-core-versions-tile
- openapi-tile
- opentelemetry-tile
- reqstool-tile
- soap-tile
- spring-boot-amqp-tile
- spring-boot-application-tile
- spring-boot-camel-tile
- spring-boot-mvc-tile
- spring-boot-oauth-tile
- spring-boot-redis-tile
- spring-boot-tile
- spring-boot-vault-tile
- sysdev-common-tile
- test-compile-tile

@talios As for the properties issue (our use case seem to be different), with respect for this PR we refer to this [new ticket](https://github.com/repaint-io/maven-tiles/issues/159) in Maven Tiles.

talios: @jimisola I saw that ticket, I see what you mean - I always propagate UP into the tile (so overrides in the project).

I've never really been a fan of composites (which are usually dependencies) as BOM imports already exist - composites just remove the _need_ to declare each dependency in each module - which, in our application which is OSGi based can cause issues if we're just adding deps unseen - I guess it also depends on how often you churn releases of tiles, or dependencies ( we use an additional tool I wrote for $work that handles version ranges/lock files) to track and check that.

Out of interesting - do you tiles include dependencies, or _just_ plugins?

lfvjimisola: Both dependencies and plugins.

talios: Finally got around to testing this (it's been a hectic fortnight).  From a fresh `archetype:generated` project, and updating the `model` version and XML preamble, adding the following maven tile "mixin":

```xml
<mixins>
  <mixin>
    <groupId>com.smxemail.tiles</groupId>
    <artifactId>com.smxemail.tiles.enforcements</artifactId>
    <version>4.0.176</version>
    <extension>xml</extension>
  </mixin>
</mixins>
```

Everything seemed to work as expected!

Compared to the tiles config:

```xml
<plugin>
  <groupId>io.repaint.maven</groupId>
  <artifactId>tiles-maven-plugin</artifactId>
  <version>2.40</version>
  <extensions>true</extensions>
  <configuration>
    <tiles>
      <tile>com.smxemail.tiles:com.smxemail.tiles.enforcements:[4.0.176]</tile>
    </tiles>
  </configuration>
</plugin>
```

I long for the day when we can have some more concise GAV specifications in Maven, but this is looking good as a transition.

The tile does transclude another tile, which obviously doesn't work here. I will craft one with its own `<mixin>` element and do some further testing.

jimisola:
One issue we have with Maven Tiles is that plugin configurations for the same plugin will overwrite each other. How is this handled with mixins?

gnodet:
> One challenge we have with Maven Tiles is that plugin configurations for the same plugin will overwrite each other. How is this handled with mixins?

They use the same mechanism as parents, so it will be like if one plugin definition is in the parent and the other one in the grand-parent.  They end up being merged.  The order will have to be clearly specified though.

lfvjimisola:
> > One challenge we have with Maven Tiles is that plugin configurations for the same plugin will overwrite each other. How is this handled with mixins?
> 
> They use the same mechanism as parents, so it will be like if one plugin definition is in the parent and the other one in the grand-parent. They end up being merged. The order will have to be clearly specified though.

When you say merged - in what way?

A (parent):

```
    <build>
        <plugins>
            <plugin>
                <groupId>plugin</groupId>
                <artifactId>plugin</artifactId>
                <configuration>
                    <a>a</a>
                    <c>a</c>
                </configuration>
            </plugin>
        </plugins>
    </build>
```

B (child)

```
    <build>
        <plugins>
            <plugin>
                <groupId>plugin</groupId>
                <artifactId>plugin</artifactId>
                <configuration>
                    <b>b</b>
                    <c>b</c>
                </configuration>
            </plugin>
        </plugins>
    </build>

```

What would be the outcome of the above?

1. The entire plugin is configured by A or B depending on order
2. Everything is merged but A has priority so
                    <a>a</a>
                    <b>b</b>
                    <c>a</c>
3. Everything is merged but B has priority so
                    <a>a</a>
                    <b>b</b>
                    <c>b</c>
4. other outcome

gnodet:
> > > One challenge we have with Maven Tiles is that plugin configurations for the same plugin will overwrite each other. How is this handled with mixins?
> > 
> > 
> > They use the same mechanism as parents, so it will be like if one plugin definition is in the parent and the other one in the grand-parent. They end up being merged. The order will have to be clearly specified though.
> 
> When you say merged - in what way?
> 
> A (parent):
> 
> ```
>     <build>
>         <plugins>
>             <plugin>
>                 <groupId>plugin</groupId>
>                 <artifactId>plugin</artifactId>
>                 <configuration>
>                     <a>a</a>
>                     <c>a</c>
>                 </configuration>
>             </plugin>
>         </plugins>
>     </build>
> ```
> 
> B (child)
> 
> ```
>     <build>
>         <plugins>
>             <plugin>
>                 <groupId>plugin</groupId>
>                 <artifactId>plugin</artifactId>
>                 <configuration>
>                     <b>b</b>
>                     <c>b</c>
>                 </configuration>
>             </plugin>
>         </plugins>
>     </build>
> ```
> 
> What would be the outcome of the above?
> 
> 1. The entire plugin is configured by A or B depending on order
> 2. Everything is merged but A has priority so
>    a
>    b
>    a
> 3. Everything is merged but B has priority so
>    a
>    b
>    b
> 4. other outcome

Maven identifies plugins by their groupId and artifactId. Since both parent and child define the same plugin, their configurations are merged.

## Configuration Merging:
For non-conflicting elements (elements present in one POM but not the other), Maven includes both elements in the final configuration.
For conflicting elements (elements with the same name in both parent and child), the child's configuration takes precedence and overrides the parent's configuration.
By default, Maven does not append or merge the content of elements unless explicitly instructed (e.g., using combine.children="append").

## Non-conflicting elements:
Parent has <a>a</a>, which is not present in the child. This element is included in the final configuration.
Child has <b>b</b>, which is not present in the parent. This element is included in the final configuration.
## Conflicting elements:
Both parent and child define <c>. The parent has <c>a</c>, and the child has <c>b</c>.
The child's <c>b</c> overrides the parent's <c>a</c>.
## Resulting Configuration:
The merged configuration for the plugin in the child project will include:
```
<a>a</a> (from parent, non-conflicting).
<b>b</b> (from child, non-conflicting).
<c>b</c> (from child, overriding parent's <c>a</c>).
```
Final Outcome
The effective configuration for the plugin in the child project (B) will be:

```xml
<build>
    <plugins>
        <plugin>
            <groupId>plugin</groupId>
            <artifactId>plugin</artifactId>
            <configuration>
                <a>a</a>
                <b>b</b>
                <c>b</c>
            </configuration>
        </plugin>
    </plugins>
</build>
```

## Notes
If you want to append configurations (e.g., combine `<c>` values from parent and child), you can use the `combine.children="append"` attribute in the child POM. For example:
```xml
<configuration combine.children="append">
    <b>b</b>
    <c>b</c>
</configuration>
```
This would require a more complex structure for `<c>` (e.g., as a list), which depends on the plugin’s configuration schema.
If you want to prevent inheritance of the parent’s configuration, you can add `<inherited>false</inherited>` to the child’s plugin definition:
```xml
<plugin>
    <groupId>plugin</groupId>
    <artifactId>plugin</artifactId>
    <inherited>false</inherited>
    <configuration>
        <b>b</b>
        <c>b</c>
    </configuration>
</plugin>
```
This would result in only `<b>b</b>` and `<c>b</c>` in the child’s configuration, ignoring `<a>a</a>` and `<c>a</c>` from the parent.

jimisola:
@gnodet Thank you for the very thorough description with examples.

Been using Maven for decades (literally!) and I just realized that I've never seen the advanced configuration inheritance with `combine.children` and `combine.self`.

Really looking forward to mixins in Maven 4.

jimisola:
@gnodet Will this be included in Maven 4 (4.0.0-rc and final)?

rmannibucau (approved):

cstamas (approved):

PR #1211, opened 2023-07-27
[MNG-7854] Warn if imported depMgt is ignored as it already exists

cstamas:
Original issue described here https://issues.apache.org/jira/browse/MNG-7852 (as 3rd bullet) and related analysis is here https://github.com/hboutemy/sigstore-maven-plugin/blob/import/analysis.md

The problem was simply that previous ("deep" BOM) already imported protobuf-java, so 2nd depMgt import was simply ignored.

This simple change just adds warning.

Here is a reproducer:
https://github.com/cstamas/MNG-7852

W/ this PR output is like this:
https://gist.github.com/cstamas/b1ceadaea9f7015d419ec1047da5b51a

---

https://issues.apache.org/jira/browse/MNG-7854

### DIFF
--- a/maven-model-builder/src/main/java/org/apache/maven/model/composition/DefaultDependencyManagementImporter.java
+++ b/maven-model-builder/src/main/java/org/apache/maven/model/composition/DefaultDependencyManagementImporter.java
@@ -30,7 +30,9 @@ import org.apache.maven.model.Dependency;
 import org.apache.maven.model.DependencyManagement;
 import org.apache.maven.model.Model;
 import org.apache.maven.model.building.ModelBuildingRequest;
+import org.apache.maven.model.building.ModelProblem;
 import org.apache.maven.model.building.ModelProblemCollector;
+import org.apache.maven.model.building.ModelProblemCollectorRequest;
 
 /**
  * Handles the import of dependency management from other models into the target model.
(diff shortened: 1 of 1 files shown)


slawekjaranowski:
What will be if I define version of some library directly in depMng and next it will be also declared in imported depMng?

I would like to override in project one of dependencies from imported dependencies.

cstamas:
This combined with https://issues.apache.org/jira/browse/MNG-5600 could be a solution

hboutemy:
first, https://issues.apache.org/jira/projects/MPH/issues/MPH-183 should be solved, because currently there is absolutely nothing to debug dependencyManagement imports, and the sigstore-maven-plugin case showed that when you have such imports chained everywhere, it's really hard to understand how the dependencyManagement went in

food for thought: Maven core logging does not log anything at all about reactor dependencies effective POM building nor dependencies resolution. That's why we rely (and we should make it clear to end-users) on `mvn help:effective-pom -Dverbose` for effective POM debugging and `mvn dependency:list` / `mvn dependency:tree -Dverbose` for dependency resolution  debugging

should we add log to Maven core for effective POM building? I'm not convinced, because it will become too chatty I fear

should we add WARN on dependencyManagement import precedence decision: I'm definitively against it, because that's not a WARN: that would be at most an INFO

perhaps we need a better documentation, because I'm convinced many details are misunderstood

cstamas:
Well, here is a counter example: I have no **conflicting imports** as I import just one:
```
<project>
  <groupId>org.cstamas</groupId>
  <artifactId>test</artifactId>
  <version>1.0.0</version>
  <packaging>jar</packaging>
  <dependencyManagement>
    <dependencies>
        <dependency>
            <groupId>com.google.cloud</groupId>
            <artifactId>libraries-bom</artifactId>
            <version>26.9.0</version>
            <type>pom</type>
            <scope>import</scope>
        </dependency>      
    </dependencies>
  </dependencyManagement>
</project>
```
and with this PR it produces all these warnings: https://gist.github.com/cstamas/110351c2eeb6b56b8ceee9818be65872

Basically this one single BOM is "self conflicting". What now?

cstamas:
I still believe, that if this warning would be emitted, no "self conflicting" BOMs would be produced like that one above.

cstamas:
Superseded by https://github.com/apache/maven/pull/1370

jira-importer:
Resolve #9066

PR #1212, opened 2023-07-27
[EXPERIMENT] Delegate to Resolver for Artifact ordering

cstamas:
Applies MRESOLVER-390 changes, and removes all sprinkled artifact ordering, delegates fully to Resolver.

If you do `mvn -X package` you can see for example that enforcer plugin classpath is built in pre-order way from tree:
https://gist.github.com/cstamas/d1cb11d96938cdb9e6da2b46eec449b6

Now, if do `mvn -X package -Daether.system.resolveDependencies.visitor=levelOrder`, then enforcer plugin classpath is built in level order:
https://gist.github.com/cstamas/6498cfb31afe74ded1471c02d982eee6

Note: to build, needs https://github.com/apache/maven-resolver/pull/322

---

https://issues.apache.org/jira/browse/MRESOLVER-390
also related https://issues.apache.org/jira/browse/MNG-6357

### DIFF
--- a/maven-core/src/main/java/org/apache/maven/RepositoryUtils.java
+++ b/maven-core/src/main/java/org/apache/maven/RepositoryUtils.java
@@ -21,6 +21,7 @@ package org.apache.maven;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Collections;
+import java.util.HashMap;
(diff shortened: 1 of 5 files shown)


KKH1198 (changes requested):
...

PR #1214, opened 2023-08-03
Fix dependabot alerts

cstamas:
(that are not applicable, but look ugly).

These tests are NOT built, but are taken from real projects several years ago, and used in POM related UTs.

Hence, for example update of velocity is completely okay, even if in "real life" it would not compile.

### DIFF
--- a/maven-core/src/test/projects/lifecycle-executor/project-with-inheritance/pom.xml
+++ b/maven-core/src/test/projects/lifecycle-executor/project-with-inheritance/pom.xml
@@ -58,7 +58,7 @@ under the License.
     <plexusInteractivityVersion>1.0-alpha-6</plexusInteractivityVersion>
     <plexusInterpolationVersion>1.1</plexusInterpolationVersion>
     <plexusPluginManagerVersion>1.0-alpha-1</plexusPluginManagerVersion>
-    <plexusUtilsVersion>1.5.8</plexusUtilsVersion>
+    <plexusUtilsVersion>3.0.16</plexusUtilsVersion>
     <plexusJetty6Version>1.6</plexusJetty6Version>
     <plexusWebdavVersion>1.0</plexusWebdavVersion>
     <wagonVersion>1.0-beta-4</wagonVersion>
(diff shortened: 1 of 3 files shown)


gnodet:
> (that are not applicable, but look ugly).
> 
> These tests are NOT built, but are taken from real projects several years ago, and used in POM related UTs.
> 
> Hence, for example update of velocity is completely okay, even if in "real life" it would not compile.

So what's the point ?  The version don't even match those used by maven...

gnodet:
Ah, the answer is in the PR title...

gnodet (approved):

cstamas:
The point is **only** to make these go away https://github.com/apache/maven/security/dependabot

Yes, these are POMs used in UTs only in some POM-related tests...

mthmulders:
Wouldn't it make more sense to exclude those projects using [dependabot.yml](https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file)?

cstamas:
Sure! 

> Wouldn't it make more sense to exclude those projects using [dependabot.yml](https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file)?

Sure, that would work as well...

cstamas:
@mthmulders any example how to do it?

mthmulders:
> @mthmulders any example how to do it?

It seems not to be as easy as I hoped: https://github.com/dependabot/dependabot-core/issues/4364, https://github.com/dependabot/dependabot-core/issues/2883.

The best I could find is a suggestion to configure the "maven" `package-ecosystem` multiple times: once for the root of the project, and then for each location that you want to exclude you specify it again but with `interval: "never"`. Not sure if that works as intended.

cstamas:
Yes, that is my understanding as well... or ignore, but then one need to enlist ALL ignored deps, that probably overlaps with Maven used ones, where we DO WANT dependabot.

It would be simplest if there would be something like `ignore_directory` or alike.

PR #1217, opened 2023-08-21
[MNG-7862] The ModelLocator should always be used when locating pom.xml

gnodet:
JIRA issue: https://issues.apache.org/jira/browse/MNG-7862
IT PR: https://github.com/apache/maven-integration-testing/pull/289

There are a few places where the `pom.xml` file name is hardcoded. There's a known interface to locate those, so those references should be cleaned.
One IT has to be adjusted (see PR) because the default value in the model is changed from `../pom.xml` to `..` for the parent `relativePath` default value.

### DIFF
--- a/maven-core/src/test/java/org/apache/maven/internal/transformation/ConsumerPomArtifactTransformerTest.java
+++ b/maven-core/src/test/java/org/apache/maven/internal/transformation/ConsumerPomArtifactTransformerTest.java
@@ -74,17 +74,17 @@ class ConsumerPomArtifactTransformerTest {
 
         @Override
         public Model getRawModel(String groupId, String artifactId) throws IllegalStateException {
-            return null;
+            throw new UnsupportedOperationException();
         }
 
         @Override
         public Model getRawModel(Path p) {
-            return null;
+            throw new UnsupportedOperationException();
         }
 
         @Override
         public Path locate(Path path) {
-            return null;
+            throw new UnsupportedOperationException();
         }
     }
 }
--- a/maven-core/src/test/java/org/apache/maven/internal/transformation/ConsumerPomArtifactTransformerTest.java
+++ b/maven-core/src/test/java/org/apache/maven/internal/transformation/ConsumerPomArtifactTransformerTest.java
@@ -81,5 +81,10 @@ class ConsumerPomArtifactTransformerTest {
         public Model getRawModel(Path p) {
             return null;
         }
+
+        @Override
+        public Path locate(Path path) {
+            return null;
+        }
     }
 }
--- a/maven-embedder/src/main/java/org/apache/maven/cli/MavenCli.java
+++ b/maven-embedder/src/main/java/org/apache/maven/cli/MavenCli.java
@@ -1349,23 +1349,20 @@ public class MavenCli {
             alternatePomFile = commandLine.getOptionValue(CLIManager.ALTERNATE_POM_FILE);
         }
 
+        File current = baseDirectory;
         if (alternatePomFile != null) {
-            File pom = resolveFile(new File(alternatePomFile), workingDirectory);
-            if (pom.isDirectory()) {
-                if (modelProcessor != null) {
-                    pom = modelProcessor.locatePom(pom);
-                } else {
-                    pom = new File(pom, "pom.xml");
-                }
-            }
+            current = resolveFile(new File(alternatePomFile), workingDirectory);
+        }
 
+        File pom;
+        if (current.isDirectory() && modelProcessor != null) {
+            pom = modelProcessor.locatePom(current);
+        } else {
+            pom = current;
+        }
+
+        if (pom.isFile()) {
             return pom;
-        } else if (modelProcessor != null) {
-            File pom = modelProcessor.locatePom(baseDirectory);
-
-            if (pom.isFile()) {
-                return pom;
-            }
         }
 
         return null;
--- a/maven-model-builder/src/main/java/org/apache/maven/model/building/ModelSource3.java
+++ b/maven-model-builder/src/main/java/org/apache/maven/model/building/ModelSource3.java
@@ -37,9 +37,9 @@ public interface ModelSource3 extends ModelSource2 {
      * to find the POM file, else if no locator is provided, a file named 'pom.xml' needs to be
      * used by the requested model source.
      *
-     * @param locator locator used to actually locate the pom file.
-     * @param relPath path of the requested model source relative to this model source POM.
-     * @return related model source or <code>null</code> if no such model source.
+     * @param locator locator used to locate the pom file
+     * @param relPath path of the requested model source relative to this model source POM
+     * @return related model source or <code>null</code> if no such model source
      */
     ModelSource3 getRelatedSource(ModelLocator locator, String relPath);
 
--- a/maven-model-builder/src/main/java/org/apache/maven/model/building/ModelSource3.java
+++ b/maven-model-builder/src/main/java/org/apache/maven/model/building/ModelSource3.java
@@ -1,0 +1,56 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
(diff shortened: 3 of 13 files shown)


maven-model-builder/src/main/java/org/apache/maven/model/building/ModelSource3.java
```diff
@@ -0,0 +1,47 @@
...
+import org.apache.maven.model.locator.ModelLocator;
+
+/**
+ * Enhancement to the {@link ModelSource2} to support locating POM files using the {@link ModelLocator}
+ * when pointing to a directory.
+ */
+public interface ModelSource3 extends ModelSource2 {
+    /**
+     * Returns model source identified by a path relative to this model source POM. Implementation <strong>MUST</strong>
+     * be able to accept <code>relPath</code> parameter values that
```
elharo, line 30: delete "be able to"

maven-model-builder/src/main/java/org/apache/maven/model/building/ModelSource3.java
```diff
@@ -0,0 +1,47 @@
...
+     * <ul>
+     * <li>use either / or \ file path separator</li>
+     * <li>have .. parent directory references</li>
+     * <li>point either at file or directory, in the latter case the provided {@code ModelLocator} will be used
```
elharo, line 34: run-on sentence. Break up

maven-model-builder/src/main/java/org/apache/maven/model/building/ModelSource3.java
```diff
@@ -0,0 +1,47 @@
...
+     * to find the POM file, else if no locator is provided, a file named 'pom.xml' needs to be used by the
+     * requested model source.</li>
+     * </ul>
+     *
+     * @param relPath is the path of the requested model source relative to this model source POM.
```
elharo, line 39: delete "is the"

elharo (commented):
Is this for v4 only? Some of the API changes look incompatible.

gnodet:
> Is this for v4 only? Some of the API changes look incompatible.

Yes, I did not really envision to backport it.  I think most of the changed methods are new in 4.0.  The exceptions are the default value in the model and the new `ModelSource3` which I introduced to avoid adding a method to an interface that could be possibly implemented by plugins / extensions.  I may have missed a possible incompatible change if a plugin/extension actually implements `ModelSource2` (as I should provide the previous behaviour in such a case).

elharo:
Probably unrelated but why in the CI output do I see :"Sending email to: rishi@ca.ibm.com"?

gnodet:
> Probably unrelated but why in the CI output do I see :"Sending email to: [rishi@ca.ibm.com](mailto:rishi@ca.ibm.com)"?

Not sure, I've disabled the s390x pipeline as it does not take into account the fact that a given branch should be tested with the maven-integration-testing branch with the same name if it exists.

elharo:
Filed https://issues.apache.org/jira/browse/MNG-7863?filter=-2

elharo:
OK, looks good overall. Is it possible to retrigger the full CI so we can see that S390 is no longer included?

maven-core/src/test/java/org/apache/maven/internal/transformation/ConsumerPomArtifactTransformerTest.java
```diff
@@ -81,5 +81,10 @@ public Model getRawModel(String groupId, String artifactId) throws IllegalStateE
         public Model getRawModel(Path p) {
             return null;
         }
+
+        @Override
+        public Path locate(Path path) {
+            return null;
```
elharo, line 87: If this is just a stub in a test that isn't called, I'd prefer `throw new UnsupportedOperationException()`

maven-embedder/src/main/java/org/apache/maven/cli/MavenCli.java
```diff
@@ -1349,23 +1349,20 @@ private File determinePom(final CommandLine commandLine, final String workingDir
...
+        if (current.isDirectory() && modelProcessor != null) {
+            pom = modelProcessor.locatePom(current);
+        } else {
+            pom = current;
+        }
 
-            if (pom.isFile()) {
-                return pom;
-            }
+        if (pom.isFile()) {
```
elharo, line 1364: If the pom locates a directory, would that be a bug that should throw an exception rather than returning null?

maven-model-builder/src/main/java/org/apache/maven/model/building/ModelSource3.java
```diff
@@ -0,0 +1,56 @@
...
+     * <ul>
+     * <li>use either / or \ file path separator</li>
+     * <li>have .. parent directory references</li>
+     * <li>point either at file or directory</li>
+     * </ul>
+     * If the given path points at a directory, the provided {@code ModelLocator} will be used
+     * to find the POM file, else if no locator is provided, a file named 'pom.xml' needs to be
+     * used by the requested model source.
+     *
+     * @param locator locator used to actually locate the pom file.
```
elharo, line 40: delete actually
delete period

maven-model-builder/src/main/java/org/apache/maven/model/building/ModelSource3.java
```diff
@@ -0,0 +1,56 @@
...
+     * @param relPath path of the requested model source relative to this model source POM.
+     * @return related model source or <code>null</code> if no such model source.
```
elharo, line 42: no period

maven-model-builder/src/main/java/org/apache/maven/model/building/ModelSource3.java
```diff
@@ -0,0 +1,56 @@
...
+     */
+    ModelSource3 getRelatedSource(ModelLocator locator, String relPath);
+
+    /**
+     * When using a ModelSource3, the method with a {@code ModelLocator} argument should
+     * be used instead.
+     *
+     * @deprecated use {@link #getRelatedSource(ModelLocator, String)} instead
```
elharo, line 50: what about when not using a ModelSource3?

maven-embedder/src/main/java/org/apache/maven/cli/MavenCli.java
gnodet: This method is called twice, a first time when discovering extensions, in which case, the `modelLocator` is not set up yet, and second time to create the actual maven request that will be processed.  So while the `modelLocator` should never return a directory, the `determinePom` method may return `null` during extension discovery.  I'll change the code to only call `determinePom` when `modelLocator` is set up correctly.

maven-model-builder/src/main/java/org/apache/maven/model/building/ModelSource3.java
gnodet: This method overrides the `ModelSource2` method.  If the source does not implement `ModelSource3`, then the `ModelSource2.getRelatedSource(String)` will be called, see https://github.com/apache/maven/pull/1217/files#diff-f590407a614fe5e60543bf824d2de0b5df48fddfcbe7c42e64ac1f6606533347R1497 .

maven-embedder/src/main/java/org/apache/maven/cli/MavenCli.java
gnodet: I tried, but the logic is expected downstream, mostly when you run Maven inside a directory with no `pom.xml`.  In such cases, a `null` value is expected.  Also, the `modelLocator` is required to always return a value, but can point to a non existent file, hence the check.

maven-model-builder/src/main/java/org/apache/maven/model/building/ModelSource3.java
gnodet: Or should I deprecate ModelSource2 in favour of ModelSource3, in which case, I could simply to make `ModelSource3` extends `ModelSource2`, that would be more clear imho.

gnodet: Such a change looks a bit risky, as some code may just expect to find `ModelSource2` as they are kept in the `ModelCache`.  I'd rather leave it that way.

elharo (approved):

jira-importer:
Resolve #9057

PR #1220, opened 2023-08-31
[MNG-7866] Improvements to the logging API usage (technical debt)

sebastien-doyon:
JIRA issue: https://issues.apache.org/jira/browse/MNG-7866

Use message formats with placeholders to avoid premature formatting and avoid the unnecessary garbage when then log level is disabled :

org.apache.maven.project.artifact.MavenMetadataSource
org.apache.maven.toolchain.DefaultToolchainsBuilder
org.apache.maven.DefaultMaven
org.apache.maven.classrealm.DefaultClassRealmManager
org.apache.maven.internal.aether.LoggingRepositoryListener
org.apache.maven.lifecycle.internal.DefaultLifecyclePluginAnalyzer
org.apache.maven.lifecycle.internal.LifecycleDependencyResolver
org.apache.maven.lifecycle.internal.builder.multithreaded.MultiThreadedBuilder
org.apache.maven.plugin.internal.DefaultMavenPluginManager
org.apache.maven.plugin.internal.DefaultPluginDependenciesResolver
org.apache.maven.plugin.prefix.internal.DefaultPluginPrefixResolver
org.apache.maven.plugin.version.internal.DefaultPluginVersionResolver
org.apache.maven.project.collector.MultiModuleCollectionStrategy
org.apache.maven.toolchain.DefaultToolchain

Guarding some logging statements with conditionals on isXXXXEnabled() to avoid unnecessary garbage when then log level is disabled. Useful when some formatting must be done outside the logger call :

org.apache.maven.classrealm.DefaultClassRealmManager
org.apache.maven.internal.impl.DefaultLog
org.apache.maven.lifecycle.DefaultLifecycles
org.apache.maven.plugin.internal.DefaultMavenPluginManager
org.apache.maven.plugin.internal.MojoLogWrapper

Remove some unneeded conditional guarding to avoid testing twice if the log level is enabled :

org.apache.maven.classrealm.DefaultClassRealmManager
org.apache.maven.lifecycle.DefaultLifecycles
org.apache.maven.plugin.prefix.internal.DefaultPluginPrefixResolver
org.apache.maven.plugin.version.internal.DefaultPluginVersionResolver

New test class :

org.apache.maven.classrealm.DefaultClassRealmManagerTest

Test class modified to reflect the new logger API usage :

org.apache.maven.toolchain.DefaultToolchainTest


maven-core/src/main/java/org/apache/maven/internal/impl/DefaultLog.java
```diff
@@ -33,12 +33,16 @@ public DefaultLog(Logger logger) {
     }
 
     public void debug(CharSequence content) {
-        logger.debug(toString(content));
+        if (isDebugEnabled()) {
```
elharo, line 36: I like this file's changes. They're nice and clean and in one place.

maven-core/src/main/java/org/apache/maven/plugin/internal/MojoLogWrapper.java
```diff
@@ -34,7 +34,9 @@ public MojoLogWrapper(Logger logger) {
     }
 
     public void debug(CharSequence content) {
-        logger.debug(toString(content));
+        if (isDebugEnabled()) {
```
elharo, line 37: I notice this duplicates a lot of content from DefaultLog above. I wonder if that's fixable somehow? Not necessarily in this PR.

maven-core/src/test/java/org/apache/maven/classrealm/DefaultClassRealmManagerTest.java
```diff
@@ -0,0 +1,164 @@
...
+        when(logger.isDebugEnabled()).thenReturn(true);
+
+        DefaultClassRealmManager classRealmManager;
+        ClassRealm classRealm;
+
+        InOrder verifier = inOrder(logger);
+
+        PlexusContainer container = new DefaultPlexusContainer();
+
+        try (MockedStatic<LoggerFactory> mockedLoggerFactory = Mockito.mockStatic(LoggerFactory.class)) {
```
elharo, line 109: In general, I'm skeptical of testing log messages. I tend to think of them as side effects that can change pretty freely, and not part of a public API contract that should be tested.

maven-core/src/main/java/org/apache/maven/classrealm/DefaultClassRealmManager.java
```diff
@@ -152,14 +150,20 @@ private ClassRealm createRealm(
             List<String> parentImports,
             Map<String, ClassLoader> foreignImports,
             List<Artifact> artifacts) {
-        Set<String> artifactIds = new LinkedHashSet<>();
+        Set<String> artifactIds = Collections.emptySet();
```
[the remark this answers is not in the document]
sebastien-doyon: I think in this case it is not a problem. The **artifactIds** reference a collection that stays local to the method, it is not escaping the class and is not an instance variable. Knowing that, the emptySet() is not modified in this state of the code. Do you agree on the safe usage?

maven-core/src/main/java/org/apache/maven/classrealm/DefaultClassRealmManager.java
```diff
@@ -301,20 +305,22 @@ private void callDelegates(
     }
 
     private Set<String> populateRealm(ClassRealm classRealm, List<ClassRealmConstituent> constituents) {
-        Set<String> includedIds = new LinkedHashSet<>();
+        Set<String> includedIds = Collections.emptySet();
```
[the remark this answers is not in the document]
sebastien-doyon: What is your concerns? The **includedIds** reference a collection that stays local to the methods **populateRealm** and **createRealm**, it is not escaping the class and is not an instance variable. The set is only modified if the logger debug log level is enabled (line 319 and 321), it is set with a LinkedHashSet on the same condition (lines 310 and 311). I see no point of creating a LinkedHashSet if the log level is not at least debug since this collection is only used for logging purpose. What do you think according to this?

maven-core/src/test/java/org/apache/maven/classrealm/DefaultClassRealmManagerTest.java
sebastien-doyon: I agree with the principle. I did this test class because in the it was specified in the contribution guidelines : "Make sure you have added the necessary tests (JUnit/[Core IT tests](https://maven.apache.org/core-its/core-it-suite/)) for your changes.". And I found at least one other test example that was testing log messages, the DefaultToolchainTest class. I don't know if any external tools may be expecting maven to log in a specific format, like Maven Daemon? If you think the tests are a bad idea, I am willing to remove, just tell me.

maven-core/src/main/java/org/apache/maven/classrealm/DefaultClassRealmManager.java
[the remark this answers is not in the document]
elharo: You're saying that at this commit, mixing mutable and immutable lists is not a problem. That's likely true. However there's no guarantee this will still be true in a commit 5 years down the road. I've seen slipping an immutable list into something that wasn't expecting it cause very painful production problems. It's just safer to always use one or the other, and make sure the declared types make it clear which is being used. Any collection not explicitly declared as immutable should not be immutable. Yes, this is a design flaw in the Java Collections API going back to Java 1.1, but it's a design flaw we have to live with.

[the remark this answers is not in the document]
sebastien-doyon: Fine with me, thanks for clarifying. Reverting this line

maven-core/src/main/java/org/apache/maven/classrealm/DefaultClassRealmManager.java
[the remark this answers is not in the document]
sebastien-doyon: Reverting this line according to avoid problems

maven-core/src/main/java/org/apache/maven/classrealm/DefaultClassRealmManager.java
```diff
@@ -152,14 +150,15 @@ private ClassRealm createRealm(
...
-        List<ClassRealmConstituent> constituents = new ArrayList<>();
-
-        if (artifacts != null) {
+        if (artifacts != null && !artifacts.isEmpty()) {
             for (Artifact artifact : artifacts) {
                 if (!isProvidedArtifact(artifact)) {
-                    artifactIds.add(getId(artifact));
+                    if (logger.isDebugEnabled()) {
+                        artifactIds.add(getId(artifact));
+                    }
```
gnodet, line 161: I would not make the computation of the artifactIds condition to the log level.

maven-core/src/main/java/org/apache/maven/classrealm/DefaultClassRealmManager.java
```diff
@@ -301,20 +300,18 @@ private void callDelegates(
...
 
         for (ClassRealmConstituent constituent : constituents) {
             File file = constituent.getFile();
 
-            String id = getId(constituent);
-            includedIds.add(id);
-
             if (logger.isDebugEnabled()) {
-                logger.debug("  Included: " + id);
+                String id = getId(constituent);
```
gnodet, line 311: Same here.  Move back the `includedIds` computation and simply remove the `isDebugEnabled()` call.

maven-core/src/main/java/org/apache/maven/classrealm/DefaultClassRealmManager.java
sebastien-doyon: Let me try to explain what was the point of this change. The artifactIds list is computed only for debugging purpose. By adding the condition on the log level it is avoiding the computation of building this list if logging is not going to happen. The same for includedIds in your other comment. They are both used only to compute the debug logs.

One other comment, while trying to reduce the complexity of this code and limit the number of isDebugEnabled()  conditions, I came to the conclusion that something in this logic is fishy. If you look closely on the createRealm() method where the artifactIds list is populated and in the populateRealm() where the includedIds is populated. The only elements that would be in the artifactIds list and not in the includedIds list are artifacts with getFile() null value. If I am right the logic of this code can be reduced a lot, both limiting the isDebugEnabled()  conditions and removing some list constructions in the populateRealm() method.

Please tell me what do you think about this idea. Did I miss something in the code? I can patch the PR with the simplified code if you agree.

sebastien-doyon: I made a commit with the proposed changes (direct link [here](maven-core/src/main/java/org/apache/maven/classrealm/DefaultClassRealmManager.java)), please comment. This is reducing the number of isDebugEnabled() blocks and the collections used only for logging informations. The business logic is unchanged and the logged information is the same.

gnodet: LGTM

gnodet (approved):

gnodet (changes requested):

maven-core/src/main/java/org/apache/maven/classrealm/DefaultClassRealmManager.java
```diff
@@ -152,16 +149,15 @@ private ClassRealm createRealm(
...
 
-        if (artifacts != null) {
+        if (artifacts != null && !artifacts.isEmpty()) {
             for (Artifact artifact : artifacts) {
                 if (!isProvidedArtifact(artifact)) {
-                    artifactIds.add(getId(artifact));
                     if (artifact.getFile() != null) {
                         constituents.add(new ArtifactClassRealmConstituent(artifact));
+                    } else if (logger.isDebugEnabled()) {
+                        logger.debug("  Excluded: {}", getId(artifact));
```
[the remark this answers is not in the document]
sebastien-doyon: I think you are right, but to be sure all path that are not adding the artifact to the constituents list are logged as excluded, I can keep the condition here and add the same else block to the "if isProvidedArtifact" has you proposed. What do you think? Alternative fix could be a simple if after the"if isProvidedArtifact" if I add a continue; after the line 158. This is a matter of taste, what do you think?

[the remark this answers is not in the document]
gnodet: What about merging the `if`s
```
                if (!isProvidedArtifact(artifact) && artifact.getFile() != null) {
                    constituents.add(new ArtifactClassRealmConstituent(artifact));
                } else if (logger.isDebugEnabled()) {
                    logger.debug("  Excluded: {}", getId(artifact));
                }
```

maven-core/src/main/java/org/apache/maven/internal/aether/LoggingRepositoryListener.java
```diff
@@ -63,48 +63,47 @@ public void metadataResolved(RepositoryEvent event) {
...
+            logger.warn(
+                    "The POM for {} is invalid, transitive dependencies (if any) will not be available: {}",
+                    event.getArtifact(),
+                    event.getException().getMessage());
         } else {
-            logger.warn(buffer + ", enable verbose output (-X) for more details");
+            logger.warn(
+                    "The POM for {} is invalid, transitive dependencies (if any) will not be available, enable "
+                            + "verbose output (-X) for more details",
+                    event.getArtifact());
```
gnodet, line 101: The code above is wrong and does not make much sense. To be consistent with other parts of this class, the whole method should look like:
```
        if (logger.isDebugEnabled()) {
            logger.warn(
                "The POM for {} is invalid, transitive dependencies (if any) will not be available: {}",
                event.getArtifact(),
                event.getException().getMessage(),
                event.getException());
        } else {
            logger.warn(
                "The POM for {} is invalid, transitive dependencies (if any) will not be available: {}",
                event.getArtifact(),
                event.getException().getMessage());
        }
```

gnodet (approved):

cstamas (approved):

jira-importer:
Resolve #9052

PR #1221, opened 2023-09-01
[MNG-7861] Remove artifacts which should be kept hidden

gnodet:
JIRA: https://issues.apache.org/jira/browse/MNG-7861
See https://github.com/apache/maven/commit/e39142b77a5a140cce50592608029c19b9716ed1#r126188891


slawekjaranowski (approved):
Should be linked with https://issues.apache.org/jira/browse/MNG-7861

PR #1228, opened 2023-09-08
Resolve circular dependencies for project org.apache.maven:maven-xml-impl

w6et:
… resolve dependencies for project org.apache.maven:maven-xml-impl

reason:Circular dependencies：
maven-xml-impl
    ->plexus-xml
        ->maven-xml-impl
plexus-xml need to exclusion maven-xml-impl
2)rename 'Implementation of Maven API XML' to 'maven-xml-impl(Implementation of Maven API XML)'

       pulling in other changes.

### DIFF
--- a/maven-xml-impl/pom.xml
+++ b/maven-xml-impl/pom.xml
@@ -26,7 +26,7 @@ under the License.
   </parent>
 
   <artifactId>maven-xml-impl</artifactId>
-  <name>maven-xml-impl(Implementation of Maven API XML)</name>
+  <name>Implementation of Maven API XML</name>
   <description>Provides the implementation classes for the Maven API XML</description>
 
   <dependencies>
--- a/maven-xml-impl/pom.xml
+++ b/maven-xml-impl/pom.xml
@@ -26,7 +26,7 @@ under the License.
   </parent>
 
   <artifactId>maven-xml-impl</artifactId>
-  <name>Implementation of Maven API XML</name>
+  <name>maven-xml-impl(Implementation of Maven API XML)</name>
   <description>Provides the implementation classes for the Maven API XML</description>
 
   <dependencies>
@@ -48,6 +48,12 @@ under the License.
       <artifactId>plexus-xml</artifactId>
       <version>${plexusXmlVersion}</version>
       <scope>provided</scope>
+      <exclusions>
+        <exclusion>
+          <groupId>org.apache.maven</groupId>
+          <artifactId>maven-xml-impl</artifactId>
+        </exclusion>
+      </exclusions>
     </dependency>
   </dependencies>
 


maven-xml-impl/pom.xml
```diff
@@ -26,7 +26,7 @@ under the License.
   </parent>
 
   <artifactId>maven-xml-impl</artifactId>
-  <name>Implementation of Maven API XML</name>
+  <name>maven-xml-impl(Implementation of Maven API XML)</name>
```
gnodet, line 29: This is not coherent with the other packages afaik.
I'd rather do that globally all all modules if we want to go that way.

maven-xml-impl/pom.xml
```diff
@@ -48,6 +48,12 @@ under the License.
       <artifactId>plexus-xml</artifactId>
       <version>${plexusXmlVersion}</version>
       <scope>provided</scope>
+      <exclusions>
+        <exclusion>
+          <groupId>org.apache.maven</groupId>
+          <artifactId>maven-xml-impl</artifactId>
+        </exclusion>
+      </exclusions>
```
gnodet, line 56: Doesn't the current module take precedence during resolution ?
What kind of issues does it actually cause ?

w6et: > Doesn't the current module take precedence during resolution ? What kind of issues does it actually cause ?
in  my landtop，windows11 ,apache-maven-3.9.4
run command 'mvn clean compile  -Dmaven.test.skip=true ' in maven home_dir(my case is D:\works\git\maven),it occured an exception as below:
[ERROR] Failed to execute goal on project maven-xml-impl: Could not resolve dependencies for project org.apache.maven:maven-xml-impl:jar:4.0.0-alpha-8-SNAPSHOT: The following artifacts could not be resolved: org.apache.maven:maven-xml-impl:jar:4.0.0-alpha-8-SNAPSHOT (absent): org.apache.maven:maven-xml-impl:jar:4.0.0-alpha-8-SNAPSHOT was not found in https://xxxxxxxxxx/repository/public during a previous attempt. This failure was cached in the local repository and resolution is not reattempted until the update interval of aliyun has elapsed or updates are forced -> [Help 1]
.......
**It looks like maven-xml-impl is needed to get plexus-xml, but maven-xml-impl is being built**

gnodet: I wonder if that's a bug somehow in the resolution... @cstamas ?

hboutemy:
I agree that there is a problem
I don't think the current PR is the right approach
as found [on the ML while discussing on the topic](https://lists.apache.org/thread/8t5szmn8fhlrklhg5dgv496k26tz3w3p), I thin the approach is to create a Maven Shared Component that provides MetaAnnotations, XML API and XML Implementation separately from Maven core
I don't know how to name this release: "Maven fundamentals"?

gnodet:
> I agree that there is a problem
> I don't think the current PR is the right approach
> as found [on the ML while discussing on the topic](https://lists.apache.org/thread/8t5szmn8fhlrklhg5dgv496k26tz3w3p), I thin the approach is to create a Maven Shared Component that provides MetaAnnotations, XML API and XML Implementation separately from Maven core
> I don't know how to name this release: "Maven fundamentals"?

While I'm not opposed to moving things, I don't think this will solve the problem.
The problem here is the cycle between maven-xml-impl / plexus-xml / maven-xml-impl.
Moving maven-xml-impl in another location won't change that problem, so while this may not be a sufficient solution, I think this PR is still needed.

gnodet (changes requested):
@w6et Could you please remove the module name change ?

w6et:
> @w6et Could you please remove the module name change ?

Done, @gnodet

gnodet (approved):

PR #1230, opened 2023-09-08
[MNG-7810] Provide a SPI for the v4 api

gnodet:
JIRA issue: https://issues.apache.org/jira/browse/MNG-7810

This PR provides an empty placeholder for the Maven 4 SPI.

The package is `org.apache.maven.api.spi` and should contain all services that can be easily provided by extensions.


jira-importer:
Resolve #9182

PR #1232, opened 2023-09-10
Move Features to api

gnodet:
- Move Features to the api


cstamas (approved):
+1000

api/maven-api-core/src/main/java/org/apache/maven/api/feature/Features.java
```diff
@@ -0,0 +1,67 @@
...
+package org.apache.maven.api.feature;
+
+import java.util.Map;
+import java.util.Properties;
+
+import org.apache.maven.api.Session;
+import org.apache.maven.api.annotations.Nullable;
+
+/**
+ * Centralized class for feature information
```
cstamas, line 28: maybe "...for Maven Core feature information" (and be clear they are/should be "final" from session perspective)

PR #1235, opened 2023-09-11
[MNG-7876] Model version analysis and downgrade

gnodet:
The modello model is used to generate a class which is able to find the minimal model version needed for a given `Model` object.  This allows a transparent downgrade during the consumer POM transformation (which can be turned off using a new flag).

JIRA issue: http://issues.apache.org/jira/browse/MNG-7876
IT PR: https://github.com/apache/maven-integration-testing/pull/291


cstamas (approved):

jira-importer:
Resolve #9039

PR #1238, opened 2023-09-11
Experiment: New Maven4 tansfer logging

cstamas:
Example of rough sketch (actual loggers are not fully done yet) for mvn4 logging:

Supports:
* quiet (as before)
* classic (as Maven3 did always log)
* classic_light (as Maven3 but only "Downloaded..." lines)
* summary

Examples (building maven-resolver-ant-tasks always with empty local repo):
* classic (2459 logged lines): https://gist.github.com/cstamas/67770cb5da1d36dc48128a3529b52905
* classic_light (1575 logged lines): https://gist.github.com/cstamas/7ff6c5c4a99c6ea4f3e3a955795c9a92
* summary (697 logged lines): https://gist.github.com/cstamas/c40d50b2f4f952f4eb7a519867756f78


maven-core/src/main/java/org/apache/maven/execution/TransferListenerConfiguration.java
```diff
@@ -0,0 +1,197 @@
...
+
+        /**
+         * Classic: as you knew Maven 3 since beginning.
+         */
+        CLASSIC,
+
+        /**
+         * Classic "light": similar to CLASSIC but without "Downloading..." lines (only "Downloaded...").
+         */
+        CLASSIC_LIGHT,
```
mbien, line 46: I like this one which sounds like a good compromise. I have the suspicion that some of the complaints might come from users which use maven without a fully featured terminal which would allow in-place updates (e.g in CI). 

The classic mode would print every dl progress update into a new line which is not optimal. This is not a unique problem to maven, git does the exacts same last time I checked.

PR #1249, opened 2023-09-19
[MNG-7820] Get rid of plexus-utils Os class

gnodet:


maven-model-builder/src/main/java/org/apache/maven/utils/Os.java
```diff
@@ -0,0 +1,213 @@
...
+     * @see <a href="https://issues.apache.org/jira/browse/HADOOP-3318">HADOOP-3318</a>
+     */
+    private static final String DARWIN = "darwin";
+
+    /**
+     * The path separator.
+     */
+    private static final String PATH_SEP = System.getProperty("path.separator");
+
+    static {
```
elharo, line 124: Please add a comment explaining why this is a static initializer instead of initializing the fields directly. I see it now, but I almost missed it; and I can believe someone down the road might remove this. 

Although now that I look even closer, I wonder if there's a problem since this calls a static method before the static initializer has completed. Maybe replace the call to isWindowsFamily with OS_NAME.contains(FAMILY_WINDOWS) instead.

maven-model-builder/src/main/java/org/apache/maven/utils/Os.java
```diff
@@ -0,0 +1,213 @@
...
+        // windows probing logic relies on the word 'windows' in the OS
+        boolean isWindows = OS_NAME.contains(FAMILY_WINDOWS);
+        boolean is9x = false;
+        boolean isNT = false;
+        if (isWindows) {
+            // there are only four 9x platforms that we look for
+            is9x = (OS_NAME.contains("95")
+                    || OS_NAME.contains("98")
+                    || OS_NAME.contains("me")
+                    // wince isn't really 9x, but crippled enough to
```
elharo, line 149: If Maven doesn't run on CE , do we care? That is, does it matter if there's a profile for CE? Could there be a CE profile in a pom.xml that is n ever activated but nonethless breaks the build if this isn't supported here?

elharo (approved):

maven-model-builder/src/main/java/org/apache/maven/utils/Os.java
gnodet: > Please add a comment explaining why this is a static initializer instead of initializing the fields directly. I see it now, but I almost missed it; and I can believe someone down the road might remove this.

It's not possible to remove the initializer of a static final constant, the compiler would reject with an error.  Anyway, I've added a comment.

> Although now that I look even closer, I wonder if there's a problem since this calls a static method before the static initializer has completed. Maybe replace the call to isWindowsFamily with OS_NAME.contains(FAMILY_WINDOWS) instead.

That's what it ends up doing [in the `isWindowsFamily` method](https://github.com/apache/maven/pull/1249/files/f4eefa52d68b4ce61026030ac0a27e57346efe5f#diff-0d69abd148e63b11a4c58601c7e7ae73fda92b61a9facc97333061be7bec1ee8R141) btw.  I'd rather avoir duplicating the logic, and the `isWindowsFamily` is the one used by the OS activation.

maven-model-builder/src/main/java/org/apache/maven/utils/Os.java
gnodet: I doubt Windows 95 is supported either: https://www.java.com/en/download/help/win95.html
However, the next line is `isNT = !is9x` so I'm not sure about moving Windows 95/CE into the NT world either...

michael-o (commented):
No opinion here.

cstamas (approved):

PR #1251, opened 2023-09-19
[MNG-7820] Get rid of plexus-utils introspection classes

gnodet:
Use maven-shared-utils reflection classes.

### DIFF
--- a/maven-core/src/main/java/org/apache/maven/plugin/ClassMap.java
+++ b/maven-core/src/main/java/org/apache/maven/plugin/ClassMap.java
@@ -27,14 +27,8 @@ import java.util.Map;
  * A cache of introspection information for a specific class instance.
  * Keys {@link Method} objects by a concatenation of the
  * method name and the names of classes that make up the parameters.
- *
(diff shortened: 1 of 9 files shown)


gnodet:
I think I can also remove the copyright because the code has already been donated to the ASF as those classes also exist [in maven-shared-utils](https://github.com/apache/maven-shared-utils/tree/master/src/main/java/org/apache/maven/shared/utils/introspection)...
@elharo do you agree ?

michael-o (commented):
No opinion here.

elharo:
I'm sorry. IANAL and I simply don't know. We need an official ruling from someone in a position at the Apache project to offer legal advice and make decisions like this.

gnodet:
> I'm sorry. IANAL and I simply don't know. We need an official ruling from someone in a position at the Apache project to offer legal advice and make decisions like this.

Well, it was really a courtesy question.  I'll use the maven-shared-utils copy of those classes which are already in the ASF so definitely no copyright / legal issue.

cstamas (approved):

jira-importer:
Resolve #9100

PR #1252, opened 2023-09-19
[MNG-7891] Support configuration for extensions

gnodet:
JIRA issue: https://issues.apache.org/jira/browse/MNG-7891
IT PR: https://github.com/apache/maven-integration-testing/pull/296

### DIFF
--- a/api/maven-api-model/src/main/mdo/maven.mdo
+++ b/api/maven-api-model/src/main/mdo/maven.mdo
@@ -3281,6 +3281,12 @@
           <description>The version of the extension.</description>
           <type>String</type>
         </field>
+        <field>
+          <name>configuration</name>
+          <version>4.1.0+</version>
+          <description>The configuration of the extension.</description>
+          <type>DOM</type>
+        </field>
       </fields>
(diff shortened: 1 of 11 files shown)


laeubi:
@gnodet I created a little example extension / project here:
- https://github.com/apache/maven-integration-testing/pull/295

cstamas (approved):
With a remark: personally am really tired to use DOM/Xpp3Dom as "config", we need something smarter here (not in this PR of course, this is just a remark/rant)

laeubi:
> With a remark: personally am really tired to use DOM/Xpp3Dom as "config"

It is `PlexusConfiguration` here ;-)

> we need something smarter

Don't know what smart you have in mind, but of course one could think about mapping this to an object annotated similar to what mojo uses (just the configuration part), but last time I tried something similar it was surprisingly hard (e.g. when it comes to interpolation/default values).

At laest this is much better than having no configuration at al or invent something that is completely extension specific (like custom properties files).

jira-importer:
Resolve #9025

PR #1253, opened 2023-09-21
[MNG-7893] Allow versioning the superpom according to the model version

gnodet:
JIRA issue: https://issues.apache.org/jira/browse/MNG-7893

Btw, I wonder if the recent changes such as https://github.com/apache/maven/pull/1139 and https://github.com/apache/maven/pull/1085 should be updated to only affect the 4.1.0 superpom...  Thoughts welcomed.

### DIFF
--- a/maven-core/src/main/java/org/apache/maven/DefaultMaven.java
+++ b/maven-core/src/main/java/org/apache/maven/DefaultMaven.java
@@ -495,6 +495,7 @@ public class DefaultMaven implements Maven {
      * @return A {@link Set} of profile identifiers, never {@code null}.
      */
     private Set<String> getAllProfiles(MavenSession session) {
+        // TODO: which version of super pom should we use here ?
         final Model superPomModel = superPomProvider.getSuperModel("4.0.0").getDelegate();
         final Set<MavenProject> projectsIncludingParents = new HashSet<>();
         for (MavenProject project : session.getProjects()) {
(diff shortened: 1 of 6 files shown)


maven-core/src/main/java/org/apache/maven/DefaultMaven.java
```diff
@@ -495,6 +495,7 @@ private void validatePrerequisitesForNonMavenPluginProjects(List<MavenProject> p
      * @return A {@link Set} of profile identifiers, never {@code null}.
      */
     private Set<String> getAllProfiles(MavenSession session) {
+        // TODO: which version of super pom should we use here ?
```
cstamas, line 498: Maybe the same version that project uses? But: what happens if reactor has multiple versions (not same)? -- IMHO should fail.

cstamas (approved):

maven-core/src/main/java/org/apache/maven/DefaultMaven.java
gnodet: I've update this part of the computation to be coherent with the way superPoms are used during model building.  The superPom of a given project is determined by the modelVersion of that project and not the modelVersion of its upper parent.  What this means is that if you write a 4.1.0 project which inherit a 4.0.0 parent, the superPom will be 4.1.0.

cstamas (approved):

PR #1258, opened 2023-09-22
Fix unneeded/wrong substring calls in expression evaluation and put back weak references in cache

gnodet:
As indicated at https://github.com/apache/maven/pull/812#discussion_r990825335

If the interpolated string is `${settings.foo}`, the expression is `settings.foo` and we pass `ettings.foo` to the `ReflectionValueExtractor`. It's working because the call just ignore whatever is on the left of the first `.`...


cstamas (approved):

slawekjaranowski:
Is it worth doing in 3.9.x also?

slawekjaranowski (approved):

gnodet:
Raised https://issues.apache.org/jira/browse/MSHARED-1304 to not forget the backport.

PR #1263, opened 2023-09-25
Use the same branch name for ITs on Jenkins

slawekjaranowski:
We have the same logic on GitHub actions.

When branch with the same name as in repository exist in integration test we should use it.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -84,14 +84,15 @@ for (String os in runITsOses) {
                     // will not trample each other plus workaround for JENKINS-52657
                     dir(isUnix() ? 'test' : "c:\\mvn-it-${EXECUTOR_NUMBER}.tmp") {
                         def WORK_DIR=pwd()
+                        def ITS_BRANCH = env.CHANGE_BRANCH != null ? env.CHANGE_BRANCH :  env.BRANCH_NAME;
                         try {
-                          echo "Checkout ITs from branch: ${env.CHANGE_BRANCH}"
+                          echo "Checkout ITs from branch: ${ITS_BRANCH}"
                           checkout([$class: 'GitSCM',
-                                  branches: [[name: env.CHANGE_BRANCH]],
+                                  branches: [[name: ITS_BRANCH]],
                                   extensions: [[$class: 'CloneOption', depth: 1, noTags: true, shallow: true]],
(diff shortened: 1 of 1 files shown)


cstamas (approved):

slachiewicz (approved):

Jenkinsfile
```diff
@@ -84,10 +84,21 @@ for (String os in runITsOses) {
...
-                                branches: [[name: "*/master"]],
-                                extensions: [[$class: 'CloneOption', depth: 1, noTags: true, shallow: true]],
-                                userRemoteConfigs: [[url: 'https://github.com/apache/maven-integration-testing.git']]])                        
+                        def ITS_BRANCH = env.CHANGE_BRANCH != null ? env.CHANGE_BRANCH :  env.BRANCH_NAME;
+                        try {
+                          echo "Checkout ITs from branch: ${ITS_BRANCH}"
+                          checkout([$class: 'GitSCM',
+                                  branches: [[name: ITS_BRANCH]],
+                                  extensions: [[$class: 'CloneOption', depth: 1, noTags: true, shallow: true]],
+                                  userRemoteConfigs: [[url: 'https://github.com/apache/maven-integration-testing.git']]])
```
gnodet, line 93: We're still using the `maven-integration-testing` GitHub repo, but it needs to point to the same repository from the PR.

gnodet (changes requested):
The repository needs to be changed to point to the repository that hosts the PR branch.

Jenkinsfile
olamy: Didn't we say earlier to use only master for this build?
Or I'm lost?

gnodet: Well, then, what's the point of this PR ? I'm lost too...

slawekjaranowski: Jobs on jenkins are not executed for PR from forked repository.

The goal is minimize a failed build PR on jenkins, when we have change in Maven and connected in ITs.

slawekjaranowski: We also have a https://github.com/apache/maven-integration-testing/tree/maven-3.8.x branch in ITs - so this branch should be used for testing maven-3.8.x and so on

slawekjaranowski: @gnodet, @olamy I still think that change is  good enough

build for #1273 pass on GH but fail on Jenkins

olamy (approved):

PR #1265, opened 2023-09-26
[MNG-7899] Various memory usage improvements 1

sebastien-doyon:
Remove unneeded StringBuilder to reduce temporary objects creation


cstamas:
maven-compat classes are really used by Maven2 plugins/extensions and are all deprecated. IMHO, they should be just left as is...

sebastien-doyon:
Thanks @cstamas, I will close this pr

jira-importer:
Resolve #9016

PR #1269, opened 2023-09-26
[MNG-7899] Various memory usage improvements 4

sebastien-doyon:
Merging the getStatus() method to optimize :

- Use the main StringBuilder to append string instead of using a
separate one
- Use the StringBuilder.append() with index to avoid String.substring(),
less temporary strings
- Reuse the FileSizeFormat object in the while loop avoiding multiple
temporary instances creation


gnodet:
@sebastien-doyon unit tests seem broken.
Also and fwiw, this looks like all this work may be broken if we somehow integrate https://github.com/apache/maven/pull/1279... we may want to investigate on that branch then...

gnodet:
> @sebastien-doyon unit tests seem broken. Also and fwiw, this looks like all this work may be broken if we somehow integrate #1279... we may want to investigate on that branch then...

Also, #1268, #1269 and #1270 really look related to the same goal, I.e. optimize the logging, so I think they should be merged in order to have a better understand of the benefits (even if they could be 3 different commits).

sebastien-doyon:
> @sebastien-doyon unit tests seem broken.

@gnodet yes, I am trying to find the problem, but without access to the surefire report files, it is hard to find the problem. Do you have access or know how to get access to the target directory of a build in github? That would help debug.

> Also and fwiw, this looks like all this work may be broken if we somehow integrate https://github.com/apache/maven/pull/1279... we may want to investigate on that branch then...

I think the PR [1279](https://github.com/apache/maven/pull/1279) is complementary with this PR. The only change in the 1279 PR that would be needed is to add back the `MessageBuilder builder(StringBuilder stringBuilder)` method to `api/maven-api-core/src/main/java/org/apache/maven/api/services/MessageBuilderFactory.java` and `maven-core/src/main/java/org/apache/maven/internal/impl/DefaultMessageBuilderFactory.java` classes.

> Also, https://github.com/apache/maven/pull/1268, https://github.com/apache/maven/pull/1269 and https://github.com/apache/maven/pull/1270 really look related to the same goal, I.e. optimize the logging, so I think they should be merged in order to have a better understand of the benefits (even if they could be 3 different commits).

If you think, I was trying to separate unrelated changes so the obvious ones could be merged while those with problems or comments could be worked out separately, like this one. Tell me if you really prefer me to merge in on PR.

sebastien-doyon:
@gnodet the last commit fixes the unit test that was broken.

maven-embedder/src/main/java/org/apache/maven/cli/transfer/ConsoleMavenTransferListener.java
```diff
@@ -135,7 +136,7 @@ public synchronized void transferFailed(TransferEvent event) {
 
     private void overridePreviousTransfer(TransferEvent event) {
         if (lastLength > 0) {
-            StringBuilder buffer = new StringBuilder(128);
+            StringBuilder buffer = new StringBuilder(lastLength + 1);
```
gnodet, line 139: This class is completely (and overly btw) synchronised, so why not using a single `StringBuilder buffer` field and use `setLength(0)` when needed.

gnodet (changes requested):

maven-embedder/src/main/java/org/apache/maven/cli/transfer/ConsoleMavenTransferListener.java
sebastien-doyon: Yes this is true, I also noticed that the synchronized block at line 71 is not needed since the transferProgressed() method is synchronized already and synchronising on an instance var have the same effect.

gnodet: And also, the `transfers` field does not need to be wrapped in a `Collections.synchronizedMap` collection.

sebastien-doyon: Good catch! I will remove the `Collections.synchronizedMap`

maven-embedder/src/main/java/org/apache/maven/cli/transfer/ConsoleMavenTransferListener.java
```diff
@@ -64,20 +65,36 @@ public synchronized void transferProgressed(TransferEvent event) throws Transfer
...
+
+                if (idx < 0) {
+                    buffer.append(resourceName);
+                } else {
+                    buffer.append(resourceName, idx + 1, resourceName.length());
                 }
+                buffer.append(" (");
+            }
+
+            buffer.append(format.formatProgress(complete, total));
```
gnodet, line 90: I wonder if the `FileSizeFormat` should be refactored to be given a `StringBuffer` to write into, instead of (or in addition to) the current methods.
But this may be for another PR...

gnodet (approved):
LGTM
There are still a couple of things to clean in this package....
 * FileSizeFormat should be moved to top level
 * it should be refactored to accept an `Appender` as argument
 * `BatchModeMavenTransferListener` should be removed
 * `Slf4jMavenTransferListener` should inherit `AbstractMavenTransferListener`
 * `AbstractMavenTransferListener` uses direct ansi sequences (this will be fixed in #1279)

@sebastien-doyon feel free to work on these issues if you're fancy

maven-embedder/src/main/java/org/apache/maven/cli/transfer/ConsoleMavenTransferListener.java
sebastien-doyon: Done in the last commit

sebastien-doyon:
@gnodet 

> There are still a couple of things to clean in this package....

> `FileSizeFormat` should be moved to top level

Done in the last commit

> it should be refactored to accept an `Appender` as argument

Done in the last commit, the `FileSizeFormat.formatProcess()` was refactored to accept an `StringBuilder` as argument

> `BatchModeMavenTransferListener` should be removed

Done in the last commit

> `Slf4jMavenTransferListener` should inherit `AbstractMavenTransferListener`

I leave this one for another PR, I suggest transform the `AbstractMavenTransferListener` to a message formatter class that would contain the common code building the text, and move to `PrintStream` code to the `ConsoleMavenTransferListener`. Maybe in the 1279 PR or a subsequent one.

PR #1270, opened 2023-09-26
[MNG-7899] Various memory usage improvements 5

sebastien-doyon:
https://issues.apache.org/jira/browse/MNG-7899
Multiple optimizations :

- renderLevel() method use static constants instead of rebuilding the
strings on each call
- replace + operator usage with more PrintStream.print() calls to reduce
temporary strings creation
- reduce usage of MessageBuilder.a() method usage with more
PrintStream.print() calls to reduce temporary strings creation
- replace the builder() method with a static import

 - [X ] In any other case, please file an [Apache Individual Contributor License Agreement](https://www.apache.org/licenses/icla.pdf).

### DIFF
--- a/api/maven-api-core/src/main/java/org/apache/maven/api/services/MessageBuilder.java
+++ b/api/maven-api-core/src/main/java/org/apache/maven/api/services/MessageBuilder.java
@@ -206,4 +206,11 @@ public interface MessageBuilder {
      */
     @Nonnull
     String build();
+
+    /**
+     * Set the buffer length.
+     *
+     * @param length the new length
+     */
+    void setLength(int length);
 }
--- a/maven-embedder/src/main/java/org/apache/maven/cli/MavenCli.java
+++ b/maven-embedder/src/main/java/org/apache/maven/cli/MavenCli.java
@@ -497,19 +497,6 @@ public class MavenCli {
         cliRequest.quiet = !cliRequest.verbose && commandLine.hasOption(CLIManager.QUIET);
         cliRequest.showErrors = cliRequest.verbose || commandLine.hasOption(CLIManager.ERRORS);
 
-        slf4jLoggerFactory = LoggerFactory.getILoggerFactory();
-        Slf4jConfiguration slf4jConfiguration = Slf4jConfigurationFactory.getConfiguration(slf4jLoggerFactory);
-
-        if (cliRequest.verbose) {
-            cliRequest.request.setLoggingLevel(MavenExecutionRequest.LOGGING_LEVEL_DEBUG);
-            slf4jConfiguration.setRootLoggerLevel(Slf4jConfiguration.Level.DEBUG);
-        } else if (cliRequest.quiet) {
-            cliRequest.request.setLoggingLevel(MavenExecutionRequest.LOGGING_LEVEL_ERROR);
-            slf4jConfiguration.setRootLoggerLevel(Slf4jConfiguration.Level.ERROR);
-        }
-        // else fall back to default log level specified in conf
-        // see https://issues.apache.org/jira/browse/MNG-2570
-
         // LOG COLOR
         String styleColor = cliRequest.getUserProperties().getProperty(STYLE_COLOR_PROPERTY, "auto");
         styleColor = commandLine.getOptionValue(COLOR, styleColor);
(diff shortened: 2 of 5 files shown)


maven-embedder/src/main/java/org/apache/maven/cli/MavenCli.java
```diff
@@ -495,19 +495,6 @@ void logging(CliRequest cliRequest) {
         cliRequest.quiet = !cliRequest.verbose && commandLine.hasOption(CLIManager.QUIET);
         cliRequest.showErrors = cliRequest.verbose || commandLine.hasOption(CLIManager.ERRORS);
 
-        slf4jLoggerFactory = LoggerFactory.getILoggerFactory();
```
gnodet, line 498: Moving this code block is necessary because Jansi needs to be initialised before the execution of this block when `MessageUtils.setColorEnabled` is called ?

sebastien-doyon: Well, the MessageUtils.setColorEnabled must be called before calling slf4jConfiguration.setRootLoggerLevel witch load for the first time the MavenSimpleLogger class. Since this modified class use now static constants, its is necessary that the colour support is set right before loading the MavenSimpleLogger class.

maven-slf4j-provider/src/main/java/org/slf4j/impl/MavenSimpleLogger.java
```diff
@@ -71,8 +78,13 @@ private void printStackTrace(Throwable t, PrintStream stream, String prefix) {
...
             stream.print(builder().strong("at"));
-            stream.print(" " + e.getClassName() + "." + e.getMethodName());
-            stream.print(builder().a(" (").strong(getLocation(e)).a(")"));
+            stream.print(" ");
+            stream.print(e.getClassName());
+            stream.print(".");
+            stream.print(e.getMethodName());
+            stream.print(" (");
+            stream.print(builder().strong(getLocation(e)));
+            stream.print(")");
```
gnodet, line 87: Are you sure this is interesting in terms of performances ? The benefit to avoid multiple buffer allocations may be balanced by the synchronisation / flushes on the underlying character/byte streams...

gnodet (changes requested):

maven-slf4j-provider/src/main/java/org/slf4j/impl/MavenSimpleLogger.java
sebastien-doyon: Yes that is right, I noticed this while doing different tests but forgot to push the changes. Pushing the commit now.

maven-slf4j-provider/src/main/java/org/slf4j/impl/MavenSimpleLogger.java
```diff
@@ -30,6 +29,14 @@
  * @since 3.5.0
  */
 public class MavenSimpleLogger extends SimpleLogger {
+
+    private static final String TRACE_RENDERED_LEVEL = builder().trace("TRACE").build();
+    private static final String DEBUG_RENDERED_LEVEL = builder().debug("DEBUG").build();
+    private static final String INFO_RENDERED_LEVEL = builder().info("INFO").build();
+    private static final String WARN_RENDERED_LEVEL =
+            builder().warning("WARNING").build();
+    private static final String ERROR_RENDERED_LEVEL = builder().error("ERROR").build();
```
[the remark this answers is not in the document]
sebastien-doyon: I see your point, I will try to find and propose a better solution a better solution. I would like to avoid recreating those strings on each `MavenSimpleLogger` instances. I know that the usual use case is that there is only on instance of `MavenSimpleLogger`, the root logger, but it is also possible to configure loggers per component or class with more efforts.

api/maven-api-core/src/main/java/org/apache/maven/api/services/MessageBuilder.java
```diff
@@ -206,4 +206,11 @@ public interface MessageBuilder {
      */
     @Nonnull
     String build();
+
+    /**
+     * Set the buffer length.
+     *
+     * @param length the new length
+     */
+    void setLength(int length);
```
gnodet, line 215: Shouldn't this be reused in #1269 somehow ?  I mean that PR mixes ansi formatting with string builders...  No big deal for now though.

maven-slf4j-provider/src/main/java/org/slf4j/impl/MavenSimpleLogger.java
```diff
@@ -103,13 +117,9 @@ protected String getLocation(final StackTraceElement e) {
         } else if (e.getFileName() == null) {
             return "Unknown Source";
         } else if (e.getLineNumber() >= 0) {
-            return String.format("%s:%s", e.getFileName(), e.getLineNumber());
+            return e.getFileName() + ":" + e.getLineNumber();
```
gnodet, line 120: Shouldn't we go further and pass the message builder as an argument to avoid concatenating strings here ?

gnodet (changes requested):
Maybe a last spot for enhancement in `getLocation` ?

api/maven-api-core/src/main/java/org/apache/maven/api/services/MessageBuilder.java
sebastien-doyon: For me it is not related, the usage of the `MessageUtils` class is not the same

maven-slf4j-provider/src/main/java/org/slf4j/impl/MavenSimpleLogger.java
sebastien-doyon: I think it is not worth the trouble, in the case where the getLocation method would concatenate the file name and the lin e number, we would need to surround `e.getFileName() + ":" + e.getLineNumber()` with the strong style, separately  or with a style set and reset. I see no support for that in the current implementation (style set/reset).

sebastien-doyon: feel free to correct me if I am wrong....

gnodet (approved):
LGTM

PR #1276, opened 2023-10-12
Reuse the same repository so that ITs can access maven snapshot jars from the repository

gnodet:
Reuse the same local repository so that IT depending on maven snapshot jars can work.
This may need to be reverted once maven is released and the ITs are upgraded to the released version instead of having to use the latest snapshots.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -112,7 +112,7 @@ for (String os in runITsOses) {
                             withEnv(["JAVA_HOME=${ tool "$jdkName" }",
                                         "PATH+MAVEN=${ tool "$jdkName" }/bin:${tool "$mvnName"}/bin",
                                         "MAVEN_OPTS=-Xms2g -Xmx4g -Djava.awt.headless=true"]) {                                               
-                                String cmd = "${runITscommand} -Dmaven.repo.local=$WORK_DIR/it-local-repo -DmavenDistro=$WORK_DIR/dists/apache-maven-bin.zip -Dmaven.test.failure.ignore"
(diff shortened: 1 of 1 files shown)


Jenkinsfile
```diff
@@ -112,7 +112,7 @@ for (String os in runITsOses) {
                             withEnv(["JAVA_HOME=${ tool "$jdkName" }",
                                         "PATH+MAVEN=${ tool "$jdkName" }/bin:${tool "$mvnName"}/bin",
                                         "MAVEN_OPTS=-Xms2g -Xmx4g -Djava.awt.headless=true"]) {                                               
-                                String cmd = "${runITscommand} -Dmaven.repo.local=$WORK_DIR/it-local-repo -DmavenDistro=$WORK_DIR/dists/apache-maven-bin.zip -Dmaven.test.failure.ignore"
+                                String cmd = "${runITscommand} -Dmaven.repo.local=$WORK_DIR/.repository -DmavenDistro=$WORK_DIR/dists/apache-maven-bin.zip -Dmaven.test.failure.ignore"
```
slawekjaranowski, line 115: I'm afraid that it will not working
- step Build / Unit Test  can be done on other node 
- step Build / Unit Test use verify goal
- each Run ITs step can be done on separate node, workspace

Please create branch in Maven repo - PR from fork not trigger build on jenkins

slawekjaranowski (changes requested):

gnodet:
Recreating PR with a branch on the main repo

PR #1277, opened 2023-10-13
Reuse the same repository so that ITs can access maven snapshot jars from the repository

gnodet:
Supersedes #1276

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -58,9 +58,6 @@ node(jenkinsEnv.nodeSelection(osNode)) {
             } finally {
                 junit testResults: '**/target/surefire-reports/*.xml,**/target/failsafe-reports/*.xml', allowEmptyResults: true
             }    
-            dir ('apache-maven/target') {
-                stash includes: 'apache-maven-bin.zip', name: 'maven-dist'
-            }
         }
     }
 }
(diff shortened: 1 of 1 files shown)


Jenkinsfile
```diff
@@ -23,7 +23,7 @@ def buildOs = 'linux'
 def buildJdk = '11'
 def buildMvn = '3.8.x'
 def runITsOses = ['linux']
-def runITsJdks = ['11', '17', '21']
+def runITsJdks = ['jdk_11_latest', 'jdk_17_latest', 'jdk_21_latest']
```
slawekjaranowski, line 26: We should update in https://github.com/apache/maven-jenkins-env/blob/master/vars/jenkinsEnv.groovy `jdkFromVersion`

gnodet: Ah, nice, I actually had no idea where those values were coming from :-)

gnodet:
@slawekjaranowski it seems to be working now :-)

gnodet:
Raised https://github.com/apache/maven-integration-testing/pull/314

Note that the IT will probably break after starting the Maven release process, as the version will change.  After the released is published, the IT should use the released version of Maven and the whole problem will be over.

slawekjaranowski:
source branch deleted

PR #1278, opened 2023-10-13
[MNG-7911] Require JDK 11 and support JDK 21 at build time

gnodet:
IT PR at https://github.com/apache/maven-integration-testing/pull/310

### DIFF
--- a/.github/workflows/maven.yml
+++ b/.github/workflows/maven.yml
@@ -42,7 +42,7 @@ jobs:
 
       - uses: actions/setup-java@v3
         with:
-          java-version: 8
+          java-version: 11
           distribution: 'temurin'
           cache: 'maven'
 
@@ -61,7 +61,7 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest, windows-latest, macOS-latest]
-        java: [8, 11, 17]
+        java: [11, 17, 21]
 
       fail-fast: false
     runs-on: ${{ matrix.os }}
--- a/maven-model-builder/src/test/java/org/apache/maven/model/building/BuildModelSourceTransformerTest.java
+++ b/maven-model-builder/src/test/java/org/apache/maven/model/building/BuildModelSourceTransformerTest.java
@@ -24,6 +24,7 @@ import java.beans.PropertyDescriptor;
 import java.nio.file.Path;
 import java.nio.file.Paths;
 import java.util.Collections;
+import java.util.List;
 import java.util.Objects;
 
 import org.apache.maven.model.Model;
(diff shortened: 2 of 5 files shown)


.github/workflows/maven.yml
```diff
@@ -61,7 +61,7 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest, windows-latest, macOS-latest]
-        java: [8, 11, 17]
+        java: [11, 17, 21]
```
mthmulders, line 64: There is no way to build on 8 **and** 21?

gnodet: I must admit I haven't checked all the intermediate mockito versions, but the one currently in use does not work with JDK 21 and the most recent one is built with JDK 11 byte code.

mthmulders: I can imagine the Mockito team 'made the jump' and dropped Java 8 altogether. This change would not _prevent_ running Maven on Java 8, would it?

gnodet: Definitely not, it's only the build time requirements that are raised, but the target byte code is still JDK 8.

mthmulders: Thanks for clarifying!

.github/workflows/maven.yml
```diff
@@ -42,7 +42,7 @@ jobs:
 
       - uses: actions/setup-java@v3
         with:
-          java-version: 8
+          java-version: 11
```
mthmulders, line 45: Should we also update the Jenkinsfile?

maven-model-builder/src/test/java/org/apache/maven/model/building/BuildModelSourceTransformerTest.java
```diff
@@ -136,6 +137,10 @@ public static boolean equalsDeep(Object m1, Object m2) {
             }
             BeanInfo bean = Introspector.getBeanInfo(m1.getClass());
             for (PropertyDescriptor prop : bean.getPropertyDescriptors()) {
+                if (("first".equals(prop.getName()) || "last".equals(prop.getName()))
```
mthmulders, line 140: How is this related to building on JDK 21?

pom.xml
```diff
@@ -642,6 +679,26 @@ under the License.
               <fail>true</fail>
             </configuration>
           </execution>
+          <execution>
+            <id>enforce-bytecode-version</id>
```
mthmulders, line 683:
```suggestion
            <id>enforce-build-jdk-version</id>
```

maven-model-builder/src/test/java/org/apache/maven/model/building/BuildModelSourceTransformerTest.java
rmannibucau: New sequence api i guess

gnodet: Exactly, see https://docs.oracle.com/en/java/javase/21/docs/api/java.base/java/util/SequencedCollection.html

Those getters are throwing `NoSuchElementException` when the list is empty.

pom.xml
gnodet: This execution overrides the one defined in the parent pom, with the difference that it adds the `ignoredScopes` configuration bit.  So it cannot be renamed...

mthmulders: Ah, wasn't aware of that. Makes sense then!

mthmulders (approved):

slawekjaranowski (approved):

.github/workflows/maven.yml
gnodet: I update the Jenkins file to build with JDK 11, 17 and 21.

mthmulders (approved):

jira-importer:
Resolve #9002

PR #1279, opened 2023-10-13
[MNG-7995] Switch to JLine to provide line editing

gnodet:
JIRA issue: https://issues.apache.org/jira/browse/MNG-7995
JLine used for:
  * output display
  * prompting

It should also be used in the future for better transfer display, i.e. rewriting `ConsoleMavenTransferListener`, still TODO, see #1302  as a possible base.


slawekjaranowski:
Will it be compatible with Maven 3.x and old maven-shared-utils?

slawekjaranowski:
eg.
```
mvn buildplan:list-phase
```

gnodet:
> eg.
> 
> ```
> mvn buildplan:list-phase
> ```

It works when using a terminal.  However, when the output is redirected to a file, the ansi sequences are not stripped correctly when using the `org.apache.maven.shared.logging.MessageUtils`.
The reason is that the ansi sequences are currently stripped inside the `System.out` and `System.err` streams.
So I'll postpone this change until https://github.com/jline/jline3/pull/871 is done and released.

gnodet:
> > eg.
> > ```
> > mvn buildplan:list-phase
> > ```
> 
> It works when using a terminal. However, when the output is redirected to a file, the ansi sequences are not stripped correctly when using the `org.apache.maven.shared.logging.MessageUtils`. The reason is that the ansi sequences are currently stripped inside the `System.out` and `System.err` streams. So I'll postpone this change until [jline/jline3#871](https://github.com/jline/jline3/pull/871) is done and released.

I've upgrade to latest JLine, which now includes Jansi.  The integration with maven-shared-utils works correctly now.

gnodet:
@slawekjaranowski would you mind having a look ?

cstamas (approved):

pom.xml
```diff
@@ -690,6 +696,9 @@ under the License.
                   <ignoredScopes>
                     <ignoredScope>test</ignoredScope>
                   </ignoredScopes>
+                  <excludes>
+                    <exclude>org.jline:jline</exclude>
+                  </excludes>
```
slawekjaranowski, line 701: Why? Which JDK is required?
And what when user have an older JDK?

gnodet: The reason is https://github.com/jline/jline3/issues/924

PR #1283, opened 2023-10-15
[MNG-7910] DefaultProfileSelector empty ctor uses immutable list

cstamas:
It uses immutable list and while it suggests it allows this below, it actually fails on constructs like this (worse even, it compiles ok, but fails runtime only): 
```
new DefaultProfileSelector()
                .addProfileActivator(new JdkVersionProfileActivator())
                .addProfileActivator(new PropertyProfileActivator())
```

As default ctor creates immutable list, but API kinda suggests this is ok. So either this PR should go in, or just remove `addProfileActivator` and make class immutable. Am fine with both (unless something else requires it to be extended).

---

https://issues.apache.org/jira/browse/MNG-7910


gnodet (approved):

jira-importer:
Resolve #9004

PR #1284, opened 2023-10-16
[MNG-7804] Fix broken MavenModelVersion generation

gnodet:
This aims at fixing master UT which have been broken (even though each individual PR was passing ok).


jira-importer:
Resolve #9186

PR #1126, opened 2023-05-24
[MNG-7792] Use a standalone asm version

gnodet:
- [MNG-7587] Upgrade to sisu 0.9.0.M2
- Use a non shaded version of asm in guice and sisu
- Fix missing asm for maven-resolver-provider
- IT PR: https://github.com/apache/maven-integration-testing/pull/290

I've extracted the use of a standalone ASM version as JDK 17 support is already covered by the upgrade to 0.9.0.M2.
To support JDK >= 19, we need ASM 9.5 and we thus need to use a standalone version of ASM and upgrade it.

This is a follow-up to https://github.com/apache/maven/pull/1119

PR #1286, opened 2023-10-17
[3.9.x] [MNG-7913] Upgrade Sisu version

cstamas:
To version that supports gleaning JSR330 component classes up to Java 19. The reasoning of this change is only to allow 3rd party Maven plugins/Maven extensions/other JSR330 components (managed by Maven embedded Sisu) to use bytecode newer that Java 14 (Sisu 0.3.5).

---

https://issues.apache.org/jira/browse/MNG-7913

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -127,7 +127,7 @@ under the License.
     <javaVersion>8</javaVersion>
     <minimalMavenBuildVersion>3.5.4</minimalMavenBuildVersion>
 
-    <sisuVersion>0.3.5</sisuVersion>
+    <sisuVersion>0.9.0.M2</sisuVersion>
     <classWorldsVersion>2.7.0</classWorldsVersion>
     <commonsCliVersion>1.5.0</commonsCliVersion>
     <commonsIoVersion>2.11.0</commonsIoVersion>


michael-o (approved):

gnodet:
> To version that supports gleaning JSR330 component classes up to Java 21. The reasoning of this change is only to allow 3rd party Maven plugins/Maven extensions/other JSR330 components (managed by Maven embedded Sisu) to use bytecode newer that Java 14 (Sisu 0.3.5).
> 
> https://issues.apache.org/jira/browse/MNG-7913

Fwiw, I think you'll also need to back port https://github.com/apache/maven/pull/1126 in order to make sure the latest ASM supporting JDK 21 is used.

cstamas:
I agree with @kwin: just keep it simple with Maven 3.9, let it go up to Java 21 but we don't want to "future proof" it, as there is maven4

kwin:
@cstamas just to clarify, M2 only supports up to Java 19 bytecode which contradicts the description of this PR. Please adjust it accordingly.

jira-importer:
Resolve #8732

PR #1295, opened 2023-10-20
[MNG-7916] Simply get rid of LifecycleMappingsXpp3Reader/Writer

gnodet:
Since the v3 model has been replaced by the v4 model, that would be in line.  We can revisit and restore full compatibility later if needed.

JIRA issue: https://issues.apache.org/jira/browse/MNG-7916


cstamas (approved):

jira-importer:
Resolve #8728

PR #1296, opened 2023-10-20
[MNG-7899] Various memory usage improvements 4-1

sebastien-doyon:
- `BatchModeMavenTransferListener` removed
- `FileSizeFormat` moved to top level class
- `FileSizeFormat.formatProcess()` refactored to accept an StringBuilder
as argument
- `FileSizeFormat` refactored `format()` methods

 - [ X] In any other case, please file an [Apache Individual Contributor License Agreement](https://www.apache.org/licenses/icla.pdf).


maven-embedder/src/main/java/org/apache/maven/cli/transfer/FileSizeFormat.java
```diff
@@ -0,0 +1,160 @@
...
+    @SuppressWarnings("checkstyle:magicnumber")
+    private String format(long size, ScaleUnit unit) {
+        if (size < 0L) {
+            throw new IllegalArgumentException("file size cannot be negative: " + size);
+        }
+
+        double scaledSize = (double) size / unit.bytes();
+
+        if (unit == ScaleUnit.BYTE) {
+            return largeFormat.format(size);
```
gnodet, line 129: The goal is to avoid those calls.  
`NumberFormat` has methods that take a `StringBuilder`...
In `NumberFormat`, I have the following declaration:
```
    public final String format(long number) {
        return format(number, new StringBuffer(),
                      DontCareFieldPosition.INSTANCE).toString();
    }
```
so we should go straight to the method that takes a `StringBuffer` to avoid unneeded allocations + toString....

gnodet (changes requested):

maven-embedder/src/main/java/org/apache/maven/cli/transfer/FileSizeFormat.java
sebastien-doyon: Would have been nice, but the NumberFormat.format() method takes only StringBuffer as parameter, the synchronized StringBuilder. They are not interchangeable  and I am not sure we should switch StringBuilder for StringBuffer for that....

gnodet: What about `new Formatter(locale, builder).format("%3.f", size)`  ? i.e. do not use the `NumberFormat`, but `Formatter` ?

sebastien-doyon: Using a Formatter, the pattern string is passed as parameter to the format() method. Doing this the pattern is recompiled each method call and not reused. This generates a lot of garbage. A more optimised solution would be to avoid using generic Format or Formatter classes and implement a custom logic specific to this particular case. Some math logic to round the numbers as needed. This would reduce even more the memory allocation.

sebastien-doyon: @gnodet you can find the refactored `FileSizeFormat` class using Math logic in the last commit

sebastien-doyon:
@gnodet your changes are cleaner. I agree with your contribution, thanks!

sebastien-doyon:
@gnodet I just fixed a format problem that made the build fail.

gnodet (approved):

jira-importer:
Resolve #9016

PR #1297, opened 2023-10-25
Use 4.0.99 as as intermediate model to keep 4.1.0 for the Maven 4.x model

gnodet:
Fix model doc generation, change the intermediate model to use 4.0.99 and align the new model to 4.1.0

This requires https://github.com/apache/maven-integration-testing/pull/315 to pass ITs...


cstamas (approved):

api/maven-api-model/src/main/mdo/maven.mdo
```diff
@@ -387,7 +388,7 @@
       </fields>
       <codeSegments>
         <codeSegment>
-          <version>4.0.0/4.1.0</version>
+          <version>4.0.0/4.0.99</version>
```
kwin, line 391: @gnodet What is this supposed to mean?

gnodet: This is closed ranged version notation.  That's defined by modello...

kwin: Ah, sorry. I get it now. I thought this element was for documentation purposes

cstamas: Sadly, modello "version range" predates (AFAIK, someone may correct me) Maven ranges, and has totally different  syntax...

PR #1299, opened 2023-11-08
[MNG-7924] Better control over and better integration with Resolver

cstamas:
Integrate better and obtain better control over Resolver. These changes did stem from "JPMS module experiment" and are considered improvement but does not implement any functionality related to JPMS module support.

Changes:
* Maven4 should stop "disconnected coexistence" of two type systems (ArtifactHandlers and Resolver ArtifactTypeRegistry), it should unify them.
* Maven4 Core should provide generic and extensible means to introduce new artifact types (fully in extension, and extension should get extended data via "roundtrip" in core/resolver)


maven-core/src/main/java/org/apache/maven/artifact/handler/manager/DefaultArtifactHandlerManager.java
```diff
@@ -22,51 +22,55 @@
...
+                    type.isIncludesDependencies(),
+                    type.getLanguage(),
+                    type.isAddedToClassPath()); // TODO: watch out for module path
+        });
     }
 
     public void addHandlers(Map<String, ArtifactHandler> handlers) {
         // legacy support for maven-gpg-plugin:1.0
-        allHandlers.putAll(handlers);
+        throw new IllegalArgumentException("you cannot do this anymore");
```
gnodet, line 69: `throw new UnsupportedOperationException("Adding handlers is not supported anymore")` ?

maven-core/src/main/java/org/apache/maven/artifact/handler/manager/DefaultArtifactHandlerManager.java
```diff
@@ -22,51 +22,55 @@
...
     }
 
     @Deprecated
     public Set<String> getHandlerTypes() {
-        return artifactHandlers.keySet();
+        return Collections.emptySet();
```
gnodet, line 74: That looks wrong. Should we either delegate to `TypeRegistry` or throw an exception ?

maven-resolver-provider/src/main/java/org/apache/maven/repository/internal/MavenRepositorySystemUtils.java
```diff
@@ -59,8 +62,37 @@ private MavenRepositorySystemUtils() {
      * the session with authentication, mirror, proxy and other information required for your environment.
      *
      * @return The new repository system session, never {@code null}.
+     * @deprecated This method is deprecated.
      */
+    @Deprecated
     public static DefaultRepositorySystemSession newSession() {
+        DefaultArtifactTypeRegistry stereotypes = new DefaultArtifactTypeRegistry();
+        stereotypes.add(new DefaultArtifactType("pom"));
```
gnodet, line 70: Should we use `DefaultTypeRegistry` and the various providers here instead ?

maven-core/src/main/java/org/apache/maven/artifact/handler/manager/LegacyArtifactHandlerManager.java
```diff
@@ -0,0 +1,72 @@
...
+import java.util.concurrent.ConcurrentHashMap;
+
+import org.apache.maven.artifact.handler.ArtifactHandler;
+import org.apache.maven.artifact.handler.DefaultArtifactHandler;
+
+/**
+ */
+@Named
+@Singleton
+public class LegacyArtifactHandlerManager implements ArtifactHandlerManager {
```
gnodet, line 36: Why do we need the `LegacyArtifactHandlerManager` _and_ the `DefaultArtifactHandlerManager` at the same time ?

rmannibucau:
Is it planned to make the properties injected from the pom "mvn 4 version" and therefore makes the type dictionnary useless?
If so +1 (means the toggles will be exposed to users and type can stick to extension and management in the build handled by end user, best of all worlds), else I'm mixed since it does not change much things if it stays internal only.

maven-core/src/main/java/org/apache/maven/artifact/handler/manager/LegacyArtifactHandlerManager.java
cstamas: This is a simple trick:
* legacyAHM is what "default" was so far (it have legacy AHs injected), is new, but nobody uses it as it is new thing (just new code "knows" about it)
* defaultAHM changed to delegate calls to typeRegistry
* typeRegistry uses as a fallback legacyAHM

This completes the circle:
* legacy code will refer to defaultAHM that goes to typeRegistry that will refer legacyAHM (as fallback)
* new code will refer to typeRegistry

This way new and legacy code both "see" the same thing: whether it is new AH (legacy) or Type (new) being added by extension for example

maven-resolver-provider/src/main/java/org/apache/maven/repository/internal/MavenRepositorySystemUtils.java
cstamas: This method is deprecated and meant NOT to be used in Maven, only outside (like MIMA). Hence, there is no Maven API present either, "only" resolver stuff, so this is good as is IMHO.

api/maven-api-core/src/main/java/org/apache/maven/api/Type.java
```diff
@@ -51,14 +51,21 @@ public interface Type {
...
+     *
+     * @return the language of this type, never {@code null}.
+     */
+    String getLanguage();
+
     /**
      * Get the file extension of artifacts of this type.
      *
-     * @return the file extension
+     * @return the file extension, never {@code null}.
```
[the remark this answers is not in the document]
cstamas: nope, they are never null for types. In maven, when ext=null, then it implies ext=typeId

api/maven-api-core/src/main/java/org/apache/maven/api/Type.java
```diff
@@ -51,14 +51,21 @@ public interface Type {
...
+     * @return the id of this type, never {@code null}.
      */
     String getId();
 
+    /**
+     * Returns the dependency type language.
+     *
+     * @return the language of this type, never {@code null}.
+     */
+    String getLanguage();
```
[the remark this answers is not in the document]
cstamas: I may remove this, as it is used ONLY in conversion of Type to AH, but am unsure is it used anywhere at all...
We can always return it using DependencyProperties where Map<String, String> can not only hold "flags" (string values of "true" and "false") but something else as well....

api/maven-api-core/src/main/java/org/apache/maven/api/Type.java
```diff
@@ -51,14 +51,21 @@ public interface Type {
      * Returns the dependency type id.
      * The id uniquely identifies this <i>dependency type</i>.
      *
-     * @return the id of this type
+     * @return the id of this type, never {@code null}.
      */
     String getId();
 
+    /**
+     * Returns the dependency type language.
```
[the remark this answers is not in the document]
cstamas: I think it is more like "universe", as for example the "war" case: means WebArchive in J2EE that is "Java universe". Unsure what the intent of this AH property was originally. Also, POMs are language="none" (so basically "any", as POM can be used even to produce native code as well, or think polyglot, so is not "language bound").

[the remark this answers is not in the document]
cstamas: https://github.com/apache/maven-resolver/blob/master/maven-resolver-api/src/main/java/org/eclipse/aether/artifact/ArtifactProperties.java#L35-L38

Will try to remove this (but Resolver ArtifactProperties could be still used to set/get this property)

api/maven-api-core/src/main/java/org/apache/maven/api/Type.java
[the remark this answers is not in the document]
cstamas: Removed language, keep type simplest as possible.

maven-resolver-provider/src/main/java/org/apache/maven/repository/internal/MavenRepositorySystemUtils.java
cstamas: Deprecation is probably wrong.... is "deprecated" from Maven POV as it wants to inject own TypeRegistry, but again, code integrating Maven Resolver (like MIMA) does not have new Maven API on classpath, and it will want to stick to "pure" resolver classes...

cstamas:
> Is it planned to make the properties injected from the pom "mvn 4 version" and therefore makes the type dictionnary useless? If so +1 (means the toggles will be exposed to users and type can stick to extension and management in the build handled by end user, best of all worlds), else I'm mixed since it does not change much things if it stays internal only.

A bit fuzzy thought but let me try:
* "Is it planned to make the properties injected from the pom "mvn 4 version" NO
* "therefore makes the type dictionnary useless?" NO
* "means the toggles will be exposed to users" YES
* "type can stick to extension and management in the build handled by end user" um, is like that even today
* "if it stays internal only" um, Type is part of public Maven API, so it cannot be internal only per definitionem

rmannibucau:
> um, is like that even today

Exactly, and I hope the refactoring enables to stop doing that, it is very misleading for end users and technically no more needed with your PR.

cstamas:
Type is part of Maven DNA, and let me try to explain why:

When you are on "resolver level" (so no Maven in vicinity, like MIMA or Resolver "demos" runs), you are dealing with "artifacts", that are really just "named files" or "files with coordinates", and coordinates are needed to lay these files out in some "space" (layout). There is NO whatsoever assumption about "what those files are". At this level resolver merely cares about getting you these files and publishing these files for you.

When you hop abstraction level higher, at Maven level, you do not deal (never!) with "artifacts" anymore, but you attribute the usage (or role) to artifact, so in Maven you deal with:
* artifact that is the project
* artifact that is the parent POM
* artifact that is maven-plugin
* artifact that is a dependency
* etc

(so, "artifact" is superset of ALL things Maven uses, but in Maven universe you KNOW and TELL (implicitly or explicitly) what you want that artifact for).

And while you have "implied types" (parent POM, maven-plugin, extension, etc are "implied types"), for dependencies you need something "extra" to express "what it is" (or "what I want with it"), and the type serves that purpose.

We could add "flags" to dependencies, but that would make Maven very very "low level", and would trigger a huge cognitive load on end users IMHO. Rather move "baby steps" with improvements, and we will see where it gets us.

rmannibucau:
I fully understand that and this is exactly why I'd like we stop doing that.

> We could add "flags" to dependencies, but that would make Maven very very "low level", and would trigger a huge cognitive load on end users IMHO. Rather move "baby steps" with improvements, and we will see where it gets us.

IMHO it is where we are today, almost no user or plugin writer fully understand the implications of a type and it is often misused.
The design of this PR opens the flags which are exactly the proper design and solution to handle the "context" you need and avoid to workaround the lack of flags with an existing marker, aka type today.

See it as it is, would you use `org.superbiz-classpath` in `groupid` to say that `org.superbiz` is in the classpath? Probably not but it is what we do in `type` cause of the lack of alternative.
With your PR we can assume it and move it to flags and I think it is what would serve the best users and us.
The only challenge will be to solve this in pom v4.0.0 (mvn 3) but ultimately since it is build flags it would just be a matter of sticking to type=extension rule so I think we would be good and help our community rather than promoting more a workaround.

rmannibucau:
@cstamas I guess current state is clear but since it is not fully embraced by the community and your PR enables to align both worlds I really think we shouldnt miss this opportunity.
If even sonatype gets it wrong it is likely not very welcomed by the community IMHO.

Side note: never said you were wrong or it was not working, just that there is so much misunderstanding that making it more obvious would be welcomed.
In terms of impl it is mainly your PR + pom adjustments and be it but it makes the user facing interactions (our "ui") way more obvious I think:
* You can to change how an artifact is consumed -> change the related flag.
* You want to resolve a particular extension -> change the extension
vs
* You want the `.jar` to ignore transitive deps? Change its `type` ("Wait, I need to rebuild the dependency?")

cstamas:
Along those lines, yes:

Packaging provides flexibility for producers, to build new (or "old", like packaging "jar" vs "takari-jar") things.

Type provides flexibility for consumers, for example if producer builds a "bundle" (w/ required extension to make sure he is producing valid OSGi bundle), a consumer still may choose to consume it as "jar", or "bundle" (if he is also in OSGi realm), etc...

cstamas:
Just a thought experiment (that takari somewhat tried to deliver): Look at the `takari-maven-component` packaging. It's goal is: producing jar artifact meant to be used within Maven.

This is so expressive... what are you building? "maven-component"s or "jar with sisu index to be used in Maven".

Basically, we could distinguish even "jar with sisu index", and if user depends on that, it could (by some plugin, surefire extension example) trigger Sisu container creation that uses INDEX to pick up components from classpath, and so on and so on... These little information crumbs can serve the purpose to make things even more smarter (without being "smart smart", but simply to behave according things our user describes to us).

Also, they express intent very very clearly.

cstamas:
And a counter example: consider maven-resolver artifacts (!):
* uses plain packaging "jar", but
* the produced JARs have Sisu index (Sisu DI, JSR330)
* the produced JARs have OSGi header (OSGi)
* the produced JARs have "Automatic-Module-Name" header (JPMS)

It is up to "consumer" how he consumes these things, which (orthogonal) way he makes use of resolver:
* on classpath as plain JAR (using RepositorySystemSupplier)
* on classpath with Sisu DI (as in Maven)
* in OSGi container as bundles
* in JPMS app (unsure is it possible, but "metadata" is there)

So type is about "consuming", and consumed thing does NOT need to be rebuilt, to consume it as this or that.

rmannibucau:
@cstamas one reason I really dislike type vs flags is type is not composable without coding (our limitation of being descriptive in our descriptor/pom - which is a good thing by itself but implies that limitation) whereas it has to be at some point in your examples.
Once again, technically we agreed a long time ago but still, users don't understand that, you seems to think they have to and I'd like to avoid them to understand all that. Ultimately it can be a semantic issue (type is not a "common" _type_) but it is a hurting thing in user land and I don't see with your PR which enables to not keep this ambiguity, why we would have to keep using it.
Ultimately your PR even enable to define custom "types" in the pom (predefined flags) so user can use it in build pom and then it is published properly with type=extension and flattened flags if desired.

maven-core/src/main/java/org/apache/maven/internal/impl/DefaultDependencyProperties.java
```diff
@@ -0,0 +1,65 @@
...
+    public DefaultDependencyProperties(Collection<String> flags) {
+        HashMap<String, String> map = new HashMap<>();
+        for (String flag : flags) {
+            map.put(flag, Boolean.TRUE.toString());
+        }
+        this.properties = Collections.unmodifiableMap(map);
+    }
+
+    public DefaultDependencyProperties(@Nonnull Map<String, String> properties) {
+        this.properties = nonNull(properties, "properties can not be null");
```
gnodet, line 51: `nonNull(properties, ...)`- > `Collections.unmodifiableMap(nonNull(properties, ...))`

maven-core/src/main/java/org/apache/maven/internal/impl/DefaultDependencyProperties.java
```diff
@@ -0,0 +1,65 @@
...
+public class DefaultDependencyProperties implements DependencyProperties {
+    private final Map<String, String> properties;
+
+    public DefaultDependencyProperties(String... flags) {
+        this(Arrays.asList(flags));
+    }
+
+    public DefaultDependencyProperties(Collection<String> flags) {
+        HashMap<String, String> map = new HashMap<>();
+        for (String flag : flags) {
```
gnodet, line 44: `nonNull(flags)` or add an `if (flags != null)` ?

maven-core/src/main/java/org/apache/maven/internal/impl/DefaultDependencyProperties.java
```diff
@@ -0,0 +1,65 @@
...
+import static org.apache.maven.internal.impl.Utils.nonNull;
+
+/**
+ * Default implementation of artifact properties.
+ */
+public class DefaultDependencyProperties implements DependencyProperties {
+    private final Map<String, String> properties;
+
+    public DefaultDependencyProperties(String... flags) {
+        this(Arrays.asList(flags));
```
gnodet, line 39: `nonNull(flags)` or add an `if (flags != null)` ?

PR #1309, opened 2023-11-16
[MNG-7947] Plugin API

gnodet:
### Switch to Guice 6.0, use guice to instantiate and inject mojos

Context: my goal for the maven API was to be completely independent of any third party library. I was working this week on migrating some parts of the Plugin API for Maven 4.  Unfortunately, the scopes (`@SessionScoped` and `@MojoExecutionScoped`) need to be annotated by a scope annotation so that they are recognised as such by Guice. So either the guice annotation or the javax.inject one.  Given `javax.inject` will slowly go away, I went straight to `jakarta.inject`, thus upgrading to Guice 6.0 (which supports both `javax.inject` and `jakarta.inject`, while Guice 7.0 only supports `jakarta.inject`).

### Maven 4 api for plugin execution

This introduces an immutable plugin descriptor (and custom lifecycle for plugins too). The Maven 3 `PluginDescriptor` contains runtime information such as the class loader, dependencies, etc...  Those are moved to the `o.a.m.api.Plugin` class which can be accessed from the `o.a.m.api.MojoExecution` along with the `MojoDescriptor` and `PluginDescriptor`.

Leveraging guice, some fields inherited from the plexus land now disappear: `instantiationStrategy` is replaced with scope annotations, `@Component` replaced by `@Inject`.


cstamas:
@mcculls please chime in

elharo (commented):
I'm not familiar with these specific annotations. However I've noticed in some cases only the simple name of the annotation matters, not the package. If that applies here, you could define Maven specific annotations with the right name rather than depending on Guice.

gnodet:
> I'm not familiar with these specific annotations. However I've noticed in some cases only the simple name of the annotation matters, not the package. If that applies here, you could define Maven specific annotations with the right name rather than depending on Guice.

Yeah, that's true for some annotations, unfortunately, guice is very specific about which ones are used, so it does not work here, see https://github.com/google/guice/blob/6.0.0/core/src/com/google/inject/internal/Annotations.java#L350

PR #1318, opened 2023-11-20
[MNG-6036] Add namespace to XmlNode

gnodet:
JIRA issue: https://issues.apache.org/jira/browse/MNG-6036

This provides support for XML namespaces inside the configuration.

### DIFF
--- a/maven-xml-impl/src/main/java/org/apache/maven/internal/xml/XmlNodeBuilder.java
+++ b/maven-xml-impl/src/main/java/org/apache/maven/internal/xml/XmlNodeBuilder.java
@@ -176,10 +176,10 @@ public class XmlNodeBuilder {
     public static XmlNodeImpl build(XMLStreamReader parser, boolean trim, InputLocationBuilderStax locationBuilder)
             throws XMLStreamException {
         boolean spacePreserve = false;
-        String pfx = null;
-        String nsUri = null;
-        String name = null;
-        String value = null;
+        String lPrefix = null;
+        String lNamespaceUri = null;
+        String lName = null;
+        String lValue = null;
         Object location = null;
         Map<String, String> attrs = null;
         List<XmlNode> children = null;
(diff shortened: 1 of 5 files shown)


maven-xml-impl/src/main/java/org/apache/maven/internal/xml/XmlNodeBuilder.java
```diff
@@ -176,6 +176,8 @@ public static XmlNodeImpl build(XMLStreamReader parser, InputLocationBuilderStax
     public static XmlNodeImpl build(XMLStreamReader parser, boolean trim, InputLocationBuilderStax locationBuilder)
             throws XMLStreamException {
         boolean spacePreserve = false;
+        String pfx = null;
```
elharo, line 179: pfx --> prefix

maven-xml-impl/src/main/java/org/apache/maven/internal/xml/XmlNodeBuilder.java
```diff
@@ -176,6 +176,8 @@ public static XmlNodeImpl build(XMLStreamReader parser, InputLocationBuilderStax
...
+        String nsUri = null;
```
elharo, line 180: nsUri --> namespace or namespaceUri

maven-xml-impl/src/main/java/org/apache/maven/internal/xml/XmlNodeWriter.java
```diff
@@ -18,49 +18,146 @@
...
+        try {
+            XMLOutputFactory factory = new com.ctc.wstx.stax.WstxOutputFactory();
+            factory.setProperty(XMLOutputFactory.IS_REPAIRING_NAMESPACES, false);
+            factory.setProperty(com.ctc.wstx.api.WstxOutputProperties.P_USE_DOUBLE_QUOTES_IN_XML_DECL, true);
+            factory.setProperty(com.ctc.wstx.api.WstxOutputProperties.P_ADD_SPACE_AFTER_EMPTY_ELEM, true);
+            XMLStreamWriter serializer = new IndentingXMLStreamWriter(factory.createXMLStreamWriter(writer));
+            write(serializer, dom);
+            serializer.close();
+        } catch (XMLStreamException e) {
+            throw new RuntimeException("Unexpected error when writing XML", e);
```
elharo, line 45: This should not be a runtime exception. XMLStreamException is better

maven-xml-impl/src/main/java/org/apache/maven/internal/xml/XmlNodeWriter.java
```diff
@@ -18,49 +18,146 @@
...
+            xmlWriter.writeCharacters(value);
+        }
+        xmlWriter.writeEndElement();
+    }
+
+    static class IndentingXMLStreamWriter extends StreamWriterDelegate {
+
+        int depth = 0;
+        boolean hasChildren = false;
+        boolean anew = true;
```
elharo, line 68: do you really need this? If so, it might need to be set to true when entering writeStartDocument

maven-xml-impl/src/main/java/org/apache/maven/internal/xml/XmlNodeWriter.java
gnodet: No problem, that method is used in `XmlNodeImpl#toString()`, so that will just move the catch block elsewhere...

maven-xml-impl/src/main/java/org/apache/maven/internal/xml/XmlNodeWriter.java
gnodet: The idea is only to be able to not ring a `\n` at the beginning when printing a snippet and not a full document.  And that's for the `XmlNodeImpl#toString()` call, once again.  Without this flag and check, we end up with a leading `\n` at the beginning: `\n<configuration></configuration>` ...

maven-xml-impl/src/main/java/org/apache/maven/internal/xml/XmlNodeWriter.java
elharo: It's a public method. It's not safe to assume it's only used by toString.
Also toString should not throw any exceptions, runtime or otherwise. If an exception can be thrown, it should be handled in a way that still returns a string.

maven-xml-impl/src/main/java/org/apache/maven/internal/xml/XmlNodeBuilder.java
gnodet: Renamed variables.

maven-xml-impl/src/main/java/org/apache/maven/internal/xml/XmlNodeBuilder.java
gnodet: Renamed variables.

CrazyHZM (approved):

cstamas (approved):

PR #1323, opened 2023-11-22
[MNG-7945] Fix profile settings being injected into consumer POM

gnodet:
JIRA issue: https://issues.apache.org/jira/browse/MNG-7945

At the same time, another bug is fixed.  When using the build/raw transformation to infer intra-reactor dependencies, the build fail if not run from the root (because those dependencies are not available).  The same problem happens when rebuilding the consumer POM.  
When this happen, a quick read of the whole reactor models is done (but only file models, not effective ones) to gather all GAV of modules being built.


maven-core/src/main/java/org/apache/maven/internal/transformation/DefaultConsumerPomBuilder.java
```diff
@@ -0,0 +1,236 @@
...
+                lookup(RepositorySystem.class),
+                lookup(RemoteRepositoryManager.class),
+                project.getRemoteProjectRepositories(),
+                ProjectBuildingRequest.RepositoryMerging.POM_DOMINANT,
+                null));
+        request.setTransformerContextBuilder(modelBuilder.newTransformerContextBuilder());
+        Properties props = new Properties();
+        props.putAll(session.getSystemProperties());
+        request.setSystemProperties(props);
+        props = new Properties();
```
elharo, line 165: rename to userProperties and systemProperties; reusing local variables is bug prone

maven-core/src/test/java/org/apache/maven/internal/transformation/ConsumerPomArtifactTransformerTest.java
```diff
@@ -60,9 +56,15 @@ void transform() throws Exception {
         try (InputStream expected = Files.newInputStream(beforePomFile)) {
             Model model = new Model(new MavenStaxReader().read(expected));
             MavenProject project = new MavenProject(model);
-            ConsumerPomArtifactTransformer t = new ConsumerPomArtifactTransformer(modelBuilder);
-            t.createConsumerPomArtifact(project, tempFile, systemSessionMock)
-                    .transform(beforePomFile, tempFile, model.getDelegate());
+            ConsumerPomArtifactTransformer t = new ConsumerPomArtifactTransformer((s, p, f) -> {
+                try (InputStream is = Files.newInputStream(f)) {
+                    return DefaultConsumerPomBuilder.transform(new MavenStaxReader().read(is));
+                } catch (Exception e) {
```
elharo, line 62: not needed, the exception fails the test

gnodet: If you mean "you should not catch the exception and let it bubble through and let the exception make the test fail", it unfortunately is not doable because of the usage of checked exceptions.  Adding an exception to the interface would go up the chain to https://github.com/apache/maven/blob/7b38457506544854cdd26aa1d49608050217b923/maven-artifact/src/main/java/org/apache/maven/artifact/Artifact.java#L82 which is part of the API and should not be changed.  So while I would happily remove it if possible, I don't see any good solution.

maven-core/src/main/java/org/apache/maven/internal/transformation/DefaultConsumerPomBuilder.java
gnodet: Fixed with a method

maven-core/src/main/java/org/apache/maven/internal/transformation/TransformedArtifact.java
```diff
@@ -67,36 +71,36 @@ public void setFile(File file) {
...
-        Path result = onChangeTransformer.get();
-        if (result == null) {
-            return null;
+        try {
+            Path result = onChangeTransformer.transform();
+            if (result == null) {
+                return null;
+            }
+            return result.toFile();
+        } catch (Exception e) {
```
elharo, line 80: Can this be a more specific exception subclass?

maven-core/src/main/java/org/apache/maven/internal/transformation/TransformedArtifact.java
```diff
@@ -67,36 +71,36 @@ public void setFile(File file) {
...
+            throw new RuntimeException(e);
```
elharo, line 81: Never throw raw runtime exceptions. Even if the overridden API doesn't allow checked exceptions, use a more specific subclass of RuntimeException, though in this case it might be better to return null

maven-core/src/main/java/org/apache/maven/internal/transformation/TransformedArtifact.java
```diff
@@ -67,36 +71,36 @@ public void setFile(File file) {
...
-                byte[] buffer = new byte[BUFFER_SIZE];
-                int read;
-                while ((read = fis.read(buffer)) != -1) {
-                    md.update(buffer, 0, read);
-                }
-            }
-            StringBuilder result = new StringBuilder();
-            for (byte b : md.digest()) {
-                result.append(String.format("%02x", b));
+    private static String sha1(Path path) throws Exception {
```
elharo, line 87: maybe IOException and perhaps another? Doesn't look like you have to throw  Exception here.

maven-core/src/main/java/org/apache/maven/internal/transformation/TransformedArtifact.java
```diff
@@ -67,36 +71,36 @@ public void setFile(File file) {
...
         }
+        StringBuilder result = new StringBuilder();
+        for (byte b : md.digest()) {
+            result.append(String.format("%02x", b));
+        }
+        return result.toString();
     }
 
-    protected abstract void transform(Path src, Path dst);
+    protected abstract void transform(Path src, Path dst) throws Exception;
```
elharo, line 103: Maybe an IOException, or maybe some redesign such that no exceptions are thrown. This doesn't need any possible exception

maven-model-builder/src/main/java/org/apache/maven/model/building/DefaultTransformerContextBuilder.java
```diff
@@ -106,6 +117,46 @@ private Model findRawModel(Path from, String groupId, String artifactId) {
...
+                        for (String module : rawModel.getModules()) {
+                            File moduleFile = defaultModelBuilder
+                                    .getModelProcessor()
+                                    .locateExistingPom(new File(pom.getParent(), module));
+                            if (moduleFile != null) {
+                                toLoad.add(moduleFile);
+                            }
+                        }
+                    } catch (ModelBuildingException e) {
+                        // gathered with problem collector
```
elharo, line 154: unusual and surprising; getters usually don't have side effects

elharo (changes requested):

maven-model-builder/src/main/java/org/apache/maven/model/building/DefaultTransformerContextBuilder.java
gnodet: Not sure where you see a getter here.

maven-core/src/main/java/org/apache/maven/internal/transformation/TransformedArtifact.java
gnodet: Not sure what you mean.  If I don't throw a checked exception, we get back to the fact you don't like throwing unchecked exceptions. I'm really lost between your conflicting statements.  The full list of checked exceptions that can be checked by the underlying code is the following: `IOException, NoSuchAlgorithmException, XMLStreamException, ModelBuildingException, ComponentLookupException`.

elharo: In general, do not throw java.lang.Exception. Use a more specific subclass that's appropriate to that point in the code. See Items 72 and 73 of Effective Java. 

But it might be possible to not throw any exceptions here, checked or runtime. It depends a little on what the method is expected to do. The trick would be to avoid reading files during the transformation and do that before and/or after the transform. Separate the transform from I/O.

maven-model-builder/src/main/java/org/apache/maven/model/building/DefaultTransformerContextBuilder.java
elharo: e.getProblems on line 155. That looks to me like a no-op that can be eliminated, but it probably isn't.

elharo (changes requested):

maven-core/src/main/java/org/apache/maven/internal/transformation/TransformedArtifact.java
gnodet: That forces the definition of lots of exceptions, just for the sake of avoiding `throws Exception`.  I think I don't really agree to this.
Anyway, I refactored the whole package to avoid exception chaining and everything is caught in a public specific exception.

cstamas (approved):

maven-core/src/main/java/org/apache/maven/internal/transformation/impl/TransformedArtifact.java
```diff
@@ -0,0 +1,148 @@
...
+    }
+
+    @Override
+    public boolean isResolved() {
+        return getFile() != null;
+    }
+
+    @Override
+    public void setFile(File file) {
+        throw new IllegalStateException("transformed artifact file cannot be set");
```
elharo, line 89: UnsupportedOperationException feels more correct here, but why is this even impossible? Why can't a transformed artifact have a file?

maven-core/src/main/java/org/apache/maven/internal/transformation/impl/DefaultConsumerPomBuilder.java
```diff
@@ -0,0 +1,230 @@
...
+    private static <T extends ModelBase.Builder> T prune(T builder, ModelBase model) {
+        builder.properties(null).reporting(null);
+        if (model.getDistributionManagement() != null
+                && model.getDistributionManagement().getRelocation() != null) {
+            // keep relocation only
+            builder.distributionManagement(DistributionManagement.newBuilder()
+                    .relocation(model.getDistributionManagement().getRelocation())
+                    .build());
+        }
+        // only keep repositories others than 'central'
```
elharo, line 219: others --> other

maven-core/src/main/java/org/apache/maven/internal/transformation/TransformedArtifact.java
```diff
@@ -38,7 +38,11 @@
  */
 abstract class TransformedArtifact extends DefaultArtifact {
 
-    private final OnChangeTransformer onChangeTransformer;
+    interface Transformer {
+        Path transform() throws Exception;
```
elharo: I'm not yet convinced that a full java.lang.Exception is desirable here. Possibly an IOException if it's writing/reading files. Possibly no exception at all should be allowed.

elharo (changes requested):

elharo:
> I am blocked by this, so can this be soon merged please? @elharo any counter arguments?

Counter-arguments to what claim? The use of exceptions in this PR does not align with Java standard practices. I don't think that's changed yet.

maven-core/src/main/java/org/apache/maven/internal/transformation/TransformedArtifact.java
gnodet: You're looking at old code.  I've removed `throws Exception` clauses from this package.

maven-core/src/main/java/org/apache/maven/internal/transformation/impl/DefaultConsumerPomBuilder.java
gnodet: Fixed

maven-core/src/main/java/org/apache/maven/internal/transformation/impl/TransformedArtifact.java
gnodet: Fixed the exception thrown.

gnodet:
> > I am blocked by this, so can this be soon merged please? @elharo any counter arguments?
> 
> Counter-arguments to what claim? The use of exceptions in this PR does not align with Java standard practices. I don't think that's changed yet.

I've done a big refactoring of the package yesterday evening.  Please have a look at the latest code.

elharo (approved):

PR #1326, opened 2023-11-25
[MNG-7934] Update to Resolver 1.9.18

cstamas:
Use latest resolver.

---

https://issues.apache.org/jira/browse/MNG-7934


jira-importer:
Resolve #8979

PR #1333, opened 2023-12-06
Dependency collection / resolution and version resolution API

gnodet:


maven-core/src/main/java/org/apache/maven/internal/impl/DefaultDependencyResolver.java
```diff
@@ -0,0 +1,215 @@
...
+
+    @Override
+    public List<Node> flatten(Session session, Node node, ResolutionScope scope) throws DependencyResolverException {
+        nonNull(session, "session can not be null");
+
+        DependencyNode root = cast(
+                        AbstractNode.class, node, "node is not an instance of " + AbstractNode.class.getName())
+                .getDependencyNode();
+        List<DependencyNode> dependencies = new ArrayList<>();
+        DependencyVisitor builder = new PreorderDependencyNodeConsumerVisitor(dependencies::add);
```
cstamas, line 74: spot where new resolver reposys method should be used

cstamas (approved):

PR #1336, opened 2023-12-07
[MNG-8176] Restrict classloader for Maven 4 plugins

gnodet:


maven-core/src/main/resources/META-INF/maven/extension.xml
```diff
@@ -187,6 +189,7 @@ under the License.
     <exportedArtifact>org.apache.maven.resolver:maven-resolver-util</exportedArtifact>
     <exportedArtifact>org.apache.maven.resolver:maven-resolver-connector-basic</exportedArtifact>
 
+    <exportedArtifact>jakarta.inject:jakarta.inject-api</exportedArtifact>
```
rmannibucau, line 192: don't think it should land there, ideally we shouldn't have it (we saw that javax was an issue already) but if imposed by some part of the stack (not sure which one since I thought we stayed on javax intentionally) we should just hide it from any consumer IMHO

gnodet: Or do you mean not adding it for Maven 3.x plugins ? That would be absolutely fine with me.

gnodet: The only other solution I could think about is to rewrite the whole DI injection (Sisu + Guice) because they are not pluggable at all. I'm not ready for that...

rmannibucau: :thinking: I don't see why:

* injections in mojo are already bridged
* abstracting injection markers (with a maven inject annotation) is easy in guice
* abstracting the injector or a lookup (like we had/have in plexus container) is quite trivial
* abstracting a scope is not crazy (wayless than what we already have)

So overall, even if I'm not sure which case(s) you want to cover, I don't see a reason to break our original hypotheis to define a clean and maven owned API and export anything jakarta related.

rmannibucau (approved):
Would be great to get rid of slf4j but clearly a big step forward

cstamas (approved):

gnodet:
> Would be great to get rid of slf4j but clearly a big step forward

The plan would be to allow some customization by the plugin later, so that they can add/remove packages and fine tune their class loader.  In most cases, slf4j is not a problem.

rmannibucau:
> In most cases, slf4j is not a problem.

Well, from my perspective we decided to use maven.api logging api and let the mojo use or not another API/binding/solution so there is no more reason to let slf4j leak (or then we should let leak other stuff too).
Slf4j was not a problem until it became and the choice was to give more guarantees to the mojo writers so default is probably better not having it and use tuning options to let it pass in unsafe mode.

But just my 2 cts indeed.

jira-importer:
Resolve #9683

PR #1338, opened 2023-12-11
[MNG-7961] Up japicmp baseline

cstamas:
From 3.8.7 to 3.8.8, but these two minors have binary incompatibilities (in ctor), so ignores updated as well.

---

https://issues.apache.org/jira/browse/MNG-7961


jira-importer:
Resolve #8945

PR #1339, opened 2023-12-12
[MNG-7959] User controlled relocations

cstamas:
With some improvements. It accepts user property with CSV entries for relocations.

To toy with it, use `-Dmaven.relocations.entries` user property, for example create `.mvn/maven.config` file with contents:
```
-Dmaven.relocations.entries=entry1,entry2,...
```
It accepts CSV (comma delimited) of entries, while entry form is as:
```
GAV>GAV
```
Where left GAV can contain `*` for any elem (so `*:*:*` would mean ALL, something you don't want). Right GAV is either fully specified, or also can contain `*`, then it behaves as "ordinary relocation": the coordinate is preserved from relocated artifact. Finally, if right hand GAV is absent (line looks like "GAV>"). the left hand matching GAV is banned fully (from resolving).

Note: the ">" means project level, while ">>" means global (whole session level, so even plugins will get relocated artifacts) relocation.

Examples:
```
-Dmaven.relocations.entries=org.foo:*:*>,org.here:*:*>org.there:*:*,javax.inject:javax.inject:1>>jakarta.inject:jakarta.inject:1.0.5
```

Meaning: 3 entries, ban `org.foo` group (exactly, so `org.foo.bar` is allowed), relocate `org.here` to `org.there` and finally **globally relocate** (see ">>") `javax.inject:javax.inject:1` to `jakarta.inject:jakarta.inject:1.0.5`

---

https://issues.apache.org/jira/browse/MNG-7959


laeubi:
@cstamas sound quite handy, I wonder if an extra file is required here (its often a bit hard to really reference "right" root as there is no official "the location of the `.mvn` folder") alos it adds maybe more stuff to the repo and hopefully user relocations should be rarely used!

So I would like to suggest to support (in addition to the file) support a syntax like this:

```
-Dmaven.relocation.banfoo=org.foo:*:*>
-Dmaven.relocation.org=org.here:*:*>org.there:*:*
-Dmaven.relocation.javax=javax.inject:javax.inject:1>jakarta.inject:jakarta.inject:1.0.5
```

so basically everything starting with `-Dmaven.relocation.` followed by any string will be merged with the content of a possible file given.

cstamas:
I think you are right:
* `-D` should be supported
* **hopefully user relocations should be rarely used** -- this one especially

Hence, will drop this "file" stuff and support only `-D` provided ones. Still, this change in general is OK as it opens the doors to other (ie. new POM model changes) relocations as well...

gnodet (approved):
I've made some changes to bring the ability to relocate artifacts with classifiers / extensions, please have a look.

jira-importer:
Resolve #8953

PR #1340, opened 2023-12-13
[MNG-7963] Clean up the build

cstamas:
Changes:
* remove hamcrest 1.3 completely
* set proc:none to prevent warnings (Sisu APT is on classpath), we do not use annotation processing, we produce Sisu index explicitly.
* remove some redundant elements from POMs
* internally switch to Sisu own annotations (Typed, PreDestroy, etc)
* sort out Runtime DI and other

---

https://issues.apache.org/jira/browse/MNG-7963


gnodet:
One requirement I have for the builds is that a subsequent run of `mvn package` does not change the jars if you run it.  Basically, running `ls -ltr **/*.jar` should give you jars which are all older than the time you launched the build.

cstamas:
Ok, was unaware of IT MavenITmng6084Jsr250PluginTest, it is only failed. Will return javax.annotation-api then...

gnodet (approved):
LGTM, nice cleanup !

maven-embedder/src/main/java/org/eclipse/sisu/plexus/PlexusXmlBeanConverter.java
```diff
@@ -43,6 +42,7 @@
 import org.codehaus.plexus.util.xml.pull.MXParser;
 import org.codehaus.plexus.util.xml.pull.XmlPullParser;
 import org.codehaus.plexus.util.xml.pull.XmlPullParserException;
+import org.eclipse.sisu.Priority;
```
olamy, line 45: out of curiosity why changing this package?

cstamas: Historically, Sisu relied on JSR250 and alike annotations, but as time passed, it implemented more and more of "own" annotations. Since a while (0.3.4 or so) it has it all in own namespace as well. If you look at history of this PR, you will see that I even removed `java.annotations:javax.annotations-api` from Maven completely, and Maven has happy about it, but we have one IT that uses it, the MavenITmng6084Jsr250PluginTest, hence I had to undo that change.

Given Sisu is "internal implementation detail" of Maven Core, I think using Sisu annotations _within_ Maven core is best/better, than relying on something else.

So this PR just makes sure that Maven Core uses Sisu annotations instead of something else, while the semantics 100% remains same as test/ITs prove.

jira-importer:
Resolve #8927

PR #1341, opened 2023-12-13
[MNG-7962] Return language to Type

cstamas:
Returns the language property to Type

---

https://issues.apache.org/jira/browse/MNG-7962


gnodet (approved):
LGTM

PR #1343, opened 2023-12-14
[MNG-7966] Augment api to provide a Map<Dependency, Path> in the resolution result

gnodet:
JIRA issue: https://issues.apache.org/jira/browse/MNG-7966

This is a bit of an ugly hack, but I don't think we should modify the Maven 3 api for the `ProjectArtifactCache`, so I introduced:
```
    interface ArtifactsSetWithResult extends Set<Artifact> {
        DependencyResolutionResult getResult();
    }
```

If that's deemed too ugly, I can modify the `ProjectArtifactCache` to add a `DependencyResolutionResult` in the `CacheRecord` maybe...


cstamas (approved):

jira-importer:
Resolve #8923

PR #1346, opened 2023-12-17
[MNG-7968] Expose Maven Version as Version (not String)

cstamas:
Change existing API to expose Version instead of String.

---

https://issues.apache.org/jira/browse/MNG-7968


api/maven-api-core/src/main/java/org/apache/maven/api/services/RuntimeInformation.java
```diff
@@ -0,0 +1,41 @@
...
+@Experimental
+public interface RuntimeInformation extends Service {
+
+    /**
+     * Returns the "runtime version", Maven version caller runs in.
+     *
+     * @return the runtime version, never {@code null}
+     */
+    @Nonnull
+    Version runtimeVersion();
```
gnodet, line 40: We already have `Session.getMavenVersion()`, I'm not sure we need a new service just for this bit.  However the session's method returns a `String`, maybe it needs to be changed to a `Version`.  If we keep this service, then the `Session.getMavenVersion` method should delegate to that Service.
I think I did not used a service, I did not see the value for that single method...

gnodet (changes requested):
Either drop the service and this PR, if we think it should be a Service, then have Session.getMavenVersion delegate to this new service.

api/maven-api-core/src/main/java/org/apache/maven/api/services/RuntimeInformation.java
cstamas: Will drop this and make `Session.getMavenVersion()` return a `Version`, as the most cases RT version is needed is:
* display purposes (asString() is there)
* compare/range check, and Version is needed for that

Having `Session.getMavenVersion()` return plain String is ok for 1st but not enough for 2nd use case.

gnodet (approved):

jira-importer:
Resolve #8921

PR #1347, opened 2023-12-17
[MNG-7972] Artifact coordinate out of "standard string"

cstamas:
Expose ArtifactCoordinate creation out of standard string.

---

https://issues.apache.org/jira/browse/MNG-7972

### DIFF
--- a/api/maven-api-core/src/main/java/org/apache/maven/api/services/ArtifactCoordinateFactory.java
+++ b/api/maven-api-core/src/main/java/org/apache/maven/api/services/ArtifactCoordinateFactory.java
@@ -37,7 +37,7 @@ public interface ArtifactCoordinateFactory extends Service {
      * Creates a coordinate.
      *
      * @param request the request holding coordinate creation parameters
-     * @return an {@code Artifact}, never {@code null}
+     * @return an {@code ArtifactCoordinate}, never {@code null}
      * @throws IllegalArgumentException if {@code request} is null or {@code request.session} is null or invalid
      */
     @Nonnull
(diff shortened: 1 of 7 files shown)


api/maven-api-core/src/main/java/org/apache/maven/api/services/ArtifactCoordinateFactory.java
```diff
@@ -43,6 +43,18 @@ public interface ArtifactCoordinateFactory extends Service {
...
+     * Creates a coordinate out of string that is formatted like:
+     * {@code <groupId>:<artifactId>[:<extension>[:<classifier>]]:<version>}
+     *
+     * @param session the session.
+     * @param coordinateString the string having "standard" coordinate.
+     * @return an {@code Artifact}, never {@code null}
+     * @throws IllegalArgumentException if {@code request} is null or {@code request.session} is null or invalid
+     */
+    @Nonnull
+    ArtifactCoordinate create(@Nonnull Session session, @Nonnull String coordinateString);
```
gnodet, line 56: In other services, when multiple "signatures" are involved, a single "request" object is used to store all those bits.  I'm not sure why we should not do the same here, i.e. add a `coordinateString` member to the `ArtifactCoordinateRequest`.

gnodet (changes requested):

api/maven-api-core/src/main/java/org/apache/maven/api/services/ArtifactCoordinateFactory.java
cstamas: moved stuff into request

gnodet (approved):
I've made some changes, please have a look too.

PR #1352, opened 2023-12-18
[MNG-7978] Decouple maven-artifact from maven-resolver-provider

cstamas:
As it is actually not needed. This also "pulls in" all of VersionParser into model version parser, so Core implementation now 100% just delegates over there.

---

https://issues.apache.org/jira/browse/MNG-7978


gnodet (approved):

jira-importer:
Resolve #8867

PR #1354, opened 2023-12-19
[MNG-7980] CLI switch to ignore transitive repositories.

cstamas:
If set, Maven will ignore all remote repositories introduced by transitive dependencies POM.

---

https://issues.apache.org/jira/browse/MNG-7980


gnodet (approved):

jira-importer:
Resolve #8864

PR #1367, opened 2024-01-09
[MNG-7999] Confine Plexus Container as much as possible

cstamas:
This is an ongoing effort to confine Plexus, but also perform a bit of cleanup in Maven Core and around. No logic changes, just replacing Plexus with Lookup (that is a thin wrapper around it), and removing unused members, redundant checks, etc. Module maven-compat omitted on purpose.

---

https://issues.apache.org/jira/browse/MNG-7999

### DIFF
--- a/maven-core/src/main/java/org/apache/maven/internal/impl/DefaultLookup.java
+++ b/maven-core/src/main/java/org/apache/maven/internal/impl/DefaultLookup.java
@@ -24,6 +24,8 @@ import javax.inject.Singleton;
 
 import java.util.List;
 import java.util.Map;
+import java.util.NoSuchElementException;
+import java.util.Optional;
 
 import org.apache.maven.api.services.Lookup;
 import org.apache.maven.api.services.LookupException;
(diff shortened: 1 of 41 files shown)


maven-core/src/main/java/org/apache/maven/lifecycle/internal/DependencyContext.java
```diff
@@ -48,7 +43,7 @@ public class DependencyContext {
 
     private final Collection<String> scopesToResolveForAggregatedProjects;
 
-    private volatile Collection<?> lastDependencyArtifacts = UNRESOLVED;
+    private volatile Collection<?> lastDependencyArtifacts = new ArrayList<>();
```
gnodet, line 46: `Collections.emptyList()`, as this collection is never written to.

cstamas: fixed

gnodet:
What about adding a specific exception inheriting `LookupException` for when the component is  not found when calling `<T> T lookup(Class<T> role)` ?

gnodet:
> What about adding a specific exception inheriting `LookupException` for when the component is not found when calling `<T> T lookup(Class<T> role)` ?

Or maybe there's no use case for that and users should use `lookupOptional` if they intend to act differently. That would make sense too.

maven-core/src/main/java/org/apache/maven/internal/impl/DefaultLookup.java
```diff
@@ -24,6 +24,8 @@
```
rmannibucau, line 24: thinking a bit out loud there but I'm starting to wonder if we should expose guice Injector instead of this container wrapper which basically propagate the status-quo in terms of number of containers in the API and have a single lookup(Type) method.
Probably something to think about before releasing this since ultimately this will need to move from internal to public cause it is a common need to be able to lookup public components programmatically (scripting case being the extreme example).

gnodet (approved):

maven-core/src/main/java/org/apache/maven/internal/impl/DefaultLookup.java
cstamas: Perso, I'd even move Lookup service to "internal service" (like CoreRealm is), not exposing it to plugins. It serves a very good purpose in Core (to hide Plexus), while plugins can achieve all these with plain JSR330 (or Maven specific JSR330-like annos if they become to existence once).

jira-importer:
Resolve #10873

PR #1370, opened 2024-01-10
[MNG-7854] Warn if imported dep is ignored

cstamas:
The artificial (or "bolted on") scope "import" behaves wildly differently that rest of Maven, causes surprise to our users. For start, we should emit "actionable" warnings about these, ultimate goal is to align this behaviour with "maven way" of working ("closest occurence wins" vs current weird "direct occurrence then first occurence wins" strategy)

---

https://issues.apache.org/jira/browse/MNG-7854

### DIFF
--- a/maven-model-builder/src/main/java/org/apache/maven/model/building/DefaultModelProblem.java
+++ b/maven-model-builder/src/main/java/org/apache/maven/model/building/DefaultModelProblem.java
@@ -136,16 +136,16 @@ public class DefaultModelProblem implements ModelProblem {
 
     @Override
     public String getMessage() {
-        String msg;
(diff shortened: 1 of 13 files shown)


gnodet:
I've added a few unit tests.  I've slightly changed things so that the warnings are only created when the dependency is not directly managed.  The warning also suggests to add a direct managed dependency to get rid of it, so it's now helpful and actionnable.

That's clearly a good first step.  The next one could be to use a "closest wins" strategy instead of this "direct wins, then first wins" weird ad-hoc strategy.

gnodet (approved):

jira-importer:
Resolve #9066

PR #1371, opened 2024-01-10
[MNG-8007] Maven TransferListener wrapper

cstamas:
A wrapper listener that can receive events even from multiple threads, and passes them to wrapped delegate listener on single thread, keeping delegate implementation simple.

---

https://issues.apache.org/jira/browse/MNG-8007


gnodet:
So the problem is that there's no synchronisation on the last resource being finished to erase the display.
This can lead to output such as:
image

cstamas:
@gnodet updated it to sync on last event

gnodet:
Looks better.  I wonder if we should have a time-based windowing for batches, rather than based on the number (or not).  Or do you think it would be better handled by the wrapped transfer listener ?

cstamas:
My experience tells to always keep simple as possible any concurrent code :smile: ...

I don't think we need any extra thing, at least until someone finds a problem here.

gnodet (approved):

maven-embedder/src/main/java/org/apache/maven/cli/transfer/SimplexTransferListener.java
```diff
@@ -0,0 +1,219 @@
...
+                            delegate.transferCorrupted(transferEvent);
+                            break;
+                        case SUCCEEDED:
+                            delegate.transferSucceeded(transferEvent);
+                            break;
+                        case FAILED:
+                            delegate.transferFailed(transferEvent);
+                            break;
+                        default:
+                            throw new IllegalArgumentException("Type: " + type);
```
cstamas, line 211: doing this does not make any sense IMO, maybe just ignore it? It will just kill the daemon thread, killing any transport logging of Maven...

gnodet (approved):

PR #1377, opened 2024-01-12
[MNG-8016] Simplify ModelCache

gnodet:
The `ModelCache` is now bloated and can be simplified a lot.

---

https://issues.apache.org/jira/browse/MNG-8016


cstamas (approved):

jira-importer:
Resolve #10836

PR #1378, opened 2024-01-13
[MNG-8015] Control the type of path where each dependency can be placed

desruisseaux:
Introduces new API described in JIRA issue [MNG-8015](https://issues.apache.org/jira/browse/MNG-8015):

* `PathType` enumeration-like class.
* `JavaPathType` subtype with values such as `MODULES`, `CLASSES`, `DOCLET`, `TAGLETS`, _etc._
* `PATH_TYPES` dependency property key (as a side-effect, the `DependencyProperties` API is modified for making it type-safe).
* `DependencyResolverRequest` API for allowing plugins to specify the `PathType` that the plugin wants.
* `DependencyResolverResult.getDispatchedPath()` method for getting the result: the paths to dependencies for each path type.

# Open issues
Commit 9c5e9ff412718e4ad29299ea79ba0d158c27d43d (which add implementation) introduces a dependency to Java 9 or later (the `ModuleDescriptor` class). If Maven 4 decides to stay executable on Java 8, this part will need to be refactored in a multi-version JAR file. If Maven 4 decides to require Java 9 or later, the `source` and `target` options in the `pom.xml` file should be replaced by a `release` option.

The implementation has no JUnit tests yet, so it may contain bug. We may want to write JUnit tests before to integrate this pull request. Waiting to see if a discussion causes a modification of this proposal before to write tests.


desruisseaux:
CI build has been successful on Java 17 and 21, but failed on Java 11 because of the use of the `{@return ...}` javadoc tag. This is a Javadoc issue only. Resolution depends on Maven PMC decision about which Java version to require.

cstamas:
If am not mistaken, [this feature](https://bugs.openjdk.org/browse/JDK-8075778) is available since Java 16? Given master is still Java 8 (build time requirement is same I think)...  I'd stick with it.

So please use Java 8 conformant javadoc...

maven-artifact/src/main/java/org/apache/maven/artifact/handler/ArtifactHandler.java
```diff
@@ -53,5 +53,16 @@ public interface ArtifactHandler {
...
+     * Specifies if the artifact contains java classes and can be added to the classpath.
+     * Whether the artifact <em>should</em> be added to the classpath depends on other
+     * dependency properties.
+     *
+     * @return if the artifact <em>can</em> be added to the class path
+     *
+     * @deprecated A value of {@code true} does not mean that the dependency <em>should</em>
+     * be placed on the classpath. See {@code JavaPathType} instead for better analysis.
+     */
+    @Deprecated
```
cstamas, line 66: IMHO whole artifactHandler/ArtifactHandlerManager should be deprecated, is there ONLY to support Maven3 plugins in Maven4, while since eee037e6763445e7901b84fce4e992ddec17df0d they are really just adapting maven4 stuff.

api/maven-api-core/src/main/java/org/apache/maven/api/DependencyProperties.java
```diff
@@ -33,27 +36,226 @@
...
     /**
-     * Boolean flag telling that dependency is meant to be placed on class path. Value of this key should be parsed with
-     * {@link Boolean#parseBoolean(String)} to obtain value.
+     * Boolean flag telling that dependency is meant to be placed on class path.
+     *
+     * @deprecated Use {@link #PATH_TYPES} and {@link JavaPathType#CLASSES} instead.
      */
-    String FLAG_CLASS_PATH_CONSTITUENT = "classPathConstituent";
+    @Deprecated
+    Key<Boolean> FLAG_CLASS_PATH_CONSTITUENT = new Key<>("classPathConstituent", Boolean.class).intern();
```
cstamas, line 217: Given this flag was added in some alpha, let's remove it, as it will remove some other stuff as well, there is not need to keep and deprecate stuff not yet in anything else than alpha. We maybe need to keep the `Type#isAddedToClassPath`, but am unsure.

api/maven-api-core/src/main/java/org/apache/maven/api/DependencyProperties.java
```diff
@@ -33,27 +36,226 @@
...
+     */
+    Key<PathType[]> PATH_TYPES = new Key<>("pathTypes", PathType[].class).intern();
+
+    /**
+     * Boolean flag telling that dependency contains all of its dependencies.
      * <p>
      * <em>Important: this flag must be kept in sync with resolver! (as is used during collection)</em>
      */
-    String FLAG_INCLUDES_DEPENDENCIES = "includesDependencies";
+    Key<Boolean> FLAG_INCLUDES_DEPENDENCIES = new Key<>("includesDependencies", Boolean.class).intern();
```
cstamas, line 209: This is the ONLY flag that needs to get down and is interpreted by Resolver. All the other flags (like `classPathConstituent` is) are totally opaque to Resolver and ignored. This flag causes dependency collection to stop.

api/maven-api-core/src/main/java/org/apache/maven/api/DependencyProperties.java
```diff
@@ -33,27 +36,226 @@
...
+        @Nonnull
+        private final Class<V> valueType;
+
+        /**
+         * Creates a new key.
+         *
+         * @param name name of the key
+         * @param valueType type of value associated to the key
+         */
+        public Key(@Nonnull final String name, @Nonnull final Class<V> valueType) {
```
gnodet, line 71: `final` in parameters is useless, please remove

api/maven-api-core/src/main/java/org/apache/maven/api/DependencyProperties.java
```diff
@@ -33,27 +36,226 @@
...
+         * as determined by the {@link #equals(Object)} method, then the key from the pool is returned. Otherwise,
+         * if no key exist in the pool for this key {@linkplain #name() name}, then this {@code Key} object is added
+         * to the pool and {@code this} is returned. Otherwise an {@link IllegalStateException} is thrown.
+         *
+         * @throws IllegalStateException if a key exists in the pool for the same name but a different class of values.
+         *
+         * @see String#intern()
+         */
+        @SuppressWarnings("unchecked")
+        public Key<V> intern() {
```
gnodet, line 126: I don't think the whole `intern` thing is needed at all.
I would suppose all references to `Key`s will be done using `DependencyProperties.XXX`, so I don't see a good use case for users to create duplicate instances.

api/maven-api-core/src/main/java/org/apache/maven/api/JavaPathType.java
```diff
@@ -0,0 +1,282 @@
...
+import org.apache.maven.api.annotations.Experimental;
+import org.apache.maven.api.annotations.Nonnull;
+
+/**
+ * The option of a Java command-line tool where to place the paths to some dependencies.
+ * A {@code PathType} can identify the class-path, the module-path, the patches for a specific module,
+ * or another kind of path. This class is like an enumeration, except that it is extensible:
+ * plugins can define their own kinds of path.
+ *
+ * <p>One path type is handled in a special way: contrarily to other options,
```
gnodet, line 36: `contrarily to` -> `unlike`

api/maven-api-core/src/main/java/org/apache/maven/api/JavaPathType.java
```diff
@@ -0,0 +1,282 @@
...
+ */
+@Experimental
+public final class JavaPathType extends PathType {
+    /**
+     * The path identified by the Java {@code --class-path} option.
+     * Used for compilation, execution and Javadoc among others.
+     *
+     * <h4>Context-sensitive interpretation</h4>
+     * A dependency with this path type will not necessarily be placed on the class-path.
+     * There is two circumstances where the dependency may be nevertheless placed somewhere else:
```
gnodet, line 56: `There is` -> `There are`
`be nevertheless` -> `nevertheless be`

api/maven-api-core/src/main/java/org/apache/maven/api/JavaPathType.java
```diff
@@ -0,0 +1,282 @@
...
+     *
+     * <ul>
+     *   <li>If {@link #MODULES} path type is also set, then the dependency can be placed either on the
+     *       class-path or on the module-path, but only one of those. The choice is up to the plugin,
+     *       possibly using heuristic rules (Maven 3 behavior).</li>
+     *   <li>If a {@link #patchModule(String)} is also set and the main JAR file was placed on the module-path,
```
gnodet, line 62: `was placed` -> `is placed`

api/maven-api-core/src/main/java/org/apache/maven/api/JavaPathType.java
```diff
@@ -0,0 +1,282 @@
...
+     */
+    public static final JavaPathType CLASSES = new JavaPathType("CLASSES", "--class-path", null);
+
+    /**
+     * The path identified by the Java {@code --module-path} option.
+     * Used for compilation, execution and Javadoc among others.
+     *
+     * <h4>Context-sensitive interpretation</h4>
+     * A dependency with this flag will not necessarily be placed on the module-path.
+     * There is two circumstances where the dependency may be nevertheless placed somewhere else:
```
gnodet, line 75: `There is` -> `There are`
`be nevertheless` -> `nevertheless be`

api/maven-api-core/src/main/java/org/apache/maven/api/JavaPathType.java
```diff
@@ -0,0 +1,282 @@
...
+     *
+     * <ul>
+     *   <li>If {@link #CLASSES} path type is also set, then the dependency <em>should</em> be placed on the
+     *       module-path, but is nevertheless compatible with placement on the class-path. Compatibility can
```
gnodet, line 79: `nevertheless` -> `also`

api/maven-api-core/src/main/java/org/apache/maven/api/JavaPathType.java
```diff
@@ -0,0 +1,282 @@
...
+     *       be achieved, for example, by repeating in the {@code META-INF/services/} directory the services
+     *       that are declared in the {@code module-info.class} file. In that case, the path type can be chosen
+     *       by the plugin.</li>
+     *   <li>If a {@link #patchModule(String)} is also set and the main JAR file was placed on the module-path,
+     *       then the test dependency will be placed on the Java {@code --patch-module} option instead of the
+     *       module-path.</li>
```
gnodet, line 85: `module-path` -> `{@code --module-path}` for consistency

api/maven-api-core/src/main/java/org/apache/maven/api/PathType.java
```diff
@@ -0,0 +1,143 @@
...
+ * <p>Path types are often exclusive. For example, a dependency should not be both
+ * on the Java class-path and on the Java module-path.</p>
+ *
+ * @see DependencyProperties#PATH_TYPES
+ * @see org.apache.maven.api.services.DependencyResolverResult#getDispatchedPaths()
+ *
+ * @since 4.0.0
+ */
+@Experimental
+public abstract class PathType {
```
gnodet, line 44: I'm not convinced we will ever have an implementation other than `JavaPathType`, in which case, both interfaces could be merged.  I'd also prefer an interface rather than an abstract class if that's possible.

api/maven-api-core/src/main/java/org/apache/maven/api/DependencyProperties.java
```diff
@@ -33,27 +36,226 @@
...
+    Set<Key<?>> keys();
+
+    /**
+     * Returns the value associated to the given key.
+     *
+     * @param <V> type of value to get
+     * @param key key of the value to get
+     * @return value associated to the given key, or {@code null} if none
+     */
+    <V> V get(@Nonnull Key<V> key);
```
gnodet, line 231: `Optional<V>` as a return value ?

api/maven-api-core/src/main/java/org/apache/maven/api/DependencyProperties.java
```diff
@@ -33,27 +36,226 @@
...
+
+    /**
+     * Returns the value associated to the given key, or the given default value if none.
+     *
+     * @param <V> type of value to get
+     * @param key key of the value to get
+     * @param defaultValue the value to return is none is associated to the given key, or {@code null}
+     * @return value associated to the given key, or {@code null} if none and the default is null
+     */
+    <V> V getOrDefault(@Nonnull Key<V> key, V defaultValue);
```
gnodet, line 241: Missing `@Nullable` on the return value and the `V` parameter to be explicit .

api/maven-api-core/src/main/java/org/apache/maven/api/Type.java
```diff
@@ -80,11 +129,16 @@ public interface Type {
...
+     * Whether the artifact <em>should</em> be added to the classpath depends on other
+     * {@linkplain #getDependencyProperties() dependency properties}.
+     *
+     * @return if the artifact <em>can</em> be added to the class path
      *
-     * @return if the artifact should be added to the class path
+     * @deprecated A value of {@code true} does not mean that the dependency <em>should</em>
+     * be placed on the classpath. See {@link JavaPathType} instead for better analysis.
      */
+    @Deprecated
```
gnodet, line 141: If deprecation is necessary, it should be noted that it will be removed before 4.0.0 GA.

api/maven-api-core/src/main/java/org/apache/maven/api/DependencyProperties.java
```diff
@@ -33,27 +36,226 @@
...
+     * the dependency will be unconditionally placed on the specified type of path
+     * and no heuristic rule will be involved.
+     *
+     * <p>It is nevertheless common to specify two or more types of path. For example,
+     * a Java library may be compatible with either the class-path or the module-path,
+     * and the user may have provided no instruction about which type to use. In such
+     * case, the plugin may apply rules for choosing a path. See for example
+     * {@link JavaPathType#CLASSES} and {@link JavaPathType#MODULES}.</p>
+     */
+    Key<PathType[]> PATH_TYPES = new Key<>("pathTypes", PathType[].class).intern();
```
gnodet, line 202: `Collection<PathType>` so that we can ensure immutability ?

gnodet (changes requested):

api/maven-api-core/src/main/java/org/apache/maven/api/DependencyProperties.java
desruisseaux: The use of collection is not fully type-safe with the `Key<V>` mechanism, because of generic type erasure. A `java.lang.Class` can represent `Collection`, but cannot represent `Collection<PathType>`. By contrast, the use of array is fully type-safe (can be represented by a `Class`), and the immutability may come in the future with [Frozen Arrays](https://openjdk.org/jeps/8261007).

A simpler alternative would be to not manage `PathType` in `DependencyProperties`, but instead as an ordinary method directly in `Dependency`. Use of collections would be type-safe there.

api/maven-api-core/src/main/java/org/apache/maven/api/PathType.java
desruisseaux: We can take an approach similar to `java.nio.file`, where `OpenOption` is an interface and `StandardOpenOption` is an `enum` implementing that interface. So no problem for making `PathType` an interface.

`JavaPathType` (or `PathType` if merged) would have been an `Enum` if it wasn't for the need to be extensible. Enumerations have to be classes, otherwise we cannot define the enumeration values in the API.

api/maven-api-core/src/main/java/org/apache/maven/api/DependencyProperties.java
desruisseaux: The `intern()` mechanism is for the `org.apache.maven.internal.impl.DefaultDependency` constructor. That constructor takes in argument an instance of `org.eclipse.aether.graph.Dependency` and copies its properties. But the Eclipse's properties are `<String, String>` entries, so the `String` keys need to be converted to `DependencyProperties.Key` instances. This is the purpose of `DependencyProperties.Key.forName(…)`, which provides `String` → `Key` conversions. The `intern()` stuff is a side-effect of the need to support that conversion. It could be removed if `DefaultDependency` wasn't merely a wrapper for `org.eclipse.aether.graph.Dependency`.

api/maven-api-core/src/main/java/org/apache/maven/api/DependencyProperties.java
```diff
@@ -33,27 +36,226 @@
...
+        public String toString() {
+            return name;
+        }
+
+        /**
+         * {@return an hash code value for this key}.
+         */
+        @Override
+        public int hashCode() {
+            return 7 + name.hashCode() + 31 * valueType.hashCode();
```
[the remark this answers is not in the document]
desruisseaux: No objection, but `name` and `valueType` cannot be null and the current code avoids implicit creation of a temporary array. But I agree that the performance impact is probably unnoticeable.

api/maven-api-core/src/main/java/org/apache/maven/api/DependencyProperties.java
```diff
@@ -33,27 +36,226 @@
...
+     * @param  <V>  type of value associated to the key
+     */
+    class Key<V> {
+        /**
+         * The keys that are defined in this {@code DependencyProperties} map.
+         * Accesses to this map shall be synchronized on the map.
+         *
+         * @see #intern()
+         */
+        private static final Map<String, Key<?>> INTERNS = new HashMap<>();
```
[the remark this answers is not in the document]
desruisseaux: I wasn't expecting a lot of concurrent accesses for this code, but no objection in using `ConcurrentHashMap`.

api/maven-api-core/src/main/java/org/apache/maven/api/DependencyProperties.java
desruisseaux: Even if we restrict the scope to standard Java tools, the extensibility is still needed for supporting the `--patch-module` option, which can be seen as class-paths specific to each module. So a new enumeration value needs to be created (but not internalized) for each module to patch. An alternative design would be to make `JavaPathType` an `Enum` and support the `--path-module` option in a separated implementation of `PathType`. Whether it would make the API simpler or not is unclear to me.

cstamas: We need extensibility, as tomorrow somebody could come up with custom package that would need some special handling... story is same as in Mave3 and ArtifactHandlers... they have to be extensible, that is the whole point.

gnodet: Would it make sense to remove the `intern()` method and redefine `forName` as:
```
        @Nullable
        public static <T> Key<T> forName(@Nonnull String name, @Nullable Class<T> defaultType, boolean intern) {
            Key<?> key;
            synchronized (INTERNS) {
                key = INTERNS.get(name);
                if (defaultType != null) {
                    if (key == null) {
                        key = new Key<>(name, defaultType);
                        if (intern) {
                            INTERNS.put(name, key);
                        }
                    } else if (key.valueType() != defaultType && intern) {
                        throw new IllegalStateException("Key " + name + " already exists for a different class of values.");
                    }
                }
            }
            return (Key) key;
        }
```
The constructor could be made `private`, so that the only entry point is the above method.

api/maven-api-core/src/main/java/org/apache/maven/api/DependencyProperties.java
gnodet: What about using `java.lang.reflect.Type` instead of `Class<?>` here and in the value converters ?

api/maven-api-core/src/main/java/org/apache/maven/api/DependencyProperties.java
desruisseaux: Yes, we can do that.

api/maven-api-core/src/main/java/org/apache/maven/api/DependencyProperties.java
desruisseaux: I think it would not avoid the "unchecked cast" compiler warning. An easier approach would be to replace this property by an ordinary `Set<PathType> getPathTypes()` method in the `Dependency` interface.

api/maven-api-core/src/main/java/org/apache/maven/api/DependencyProperties.java
desruisseaux: Actually the above code does not fit exactly the `DefaultDependencies` needs, because it forces the key to be of the type specified by `defaultType`. This is the correct thing to do for above method because its return value is `Key<T>`. However, `DefaultDependencies` needs to first check for a key with whatever type an existing key may have. So we cannot escape the need to have two methods, one with return value `Key<?>` and another one with return value `Key<V>`.

I will replace the existing `forName` method with a simpler one like below, checking only for existing key:

```
Optional<Key<?>> forName(String);
```

As a replacement of `intern()`, I considered adding the following method:

```
<V> Key<V> createAndCache(String name, Class<V> valueType);
```

However it would not work if some plugin wants to define its own subclass of `Key`, while `intern()` allows this flexibility. I'm not sure if it is a flexibility worth to have however.

api/maven-api-core/src/main/java/org/apache/maven/api/PathType.java
desruisseaux: Makes `PathType` an interface, `JavaPathType` an enumeration implementing that interface, and the case of the `--patch-module` option an inner class named `Modular` also implementing the `PathType` interface. The latter class generalizes `--path-module` support to other kinds of module-dependent option if some appear in the future.

rmannibucau: Not sure we need the enumeration at all, java is not in a closed set of paths (just check out war, ear, osgi for common examples but this is still just a subset) so in this design type is really a banalised value core does not even have to be aware about at the end of the refactoring, only mojos (even if shared somewhere but not core) so an interface is okish even if it is really just a `String` and not name+option.
The link to an option is dependent of the context (=mojo) so would rather belong to the impl (integration layer since we wouldnt have an interface for a String I guess) - this is one reason I was not ultra enthusiast to use type for that, means we need to ensure we can duplicate dependencies with multiple types now even if all are just `jar`.

From a design standpoint this is too opiniated to stay in core IMHO as mentionned on the list so maybe keep it an `interface {name}` or merge whatever works then work on moving it outside core (not 100% sure maven-shared is still a thing for plugins but looks like it would be a saner location or we just create a repo for java plain command line options maybe).

desruisseaux:
Applied most of above comments (still thinking for a solution about the two remaining ones). The build continues to fail on Java 11, again because of Javadoc, but this new error is more annoying. The error message is _"Header used out of sequence: `<H4>`"_. This error happens when HTML title headers are used in Javadoc for clear separation of sections. Javadoc verifies that the numbers are used in sequential order: `<H4>` inside `<H3>`, themselves inside `<H2>`, _etc_. A difficulty is that Javadoc uses `<H1>`, `<H2>`, _etc._ for its own purpose, so the first level available in the Javadoc of a method starts at `<H4>`. Javadoc verifies that we use the proper level, which is good. The problem is that the expected level changed somewhere between Java 11 and Java 17 (I think it was around Java 13). Java 13 (approximately) and later expects `<H4>`. Java 11 expected something else.  All versions raise an error when the HTML header level is not what they expected.

So it is impossible (as far as I know) to have a code compilable on both Java 11 and Java 17 when HTML headers are used in Javadoc and Javadoc checks are enabled. If compilation or Javadoc generation passes on one, it fails on the other. Workaround can be one of the following:

* Do not use HTML headers.
* Disable Javadoc checks.
* Requires Java 17 for compilation but target Java 11 or 8 with `--release` option.

Which approach should we take? The last one would be the one allowing most gradual transition to newest Java versions (by allowing the use of most recent conventions right now) while preserving compatibility.

api/maven-api-core/src/main/java/org/apache/maven/api/PathType.java
desruisseaux: The `JavaPathType` focuses on the standard Java tools. In order to avoid being opinionated, the current strategy is to take the [Java Development Kit Version 21 Tool Specifications](https://docs.oracle.com/en/java/javase/21/docs/specs/man/index.html) and put all options for all kinds of path in the `JavaPathType` enumeration. So this is non-arbitrary from a Java standard point of view.

If there is a demand for other popular types of path, I suppose that they could be added to this enumeration as well. But even if they are not added, all plugins have the possibility to provide their own `PathType` enumeration if they want.

The `PathType.option()` method could indeed be controversial. The reason for this API is because some options are formatted is a special way. Currently, `--patch-module` is the only example I'm aware of, but more may exist. This architecture makes possible to handle those special cases in dedicated `PathType` implementations. The presence of this API in a shared place is because _Java Development Kit Version 21 Tool Specifications_ uses those options consistently in all tools as far as I know. For `PathType` defined in plugins, the plugin has the possibility to provide the right option for the tool that they target.

Note: `JavaPathType` is currently not the only Java-specific interface in core. There is also `JavaToolchain`.

rmannibucau: > So this is non-arbitrary from a Java standard point of view.

well, with examples I gave it is, you target a few java standalone tools - agree it is the most used in builds but it is not a reason IMHO and stays bad to link core to spceific plugins so much (we already suffer from that).

> I suppose that they could be added to this enumeration as well

So what you propose is to include compiler, javadoc, jlink, jpackage etc into maven core and drop plugins (I know it is not what you wrote but it is what it means otherwise the enum is not correct as explained before).

> The PathType.option() method could indeed be controversial.

It is well understood but here again linked to specific tools to in maven it belongs to plugins or their transitive deps ("shared" module), not to core which is just a plugin orchestrator from a design standpoint, less we have there better we'll be IMHO.

desruisseaux: > well, with examples I gave it is

No, Java is a registered trademark protected by laws. Only Oracle controls what can be call "Java® specification". Everything else is non-standard, unless agreed by Oracle (e.g. through JSP). It may be widely-used, but it is not standard.

> what you propose is to include compiler, javadoc, jlink, jpackage etc into maven core and drop plugins 

No. There is a plugin for `javac`, another plugin for `javadoc`, _etc._ But all those plugins share the notion of class-path, module-path, _etc._ The path type is a cross-plugins concept, which is why it needs to be defined in a common place. That's all, only the common stuff is defined there.

rmannibucau: Well you go nowhere there since some of the brands i mzntionned were owned by the same guys and ultimately if you are reasoning like that maven is no limited to java so same state for me.

The cross plugin need was ack, no need to discuss it and you got the reason why it doesnt belong to core with an example that you would pull all the code of all plugins in core, even osgi for the same reason.
We just need it in a lib and find a neat way to change type/option mapping per plugin as explained on the list - same module.

desruisseaux: There is an official thing called _Java® tools specification_. Its content is the official standard according those who legally own the "Java" name for the programming language. Everything else is not Java. It may be called "OpenJDK", "Eclipse" or "Adoptium", but not Java. Whether the extra stuff is done by the same peoples or not is irrelevant to the _Java® tools specification_.

The _Java® tools specification_ is only taken as a base. Any library or plugin can define its own `PathType`. The _Java® tools specification_ is considered a base worth to be in shared code because some of its concepts, such as class-path, are shared by almost all tools. This is not the case of WAR, EAR, OSGi, etc.

We are not going to move plugins in core. We put there only the parts that are common to almost all Java tools, which is path management now, maybe `--add-exports` management later. Everything else stay in their plugins.

If you want Maven core to be language neutral and move all Java-specific stuff in a separated module shared by all Java plugins, I have no objection. But this is a separated debate (`JavaPathType` is not the only Java-specific thing currently in core).

rmannibucau: > There is an official thing called Java® tools specification

True but this is unrelated to this PR by the fact maven does not aim to target this scope but way more so question is do we put all we intend to support in core or not.
For me answer can only be no otherwise core would quickly be bloated and we already saw it does not scale in time (in terms of maintenance).
So overall you can reference any trademark but this stays out of topic for me, topic is what is the scope of core and it is the pipeline only for me and hopefully we'll be able to drop mvn 2 legacy remaining parts not respecting that historically.

> This is not the case of WAR, EAR, OSGi, etc.

Once again it is out of topic but there are a lot of shared parts there too so if you reasoning is right then we must embrace that too (it is default maven plugins too after all).

> If you want Maven core to be language neutral and move all Java-specific stuff in a separated module shared by all Java plugins, I have no objection.

Exactly.

> But this is a separated debate (JavaPathType is not the only Java-specific thing currently in core).

Not really, the point is that `JavaPathType` has not to exist in core but in `maven-java-tool-shared` module (whatever name it is). Core can stick to `String` since `option` doesnt always makes sense.

desruisseaux: It has never been proposed to put everything in core, but **only the base shared by practically all Java tools**. This base is defined by the _Java® tools specification_ (or a subset of it), not OSGi, not WAR, not EAR. If you want to add some extra paths shared by the latter, it can be discussed. But as starting point, the _Java® tools specification_ provides a standard definition of what a base is.

I have no objection in moving Java-specific stuff in a separated module. But it cannot be done as long as methods such as `isAddedToClasspath()` are in core, even if they are legacy (otherwise we cannot implement them as a redirection to new API). If those methods move, I will follow. But my impression is that Maven has close ties to Java, both in its API and inherited from `org.eclipse.aether` API.

api/maven-api-core/src/main/java/org/apache/maven/api/DependencyProperties.java
desruisseaux: Replaced `PathType[]` by `Set<PathType>`. It introduces a compiler warning (disabled with `@SuppressWarnings`), but only in a code that we control. So hopefully, this is okay.

api/maven-api-core/src/main/java/org/apache/maven/api/PathType.java
cstamas: > Not really, the point is that JavaPathType has not to exist in core but in maven-java-tool-shared module (whatever name it is). Core can stick to String since option doesnt always make sense.

Core _always had the "stuff" to describe WHAT you use (Artifact, ArtifactHandler)_. Plugins in turn _have the stuff/logic that knows HOW to use them_. This PR is to extend the vocabulary for "what" part. Your reasoning to me sounds like even Artifact should not be in core, and every plugin should define for itself what "artifact" is.

Java was always part of the core, so do not come now with some "shared java lang library" please :smile:  THAT is totally and utterly off topic for this PR (to make Maven fully and totally lang agnostic).

rmannibucau: @cstamas it is what it is, most of maven stuff related to java is inherited from a time maven was more bloated but I don't find a single point to pursue that path, even now, in particular cause each plugin needs its own configuration now (it was explained on the list) so more we put in core more we'll bring whatever the original will.

> only the base shared by practically all Java tools

Does not mean much sadly and is either a too small subset or a too big subset but here we don't reach any goal with something in between.

Just let move it to a shared module and move from there.

On my side I'm more than fine dropping `isAddedToClasspath` is maven 4 API, it does not make much sense with your PR (:tada:).

> THAT is totally and utterly off topic for this PR

fine with that while it does not end in any release, can be done after while only snapshots for me

gnodet:
@desruisseaux I've pushed a small rework of this PR on top of #1391 on my fork.
The squashed and rebased commit can be found at https://github.com/gnodet/maven/commit/c535260586d7eb34c56e836b8385f97ee6c3b9a0
The branch merged fork is https://github.com/gnodet/maven/commits/explicit-module-path

The main change is that I removed the `DependenciesProperties` in the other PR, so there are some adjustments which I'm not completely sure are working.... But the idea is that `Type` is extensible.  I think `PathType` should be extensible too, but I could not find any location where that would be needed (apart when creating custom `Type`s)....

desruisseaux:
Thanks, it look good to me. I will try to add a few JUnit tests today or tomorrow.

desruisseaux:
Added Javadoc on shortcut methods in `Session` interface. Sometime I had to guess, and there is questions previously flagged by `@todo` tag in b72d1ce. The test case gives the impression that the answer to those questions is that the dependency given in argument is included in the returned list, but it would be nice to said that in the javadoc if it is the case.

Added a convenience method:

```java
@Nonnull
Map<PathType, List<Path>> resolveDependencies(
        @Nonnull DependencyCoordinate dependencyCoordinate,
        @Nonnull PathScope scope,
        @Nonnull Collection<PathType> desiredTypes);
```

Expanded an existing test case: after fetching the dependency paths, fetch again the same dependency paths but this time dispatched by type.

* First invoke above convenience method with `Arrays.asList(JavaPathType.CLASSES, JavaPathType.MODULES)`:
  * The `pom` dependency is excluded, because it is placed neither on the class-path or module-path.
  * The `junit` dependency appears on the module-path because it has as `Automatic-Module-Name` attribute.
  * All other dependencies are on the class-path.
* Invoke the same method again, but this time with only `Arrays.asList(JavaPathType.CLASSES)`:
  * The `junit` dependency is now on the class-path.
  * This is the mechanism by which a plugin can tell that it is not interested in JPMS.

**Note:** A test if failing in the "Maven Core ITs suite" module:

```
Error: MavenITmng5135AggregatorDepResolutionModuleExtensionTest.testit:61-
AbstractMavenIntegrationTestCase.assertTrue:359 [test-classes, classes]
==> expected: <true> but was: <false>
```

I don't know what is happening here. A search for `AbstractMavenIntegrationTestCase` in my repository clone found nothing.

gnodet:
@desruisseaux fwiw, I've merge master branch into this branch at https://github.com/apache/maven/tree/explicit-module-path

desruisseaux:
Thanks. Closing this pull request in favour of #1401 then.

jira-importer:
Resolve #9864

PR #1380, opened 2024-01-13
Simplify graph

gnodet:
Simplify graph.
First step before making it concurrent if possible.


cstamas (approved):

gnodet:
> I dislike last two commits, we are again getting over the fence here. Keep it simple, but not super-duper-1brc-smart 😄

And it's actually slower. It was a failed experiment... I'll revert to the simple graph.

cstamas (approved):

PR #1383, opened 2024-01-17
Make ArtifactFilter into a Predicate<Artifact>

cstamas:


gnodet:
If they are needed, shouldn't all these classes be reimplemented on top of the v4 api instead ?

PR #1384, opened 2024-01-17
[MNG-8006] SPI to contribute to effective properties and more

cstamas:
Create _aether internal_ "extenders" components to serve several purposes:
* extender to meddle with exec request, for example to mediate between effective properties SPI (exec request extender)
* extender to meddle with prepped session and related bits, for example to move out deprecated code from session factory (session extender)

This PR implements M4 API SPI for contributing effective properties and exposes method for effective properties on `Session`.
* `Session` should expose properties service "ready" for interpolation, otherwise each interpolating code (plugin, etc) would need to construct the "effective" ones themselves, repeating same steps over and over again.
* SPI can contribute to user properties
* `Session` exposed properties are all immutable (sans effective, that is computed on each call).
* and lastly, a cleanup: all packages within `o.a.m.internal.aether` are now package protected, to tighten internal encapsulation. New core package `o.a.m.resolver` introduced that contains 2 publicly accessible classes.

---

https://issues.apache.org/jira/browse/MNG-8006
https://issues.apache.org/jira/browse/MNG-8021


maven-core/src/main/java/org/apache/maven/session/RepositorySystemSessionExtender.java
```diff
@@ -0,0 +1,39 @@
...
+/**
+ * Component able to "extend" session in some way.
+ *
+ * @since 4.0.0
+ */
+public interface RepositorySystemSessionExtender {
+    void extend(
+            MavenExecutionRequest mavenExecutionRequest,
+            SessionBuilder sessionBuilder,
+            MirrorSelector mirrorSelector,
```
gnodet, line 36: Those 3 last parameters looks really coming out of nowhere.
In particular, I think they can be accessed from the `sessionBuilder` as it has getters for those and they are set earlier, so I'd remove those 3 params.

gnodet (changes requested):

maven-core/src/main/java/org/apache/maven/session/RepositorySystemSessionExtender.java
cstamas: sessionBuilder does not have these, you told to remove them :smile:  so we are going back and forth... anyway, this is strictly "internal" component (see package), and let me push more.

cstamas:
Fixing the UT (changed the source of props) is now beyond my current capabls :smile: but @gnodet i think you see what the goal is: have some _internal_ stuff that mediates.

gnodet (commented):
I think we need an IT.  I have some doubts on the fact that properties set by an `EffectivePropertyContributor` will end up in the maven session and made available for project interpolation...
They are currently set in the repository session's `configProperties`, but I don't think they are pushed up to user properties and in the maven session.

api/maven-api-core/src/main/java/org/apache/maven/api/Project.java
```diff
@@ -133,4 +134,10 @@ default String getId() {
 
     @Nonnull
     List<RemoteRepository> getRemotePluginRepositories();
+
+    /**
+     * Returns the effective project properties.
+     */
+    @Nonnull
+    Map<String, String> getProperties();
```
gnodet, line 142: I don't think introducing a mutable aspect here is a good idea.  This map should be made immutable imho.
Modifying project properties can be done using the `ProjectManager`.

api/maven-api-core/src/main/java/org/apache/maven/api/Session.java
```diff
@@ -73,15 +74,12 @@ public interface Session {
...
-     * <p>
-     * These properties are "effective" in a way they are built obeying all "user overrides" and others.
+     * Gets the effective properties to use for interpolation.
      *
-     * @return the config properties, never {@code null}
+     * @return the effective properties, never {@code null}
      */
     @Nonnull
-    Map<String, Object> getEffectiveProperties();
+    Map<String, String> getEffectiveProperties(@Nullable Project project);
```
gnodet, line 83: We should specify if this map is mutable or not, and if mutable, if a new map is created each time.

cstamas: Just did that with last two "javadoc" commit: ... each call returns "new copy" of ...

api/maven-api-core/src/main/java/org/apache/maven/api/services/Properties.java
```diff
@@ -0,0 +1,43 @@
...
+import org.apache.maven.api.annotations.Nullable;
+
+/**
+ * @since 4.0.0
+ */
+@Experimental
+public interface Properties extends Service {
+
+    /**
+     * Creates a {@link org.apache.maven.api.Project} from a POM file.
```
gnodet: Irrelevant copy/paste

api/maven-api-core/src/main/java/org/apache/maven/api/Project.java
gnodet: Ah, it's already immutable.  It should be written in the javadoc.  And maybe point to `ProjectManager#setProperty` ?

api/maven-api-core/src/main/java/org/apache/maven/api/services/Properties.java
cstamas: you did not refresh? You create comments that are outdated from start :smile:

api/maven-api-core/src/main/java/org/apache/maven/api/Session.java
```diff
@@ -73,15 +74,12 @@ public interface Session {
...
 
     /**
-     * Gets the immutable effective properties to use for interpolation. The effective properties are collected from
-     * {@link #getSystemProperties()} and {@link #getUserProperties()} in this order, and more.
-     * <p>
-     * These properties are "effective" in a way they are built obeying all "user overrides" and others.
+     * Gets the effective properties to use for interpolation.
      *
-     * @return the config properties, never {@code null}
+     * @return the effective properties, never {@code null}
```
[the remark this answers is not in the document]
cstamas: Added

api/maven-api-core/src/main/java/org/apache/maven/api/Session.java
```diff
@@ -55,23 +56,32 @@ public interface Session {
...
+     * Returns immutable system properties to use for interpolation. The system properties are collected from the
+     * runtime environment such as {@link System#getProperties()} and environment variables.
      *
      * @return the system properties, never {@code null}
      */
     @Nonnull
     Map<String, String> getSystemProperties();
 
+    /**
+     * Each invocation computes a new map of effective properties. To be used in interpolation.
```
gnodet, line 77: I think we need to explain here what "effective properties" are.

cstamas: Done

gnodet (approved):

jira-importer:
Resolve #10831

PR #1385, opened 2024-01-18
[MNG-8005] Fix workspace reader drop bug

cstamas:
Maven4 seems sets but then drops/overrides some workspace readers. Partially due hazy borders who or what manages them, as role is split between session factory and session factory caller.

Changes:
* session factory does NOTHING re workspace readers, it becomes fully the caller duty
* two spots calling session factory (default maven, extension bootstrap) are now fully in charge to properly setup workspace readers

Co-authored-by: Jonas Rutishauser <jonas.rutishauser@alumni.ethz.ch>

---

https://issues.apache.org/jira/browse/MNG-8005

### DIFF
--- a/maven-embedder/src/main/java/org/apache/maven/cli/internal/BootstrapCoreExtensionManager.java
+++ b/maven-embedder/src/main/java/org/apache/maven/cli/internal/BootstrapCoreExtensionManager.java
@@ -100,16 +100,9 @@ public class BootstrapCoreExtensionManager {
     public List<CoreExtensionEntry> loadCoreExtensions(
             MavenExecutionRequest request, Set<String> providedArtifacts, List<CoreExtension> extensions)
             throws Exception {
-        MavenChainedWorkspaceReader chainedWorkspaceReader = new MavenChainedWorkspaceReader();
-        if (request.getWorkspaceReader() != null) {
-            chainedWorkspaceReader.addReader(request.getWorkspaceReader());
-        }
-        if (ideWorkspaceReader != null) {
-            chainedWorkspaceReader.addReader(ideWorkspaceReader);
-        }
         try (CloseableSession repoSession = repositorySystemSessionFactory
                 .newRepositorySessionBuilder(request)
-                .setWorkspaceReader(chainedWorkspaceReader)
+                .setWorkspaceReader(new MavenChainedWorkspaceReader(request.getWorkspaceReader(), ideWorkspaceReader))
                 .build()) {
             List<RemoteRepository> repositories = RepositoryUtils.toRepos(request.getPluginArtifactRepositories());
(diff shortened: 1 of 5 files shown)


maven-embedder/src/main/java/org/apache/maven/cli/internal/BootstrapCoreExtensionManager.java
```diff
@@ -77,24 +79,29 @@ public class BootstrapCoreExtensionManager {
...
         this.parentRealm = container.getContainerRealm();
+        this.ideWorkspaceReader = ideWorkspaceReader;
     }
 
     public List<CoreExtensionEntry> loadCoreExtensions(
             MavenExecutionRequest request, Set<String> providedArtifacts, List<CoreExtension> extensions)
             throws Exception {
         try (CloseableSession repoSession = repositorySystemSessionFactory
                 .newRepositorySessionBuilder(request)
+                .setWorkspaceReader(ideWorkspaceReader)
```
gnodet, line 104: Won't that override the existing chained workspace reader ? is that the intent ?

cstamas: With this PR session factory does NOT set any workspace reader anymore (read PR description and see code), it is _caller duty_

gnodet: Yes, I don't question the change or responsibilities, but a possible change in behaviour.  
The caller is `BootstrapCoreExtensionManager` and now responsible to set the workspace reader on the session.  What I'm wondering is if this leads back to the same behaviour... I think before this PR, the workspace reader is to set a chained one, now, it's only the `ide` one...

cstamas: Oh, you mean IF request has one as well? Hm, let me recheck is that possible...

cstamas: yes, clearly IS possible (as copy of "origina" request is used), fixing...

cstamas: Fixed.

But your comment is wrong: session factory _never used_ chained WSReader, it was ONLY the default maven using it. Before this PR, the session could be inited with request WS (if non null) OR the ide WS (that may be null).

cstamas: Fixed and also simplified, now chained WS omits possible null entries in collection, thus all these become one liner.

gnodet (approved):

PR #1386, opened 2024-01-19
[MNG-8014] Fix multithreaded builder

gnodet:


maven-xml-impl/src/main/java/org/apache/maven/internal/xml/ImmutableCollections.java
```diff
@@ -0,0 +1,601 @@
...
+package org.apache.maven.internal.xml;
+
+import java.io.Serializable;
+import java.util.*;
+import java.util.function.BiFunction;
+import java.util.function.Function;
+import java.util.function.Predicate;
+import java.util.function.UnaryOperator;
+
+class ImmutableCollections {
```
cstamas, line 28: why? Why over Collections?

cstamas (approved):
LGTM

maven-xml-impl/src/main/java/org/apache/maven/internal/xml/ImmutableCollections.java
gnodet: The same classes are used in the immutable classes generated from the modello schemas.
Those are similar to the new `Map.of()` and `List.of()` in recent JDKs.
The main benefit is that there's no copy if you ask for an immutable collection if you pass an already immutable collection (this list is always wrapped in JDK 8) and there's no safe copy, i.e. the unmodifiable collection is just a wrapper on top of the collection, so it's still can be modified, so one needs to do a copy first.
Bottom line, the current code is actually using `Collections.unmodifiableMap(new HashMap<>(attributes))` which is quite expensive.

If we were to raise the runtime requirement to JDK 9, I'd happily get rid of those.

jira-importer:
Resolve #10839

PR #1388, opened 2024-01-21
[MNG-8024] Make WrapperProperties and WrapperList serizalizable.

cstamas:
This solves the two, but there are more...

---

https://issues.apache.org/jira/browse/MNG-8024

### DIFF
--- a/src/mdo/java/WrapperList.java
+++ b/src/mdo/java/WrapperList.java
@@ -27,16 +27,21 @@ import java.util.function.Function;
 import java.util.function.Supplier;
 
 class WrapperList<T, U> extends AbstractList<T> {
-    private final Supplier<List<U>> getter;
-    private final Consumer<List<U>> setter;
-    private final Function<U, T> mapper;
-    private final Function<T, U> revMapper;
+    public interface SerializableConsumer<T> extends java.util.function.Consumer<T>, java.io.Serializable {}
+    public interface SerializableFunction<U, T> extends java.util.function.Function<U, T>, java.io.Serializable {}
+    public interface SerializableSupplier<T> extends java.util.function.Supplier<T>, java.io.Serializable {}
 
-    WrapperList(List<U> list, Function<U, T> mapper, Function<T, U> revMapper) {
+
+    private final SerializableSupplier<List<U>> getter;
+    private final SerializableConsumer<List<U>> setter;
+    private final SerializableFunction<U, T> mapper;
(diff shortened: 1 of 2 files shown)


gnodet:
We need to extend the unit test https://github.com/apache/maven/blob/master/maven-model/src/test/java/org/apache/maven/model/SerializationTest.java to exercise the actual problem.

src/mdo/java/WrapperList.java
```diff
@@ -27,16 +27,21 @@
 import java.util.function.Supplier;
 
 class WrapperList<T, U> extends AbstractList<T> {
```
gnodet, line 29: What's the point is the `WrapperList` itself does not implement `Serializable` ?

gnodet (changes requested):
We need a clear unit test.  Basic serialisation is already supported, so something's wrong...

gnodet:
> We need a clear unit test. Basic serialisation is already supported, so something's wrong... The `WrapperList` and `WrapperProperties` should not actually be used during serialisation, as those objects are not present in fields.

I wonder if the problem is when serializing part of the model instead of the whole `Model`.

cstamas:
Yes, that is exactly what is happening, the incremental build basically taks "bits" of model and would like to serialize those with DataOutputStream...

gnodet:
> Yes, that is exactly what is happening, the incremental build basically taks "bits" of model and would like to serialize those with DataOutputStream...

I'm not sure we should support this use case, as I think the whole model will be serialised anyway.  The lambdas in the list/properties objects do retain a reference to the parent up to the Model.  This is needed in order to implement the modification methods while keeping the original reference.

So while this may actually make things work, I'm not completely sure we should do it...  It may be easier to simplify check if there's a `getDelegate()` method, and if so, serialize that object instead.

gnodet:
> > Yes, that is exactly what is happening, the incremental build basically taks "bits" of model and would like to serialize those with DataOutputStream...
> 
> I'm not sure we should support this use case, as I think the whole model will be serialised anyway. The lambdas in the list/properties objects do retain a reference to the parent up to the Model. This is needed in order to implement the modification methods while keeping the original reference.
> 
> So while this may actually make things work, I'm not completely sure we should do it... It may be easier to simplify check if there's a `getDelegate()` method, and if so, serialize that object instead.

Another way would be to generate writeObject / readObject methods that will actually delegate to... the delegates from the v4 model.

PR #1389, opened 2024-01-23
[MNG-8037] Restrict project to the entity being built and make it immutable

gnodet:
Follow-up to https://github.com/apache/maven/pull/1387
JIRA issue: https://issues.apache.org/jira/browse/MNG-8037

### DIFF
--- a/api/maven-api-core/src/main/java/org/apache/maven/api/Project.java
+++ b/api/maven-api-core/src/main/java/org/apache/maven/api/Project.java
@@ -34,13 +34,11 @@ import org.apache.maven.api.model.Model;
  * need to do so using the {@link org.apache.maven.api.services.ProjectManager}
  * service.
  * <p>
- * Projects are created using the {@code ProjectBuilder} either from a POM file
- * (usually named {@code pom.xml}) on the file system or by loading the POM
- * associated to an artifact coordinate in a repository.  In the first case,
- * the {@link #getPomPath()} will point to the POM file and the
+ * Projects are created using the {@code ProjectBuilder} from a POM file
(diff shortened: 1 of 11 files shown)


gnodet:
> Just FTR as we discussed:
> 
> * IMHO baseDir != "pom file parent"
> * especially as a project may (could) have baseDir but no POM file (but have model provided)
> * possible use case: generate some POM (ie is in tmp directory), create a Project that is "placed" on top of some existing module, and have it built?

So we're looking at Maven being used as a library, right ? That's clearly not supported from the CLI.
Right, now, in the Maven 4 API (let's limit to that use case) the only way to obtain a `Project` is by using the `ProjectBuilder` and the only way to use it is to give either a `Path`, `Source`, `Artifact` or `ArtifactCoordinate`.
My understanding (it closely maps the v3 api) is that it you pass an ArtifactCoordinate (or Artifact), that artifact is resolved (it must point to a POM), then loaded.  I need to double check, but this results in a `Project` which has no POM file and no BaseDir.
I think that's the current situation, i.e.: either you have both or none, but it's linked.  The POM file cannot point to the local repository, it's *always* a build POM in your project file system, and it always define the base directory.

I'll also double check the `DefaultMavenProjectBuilder`.  But I'm sure the use case you describe is currently supported.

gnodet:
@cstamas I've removed the mutable bits from the `Project`, removed the optionals on `pomPath` and `basedir`.  I think we'll miss a `ModelBuilder` service to get effective models for artifacts, but I'll add those in a separate PR.

cstamas (approved):

PR #1392, opened 2024-01-25
[MNG-8026] Maven drives regarding scopes

cstamas:
Maven should be fully in control regarding dependency scope definitions and their interpretations (and transformations). 
This PR makes following changes:
* no change in behaviour (all works as before), but
* maven codebase contains now all the code related to scopes
* no resolver code is in use that was deprecated (as it was wrong to have it in resolver), see issue [MRESOLVER-471](https://issues.apache.org/jira/browse/MRESOLVER-471)
* this also stops scope definitions duplication as well, they are now defined in one single project: Maven

---

https://issues.apache.org/jira/browse/MNG-8026


gnodet (approved):

jira-importer:
Resolve #10871

PR #1393, opened 2024-01-25
[MNG-7954] New dependency injection mechanism

gnodet:
JIRA issue: https://issues.apache.org/jira/browse/MNG-7954

This PR provides a minimalistic dependency injection library (the code is initially based on the ActiveJ project).
This provides
 * our own annotations for the user facing API in [`org.apache.maven.api.di` package](https://github.com/gnodet/maven/tree/dependency-injection/api/maven-api-di/src/main/java/org/apache/maven/api/di)
 * no dependency (especially ASM)
 * a simple programmatic access (not to be exposed, but used to actually populate the bean container) in [`org.apache.maven.di` package](https://github.com/gnodet/maven/tree/dependency-injection/maven-di/src/main/java/org/apache/maven/di)
 * an [index creation using APT](https://github.com/gnodet/maven/blob/dependency-injection/maven-di/src/main/java/org/apache/maven/di/processor/IndexAnnotationProcessor.java)
 
This is currently plugged into the plugin creation for plugins using the Maven 4 API.
It's currently not used for extensions or for the core maven components.  This would require extra support and a plexus compatibility layer (which is kinda what we want to get rid of).

### DIFF
--- a/api/maven-api-di/src/main/java/org/apache/maven/api/di/Scope.java
+++ b/api/maven-api-di/src/main/java/org/apache/maven/api/di/Scope.java
@@ -22,20 +22,10 @@ import java.lang.annotation.Documented;
 import java.lang.annotation.Retention;
 import java.lang.annotation.Target;
 
-import static java.lang.annotation.ElementType.METHOD;
-import static java.lang.annotation.ElementType.TYPE;
+import static java.lang.annotation.ElementType.ANNOTATION_TYPE;
 import static java.lang.annotation.RetentionPolicy.RUNTIME;
 
-/**
- * Indicates that the annotated bean has a lifespan limited to a given mojo execution,
- * which means each mojo execution will result in a different instance being injected.
- *
- * TODO: this is currently not implemented
- *
- * @since 4.0.0
- */
-@Scope
-@Documented
+@Target(ANNOTATION_TYPE)
 @Retention(RUNTIME)
-@Target({TYPE, METHOD})
-public @interface MojoExecutionScoped {}
+@Documented
+public @interface Scope {}
--- a/maven-di/src/main/java/org/apache/maven/di/Injector.java
+++ b/maven-di/src/main/java/org/apache/maven/di/Injector.java
@@ -38,6 +38,8 @@ public interface Injector {
 
     <T> Injector bindInstance(Class<T> cls, T instance);
 
+    <T> Injector bindInstanceAndInject(Class<T> cls, T instance);
+
     //
     // Bean access
     //
--- a/maven-di/src/main/java/org/apache/maven/di/Injector.java
+++ b/maven-di/src/main/java/org/apache/maven/di/Injector.java
@@ -16,26 +16,33 @@
  * specific language governing permissions and limitations
  * under the License.
  */
-package org.apache.maven.api.di;
+package org.apache.maven.di;
 
-import java.lang.annotation.Documented;
-import java.lang.annotation.Retention;
-import java.lang.annotation.Target;
+import java.lang.annotation.Annotation;
 
-import static java.lang.annotation.ElementType.METHOD;
-import static java.lang.annotation.ElementType.TYPE;
-import static java.lang.annotation.RetentionPolicy.RUNTIME;
+import org.apache.maven.di.impl.InjectorImpl;
 
-/**
- * Indicates that the annotated bean has a lifespan limited to a given mojo execution,
- * which means each mojo execution will result in a different instance being injected.
- *
- * TODO: this is currently not implemented
- *
- * @since 4.0.0
- */
-@Scope
-@Documented
-@Retention(RUNTIME)
-@Target({TYPE, METHOD})
-public @interface MojoExecutionScoped {}
+public interface Injector {
+
+    //
+    // Builder API
+    //
+
+    static Injector create() {
+        return new InjectorImpl();
+    }
+
+    Injector bindScope(Class<? extends Annotation> scopeAnnotation, Scope scope);
+
+    Injector bindImplicit(Class<?> cls);
+
+    <T> Injector bindInstance(Class<T> cls, T instance);
+
+    //
+    // Bean access
+    //
+
+    <T> T getInstance(Class<T> key);
+
+    <T> T getInstance(Key<T> key);
+}
--- a/maven-di/src/main/java/org/apache/maven/di/Injector.java
+++ b/maven-di/src/main/java/org/apache/maven/di/Injector.java
@@ -38,11 +38,11 @@ public interface Injector {
 
     <T> Injector bindInstance(Class<T> cls, T instance);
 
-    <T> Injector bindInstanceAndInject(Class<T> cls, T instance);
-
     //
     // Bean access
     //
+
+    <T> void injectInstance(T instance);
 
     <T> T getInstance(Class<T> key);
 
--- a/api/maven-api-core/src/main/java/org/apache/maven/api/di/SessionScoped.java
+++ b/api/maven-api-core/src/main/java/org/apache/maven/api/di/SessionScoped.java
@@ -22,8 +22,6 @@ import java.lang.annotation.Documented;
 import java.lang.annotation.Retention;
 import java.lang.annotation.Target;
 
-import jakarta.inject.Scope;
-
 import static java.lang.annotation.ElementType.METHOD;
 import static java.lang.annotation.ElementType.TYPE;
 import static java.lang.annotation.RetentionPolicy.RUNTIME;
@@ -31,6 +29,8 @@ import static java.lang.annotation.RetentionPolicy.RUNTIME;
 /**
  * Indicates that annotated component should be instantiated before session execution starts
  * and discarded after session execution completes.
+ *
+ * TODO: this is currently not implemented
  *
  * @since 4.0.0
  */
--- a/api/maven-api-core/src/main/java/org/apache/maven/api/plugin/annotations/Execute.java
+++ b/api/maven-api-core/src/main/java/org/apache/maven/api/plugin/annotations/Execute.java
@@ -35,7 +35,7 @@ import org.apache.maven.api.annotations.Nonnull;
  */
 @Experimental
 @Documented
-@Retention(RetentionPolicy.CLASS)
+@Retention(RetentionPolicy.RUNTIME)
 @Target(ElementType.TYPE)
 @Inherited
 public @interface Execute {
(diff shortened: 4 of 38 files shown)


api/maven-api-di/src/main/java/org/apache/maven/api/di/Scope.java
```diff
@@ -0,0 +1,33 @@
...
+import java.lang.annotation.Target;
+
+import static java.lang.annotation.ElementType.ANNOTATION_TYPE;
+import static java.lang.annotation.RetentionPolicy.RUNTIME;
+
+@Target(ANNOTATION_TYPE)
+@Retention(RUNTIME)
+@Documented
+public @interface Scope {
+    boolean threadsafe() default true;
```
rmannibucau, line 32: Needed?

api/maven-api-di/src/main/java/org/apache/maven/api/di/Transient.java
```diff
@@ -0,0 +1,31 @@
...
+import java.lang.annotation.Retention;
+import java.lang.annotation.Target;
+
+import static java.lang.annotation.ElementType.METHOD;
+import static java.lang.annotation.RetentionPolicy.RUNTIME;
+
+@Target(METHOD)
+@Retention(RUNTIME)
+@Documented
+public @interface Transient {}
```
rmannibucau, line 31: Needed? why not use a scope (prototype, default, whatever name)

maven-di/src/main/java/org/apache/maven/api/di/ProvidesIntoSet.java
```diff
@@ -0,0 +1,38 @@
...
+
+/**
+ * This is a built-in shortcut for {@link Provides provider method} that provides its result as a singleton set
+ * and adds a {@link Multibinders#toSet() set multibinder} for the provided set key to the module.
+ *
+ * @see AbstractModule
+ */
+@Target(METHOD)
+@Retention(RUNTIME)
+public @interface ProvidesIntoSet {}
```
rmannibucau, line 38: guice leak? not sure we need this one

maven-di/src/main/java/org/apache/maven/api/di/Scopes.java
```diff
@@ -0,0 +1,38 @@
...
+
+/**
+ * This is a special annotation that allows you to declare bindings in nested scopes
+ * using {@link Provides provider methods}.
+ * <p>
+ * Note that it does not allow you to use stateful scope annotations.
+ */
+@Retention(RUNTIME)
+@Target(ElementType.METHOD)
+public @interface Scopes {
```
rmannibucau, line 36: needed?

maven-di/src/main/java/org/apache/maven/di/Injector.java
```diff
@@ -0,0 +1,613 @@
...
+
+    final AtomicReferenceArray[] scopeCaches;
+
+    private static Supplier<UnaryOperator<CompiledBinding<?>>> bytecodePostprocessorFactory = UnaryOperator::identity;
+
+    /**
+     * Enables specialization of compiled bindings. <b>Depends on {@code ActiveJ-Specializer} module</b>
+     */
+    @SuppressWarnings("unchecked")
+    public static void useSpecializer() {
```
rmannibucau, line 94: do we use it? if so we should either require it (no reflection) or drop it no?

maven-di/src/main/java/org/apache/maven/di/ResourceLocator.java
```diff
@@ -0,0 +1,58 @@
...
+import org.apache.maven.api.annotations.Nullable;
+
+public interface ResourceLocator {
+    <T> T getInstance(Key<T> key);
+
+    default <T> T getInstance(Class<T> type) {
+        return getInstance(Key.of(type));
+    }
+
+    <T> @Nullable T getInstanceOrNull(Key<T> key);
```
rmannibucau, line 32: guess all the next getInstances can be replaced by getOptionalInstance and the cases be managed by `Optional` directly instead of duplicating API?

maven-di/src/main/java/org/apache/maven/di/binding/Bindings.java
```diff
@@ -0,0 +1,609 @@
...
+                    : new CompiledBinding<R>() {
+                        @Override
+                        public R getInstance(AtomicReferenceArray[] scopedInstances, int synchronizedScope) {
+                            return constructor.create();
+                        }
+                    };
+        }
+    }
+
+    public static class BindingToConstructor1<R, T1> extends Binding<R> {
```
rmannibucau, line 171: do we need all these flavors?

maven-di/src/main/java/org/apache/maven/di/impl/AbstractCompiledBinding.java
```diff
@@ -0,0 +1,60 @@
...
+
+    @SuppressWarnings("unchecked")
+    @Override
+    public final R getInstance(AtomicReferenceArray[] scopedInstances, int synchronizedScope) {
+        AtomicReferenceArray array = scopedInstances[scope];
+        R instance = (R) array.get(index);
+        if (instance != null) {
+            return instance;
+        }
+        if (synchronizedScope == scope) {
```
rmannibucau, line 42: im tempted to drop this notion and just always lock, wdyt?

maven-di/src/main/java/org/apache/maven/di/module/UniqueQualifierImpl.java
```diff
@@ -0,0 +1,50 @@
...
+ * under the License.
+ */
+package org.apache.maven.di.module;
+
+import java.lang.annotation.Annotation;
+
+import org.apache.maven.api.annotations.Nullable;
+
+@SuppressWarnings("ClassExplicitlyAnnotation")
+public final class UniqueQualifierImpl implements UniqueQualifier {
```
rmannibucau, line 26: cant we reuse it as binding too? (a bit like cdi does with annotation literals)

rmannibucau (commented):
No strong issue for me but I'd like we review these points before merging then if it is ok just go:

* do we want to copy all part of guice, guess in maven we just need a subtype
* do we need bindinset notion, guess we can just lookup sets from a type now
* do we need all these flavors of structures (tuple<N> for ex), using a list wouldnt be dramatic for us I think
* do we want all these notions in modules/binders

overall idea is to drop all we dont need from guice if possible to maintain the minimum we need

hope it makes sense

gnodet:
> No strong issue for me but I'd like we review these points before merging then if it is ok just go:
> 
> * do we want to copy all part of guice, guess in maven we just need a subtype

There's no guice dependency here.

> * do we need bindinset notion, guess we can just lookup sets from a type now
> * do we need all these flavors of structures (tuple for ex), using a list wouldnt be dramatic for us I think
> * do we want all these notions in modules/binders 
> overall idea is to drop all we dont need from guice if possible to maintain the minimum we need

Agreed.  There are a bunch of things that are unneeded to us.  That's because the library has been designed with programmatic definition in mind (using the ModuleBuilder DSL) while we need a declarative definition. So I do plan to trim down to what we really need. 
But that's just a first draft and I need to try and plug this DI engine into plugin loading mechanism and see how it goes.  If I can port a few core plugins to this DI engine, I'll then start trimming it down.

api/maven-api-di/src/main/java/org/apache/maven/api/di/Scope.java
gnodet: I don't think so, I plan to remove it.

maven-di/src/main/java/org/apache/maven/api/di/ProvidesIntoSet.java
gnodet: I don't think. Planned for removal, that's why it has not been added to `api/maven-api-di` for now.

maven-di/src/main/java/org/apache/maven/api/di/Scopes.java
gnodet: No, same as above.

maven-di/src/main/java/org/apache/maven/di/Injector.java
gnodet: I don't think we need, so that will be removed.

rmannibucau:
> There's no guice dependency here.

Well some part of the code were highly inspired from Guice, this is what I meant.

> So I do plan to trim down to what we really need.

Perfect for me. Thanks a lot!

api/maven-api-di/src/main/java/org/apache/maven/api/di/Transient.java
cstamas: I believe this is true to Maven Core components, as for example Resolver components could not be (nor should be due model) session scoped...

cstamas:
Am just worried that we end up again as we did with old plexus: was our, but began to rot from one point... I'd be happies if we could simply circumvent this by something like https://github.com/eclipse/sisu.inject/issues/103

gnodet:
> Am just worried that we end up again as we did with old plexus: was our, but began to rot from one point... I'd be happies if we could simply circumvent this by something like [eclipse/sisu.inject#103](https://github.com/eclipse/sisu.inject/issues/103)

Feel free to give it a try, but I could not find an easy way without rewriting everything...

cstamas:
Do we want something like `@Description` is? Or it was added to Sisu only as Plexus XML had "component description"? In fact, am unaware of any use of it, but still...

rmannibucau:
Think we should start minimalistic, maven IoC need is actually pretty low - mainly bean definition + overriding - so less we keep better it is in terms of maintenance.
Rewriting from scratch gives the opportunity to drop guice and sisu at some point too so can be quite promishing if we manage to make plexus-world optional (realm = TCCL) but this is likely mvn 6 or 7 ;).

gnodet:
> Think we should start minimalistic, maven IoC need is actually pretty low - mainly bean definition + overriding - so less we keep better it is in terms of maintenance. Rewriting from scratch gives the opportunity to drop guice and sisu at some point too so can be quite promishing if we manage to make plexus-world optional (realm = TCCL) but this is likely mvn 6 or 7 ;).

Yeah, I think I'll even drop support for scopes for now, we don't really need it for plugins.  Those are only needed when dealing with extensions or maven-core, as plugins are always bound to the mojo execution.

cstamas:
Unless, as you mention, is an extension... and we do have quite some of those.

rmannibucau:
Yes think extensions are in scope there and don't think scope is the biggest code (maybe guice made it verbose but the CDI way is actually quite light - `Context` usage in subclass proxies which can be done at build time, even if it is just dumping asm one at build time and not using an annot proc or sthg more build oriented). But if we want to split we can maybe start by param injection only.

cstamas (approved):
LGTM, nothing to add really

rmannibucau (approved):

PR #1394, opened 2024-01-29
[MNG-8035] Update to Resolver 2.0.0-alpha-7

cstamas:
Update to Resolver 2.0.0-alpha-7, sort out TODO.

This is also one simple change "sneaked in" for RAT, to ignore local `.mvn` that is unrelated to all this, but does not make any harm. In turn, it allows dev local checkout to have various stuff set without
decorating those with headers or fighing RAT plugin.

---

https://issues.apache.org/jira/browse/MNG-8035


gnodet (approved):

jira-importer:
Resolve #9834

PR #1395, opened 2024-01-29
[MNG-8029] improve documentation about mirror settings

hboutemy:
https://issues.apache.org/jira/browse/MNG-8029

### DIFF
--- a/maven-settings/src/main/mdo/settings.mdo
+++ b/maven-settings/src/main/mdo/settings.mdo
@@ -28,9 +28,9 @@
   <description>
     <![CDATA[
     <p>This is a reference for the user-specific configuration for Maven.</p>
-    <p>Includes things that should not be distributed with the pom.xml file, such as developer identity, along with
+    <p>Includes things that should not be distributed with the <code>pom.xml</code> file, such as developer identity, along with
     local settings, like proxy information.</p>
-    <p>The default location for the settings file is <i>~/.m2/settings.xml</i></p>
+    <p>The default location for the user settings file is <code>~/.m2/settings.xml</code></p>
     ]]>
   </description>
   <defaults>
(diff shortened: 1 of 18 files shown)


maven-settings/src/main/mdo/settings.mdo
```diff
@@ -621,16 +626,24 @@
           <version>1.1.0+</version>
           <type>String</type>
           <defaultValue>default</defaultValue>
-          <description>The layout of the mirror repository. Since Maven 3.</description>
+          <description>
+            <![CDATA[
+            The layout of the mirror repository.
+            <br><b>Since</b>: Maven 3
```
gnodet, line 632: Why not using `@since Maven 3` instead ? (same for the other `<br><b>Since</b>: Maven xxx` pattern)

hboutemy: because I just reused the existing pattern
I don't know if `@since` works with Modello xdoc, need to check: doing the pattern change in every `.mdo` could be a good thing for another PR

gnodet: It clearly does not render very well. It would be nice if modello could be extended to understand javadoc syntax, as it's already used and is not displayed really nicely:
  https://maven.apache.org/ref/4.0.0-alpha-12/maven-settings/settings.html#proxy

gnodet: Maybe using a small library such as https://github.com/chhorz/javadoc-parser ?

gnodet: https://github.com/codehaus-plexus/modello/pull/414

Renders as 
image
 instead of 
image

hboutemy: I'll merge this PR with the existing pattern
dropping the existing pattern to replace by better option is nice, but it's not the current objective: I'll be happy that the content is changed later

jira-importer:
Resolve #10827

PR #1396, opened 2024-02-01
[MNG-7906] Add unit test

gnodet:
This PR currently provides 3 unit tests for [MNG-7906](https://issues.apache.org/jira/browse/MNG-7906).
The first two tests (`testManagedDependencyBeforeImport` and `testManagedDependencyAfterImport`) work in an expected way, i.e. the managed dependency defined in the POM is the one used.
The third one (`testManagedDependencyTwoImports`) is the problematic one, showing that the algorithm used to resolve conflicting managed dependencies is almost (apart from the two first tests) a "first wins" and not a "last wins" or even better a "closest then last wins".

It was proposed to add a WARNING when such conflict happens.


gnodet:
I've added the improved unit tests to #1370 instead.

jira-importer:
Resolve #9009

PR #1398, opened 2024-02-05
[MNG-8041] A quick draft about problem

cstamas:
This is just a attempt to fix reproducer, and to see what ITs have to say.

---

https://issues.apache.org/jira/browse/MNG-8041

### DIFF
--- a/maven-core/src/main/java/org/apache/maven/lifecycle/internal/LifecycleDependencyResolver.java
+++ b/maven-core/src/main/java/org/apache/maven/lifecycle/internal/LifecycleDependencyResolver.java
@@ -240,8 +240,11 @@ public class LifecycleDependencyResolver {
         scopesToCollect = new HashSet<>(scopesToCollect);
         scopesToCollect.addAll(scopesToResolve);
 
-        DependencyFilter collectionFilter = new ScopeDependencyFilter(null, negate(scopesToCollect));
-        DependencyFilter resolutionFilter = new ScopeDependencyFilter(null, negate(scopesToResolve));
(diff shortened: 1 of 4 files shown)


maven-core/src/main/java/org/apache/maven/lifecycle/internal/LifecycleDependencyResolver.java
```diff
@@ -240,8 +240,11 @@ private Set<Artifact> getDependencies(
         scopesToCollect = new HashSet<>(scopesToCollect);
         scopesToCollect.addAll(scopesToResolve);
 
-        DependencyFilter collectionFilter = new ScopeDependencyFilter(null, negate(scopesToCollect));
-        DependencyFilter resolutionFilter = new ScopeDependencyFilter(null, negate(scopesToResolve));
+        Collection<String> collectScopesToExclude = negate(scopesToCollect);
+        Collection<String> resolveScopesToExclude = negate(scopesToResolve);
+
+        DependencyFilter collectionFilter = new ScopeDependencyFilter(null, collectScopesToExclude);
+        DependencyFilter resolutionFilter = new ScopeDependencyFilter(null, resolveScopesToExclude);
```
gnodet, line 247: can we just use the scopes to collect / resolve instead of the ones to exclude and give them as a negative test in the filter ?
i.e.:
```
DependencyFilter collectionFilter = new ScopeDependencyFilter(scopesToCollect);
DependencyFilter resolutionFilter = new ScopeDependencyFilter(scopesToResolve);
```

jira-importer:
Resolve #10874

PR #1399, opened 2024-02-06
[MNG-8043] Dependency properties should be provided by Maven

cstamas:
Followup of MRESOLVER-484:
* move off deprecated resolver bits
* introduce non-deprecated replacements
* move existing to code to those new bits

---

https://issues.apache.org/jira/browse/MNG-8043


api/maven-api-core/src/main/java/org/apache/maven/api/DependencyScope.java
```diff
@@ -92,7 +92,7 @@ public enum DependencyScope {
      * System scope.
      * <p>
      * Important: this scope {@code id} MUST BE KEPT in sync with label in
-     * {@code org.eclipse.aether.util.artifact.Scopes#SYSTEM}.
+     * {@code org.eclipse.aether.util.artifact.DependencyScopes#SYSTEM}.
```
gnodet, line 95: we should add a provider spi in resolver, to avoid this "must be KEPT in sync" ...

cstamas: https://issues.apache.org/jira/browse/MRESOLVER-491

maven-resolver-provider/src/main/java/org/apache/maven/repository/internal/type/DefaultType.java
```diff
@@ -95,8 +101,8 @@ public Map<String, String> getProperties() {
         Map<String, String> properties = new HashMap<>();
         properties.put(ArtifactProperties.TYPE, this.id);
         properties.put(ArtifactProperties.LANGUAGE, this.language.id());
-        properties.put(ArtifactProperties.INCLUDES_DEPENDENCIES, String.valueOf(includesDependencies));
-        properties.put(ArtifactProperties.CONSTITUTES_BUILD_PATH, String.valueOf(buildPathConstituent));
+        properties.put(MavenArtifactProperties.INCLUDES_DEPENDENCIES, String.valueOf(includesDependencies));
+        properties.put(MavenArtifactProperties.CONSTITUTES_BUILD_PATH, String.valueOf(buildPathConstituent));
         return Collections.unmodifiableMap(properties);
```
gnodet, line 106: This map is immutable and could be computed in the `DefaultType` constructor

cstamas: fixed

slawekjaranowski (approved):

gnodet:
@cstamas I've pushed an additional commit that simplify things a lot, let me know what you think, we can revert if needed.

cstamas:
@gnodet look good

jira-importer:
Resolve #10864

PR #1400, opened 2024-02-06
DependencyScopeManager

cstamas:
This is WIP


gnodet:
While I totally get the idea to refactor those bits in order to have a single source for scopes, this seems overly complicated and I don't see why the original enum is not sufficient.  The manager adds an additional layer for no benefit, as it does not provide any value over the enum definition.

PR #1401, opened 2024-02-07
[MNG-8015] Control the type of path where each dependency can be placed

gnodet:
#1378  merged with master and pushed to maven so that we can actually collaborate on it (as the branch for #1378 is not writable)

### DIFF
--- a/api/maven-api-core/src/main/java/org/apache/maven/api/ExtensibleEnum.java
+++ b/api/maven-api-core/src/main/java/org/apache/maven/api/ExtensibleEnum.java
@@ -22,6 +22,8 @@ import org.apache.maven.api.annotations.Experimental;
 import org.apache.maven.api.annotations.Nonnull;
 
 /**
+ * Interface that defines some kind of enums that can be extended by Maven plugins or extensions.
+ *
  * Implementation must have {@code equals()} and {@code hashCode()} implemented, so implementations of this interface
  * can be used as keys.
  *
--- a/api/maven-api-core/src/main/java/org/apache/maven/api/JavaPathType.java
+++ b/api/maven-api-core/src/main/java/org/apache/maven/api/JavaPathType.java
@@ -1,0 +1,319 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.maven.api;
+
+import java.io.File;
+import java.nio.file.Path;
+import java.util.Objects;
+import java.util.Optional;
+import java.util.StringJoiner;
+
+import org.apache.maven.api.annotations.Experimental;
+import org.apache.maven.api.annotations.Nonnull;
+
+/**
+ * The option of a Java command-line tool where to place the paths to some dependencies.
+ * A {@code PathType} can identify the class-path, the module-path, the patches for a specific module,
+ * or another kind of path.
+ *
+ * <p>One path type is handled in a special way: unlike other options,
+ * the paths specified in a {@code --patch-module} Java option is effective only for a specified module.
+ * This type is created by calls to {@link #patchModule(String)} and a new instance must be created for
+ * every module to patch.</p>
+ *
+ * <p>Path types are often exclusive. For example, a dependency should not be both on the Java class-path
+ * and on the Java module-path.</p>
+ *
+ * @see org.apache.maven.api.services.DependencyResolverResult#getDispatchedPaths()
+ *
+ * @since 4.0.0
+ */
+@Experimental
+public enum JavaPathType implements PathType {
+    /**
+     * The path identified by the Java {@code --class-path} option.
+     * Used for compilation, execution and Javadoc among others.
+     *
+     * <p><b>Context-sensitive interpretation:</b>
+     * A dependency with this path type will not necessarily be placed on the class-path.
+     * There are two circumstances where the dependency may nevertheless be placed somewhere else:
+     * </p>
+     * <ul>
+     *   <li>If {@link #MODULES} path type is also set, then the dependency can be placed either on the
+     *       class-path or on the module-path, but only one of those. The choice is up to the plugin,
+     *       possibly using heuristic rules (Maven 3 behavior).</li>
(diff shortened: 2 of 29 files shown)


gnodet:
@desruisseaux fwiw, the failing test is [an integration test](https://github.com/apache/maven-integration-testing/blob/master/core-it-suite/src/test/java/org/apache/maven/it/MavenITmng5135AggregatorDepResolutionModuleExtensionTest.java). which belongs to the https://github.com/apache/maven-integration-testing repository and not https://github.com/apache/maven ...

cstamas:
By looking at this PR, one thing occurred to me: today resolver-only users (those integrating Resolver w/o Maven) are able to fully resolve things as Maven would. If any of these become implemented in maven-core (as opposed to maven-resolver-provider), this "functional equivalence" is gone: JPMS will become handled by Maven, but Resolver integrators will not be able to resolve projects that do use these new functions. We need to keep eye on this fact.

gnodet:
> By looking at this PR, one thing occurred to me: today resolver-only users (those integrating Resolver w/o Maven) are able to fully resolve things as Maven would. If any of these become implemented in maven-core (as opposed to maven-resolver-provider), this "functional equivalence" is gone: JPMS will become handled by Maven, but Resolver integrators will not be able to resolve projects that do use these new functionalities. We need to keep eye on this fact.

Ideally, we'd come up with a way to use Maven 4 API outside of maven...

cstamas:
> Ideally, we'd come up with a way to use Maven 4 API outside of maven...

That sound awkward to me :smile: like saying "we will make servlet-api usable without servlet engine", so something will need to implement it. And if not Maven, what?

rmannibucau:
While I understand the technical relevance we should also likely discourage it since one of the goal is to own and not conflict without anything possible.
That said, this PR is way outside the scope of resolver users outside maven IMHO, it is very opiniated so shouldn't be a big blocker we cant compensate by a few standalone impl in practise as it is already the case, no?

desruisseaux:
I'm not familiar with the relationship between Maven and Resolver. But if the `maven-resolver-provider` is used outside Maven with the expectation that it provides the same result, maybe we should move the PR there? It is not opinionated, as it is based on the intersection of two supplied information:

* Which paths a JAR can be placed (information supplied by the JAR producer).
* Which paths a plugin wants to use (information supplied by the JAR consumer).

The only heuristic part is what to do when the intersection of the above contains 2 or more paths.

desruisseaux:
As a real world use case for testing the Maven improvements, there is the following scenario: a project depends on JUnit 5 with `compile` scope (because the project is a conformance test kit). Everything is JPMS: the project, JUnit 5 and their dependencies. JUnit 5 is itself splitted in many modules, including the `junit-jupiter-api`, `junit-jupiter-engine` and `junit-platform-commons` modules.

For a mysterious reason, Maven 3 places `junit-platform-commons` dependency on the module path, but `junit-jupiter-engine` on the class-path (verified with `mvn test --debug`). The consequence is that JUnit 5 cannot be launch. The JVM refuses to start with the following error message: _"class org.junit.platform.engine.discovery.DiscoverySelectors (in unnamed module @0x3b2c72c2) cannot access class org.junit.platform.commons.util.Preconditions (in module org.junit.platform.commons) because module org.junit.platform.commons does not export org.junit.platform.commons.util to unnamed module @0x3b2c72c2"_. Translation: **JUnit 5 cannot access its own internal classes!**

The reason why JUnit 5 cannot access its own classes is because `junit-platform-commons` allows access to its packages only to other JUnit 5 modules. It can be seen with the following command:

```bash
jar --describe-module -f junit-platform-commons-1.10.2.jar
```

Output snippet (reformatted):

```
qualified exports org.junit.platform.commons.util to
     org.junit.jupiter.api
     org.junit.jupiter.engine
     etc...
```

Because Maven puts `junit-platform-commons` on the module-path, access restrictions to that JAR are enforced. But because Maven puts `junit-jupiter-engine` on the class-path, that module is unnamed and consequently not recognized as a module in above list of modules allowed to access the `junit-platform-commons` internal packages. Thus the failure to launch JUnit.

A workaround is to add the following lines in Surefire configuration for breaking module encapsulation:

```xml
<argLine>
  --add-exports org.junit.platform.commons/org.junit.platform.commons.util=ALL-UNNAMED
  --add-exports org.junit.platform.commons/org.junit.platform.commons.logging=ALL-UNNAMED
</argLine>
```

We should not be forced to apply such workaround. This real use case is another demonstration of the need to improve JPMS handling compared to what Maven 3 does. For testing if this pull request achieves that goal in Maven 4, we can test if it allows us to remove the above hack from [GeoAPI conformance pom.xml file](https://github.com/opengeospatial/geoapi/blob/9e4ff919ef56e52e05767cc85c88e5795081c883/geoapi-conformance/pom.xml#L133).

rmannibucau:
@desruisseaux I'm not sure I understand the proposal, the "workaround" is needed per java, do you want surefire to do it implicitly or just use the right path? Isn't it solved by the type work you did and Guillaume reuses in maven 4? (trying to convert the statement in actions)

desruisseaux:
The workaround is not needed per Java. It is needed because Maven 3 mixes class-path and module-path in an apparently random way. It is an example of case where putting the JAR on the class-path breaks the application. Just putting the `junit-jupiter-engine.jar` on the module-path, as it should be, fixes the issue and removes completely the need for the workaround.

This problem should indeed by solved by the path work. I reported this story as a real use case for testing that the proposal works. The criterion for considering that test as successful is to be able to remove any hacks in above-cited `pom.xml` file.

rmannibucau:
Ok, so no action needed.

side note: it is not random but an heuristic so does not always work indeed - don't ask me why default is not 100% the classpath, it is statically saner for 3rd parties even if it enforces more work for jpms apps, but we see the light so we'll drop it soon :crossed_fingers: .

gnodet:
Should we go ahead and merge that one ? It does not really affects the current behaviour, so if we even find something missing, we can still fix it later...

cstamas (approved):

gnodet:
> Fine for me. I added [one more commit](https://github.com/Geomatys/maven/commit/75d8e88046e38cb30134c34eb731c226f4dfce87) for allowing plugins to log warning to users when a dependency cannot be placed on any path.
> 
> After the merge, I would like to try the new API in `maven-compiler-plugin` and create a new merge request for it.

Sounds good.  You'll need to build the following branches:
* https://github.com/apache/maven-plugin-tools/pull/242
* https://github.com/apache/maven-plugin-testing/pull/39
* https://github.com/apache/maven-compiler-plugin/pull/147

The above maven-compiler-plugin PR can be used as a basis, as it is already has all the migration to the v4 api.  If the underlying plexus-compiler API does not support the needed features to fully leverage the new jpms support, a possibility would be to change the m-compiler-p to simply wrap the `javac` compiler instead of trying to abstract all implementations.  The other one is the eclipse compiler, not sure they are really others.  If needed, we'd end up with two plugins, but they may ease the mapping of various options...    We discussed that with @cstamas ...

cstamas:
WooHoo! :fireworks:

desruisseaux:
Thanks! I will try to start playing with the compiler plugin this weekend.

desruisseaux:
> If the underlying plexus-compiler API does not support the needed features to fully leverage the new jpms support, a possibility would be to change the m-compiler-p to simply wrap the javac compiler instead of trying to abstract all implementations.

I have not explored the Plexus compiler capabilities, but indeed I propose to use `javax.tools.JavaCompiler` instead. It is an interface, so it would hopefully be possible to create a wrapper around the Eclipse compiler. Maybe Eclipse would even do that themselves, since `javax.tools.JavaCompiler` is standard API.

PR #1402, opened 2024-02-07
Remove empty dependency scope from the api

gnodet:


api/maven-api-core/src/main/java/org/apache/maven/api/DependencyScope.java
```diff
@@ -49,9 +49,12 @@ public enum DependencyScope {
     NONE("none", false),
 
     /**
-     * Empty scope.
+     * Undefined. When no scope is explicitly given, UNMANAGED will be used, but its meaning will depend on
```
cstamas, line 52: UNDEFINED will be useed

cstamas (approved):

PR #1404, opened 2024-02-07
[MNG-8045] Use DependencyGraphDumper instead of own implementation

slawekjaranowski:
https://issues.apache.org/jira/browse/MNG-8045


cstamas (approved):

jira-importer:
Resolve #10818

PR #1406, opened 2024-02-08
Hack for reactor reduction

cstamas:
This is just a hack, to test my theory. And it makes issue of the reproducer gone, it is working "as expected" and is not broken.
https://lists.apache.org/thread/sgcqyz5trxngxb7r64hjnccb8btxhbrl

### DIFF
--- a/maven-core/src/main/java/org/apache/maven/project/MavenProject.java
+++ b/maven-core/src/main/java/org/apache/maven/project/MavenProject.java
@@ -32,6 +32,7 @@ import java.util.Map;
 import java.util.Objects;
 import java.util.Properties;
 import java.util.Set;
+import java.util.concurrent.atomic.AtomicBoolean;
 
 import org.apache.maven.RepositoryUtils;
 import org.apache.maven.artifact.Artifact;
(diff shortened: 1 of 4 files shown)


maven-core/src/main/java/org/apache/maven/project/MavenProject.java
```diff
@@ -182,6 +183,18 @@ public class MavenProject implements Cloneable {
 
     private final Set<String> lifecyclePhases = Collections.synchronizedSet(new LinkedHashSet<>());
 
+    private final AtomicBoolean projectNeedsBuild = new AtomicBoolean(true);
```
gnodet, line 186: this should only be set at the beginning, why using an `AtomicBoolean` ?

cstamas: I wanted just to ensure that "transition" from true->false can happen only once, and there is no way to "reset" it

cstamas: Again, this is just a hack to prove my theory, this flag is omitted from clone etc...

PR #1410, opened 2024-02-13
[MNG-8053] Profile activation by packaging in the POM

gnodet:
JIRA issue: https://issues.apache.org/jira/browse/MNG-8053

Following up https://issues.apache.org/jira/browse/MNG-6609, we should now make the packaging activation for profiles a first class citizen in the model.

### DIFF
--- a/api/maven-api-model/src/main/mdo/maven.mdo
+++ b/api/maven-api-model/src/main/mdo/maven.mdo
@@ -2771,6 +2771,14 @@
             <type>ActivationFile</type>
           </association>
         </field>
+        <field>
+          <name>packaging</name>
+          <version>4.1.0+</version>
+          <description>Specifies that this profile will be activated based on the project's packaging.</description>
+          <association>
+            <type>ActivationPackaging</type>
+          </association>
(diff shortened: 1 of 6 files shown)


api/maven-api-model/src/main/mdo/maven.mdo
```diff
@@ -2771,6 +2771,14 @@
             <type>ActivationFile</type>
           </association>
         </field>
+        <field>
+          <name>packaging</name>
+          <version>4.1.0+</version>
+          <description>Specifies that this profile will be activated based on the project's packaging.</description>
+          <association>
+            <type>ActivationPackaging</type>
```
slawekjaranowski, line 2779: Do we need a next type with only one field?

gnodet: Oh, are you suggesting to add a single new field `packaging`, same as the existing `jdk` one ?

slawekjaranowski (approved):

jira-importer:
Resolve #10814

PR #1412, opened 2024-02-13
[MNG-8050] emit warn in case of repo id clashes between settings and POM

kwin:

       pulling in other changes.

### DIFF
--- a/maven-model-builder/src/main/java/org/apache/maven/model/validation/DefaultModelValidator.java
+++ b/maven-model-builder/src/main/java/org/apache/maven/model/validation/DefaultModelValidator.java
@@ -41,10 +41,7 @@ import java.util.function.UnaryOperator;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 import java.util.stream.Collectors;
-<<<<<<< Upstream, based on master
 import java.util.stream.StreamSupport;
-=======
->>>>>>> 4f3fe69 [MNG-8050] emit warn in case of repo id clashes between settings and POM
 
 import org.apache.maven.api.model.Activation;
 import org.apache.maven.api.model.ActivationFile;
--- a/maven-model-builder/src/main/java/org/apache/maven/model/validation/DefaultModelValidator.java
+++ b/maven-model-builder/src/main/java/org/apache/maven/model/validation/DefaultModelValidator.java
@@ -24,6 +24,7 @@ import javax.inject.Singleton;
 
 import java.io.File;
 import java.util.Arrays;
+import java.util.Collection;
 import java.util.Collections;
 import java.util.Deque;
 import java.util.HashMap;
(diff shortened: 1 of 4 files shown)


maven-model-builder/src/main/java/org/apache/maven/model/validation/DefaultModelValidator.java
```diff
@@ -576,6 +578,36 @@ public void validateEffectiveModel(Model ma, ModelBuildingRequest request, Model
...
+            List<org.apache.maven.model.Profile> activeExternalProfiles,
+            Model ma,
+            ModelBuildingRequest request,
+            ModelProblemCollector problems) {
+        org.apache.maven.api.model.Model m = ma.getDelegate();
+        // check for id clashes in repositories
+        for (Profile profile : activeExternalProfiles.stream()
+                .map(org.apache.maven.model.Profile::getDelegate)
+                .collect(Collectors.toList())) {
+            for (Repository repository : profile.getRepositories()) {
```
kwin, line 592: TODO: Also check for clashing plugin repositories

cstamas:
@gnodet  ?

gnodet:
@kwin Should we check that the repositories are actually not the same ? What's the problem is they point to the same repo ? If the user simply wants to override the policy in the settings ?

BTW, I think the mapping from an `id` to a repository URI  should actually be defined in the local repository.  I think two projects using the same `id` but pointing to different repositories might be problematic too.

kwin:
@gnodet I now only emit warning if the URLs differ (https://github.com/apache/maven/commit/cd7aba6af6ff80a394e12b975a293fa07cb813c9)

> I think the mapping from an id to a repository URI should actually be defined in the local repository

Not sure how this can ever be achieved. I still want to be able to clone arbitrary projects and build it immediately without needing to tweak either my `settings.xml` or some other Maven configuration file.

gnodet:
> @gnodet I now only emit warning if the URLs differ ([cd7aba6](https://github.com/apache/maven/commit/cd7aba6af6ff80a394e12b975a293fa07cb813c9))
> 
> > I think the mapping from an id to a repository URI should actually be defined in the local repository
> 
> Not sure how this can ever be achieved. I still want to be able to clone arbitrary projects and build it immediately without needing to tweak either my `settings.xml` or some other Maven configuration file.

@cstamas can the resolver provide any help here ?

cstamas:
Sadly nope.

IMHO the best we can do is to accept the facts:
* repository ID is the "key" (so assume URL is "just" an attribute) and limit ourselves onto ID clash checking
* maybe as improvement, warn (optionally fail) the build, if there are two different IDs with _same_ URLs?
* cases when same ID may be used for different reposes (or when different IDs are used for same repository) mostly stands for us, people that check out various project, applying various "repo naming conventions" (as for example, ideally within one company or even forge, this should be unified) is we simply cannot handle.

kwin:
@gnodet I think we should also warn if the URL is equal but the `<enabled>` flags for either `<release>` or `<snapshot>` differ. Otherwise this might lead to subtle issues which are hard to debug.

kwin:
> I think we should also warn if the URL is equal but the <enabled> flags for either <release> or <snapshot> differ. Otherwise this might lead to subtle issues which are hard to debug.

I decided to ignore those subtle differences for now.

PR #1419, opened 2024-02-20
[MNG-8006] Switch property contributors to use a lazy lookup

gnodet:
... else, the map seems empty and not dynamically populated.


cstamas (approved):

gnodet:
This will actually allow https://github.com/trustin/os-maven-plugin/pull/74 to work

PR #1421, opened 2024-02-23
[MNG-5726] Support regular expression matching in profile activation for

kwin:
OS version

This requires using the reserved prefix "regex:" in the version element.

       pulling in other changes.

### DIFF
--- a/maven-model-builder/src/main/java/org/apache/maven/model/profile/activation/OperatingSystemProfileActivator.java
+++ b/maven-model-builder/src/main/java/org/apache/maven/model/profile/activation/OperatingSystemProfileActivator.java
@@ -21,6 +21,8 @@ package org.apache.maven.model.profile.activation;
 import javax.inject.Named;
 import javax.inject.Singleton;
 
+import java.util.Locale;
+
 import org.apache.maven.model.Activation;
 import org.apache.maven.model.ActivationOS;
 import org.apache.maven.model.Profile;
@@ -36,6 +38,8 @@ import org.apache.maven.utils.Os;
 @Named("os")
 @Singleton
 public class OperatingSystemProfileActivator implements ProfileActivator {
+
+    private static final String REGEX_PREFIX = "regex:";
 
     @Override
     public boolean isActive(Profile profile, ProfileActivationContext context, ModelProblemCollector problems) {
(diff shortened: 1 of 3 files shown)


gnodet:
Would it make sense to also support the easier `glob:` syntax then ?

kwin:
@gnodet Is there general support for matching globs built into Java (apart from Files NIO)? I haven't found anything and quick googling revealed https://stackoverflow.com/questions/1247772/is-there-an-equivalent-of-java-util-regex-for-glob-type-patterns. So I would say, regex is good enough for now.

gnodet:
> @gnodet Is there general support for matching globs built into Java (apart from Files NIO)? I haven't found anything and quick googling revealed https://stackoverflow.com/questions/1247772/is-there-an-equivalent-of-java-util-regex-for-glob-type-patterns. So I would say, regex is good enough for now.
> 
> (I don't want to rely on external libraries like plexus-utils or maven-utils for such stuff TBH)

Fwiw, a workaround is already used in maven: https://github.com/apache/maven/blob/master/maven-core/src/main/java/org/apache/maven/artifact/resolver/filter/ExclusionArtifactFilter.java
Maybe we can move the code in its own class in maven-core and reuse it ?

kwin:
@gnodet Interesting. We can support `glob:` later once that refactoring is done. For now I would go with just `regex:`. Any concerns with that?

gnodet:
> @gnodet Interesting. We can support `glob:` later once that refactoring is done. For now I would go with just `regex:`. Any concerns with that?

None...

gnodet (approved):

maven-model-builder/src/main/java/org/apache/maven/model/profile/activation/OperatingSystemProfileActivator.java
```diff
@@ -53,17 +57,21 @@ public boolean isActive(Profile profile, ProfileActivationContext context, Model
 
         boolean active = ensureAtLeastOneNonNull(os);
 
+        String actualOsName = context.getSystemProperties().get("os.name").toLowerCase(Locale.ENGLISH);
```
gnodet, line 60: Why not `Os.OS_NAME` and related ?

kwin: To streamline the testing approach with the other activators. `Os.OS_NAME` cannot be easily overwritten.

kwin: Compare with https://github.com/apache/maven/blob/ab6ec5bd74af20ab429509eb56fc8e3dff4c7fc7/maven-model-builder/src/main/java/org/apache/maven/model/profile/activation/JdkVersionProfileActivator.java#L64 and https://github.com/apache/maven/blob/ab6ec5bd74af20ab429509eb56fc8e3dff4c7fc7/maven-model-builder/src/test/java/org/apache/maven/model/profile/activation/JdkVersionProfileActivatorTest.java#L50

jira-importer:
Resolve #6587

PR #1429, opened 2024-02-27
[MNG-8225] Fully concurrent builder for Maven 4

gnodet:
JIRA issue: [MNG-8225](https://issues.apache.org/jira/browse/MNG-8225)

This PR aims at defining a complete new lifecycle for Maven 4.  This lifecycle is disabled by default and needs to be turned on using `mvn --builder=concurrent`.

Note that the builder is fully concurrent, so that a project may not be executed in a single thread and continuously, rather, all mojos will be scheduled independently.  The builder displays a nice summary at the beginning (currently at `info` level).

## Status

To avoid [MNG-7335](https://issues.apache.org/jira/browse/MNG-7335), [MNG-6843](https://issues.apache.org/jira/browse/MNG-6843) and all those long standing related bugs, the new lifecycle/builder warns when using non Maven 4 mojos (kinda like we have with the multithreaded builder which warns when mojos are not flagged as thread safe).

### DIFF
--- a/api/maven-api-core/src/main/java/org/apache/maven/api/Lifecycle.java
+++ b/api/maven-api-core/src/main/java/org/apache/maven/api/Lifecycle.java
@@ -56,6 +56,7 @@ public interface Lifecycle extends ExtensibleEnum {
     // ======================
     String BEFORE = "before:";
     String AFTER = "after:";
+    String AT = "at:";
 
     /**
      * Name or identifier of this lifecycle.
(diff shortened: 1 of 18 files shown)


cstamas:
Can you point me where "concurrent" is enabled by def? I cannot find that spot

gnodet:
> Can you point me where "concurrent" is enabled by def? I cannot find that spot

It was enabled in earlier versions, I disabled it a few days ago.

cstamas (approved):

jira-importer:
Resolve #9778

PR #1432, opened 2024-03-01
[MNG-8062] Consider session close result into build outcome

cstamas:
Session close may throw as well, factor that into global build outcome.

---

https://issues.apache.org/jira/browse/MNG-8062

### DIFF
--- a/maven-core/src/main/java/org/apache/maven/DefaultMaven.java
+++ b/maven-core/src/main/java/org/apache/maven/DefaultMaven.java
@@ -222,7 +222,16 @@ public class DefaultMaven implements Maven {
 
             legacySupport.setSession(session);
 
-            return doExecute(request, session, result, chainedWorkspaceReader);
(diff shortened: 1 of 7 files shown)


gnodet:
I'm not sure I'm seeing the purpose of the new event, as the repository session is closed in a related code block.  In addition, this looks quite redundant with the `AbstractMavenLifecycleParticipant.afterSessionEnd` event...

gnodet:
> I'm not sure I'm seeing the purpose of the new event, as the repository session is closed in a related code block. In addition, this looks quite redundant with the `AbstractMavenLifecycleParticipant.afterSessionEnd` event...

Also, considering the actual problem, which is https://github.com/apache/maven-resolver/pull/437, I wonder if the real need is to allow plugins to register callbacks to run at the end of the build somehow.  
Fwiw, I'm not sure that the actual install/deploy of artifacts should be performed *while the session is being closed*.  I would assume such important task has to be done before actually closing the session.

PR #1434, opened 2024-03-01
[MNG-8012] Warn if in-reactor BOM import happens

cstamas:
This should be in fact prevented.
Also, model builder missed File uses are corrected.

---

https://issues.apache.org/jira/browse/MNG-8012

### DIFF
--- a/maven-model-builder/src/main/java/org/apache/maven/model/building/DefaultModelBuilder.java
+++ b/maven-model-builder/src/main/java/org/apache/maven/model/building/DefaultModelBuilder.java
@@ -1847,6 +1847,10 @@ public class DefaultModelBuilder implements ModelBuilder {
 
             if (importSource instanceof FileModelSource && request.getRootDirectory() != null) {
                 URI sourceUri = ((FileModelSource) importSource).getLocationURI();
+                // URI.relativize does the job for us, as if the scheme and authority components of the two URIs
+                // are not identical, or if the path of this URI is not a prefix of the path of the passed URI,
+                // then the passed in URI is returned. Hence, if relativize succeeds, we should report problem,
+                // as the POM originates from within the reactor.
                 if (request.getRootDirectory().toUri().relativize(sourceUri) != sourceUri) {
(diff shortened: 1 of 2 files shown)


maven-model-builder/src/main/java/org/apache/maven/model/building/DefaultModelBuilder.java
```diff
@@ -1830,6 +1845,15 @@ private org.apache.maven.api.model.DependencyManagement doLoadDependencyManageme
                 return null;
             }
 
+            if (importSource instanceof FileModelSource && request.getRootDirectory() != null) {
+                URI sourceUri = ((FileModelSource) importSource).getLocationURI();
+                if (request.getRootDirectory().toUri().relativize(sourceUri) != sourceUri) {
```
gnodet, line 1850: Fwiw, this test is unintuitive... a comment would be welcomed.  I had to check the javadoc...

cstamas: Fixed

gnodet: I don't want to be pushy, but what about the following ?
```
Path sourcePath = ((FileModelSource) importSource).getPath();
if (sourcePath.startsWith(request.getRootDirectory()) {
```

cstamas: Pushed, so who's pushy now?

gnodet (approved):
LGTM

reckart:
I believe this implementation is incomplete. It does not discern between a **BOM** (a module which manages the versions of the modules **in** the reactor) and a **dependency management module** (a module which manages the versions of external libraries used in the reactor).

* example of a BOM: https://central.sonatype.com/artifact/org.springframework/spring-framework-bom
* example of a dependency management module: https://central.sonatype.com/artifact/org.springframework.boot/spring-boot-dependencies

PR #1435, opened 2024-03-06
[MNG-8069] add failing projects if a project is banned from reactor

laeubi:
Currently if a multimodule build fails (either fast or at the end), the user gets a message in the log that reads:

"This project has been banned from the build due to previous failures."

this is correct but does not give a hint what has failed and one needs to investigate the build logs to get more details.

This now instead shows as part of the message the actual project that make it impossible to build this  one and is the reason for this project being banned.

 - [x] Make sure there is a [JIRA issue](https://issues.apache.org/jira/browse/MNG-8069) filed
       pulling in other changes.

### DIFF
--- a/maven-embedder/src/main/java/org/apache/maven/cli/event/ExecutionEventLogger.java
+++ b/maven-embedder/src/main/java/org/apache/maven/cli/event/ExecutionEventLogger.java
@@ -20,6 +20,7 @@ package org.apache.maven.cli.event;
 
 import java.io.File;
 import java.nio.file.Path;
+import java.util.Comparator;
 import java.util.List;
 import java.util.Objects;
 
@@ -28,8 +29,10 @@ import org.apache.maven.execution.BuildFailure;
 import org.apache.maven.execution.BuildSuccess;
 import org.apache.maven.execution.BuildSummary;
 import org.apache.maven.execution.ExecutionEvent;
+import org.apache.maven.execution.MavenExecutionRequest;
 import org.apache.maven.execution.MavenExecutionResult;
 import org.apache.maven.execution.MavenSession;
+import org.apache.maven.execution.ProjectDependencyGraph;
 import org.apache.maven.plugin.MojoExecution;
 import org.apache.maven.plugin.descriptor.MojoDescriptor;
 import org.apache.maven.project.MavenProject;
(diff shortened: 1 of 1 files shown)


gnodet:
Wouldn't it be better to add the cause when the project is banned rather than trying to reconstruct it later ? I wonder if we should enhance the [`blacklist()` method](https://github.com/apache/maven/blob/0b3246381d12e69deef63888d0db274fb49c6697/maven-core/src/main/java/org/apache/maven/lifecycle/internal/ReactorBuildStatus.java#L49) to be given a cause...

Also, what about providing two slightly  different messages, I think it's interesting to know if the project has been skipped because one project has failed (without the fail-at-end flag, do we actually need to display the failing projects for each skipped project, even unrelated ?) or it's actually an upstream project which caused the failure ?

laeubi:
The problem is that if you have a multimodule build (lets say 300 modules) and use multithreading (so build messages are interleaved), and one of the projects fail in the middle you have a huge logfile full of "banned" messages but you need to find out what/where it actually fails, so the message without any context is quite useless and even if you find the erroneous project its not very helpful. So I though showing it at tat time might be more useful.

Another approach would be to simply not display the message at all, and instead give a summary at the end of build, e.g. ```
The following projects where not build due to previous failures:
- a (depends on x that failed)
- b
- c (depends on b that was banned)
```

sadly currently this information is somehow spread across some places and I'm not sure why e.g. a listener was used instead of printing the message directly on `blacklist()` ...

regarding `fae`/`ff` I think we can add such message of course but I didn't wanted to complicate things for a first draft, if you think its usefull I think it is quite easy to add,

File: maven-embedder/src/main/java/org/apache/maven/cli/event/ExecutionEventLogger.java [hunk 1]
```diff
@@ -243,10 +246,44 @@ public void projectSkipped(ExecutionEvent event) {
...
             logger.info("");
             infoLine('-');
+            MavenSession session = event.getSession();
+            MavenExecutionResult result = session.getResult();
+            ProjectDependencyGraph projectDependencyGraph = session.getProjectDependencyGraph();
+            List<MavenProject> upstreamProjects;
+            if (MavenExecutionRequest.REACTOR_FAIL_AT_END.equals(session.getReactorFailureBehavior())
+                    && projectDependencyGraph != null) {
+
+                // the project is blacklisted only so one of its upstreams must have failed here...
```
elharo, line 256: I can't quite follow this comment, please rephrase

maven-embedder/src/main/java/org/apache/maven/cli/event/ExecutionEventLogger.java
```diff
@@ -243,10 +246,44 @@ public void projectSkipped(ExecutionEvent event) {
...
+                upstreamProjects = projectDependencyGraph.getUpstreamProjects(event.getProject(), true);
+            } else {
+                // any other failure must have lead to this so any projects is eligible
```
elharo, line 259: are eligible

jira-importer:
Resolve #10800

gnodet (commented):
## AI Code Review — PR #1435

**Verdict: ⚠️ Bug found**

This PR adds useful failure-reason context to the "project banned" skip message in multi-module reactor builds. The approach of listing failed upstream projects is reasonable and addresses a real usability gap (MNG-8069).

However, there is a definite bug in the error message construction.

### Findings

| 1 | 🔴 High | `ExecutionEventLogger.java:278` | **Bug: variable typo** — `message += ": " + message;` concatenates `message` to itself instead of appending it to `reason`. The intent is clearly `reason += ": " + message;` — to build a string like `"LifecycleExecutionException: some error"`. As written, `reason` only ever contains the bare exception class name (e.g., `"LifecycleExecutionException"`) and the message text is lost. |
| 2 | 🟡 Medium | `ExecutionEventLogger.java:275-280` | **Consequence of bug** — Because `reason` never gets the message text, the output for banned projects will show only the exception class name with no description, significantly reducing the usability improvement this PR aims to provide. |

### Context
- **Targets 3.9.x** — File path is valid for `maven-3.9.x` but has moved to `impl/maven-cli/` on master.
- **gnodet's architectural feedback** (March 2024) suggested capturing the cause at blacklist time rather than reconstructing it. This was not addressed.
- **The fix for the bug is simple:** change line 278 from `message += ": " + message;` to `reason += ": " + message;`.

---

gnodet (changes requested):
## Code Review

The PR adds reactor failure handling for banned projects, but contains a confirmed bug:

1. **Variable assignment bug** (`ExecutionEventLogger.java:282`): `message += ": " + message` has two defects: (a) it self-concatenates `message` with itself (e.g., `"error text"` becomes `"error text: error text"`), and (b) it updates the local `message` variable which is never read again, instead of appending to `reason`. As a result, the logger at line 285 outputs only the bare exception class name (e.g., `"LifecycleExecutionException"`) without the actual error description. The intended code is:

   ```suggestion
   reason += ": " + message;
   ```

2. **Minor typo** (`ExecutionEventLogger.java:259`): Comment reads "must have lead to this" — should be "must have led to this" (past participle).

PR #1437, opened 2024-03-10
[MNG-8072] add ConsumerPomFile methods

laeubi:
Currently there is some vague definition of a "build pom" versus a "consumer pom":

- https://maven.apache.org/studies/consumer-pom/
- https://cwiki.apache.org/confluence/display/MAVEN/Build+vs+Consumer+POM

there are also some plugins around to work with this e.g.

- 
- https://tycho.eclipseprojects.io/doc/latest/tycho-packaging-plugin/update-consumer-pom-mojo.html
- https://www.mojohaus.org/flatten-maven-plugin/plugin-info.html

but there are some issues:

1. Maven makes some assumptions about its "file", e.g setting the file
2. changes the basedir
3. Even though there is a setPomFile method that do not change the
4. basedir e.g. a parent ref by default would be broken
5. once we change the file, there is a mismatch between the pom on disk
6. and the model
7. it is not possible for a plugin to know the original file and the
8. consumer file, e.g. if I want to deploy the original file e.g. with classifier "pom-build"
9. because of this usually "consumer pom plugins" generate the the new
10. pom to a new file in the project root, where it actually not belongs to and leaves the file even after mvn clean if no special actions are taken.

This enhance the MavenProject with one more method set/getConsumerPom that fills this gap, then plugins working on the consumer pom can set the file there and where it then is used by other plugins.

 - [x] Make sure there is a [JIRA issue](https://issues.apache.org/jira/browse/MNG-8072) filed
       pulling in other changes.

### DIFF
--- a/maven-core/src/main/java/org/apache/maven/project/MavenProject.java
+++ b/maven-core/src/main/java/org/apache/maven/project/MavenProject.java
@@ -103,6 +103,8 @@ public class MavenProject implements Cloneable {
 
     private File file;
 
+    private File consumerPomFile;
+
     private File basedir;
 
     private Set<Artifact> resolvedArtifacts;
@@ -254,6 +256,17 @@ public class MavenProject implements Cloneable {
      */
     public void setPomFile(File file) {
         this.file = file;
+    }
+
+    public void setConsumerPomFile(File file) {
+        this.consumerPomFile = file;
+    }
+
+    public File getConsumerPomFile() {
+        if (consumerPomFile == null) {
+            return getFile();
+        }
+        return consumerPomFile;
     }
 
     public File getBasedir() {
@@ -784,6 +797,7 @@ public class MavenProject implements Cloneable {
      * @deprecated Please use {@link MavenProjectHelper}
      * @throws DuplicateArtifactAttachmentException will never happen but leave it for backward compatibility
      */
+    @Deprecated
     public void addAttachedArtifact(Artifact artifact) throws DuplicateArtifactAttachmentException {
         // if already there we remove it and add again
         int index = attachedArtifacts.indexOf(artifact);
@@ -1026,6 +1040,7 @@ public class MavenProject implements Cloneable {
 
         // copy fields
         file = project.file;
+        this.consumerPomFile = project.consumerPomFile;
         basedir = project.basedir;
 
         // don't need a deep copy, they don't get modified or added/removed to/from - but make them unmodifiable to be


rmannibucau:
Hi @laeubi , wonder how consistent it is with https://github.com/apache/maven/blob/master/maven-core/src/main/java/org/apache/maven/internal/transformation/impl/DefaultConsumerPomArtifactTransformer.java#L94, shouldn't it hook there? We should also ensure the API is clearly defined for that part if we enable the consumer pom to not be handled automatically - guess we can let it be automatic for now since you can decorate the transformer but not 100% sure of the use case - typically flatten does not need this new API.

laeubi:
I currently only looking at [3.9.x](https://github.com/apache/maven/tree/maven-3.9.x) where I can't see this is used right now?

Also here the problem is that it only ever will allow one transformer, iterative transformation is not possible and I can't access the artifact / file until it is finalized already.

And integration would be quite trivial, e.g in this case simply replace `project.getFile()` by `project.getConsumerPomFile()` and you are done :-)

rmannibucau:
> I currently only looking at [3.9.x](https://github.com/apache/maven/tree/maven-3.9.x) where I can't see this is used right now?

We should always start from master otherwise the method will be lost in a coming version and will be pointless for almost all users I fear.

> Also here the problem is that it only ever will allow one transformer, iterative transformation is not possible and I can't access the artifact / file until it is finalized already.

Guess it is a matter of extension wiring but ultimately we can discuss moving the transformation but doing it before limits the side effects of the new models which is probably good but my point was more to not add a new consumerPom file which is not about consumer pom and just yet another attached artifact - we already have this feature without this new method.

> simply replace project.getFile() by project.getConsumerPomFile() and you are done :-)

Waiting your PR on all existing plugins using `getFile()` ;). Switching a method is not that trivial there. For example you break flatten plugin.

laeubi:
> but my point was more to not add a new consumerPom file which is not about consumer pom and just yet another attached artifact - we already have this feature without this new method.

But you can't modify an artifact (or should not) + there is no way for a plugin to participate as part of the mojo-execution with an extension but this is required as usually you want ti to be configured and you want to control order, e.g. usually one want maybe first tycho inject some things, then clear others with flatten-maven, maybe then finally something what maven do by default....

> Waiting your PR on all existing plugins using getFile() ;). Switching a method is not that trivial there. For example you break flatten plugin.

It won't break anything ... plugins not supporting this will simply never set this and then getConsumerPom() will return the "traditional" file like today.

rmannibucau:
> there is no way for a plugin to participate as part of the mojo-execution

Yes, this is why I mentionned we can need to revisit a but the current API. Once again my main concern was that the "consumerPom" API has no link with consumer pom at all and we must converge to maven 4 (even if we backport to 3.9) cause otherwise your API will be dead, even in 3.x when we'll backport things.
Ultimately it looks more - in v3 world - to attached artifacts so it is doable without the new API to me - short term.
Long term we need to find a better lifecycle but can need some list discussion?

> It won't break anything ... plugins not supporting this will simply never set this and then getConsumerPom() will return the "traditional" file like today. Also not "all" plugins need to be adjusted, just these that care about build vs consumer pom...

exactly so if you do you are no more compatible with the chain you describe earlier since the right pom will not be consumed downside so we need to be vigilant and see if it is an issue to break all the existing ecosystem or if we don't care plugins using this new (to be defined) API will not be compatible with existing plugins.

laeubi:
Whatever we choose "will not be compatible with existing plugins", but without anything offered no plugin can be adjusted so we are in an endless cycle...

Also feel free to propose an alternative I just want:

1. simple and 3.9.x compatible I don't want to wait for glorious future of Maven 4 where everything will be perfect an clean
2. must work incrementally (e.g. plugin A can create an initial consumer pom and plugin B can further enhance/adjust that)
3. no need for extensions, delegations, side effects (e.g. modify the original pom location is an undesired side-effect)

rmannibucau:
1. for me it is the promise whatever great you do you will not be able to use it so "-1" (dont see it as a blocker more as a "you can but dont cry later") to work on anything not on master
2. think it is thinking wrong, consumer pom should be automatically provided by the transformation of the transformed pom which guarantee the v4.0.0 compat - if it is incremental we loose that. So I think we should work on the native model then let the consumer pom be automatically handled
3. +1

laeubi:
1. You should not assume maven-core(-plugins) inability to adapt to never features to everyone, e.g. Tycho always uses latest maven and latest features
2. "by default" Maven can't know what is desired for a consumer pom, and different users have different demands. It might can do some additional transformations (e.g. map 'new' constructs > 'old' constructs) but for example the initial model for Tycho is almost empty, then we will it dynamically of course not reflected in file as we don't want to dirty the build files of the user, what even maybe not exits at all as it is a pomless module... then tycho-update-consumer-pom performs some transformation (e.g. mapping p2 > real maven coordinates) and writes a new file, I can't see how maven could ever handle this automatically given that
3. where current solution has undesired side-effect and only looks at `project.getFile()` ....

rmannibucau:
> You should not assume maven-core(-plugins) inability to adapt to never features to everyone, e.g. Tycho always uses latest maven and latest features

I didn't at all, even the opposite if you read carefully. I just highlight you create a concurrent feature of something existing in latest version of the project - indeed it needs a few adjustment but it does not need the proposed PR.

> "by default" Maven can't know what is desired for a consumer pom

Then you don't speak about consumer pom which is well defined as a consummable pom by downstream builds (v4.0.0), anything else is not consumer pom and does not belong to core IMO - I suspect even if not sure you want some tycho meta but this is not about consumer pom IMHO, potentially main pom but consumer has another intent.

>  I can't see how maven could ever handle this automatically given that..

Consumer pom is built form the model, if you fill the model you get a valid consumer pom.

> where current solution has undesired side-effect and only looks at project.getFile()

In maven 3 there isnt really a notion of consumer pom (we had some proto but no real API) so it is mainly about a single v4.0.0 pom you were able to mutate. Indeed it has side effects but similarly your consumer pom will have exactly the same since it has the exact same design and all plugins will not always agree on that (current pitfall).
This is why it is important to use the model IMHO and let the tranformer create a valid shareable and backward compat consumer pom.

laeubi:
> Consumer pom is built form the model, if you fill the model you get a valid consumer pom.

Just look here:

https://github.com/apache/maven/blob/33788fd7fc33932639d1a600eada36e3cd015bfd/maven-core/src/main/java/org/apache/maven/internal/transformation/impl/DefaultConsumerPomArtifactTransformer.java#L79

and here:

https://github.com/apache/maven/blob/33788fd7fc33932639d1a600eada36e3cd015bfd/maven-core/src/main/java/org/apache/maven/internal/transformation/impl/DefaultConsumerPomArtifactTransformer.java#L109

and finally here:

https://github.com/apache/maven/blob/33788fd7fc33932639d1a600eada36e3cd015bfd/maven-core/src/main/java/org/apache/maven/internal/transformation/impl/DefaultConsumerPomBuilder.java#L201

for me it looks like it is reading this file, not the (in memory) pom ...

> Then you don't speak about consumer pom which is well defined as a consummable pom by downstream builds (v4.0.0), anything else is not consumer pom and does not belong to core IMO - I suspect even if not sure you want some tycho meta but this is not about consumer pom IMHO, potentially main pom but consumer has another intent.

I'm defiantly talking about consumer pom, because consumers ( == plain maven) wants to have a pom that they can consume without Tycho (Tycho itself don't need a physical pom.xml). So if you build some jar with Tycho you want it to be consumable after you deployed it to central by others (non Tycho users) like a traditionally hand written pom.xml

rmannibucau:
> for me it looks like it is reading this file, not the (in memory) pom ...

Right, it is an issue for any plugin contirbuting dependencies but we can solve it easily keeping a single consumer pom logic instead of duplicating it.

> So if you build some jar with Tycho you want it to be consumable after you deployed it to central by others (non Tycho users) like a traditionally hand written pom.xml

Same than if you build with model v > 4.0.0 you want to be consumed by v4.0.0 consumers (older maven, coursier, gradle, plain ivy etc), so the same solution should converge IMHO.

laeubi:
That's why I started this PR, it maybe *could* be solved, but it currently is not solved ... so as soon as there is a solution this is probably obsolete ;-)

elharo (changes requested):
needs tests

jira-importer:
Resolve #10863

gnodet (commented):
## AI Code Review — PR #1437

**Verdict: 💬 COMMENT**

This PR adds `getConsumerPomFile()`/`setConsumerPomFile()` to `MavenProject` with a fallback to `getFile()` when no consumer POM is set, plus adds a missing `@Deprecated` annotation on `addAttachedArtifact`. The API concept is reasonable and the code follows existing patterns in the class.

### Finding

| 1 | 🟡 Medium | `MavenProject.java` | **No tests** — elharo requested tests in a prior CHANGES_REQUESTED review. The fallback behavior (`getConsumerPomFile()` returns `getFile()` when null) in particular should be tested. |

### Context
- **Targets 3.9.x** — The file path is valid for `maven-3.9.x` but has moved to `impl/maven-core/` on master (MNG-8346).
- **Unresolved design discussion** — rmannibucau raised concerns about whether this belongs in 3.9.x core vs being handled by the existing consumer-pom transformer in Maven 4. This question was never resolved.
- **Maven 4 uses a different approach** — On master, `ConsumerPomArtifactTransformer` uses attached artifacts with a "consumer" classifier. The mutable-field approach here is architecturally different.
- **Code quality is fine** — The getter/setter pattern matches existing methods in `MavenProject`, the parameter naming follows conventions, and the `@Deprecated` annotation fix is a correct housekeeping improvement.

### Recommendation
Add tests (especially for the null-fallback behavior) and get committer consensus on whether this API is desired for 3.9.x given the Maven 4 divergence.

---

gnodet (commented):
## Code Review

This PR adds `getConsumerPomFile()`/`setConsumerPomFile()` to MavenProject on the 3.9.x branch. The concept is reasonable and the fallback behavior (returning `getFile()` when no consumer POM is set) is a sound backward-compatibility design. The `@Deprecated` annotation addition on `addAttachedArtifact` is correct housekeeping. However, two issues remain:

1. **No tests provided** (`MavenProject.java:258`): The outstanding CHANGES_REQUESTED from elharo (March 2024) requesting tests has not been addressed in over two years. At minimum, the null-fallback behavior, set/get round-trip, and `deepCopy` propagation should be tested.

2. **Missing Javadoc and @since** (`MavenProject.java:258`): The neighboring `setPomFile()` method has documentation and a `@since 3.2.4` tag. The new `setConsumerPomFile()`/`getConsumerPomFile()` methods should follow the same convention, including documentation of the fallback behavior and a `@since` tag.

Note: The design discussion between rmannibucau and laeubi about whether this API belongs in 3.9.x (given Maven 4's architecturally different approach via `ConsumerPomArtifactTransformer`) was never resolved.

PR #1441, opened 2024-03-15
[MNG-8084] Make the v4 api usable outside the Maven runtime

gnodet:
First step at making the Maven 4 API usable outside a full maven process.
This PR includes:
  * move Maven 4 api implementation classes from `maven-core` into a new `maven-api-impl` module
  * implement native services for settings and toolchains builders, make v3 implementation wrap the new ones
  * various API enhancements related to the two services above

The Maven 4 API is not completely usable outside Maven yet, but here's a [working example](https://github.com/gnodet/maven/blob/lw-api/maven-api-impl/src/test/java/org/apache/maven/internal/impl/standalone/TestApiStandalone.java#L39-L49):
```
        Session session = ApiRunner.createSession();

        ArtifactCoordinate coord = session.createArtifactCoordinate("org.apache.maven:maven-api-core:4.0.0-alpha-13");
        Map.Entry<Artifact, Path> res = session.resolveArtifact(coord);
        assertNotNull(res);
        assertNotNull(res.getValue());
        assertTrue(Files.exists(res.getValue()));

        Node node = session.collectDependencies(session.createDependencyCoordinate(coord));
        assertNotNull(node);
        assertEquals(8, node.getChildren().size());
```


cstamas:
IIUC, this moves things out from maven-core to somewhere at line of maven-resolver-provider?

gnodet:
> IIUC, this moves things out from maven-core to somewhere at line of maven-resolver-provider?

Yes, the dependencies are:
  * maven-api-core (so the whole maven api)
  * maven-di (the dependency injection framework)
  * maven-resolver-provider
  * and obviously the resolver

The next step would be to make maven-resolver-provider and maven-model-builder built on top of the maven 4 model to make it even more lightweight.

cstamas (approved):

jira-importer:
Resolve #10798

PR #1442, opened 2024-03-15
[MNG-8088] Using Resolver with MRESOLVER-512

cstamas:
This makes all the new scopes implemented out of the box, but MNG-8041 still stands (is unsolved).

Related PR https://github.com/apache/maven-resolver/pull/445

---

https://issues.apache.org/jira/browse/MRESOLVER-512
https://issues.apache.org/jira/browse/MNG-8088


cstamas:
@gnodet ported over changes happened in https://github.com/apache/maven-resolver/pull/445 : removal of system scope handler.

cstamas:
Added both scope manager config to maven4. 

IMO, when mvn4 builds mvn3 project (model Ver 4.0.0), it should use mvn3 scope manager, but when uses newer model, it should use mvn4 scope manager.

gnodet (approved):

jira-importer:
Resolve #10791

PR #1443, opened 2024-03-16
Document requiredMavenVersion in plugin descriptor (since Maven 3.0.2)

hboutemy:
https://issues.apache.org/jira/browse/MNG-4840

but not integrated in documentation until now https://maven.apache.org/ref/3.9.6/maven-plugin-api/plugin.html


hboutemy:
wrong target branch...

PR #1444, opened 2024-03-16
[MNG-4840] document requiredMavenVersion in plugin descriptor

hboutemy:
https://issues.apache.org/jira/browse/MNG-4840

but not integrated in documentation until now https://maven.apache.org/ref/3.9.6/maven-plugin-api/plugin.html

### DIFF
--- a/maven-plugin-api/src/main/java/org/apache/maven/plugin/descriptor/PluginDescriptor.java
+++ b/maven-plugin-api/src/main/java/org/apache/maven/plugin/descriptor/PluginDescriptor.java
@@ -77,6 +77,7 @@ public class PluginDescriptor extends ComponentSetDescriptor implements Cloneabl
 
     private String description;
 
+    // MNG-4840
     private String requiredMavenVersion;
 
     private Plugin plugin;
@@ -295,10 +296,22 @@ public class PluginDescriptor extends ComponentSetDescriptor implements Cloneabl
         return description;
     }
 
+    /**
+     * Set required Maven version.
+     *
+     * @param requiredMavenVersion Maven version required by the plugin
+     * @since 3.0.2
+     */
     public void setRequiredMavenVersion(String requiredMavenVersion) {
         this.requiredMavenVersion = requiredMavenVersion;
     }
 
+    /**
+     * Get required Maven version.
+     *
+     * @return the Maven version required by the plugin
+     * @since 3.0.2
+     */
     public String getRequiredMavenVersion() {
         return requiredMavenVersion;
     }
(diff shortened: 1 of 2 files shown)


kwin:
@hboutemy I am confused by this PR. MNG-4840 iss not about plugin descriptors but about POMs. The required Maven version in plugin descriptors has been added in just recently and is only evaluated in Maven4+.

hboutemy:
I see you added the field to .mdo in https://github.com/apache/maven/pull/832/files#diff-62346c3118817c527a41027bab54349b463add4c91df98f95984012304378c87

the fact is that mdo was not used to generate code, but only to document XML against existing hand-written PluginDescriptor class https://github.com/apache/maven/blob/maven-3.9.x/maven-plugin-api/src/main/java/org/apache/maven/plugin/descriptor/PluginDescriptor.java
= something that IIUC changed in Maven 4

then the missing field in 3.9 .mdo was "just" a documentation bug that I fixed in this PR: now I understand that it became a missing field in code in Maven 4, and you already added it in #832 :)

kwin:
As you can see in https://github.com/apache/maven/blob/eae39474cbb5a88048cea76709321f20e2f42b0b/maven-plugin-api/src/main/java/org/apache/maven/plugin/descriptor/PluginDescriptorBuilder.java#L42-L106 this field was never evaluated by the parser (although already contained in the PluginDescriptor object). Therefore effectively prior to Maven 4 it is only ever set via https://github.com/apache/maven/blob/eae39474cbb5a88048cea76709321f20e2f42b0b/maven-core/src/main/java/org/apache/maven/plugin/internal/DefaultMavenPluginManager.java#L191 through the according POM field.
Therefore I would keep documenting that this is only supported with Plugin Descriptor v1.1 and Maven 4 only. Prior it was just a shortcut for accessing the plugin's POM prerequisites element.

hboutemy:
don't hesitate to update description: yes, history is complex between field addition and its usage

hboutemy:
history is complex: when the fiels has been introduced in 3.0.2, it has been filled https://github.com/apache/maven/commit/40fb1880ae8ea832e822e975168919bc56239dca#diff-8a738ae425b439d6a5288b118ce793df13ee73db9cfcb183f7b41633e72abad9L141

not the place where I'd expect it, but used
I don't know if this is still done in Maven 3.9: I'd be interested to see how the change in Maven 4 did manage the lacking field

hboutemy:
still there in Maven 3.9.x https://github.com/apache/maven/blob/maven-3.9.x/maven-core/src/main/java/org/apache/maven/plugin/internal/DefaultMavenPluginManager.java#L189

oh, now I see https://issues.apache.org/jira/browse/MNG-7570
the field is in plugin descriptor since 3.0.2, but not filled from plugin descriptor XML but from plugin artifact's pom.xml!!!

then now I understand that documenting the field in descriptor XML is a mistake
oh, I did not get that, it's something to be clarified in the field javadoc for 3.9.x and probably mdo

hboutemy:
I'll propose a PR to fix my mistakes

jira-importer:
Resolve #6268

PR #1445, opened 2024-03-16
[MNG-4840] document requiredMavenVersion in plugin descriptor

hboutemy:
same as #1444 for Maven 4


jira-importer:
Resolve #6268

PR #1446, opened 2024-03-18
[MNG-8081] interpolate available properties during default profile selection (Maven 4.x)

mbenson:
Evaluate profile activations after having interpolated embedded property constructs.

 - [X] Make sure there is a [JIRA issue MNG-8081](https://issues.apache.org/jira/browse/MNG-8081) filed
       pulling in other changes.

ICLA on file

### DIFF
--- a/maven-api-impl/src/main/java/org/apache/maven/internal/impl/model/DefaultModelBuilder.java
+++ b/maven-api-impl/src/main/java/org/apache/maven/internal/impl/model/DefaultModelBuilder.java
@@ -409,8 +409,8 @@ public class DefaultModelBuilder implements ModelBuilder {
             }
 
             @Override
-            public org.apache.maven.api.model.Profile apply(org.apache.maven.api.model.Profile p) {
-                return org.apache.maven.api.model.Profile.newBuilder(p)
+            public Profile apply(Profile p) {
+                return Profile.newBuilder(p)
                         .activation(transformActivation(p.getActivation()))
                         .build();
             }
--- a/maven-api-impl/src/main/java/org/apache/maven/internal/impl/model/DefaultModelBuilder.java
+++ b/maven-api-impl/src/main/java/org/apache/maven/internal/impl/model/DefaultModelBuilder.java
@@ -35,7 +35,9 @@ import java.util.concurrent.Callable;
 import java.util.concurrent.atomic.AtomicReference;
 import java.util.function.BiConsumer;
 import java.util.function.Supplier;
+import java.util.function.UnaryOperator;
 import java.util.stream.Collectors;
+import java.util.stream.Stream;
 
 import org.apache.maven.api.VersionRange;
 import org.apache.maven.api.annotations.Nullable;
@@ -50,6 +52,7 @@ import org.apache.maven.api.model.Dependency;
 import org.apache.maven.api.model.DependencyManagement;
 import org.apache.maven.api.model.Exclusion;
 import org.apache.maven.api.model.InputLocation;
+import org.apache.maven.api.model.InputLocationTracker;
 import org.apache.maven.api.model.InputSource;
 import org.apache.maven.api.model.Model;
 import org.apache.maven.api.model.Parent;
@@ -80,8 +83,11 @@ import org.apache.maven.api.services.xml.XmlReaderRequest;
 import org.apache.maven.internal.impl.InternalSession;
 import org.apache.maven.internal.impl.resolver.DefaultModelCache;
 import org.apache.maven.internal.impl.resolver.DefaultModelResolver;
+import org.apache.maven.model.v4.MavenTransformer;
 import org.codehaus.plexus.interpolation.InterpolationException;
+import org.codehaus.plexus.interpolation.Interpolator;
 import org.codehaus.plexus.interpolation.MapBasedValueSource;
+import org.codehaus.plexus.interpolation.RegexBasedInterpolator;
 import org.codehaus.plexus.interpolation.StringSearchInterpolator;
 import org.eclipse.aether.impl.RemoteRepositoryManager;
 
@@ -377,54 +383,75 @@ public class DefaultModelBuilder implements ModelBuilder {
 
     private List<Profile> interpolateActivations(
             List<Profile> profiles, DefaultProfileActivationContext context, DefaultModelProblemCollector problems) {
-        List<Profile> newProfiles = null;
-        for (int index = 0; index < profiles.size(); index++) {
-            Profile profile = profiles.get(index);
-            Activation activation = profile.getActivation();
-            if (activation != null) {
-                ActivationFile file = activation.getFile();
-                if (file != null) {
-                    String oldExists = file.getExists();
-                    if (isNotEmpty(oldExists)) {
+        if (profiles.stream()
+                .map(org.apache.maven.api.model.Profile::getActivation)
+                .noneMatch(Objects::nonNull)) {
+            return profiles;
+        }
+        final Interpolator xform = new RegexBasedInterpolator();
+        xform.setCacheAnswers(true);
+        Stream.of(context.getUserProperties(), context.getSystemProperties())
+                .map(MapBasedValueSource::new)
+                .forEach(xform::addValueSource);
+
+        class ProfileInterpolator extends MavenTransformer implements UnaryOperator<Profile> {
+            ProfileInterpolator() {
+                super(s -> {
+                    if (isNotEmpty(s)) {
                         try {
-                            String newExists = interpolate(oldExists, context);
-                            if (!Objects.equals(oldExists, newExists)) {
-                                if (newProfiles == null) {
-                                    newProfiles = new ArrayList<>(profiles);
-                                }
-                                newProfiles.set(
-                                        index, profile.withActivation(activation.withFile(file.withExists(newExists))));
-                            }
+                            return xform.interpolate(s);
                         } catch (InterpolationException e) {
-                            addInterpolationProblem(problems, file, oldExists, e, "exists");
-                        }
-                    } else {
-                        String oldMissing = file.getMissing();
-                        if (isNotEmpty(oldMissing)) {
-                            try {
-                                String newMissing = interpolate(oldMissing, context);
-                                if (!Objects.equals(oldMissing, newMissing)) {
-                                    if (newProfiles == null) {
-                                        newProfiles = new ArrayList<>(profiles);
-                                    }
-                                    newProfiles.set(
-                                            index,
-                                            profile.withActivation(activation.withFile(file.withMissing(newMissing))));
-                                }
-                            } catch (InterpolationException e) {
-                                addInterpolationProblem(problems, file, oldMissing, e, "missing");
-                            }
+                            problems.add(Severity.ERROR, ModelProblem.Version.BASE, e.getMessage(), e);
                         }
                     }
+                    return s;
+                });
+            }
+
+            @Override
+            public org.apache.maven.api.model.Profile apply(org.apache.maven.api.model.Profile p) {
+                return org.apache.maven.api.model.Profile.newBuilder(p)
+                        .activation(transformActivation(p.getActivation()))
+                        .build();
+            }
+
+            @Override
+            protected ActivationFile.Builder transformActivationFile_Missing(
+                    Supplier<? extends ActivationFile.Builder> creator,
+                    ActivationFile.Builder builder,
+                    ActivationFile target) {
+                String path = target.getMissing();
+                String xformed = transformPath(path, target, "missing");
+                return xformed != path ? (builder != null ? builder : creator.get()).missing(xformed) : builder;
+            }
+
+            @Override
+            protected ActivationFile.Builder transformActivationFile_Exists(
+                    Supplier<? extends ActivationFile.Builder> creator,
+                    ActivationFile.Builder builder,
+                    ActivationFile target) {
+                final String path = target.getExists();
+                final String xformed = transformPath(path, target, "exists");
+                return xformed != path ? (builder != null ? builder : creator.get()).exists(xformed) : builder;
+            }
+
+            private String transformPath(String path, ActivationFile target, String locationKey) {
+                if (isNotEmpty(path)) {
+                    try {
+                        return profileActivationFilePathInterpolator.interpolate(path, context);
+                    } catch (InterpolationException e) {
+                        addInterpolationProblem(problems, target, path, e, locationKey);
+                    }
                 }
+                return path;
             }
         }
-        return newProfiles != null ? newProfiles : profiles;
+        return profiles.stream().map(new ProfileInterpolator()).toList();
     }
 
     private static void addInterpolationProblem(
             DefaultModelProblemCollector problems,
-            ActivationFile file,
+            InputLocationTracker target,
             String path,
             InterpolationException e,
             String locationKey) {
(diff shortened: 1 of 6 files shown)


maven-model-builder/src/main/java/org/apache/maven/model/profile/DefaultProfileSelector.java
```diff
@@ -44,15 +54,53 @@
...
+        return m.entrySet().stream()
+                .collect(Collectors.toMap(e -> e.getKey(), e -> e.getValue(), (l, r) -> r, Properties::new));
+    }
+
     private final List<ProfileActivator> activators;
+    private ModelInterpolator interpolator;
 
     public DefaultProfileSelector() {
-        this.activators = new ArrayList<>();
+        this(new ArrayList<>(), new ModelInterpolator() {
```
gnodet, line 66: What about `null` instead ?

maven-model-builder/src/test/java/org/apache/maven/model/profile/DefaultProfileSelectorTest.java
```diff
@@ -0,0 +1,86 @@
...
+    private ModelInterpolator interpolator;
+
+    @BeforeEach
+    public void setup() {
+        interpolator = Mockito.mock(ModelInterpolator.class);
+        selector = new DefaultProfileSelector(Collections.singletonList(new PropertyProfileActivator()), interpolator);
+    }
+
+    @Test
+    public void testProfileActivationInterpolation() {
```
gnodet, line 50: This test seems to work if the mock interpolator simply returns the input model, thereby indicating that the test is wrong or that the feature is already implemented.

gnodet (changes requested):

maven-model-builder/src/main/java/org/apache/maven/model/profile/DefaultProfileSelector.java
```diff
@@ -108,6 +168,40 @@ public List<Profile> getActiveProfiles(
...
+
+        ModelBuildingRequest mbr = new DefaultModelBuildingRequest()
+                .setActiveProfileIds(context.getActiveProfileIds())
+                .setInactiveProfileIds(context.getInactiveProfileIds())
+                .setSystemProperties(asProperties(context.getSystemProperties()))
+                .setUserProperties(asProperties(context.getUserProperties()))
+                .setTwoPhaseBuilding(true)
+                .setValidationLevel(ModelBuildingRequest.VALIDATION_LEVEL_MINIMAL);
+
+        model = interpolator.interpolateModel(
```
gnodet, line 193: I wonder if using directly the `MavenTransformer` class would be easier instead of recreating a `ModelBuildingRequest` here....

maven-model-builder/src/test/java/org/apache/maven/model/profile/DefaultProfileSelectorTest.java
mbenson: You're right, in that the test is only setting `property.name` on the activation, thus the property simply being set would cause the profile to be selected. I will address this first.

mbenson:
Second iteration:
 * test case fails without updated code
 * using `MavenTransformer` for interpolation

gnodet:
The code looks good.

However, what's the real use case for that ? It seems actually unrelated to MNG-5235, in a sense that the behaviour on this project is not modified by this PR. My understanding is that the use case for MNG-5235 is to activate profile based on project properties, while this PR is about interpolating activation property values (but not based on project properties).  Also note that properties defined from activated external profiles are already taken into account, see [MNG-2276](https://issues.apache.org/jira/browse/MNG-2276).
Should this require a different JIRA issue ?  I'm not really sure to understand the use case, specifically why you'd want the activation part to possibly change from a maven run to another ?

gnodet:
> My use case is to be able to compare environment variables. Specifically, with GitLab CI, if `CI_COMMIT_REF_NAME` has the same value as `CI_DEFAULT_BRANCH`, then I want to enable stricter enforcer checks. For my purposes I'd rather the project make assumptions about the CI than vice versa.

Not sure how you'll model that with the limited support we have right now, do you ?
Maybe we should extend the model activation to support a simple language which provide basic operations such as `=` and `!=`.  Something like:
```
<property>
  <expression>${env. CI_COMMIT_REF_NAME} = ${env. CI_DEFAULT_BRANCH}</expression>
</property>
```
At some point, it may be easier to provide a custom Maven extension that would compute a system property based on those two variables so you can easily activate a profile based on the computed property.  Maven 4 added `org.apache.maven.api.spi.PropertyContributor` for such use case (to have a cleaner api, but it's possible in Maven 3 using the `AbstractMavenLifecycleParticipant`.  See https://github.com/trustin/os-maven-plugin/tree/master and https://github.com/trustin/os-maven-plugin/pull/74 
An example

mbenson:
Hmm, my approach was, e.g.:

```
<activation>
  <property>
    <name>CI_COMMIT_REF_NAME</name>
    <value>${env.CI_DEFAULT_BRANCH}</value>
  </property>
</activation>
```

...which seems fine to me 🤷

rmannibucau:
personally i do it in the sh part of the yaml:

````
case ${DEPLOY_ENV:-dev} in
  dev)
    export SCRIPTS_DEPLOY_DEPLOYMENT_PROFILES="-Pshared-dev -Dmonitoring.enabled=true"
    export SCRIPTS_DEPLOY_DEPLOYMENT_ALVEOLUS="minikube"
    ;;
  *)
    export SCRIPTS_DEPLOY_DEPLOYMENT_PROFILES="-P$DEPLOY_ENV"
    export SCRIPTS_DEPLOY_DEPLOYMENT_ALVEOLUS="${SCRIPTS_DEPLOY_DEPLOYMENT_ALVEOLUS:-all-in-one}"
    ;;
esac

# this shared script just runs mvn with the SCRIPTS_DEPLOY_... options but the magic literally happens in previous case and is open bar on the CI
/usr/local/bin/deploy.sh
````

No assumption needed all is detected in the build pipeline (where it belongs) and project itself just enables to activate it or not.

**Comment by mbenson** [comment]
@rmannibucau but your pipeline has to know/hope that the project has such a `shared-dev` profile. For my purposes my pipeline wouldn't know what such a profile might be named, but my org's projects do know what CI they run on (in the worst case, migrating to a different CI--which won't actually happen, but if it did--it would be pretty easy to make XXX CI emulate GitLab CI by setting the same environment variables).

**Comment by gnodet** [comment]
> Hmm, my approach was, e.g.:
> 
> ```
> <activation>
>   <property>
>     <name>env.CI_COMMIT_REF_NAME</name>
>     <value>${env.CI_DEFAULT_BRANCH}</value>
>   </property>
> </activation>
> ```
> 
> ...which seems fine to me 🤷

Yes, obviously that'll work :-)

File: maven-model-builder/src/main/java/org/apache/maven/model/profile/DefaultProfileSelector.java [hunk 4]
```diff
@@ -76,16 +85,24 @@ public List<org.apache.maven.api.model.Profile> getActiveProfilesV4(
     @Override
     public List<Profile> getActiveProfiles(
             Collection<Profile> profiles, ProfileActivationContext context, ModelProblemCollector problems) {
+
+        if (profiles.stream().map(Profile::getId).distinct().count() < profiles.size()) {
```
gnodet, line 89: What is this test about ? That seems unrelated to this PR.  Should this be rather validated (with a warning printed) earlier in the process ? During settings validation and project validation ?

rmannibucau:
@mbenson see it the other way, `if (magic) -DdoItSlow` then you can use this new fake variable accordingly. You can migrate but the easiest is a custom base image IMHO which does this for you and abstracts this logic.

mbenson:
@rmannibucau I think it's okay if different users want to interact with Maven in different ways. Your preference does not invalidate mine.

mbenson:
@gnodet I agree that MNG-5235 has a different ask than this PR addresses. I will open a new issue and rewrite this commit message accordingly.

rmannibucau:
@mbenson well a mini language got rejected multiple times on the list - this is where this sh solution comes from. ultimately I'm for having a multipass filtering to filter as much as we can but my personal preference would be to do it on the full model and not specifically for profiles. That said, while v4 has the exact same code than v3.9 I'm happy.

gnodet:
> @mbenson well a mini language got rejected multiple times on the list - this is where this sh solution comes from. ultimately I'm for having a multipass filtering to filter as much as we can but my personal preference would be to do it on the full model and not specifically for profiles. That said, while v4 has the exact same code than v3.9 I'm happy.

It seems it can be solved really easily though:
  https://github.com/kpiwko/el-profile-activator-extension/blob/master/src/main/java/com/redhat/jboss/maven/elprofile/ElProfileActivator.java

gnodet:
> > @mbenson well a mini language got rejected multiple times on the list - this is where this sh solution comes from. ultimately I'm for having a multipass filtering to filter as much as we can but my personal preference would be to do it on the full model and not specifically for profiles. That said, while v4 has the exact same code than v3.9 I'm happy.
> 
> It seems it can be solved really easily though: https://github.com/kpiwko/el-profile-activator-extension/blob/master/src/main/java/com/redhat/jboss/maven/elprofile/ElProfileActivator.java

That makes we wonder if this PR targets the wrong location.  Should we modify the [`PropertyProfileActivator`](https://github.com/apache/maven/blob/83949a73841683fc5a3347d9c93a71737ed2911b/maven-model-builder/src/main/java/org/apache/maven/model/profile/activation/PropertyProfileActivator.java) instead ?

mbenson:
That was what I originally considered, but I quickly came to the opinion that as soon as we did it that way, someone would want to use some such property to match a Java version, etc., and that this would be a generally applicable way to solve a broad cross section of similar use cases.

gnodet:
> That was what I originally considered, but I quickly came to the opinion that as soon as we did it that way, someone would want to use some such property to match a Java version, etc., and that this would be a generally applicable way to solve a broad cross section of similar use cases.

How could that be considered a bad thing ?

mbenson:
> > That was what I originally considered, but I quickly came to the opinion that as soon as we did it that way, someone would want to use some such property to match a Java version, etc., and that this would be a generally applicable way to solve a broad cross section of similar use cases.
> 
> How could that (solving similar use cases) be considered a bad thing ?

That's what I wanted to know 😮 
i.e., I think this approach takes care of them in one place before everybody comes asking for similar changes to be made everywhere.

mbenson:
From email thread:

Guillaume Nodet <gnodet@apache.org> wrote:
> I thought you were referring to having a small language in the
property activation...

I hadn't noted the discussion when it took place before, but fwiw I also lean in the direction that an embedded language might be going too far. At the same time EL, as in the extension you mentioned, seems a fine choice, but feels fine to leave that as an extension.

Matt

gnodet:
> From email thread:
> 
> Guillaume Nodet [gnodet@apache.org](mailto:gnodet@apache.org) wrote:
> 
> > I thought you were referring to having a small language in the
> > property activation...
> 
> I hadn't noted the discussion when it took place before, but fwiw I also lean in the direction that an embedded language might be going too far. At the same time EL, as in the extension you mentioned, seems a fine choice, but feels fine to leave that as an extension.
> 
> Matt

@mbenson  I just spotted that the model builder performs some interpolation for file base activation in https://github.com/apache/maven/blob/5029cc238ce5ce03b0fb431a278889f8b788c0a7/maven-model-builder/src/main/java/org/apache/maven/model/building/DefaultModelBuilder.java#L896-L943
Would it make sense to move this code in this block to interpolate ?  It seems more intuitive that the two interpolations would actually be performed at the same time...

mbenson:
That is possible, sure. I'll take a look in a bit.

mbenson:
Here is the version that enhances `DefaultModelBuilder`. Can anyone offer ideas on the likely shape of unit tests for this approach?

maven-model-builder/src/main/java/org/apache/maven/model/building/DefaultModelBuilder.java
```diff
@@ -897,69 +904,89 @@ private List<org.apache.maven.api.model.Profile> interpolateActivations(
...
+                        return profileActivationFilePathInterpolator.interpolate(path, context);
+                    } catch (InterpolationException e) {
+                        addInterpolationProblem(problems, target, path, e, locationKey);
+                    }
+                }
+                return path;
+            }
         }
-        return newProfiles != null ? newProfiles : profiles;
+        return profiles.stream().map(new ProfileInterpolator()).collect(Collectors.toList());
```
gnodet, line 973: `.collect(Collectors.toList())` -> `.toList()`

gnodet (approved):

mbenson:
By way of [some] testing, I went ahead and expanded the profile activation model validation to consider the whole activation, similarly to how we are now interpolating the whole activation.

Here I could be construed as misusing the `MavenTransformer` to walk the activation; if this is deemed offensive we could consider generating a similar `MavenWalker` class.

rmannibucau (approved):

mbenson:
Thanks @gnodet and @rmannibucau for the approvals! What is the next step?

cstamas (approved):

gnodet:
> Thanks @gnodet and @rmannibucau for the approvals! What is the next step?

I'm nearly done with http://github.com/apache/maven/pull/1457, so as soon as that one is merged, we'll be able to port this PR to the new model builder easily.

mbenson:
@gnodet do you have any thoughts on the best way to backport this feature to Maven 3? https://github.com/apache/maven/pull/1447 is my original PR, but a wildly divergent implementation may cause headaches in future. I wonder if it would make sense [from a future merge perspective] to stabilize the v3 impl and include the same commit in this branch prior to being replaced by the actual v4 implementation. 🤔

gnodet:
@mbenson I've rebased the PR and ported it to the new v4 model builder, please have a look

mbenson:
hi @gnodet ... so with the implementation having been largely duplicated from model-builder into api-impl, would the similar code remain in both places indefinitely, or will the model-builder code eventually point directly to api-impl?

gnodet:
> hi @gnodet ... so with the implementation having been largely duplicated from model-builder into api-impl, would the similar code remain in both places indefinitely, or will the model-builder code eventually point directly to api-impl?

The code in model-builder should be deprecated, it will be kept for compatibility with 3.x plugins for some time, and then removed.  Not sure about the exact dates though.  The point is that model-builder is not really used anymore in beta-1.

maven-api-impl/src/main/java/org/apache/maven/internal/impl/model/DefaultModelBuilder.java
```diff
@@ -377,69 +383,88 @@ private Model readEffectiveModel(
...
-                ActivationFile file = activation.getFile();
-                if (file != null) {
-                    String oldExists = file.getExists();
-                    if (isNotEmpty(oldExists)) {
+        if (profiles.stream()
+                .map(org.apache.maven.api.model.Profile::getActivation)
+                .filter(Objects::nonNull)
+                .findFirst()
+                .isEmpty()) {
+            return profiles;
```
gnodet, line 391: `.noneMatch(Objects::nonNull)` instead ?

maven-model-builder/src/main/java/org/apache/maven/model/building/DefaultModelBuilder.java
```diff
@@ -897,69 +904,91 @@ private List<org.apache.maven.api.model.Profile> interpolateActivations(
...
-            if (activation != null) {
-                org.apache.maven.api.model.ActivationFile file = activation.getFile();
-                if (file != null) {
-                    String oldExists = file.getExists();
-                    if (isNotEmpty(oldExists)) {
+        if (profiles.stream()
+                .map(org.apache.maven.api.model.Profile::getActivation)
+                .filter(Objects::nonNull)
+                .findFirst()
+                .isEmpty()) {
```
gnodet, line 911: use `anyMatch` here too

mbenson:
Incorporated `noneMatch` as suggested. IDK how I missed that pair of `Stream` methods all this time. 😕

mbenson:
@gnodet again, thanks!

jira-importer:
Resolve #6907

jira-importer:
Resolve #6907

jira-importer:
Resolve #10796

jira-importer:
Resolve #10796

PR #1447, opened 2024-03-18
[MNG-8081] Interpolate available properties during default profile selection (Maven 3.9.x)

mbenson:
Using the available `ModelInterpolator`, evaluate profile activations after having interpolated any embedded property constructs among configurations of activations only.

       pulling in other changes.

ICLA on file

### DIFF
--- a/maven-model-builder/src/test/java/org/apache/maven/model/validation/DefaultModelValidatorTest.java
+++ b/maven-model-builder/src/test/java/org/apache/maven/model/validation/DefaultModelValidatorTest.java
@@ -762,20 +762,17 @@ public class DefaultModelValidatorTest extends TestCase {
                 result.getWarnings().get(0));
     }
 
-    public void repositoryWithExpression() throws Exception {
+    public void testRepositoryWithExpression() throws Exception {
         SimpleProblemCollector result = validateRaw("raw-model/repository-with-expression.xml");
-        assertViolations(result, 0, 1, 0);
-        assertEquals(
-                "'repositories.repository.[repo].url' contains an expression but should be a constant.",
-                result.getErrors().get(0));
+        assertViolations(result, 0, 0, 0);
     }
 
-    public void repositoryWithBasedirExpression() throws Exception {
+    public void testRepositoryWithBasedirExpression() throws Exception {
         SimpleProblemCollector result = validateRaw("raw-model/repository-with-basedir-expression.xml");
         assertViolations(result, 0, 0, 0);
     }
 
-    public void profileActivationWithAllowedExpression() throws Exception {
+    public void testProfileActivationWithAllowedExpression() throws Exception {
         SimpleProblemCollector result = validateRaw(
                 "raw-model/profile-activation-file-with-allowed-expressions.xml",
                 mbr -> mbr.setUserProperties(new Properties() {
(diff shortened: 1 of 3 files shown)


maven-model-builder/src/main/java/org/apache/maven/model/profile/DefaultProfileSelector.java
```diff
@@ -44,29 +54,56 @@
...
+    private static Properties asProperties(Map<String, String> m) {
+        return m.entrySet().stream()
+                .collect(Collectors.toMap(e -> e.getKey(), e -> e.getValue(), (l, r) -> r, Properties::new));
+    }
+
     @Inject
     private List<ProfileActivator> activators = new ArrayList<>();
 
+    @Inject
+    private ModelInterpolator interpolator = new ModelInterpolator() {
```
rmannibucau, line 66: `@Inject` = new? :thinking:

maven-model-builder/src/main/java/org/apache/maven/model/profile/DefaultProfileSelector.java
```diff
@@ -89,6 +126,35 @@ public List<Profile> getActiveProfiles(
...
+
+        UnaryOperator<Profile> activatableProfile = p -> {
+            Profile result = new Profile();
+            result.setId(p.getId());
+            result.setActivation(p.getActivation());
+            return result;
+        };
+        model.setProfiles(original.stream().map(activatableProfile).collect(Collectors.toList()));
+
+        ModelBuildingRequest mbr = new DefaultModelBuildingRequest()
```
rmannibucau, line 142: guess it would be saner to use the actual loaded model we already have at that time and avoid `asProperties` too since request is already there

can be done in `DefaultModelBuilder` probably

side note: we must somehow align it on `master` and not create a behavior in v3 which is different from v4 so for me ensuring v4 behavior (merge) and impl alignment is a prerequisite to this PR.

maven-model-builder/src/main/java/org/apache/maven/model/profile/DefaultProfileSelector.java
```diff
@@ -89,6 +126,35 @@ public List<Profile> getActiveProfiles(
...
+                .setSystemProperties(asProperties(context.getSystemProperties()))
+                .setUserProperties(asProperties(context.getUserProperties()))
+                .setTwoPhaseBuilding(true)
+                .setValidationLevel(ModelBuildingRequest.VALIDATION_LEVEL_MINIMAL);
+
+        interpolator
+                .interpolateModel(model, context.getProjectDirectory(), mbr, problem -> {})
+                .getProfiles();
+
+        return model.getProfiles().stream().collect(Collectors.toMap(Profile::getId, UnaryOperator.identity()));
```
rmannibucau, line 155: identity belongs to Function IIRC

maven-model-builder/pom.xml
```diff
@@ -87,6 +87,11 @@ under the License.
       <artifactId>xmlunit-matchers</artifactId>
       <scope>test</scope>
     </dependency>
+    <dependency>
```
rmannibucau, line 90: :pray: why mocking there? this is a hell anywhere in maven code base and ultimately the related tests are useless in general so let's try to avoid it

rmannibucau (changes requested):
can we ensure the alignment on v4 before working on v3 :pray: ?

gnodet (changes requested):
This is wrong imho, the interpolation should be done in the `ModelBuilder` during [profile activation interpolation](https://github.com/apache/maven/blob/maven-3.9.x/maven-model-builder/src/main/java/org/apache/maven/model/building/DefaultModelBuilder.java#L427-L450)

mbenson:
Sounds like this approach would keep the v3/v4 impls more similar. I'll work on this.

gnodet (approved):

mbenson:
Thanks @gnodet !

jira-importer:
Resolve #6907

PR #1448, opened 2024-03-21
[MNG-8052] New lifecycle for Maven 4

gnodet:
JIRA issue: [MNG-8052](https://issues.apache.org/jira/browse/MNG-8052)

The lifecycle is now defined as a tree of phases, each phase being split with a `before:` phase and a `after:` phase.  Each phase is given a list of timing constraints: a phase from the same lifecycle (`compile` must be executed after `sources`), a phase in project dependencies (`compile` must execute after `compile-only` project dependencies have reached the `ready` phase), or a set of children project dependencies (to re-define aggregators, not yet implemented).
The default lifecycle is defined in the [`DefaultLifecycleRegistry`](https://github.com/gnodet/maven/blob/dynamic-phases/maven-core/src/main/java/org/apache/maven/internal/impl/DefaultLifecycleRegistry.java#L349-L379)

In addition to the `before:` and `after:` prefixes, an ordering can be defined by appending an integer inside brackets, for example, `after:integration-test[1000]`.

Note that there are a few changes with the Maven 3 default lifecycle: it's a graph, so `sources` does not always execute after `resources`, nor `compile` after `resources`.  Also, unit tests and integration tests have been moved to the `verify` phase which is run last inside the `build` phase, but not in the `package` phase.  The goal is to have a phase (here, `package` which can run all the reactor with no tests).  In order to be compatible, old phases are mapped to new ones using [aliases](https://github.com/gnodet/maven/blob/lifecycle/maven-core/src/main/java/org/apache/maven/internal/impl/DefaultLifecycleRegistry.java#L381-L393).  The `install` and `deploy` phases now depends on `package`, but not `verify` (and `deploy` does not require `install`).  This currently has no effect when calling `mvn deploy`, as the ordered _list_ of phases is still used to compute the build plan (see #1429).

It's missing the ability to create some scheduling constraints in the POM and to define custom phases.  All executions in a given phase such as `compile` or `after:sources` are executed sequentially, but it would be nice to be able to execute them in different subphases, so that they could be executed concurrently.

This PR is required for #1429

### DIFF
--- a/api/maven-api-core/src/main/java/org/apache/maven/api/Lifecycle.java
+++ b/api/maven-api-core/src/main/java/org/apache/maven/api/Lifecycle.java
@@ -21,9 +21,11 @@ package org.apache.maven.api;
 import java.util.Collection;
 import java.util.List;
 import java.util.Optional;
+import java.util.stream.Stream;
 
 import org.apache.maven.api.annotations.Experimental;
 import org.apache.maven.api.annotations.Immutable;
+import org.apache.maven.api.annotations.Nonnull;
 import org.apache.maven.api.model.Plugin;
 
 /**
@@ -41,13 +43,19 @@ import org.apache.maven.api.model.Plugin;
 @Immutable
 public interface Lifecycle extends ExtensibleEnum {
 
+    // =========================
+    // Maven defined lifecycles
+    // =========================
     String CLEAN = "clean";
-
     String DEFAULT = "default";
-
     String SITE = "site";
-
     String WRAPPER = "wrapper";
+
+    // ======================
+    // Phase qualifiers
+    // ======================
+    String BEFORE = "before:";
+    String AFTER = "after:";
 
     /**
      * Name or identifier of this lifecycle.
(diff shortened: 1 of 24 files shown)


oehme:
@gnodet One of our tests broke after this change made it into the nightlies - We test a custom lifecycle like 
``` 
<role>org.apache.maven.lifecycle.Lifecycle</role>
<role-hint>custom_hint</role-hint>
<implementation>org.apache.maven.lifecycle.Lifecycle</implementation>
<configuration>
  <id>custom_lifecycle</id>
  <phases>
    <phase>phase1</phase>
    <phase>phase2</phase>
    <phase>phase3</phase>
  </phases>
...
``` 

Before this change, the phases were executed in the correct order, i.e. calling `phase3` would first run `phase1` and `phase2`. After this change, the order is now reversed.

Looking through the Maven integration tests, there doesn't seem to be a test case for custom lifecycles, so this went undiscovered.

gnodet:
> @gnodet One of our tests broke after this change made it into the nightlies - We test a custom lifecycle like
> 
> ```
> <role>org.apache.maven.lifecycle.Lifecycle</role>
> <role-hint>custom_hint</role-hint>
> <implementation>org.apache.maven.lifecycle.Lifecycle</implementation>
> <configuration>
>   <id>custom_lifecycle</id>
>   <phases>
>     <phase>phase1</phase>
>     <phase>phase2</phase>
>     <phase>phase3</phase>
>   </phases>
> ...
> ```
> 
> Before this change, the phases were executed in the correct order, i.e. calling `phase3` would first run `phase1` and `phase2`. After this change, the order is now reversed.
> 
> Looking through the Maven integration tests, there doesn't seem to be a test case for custom lifecycles. And since maven-core no longer uses components.xml itself, this went undiscovered.

Could you raise a JIRA issue and provide a small project to reproduce the problem ?

jira-importer:
Resolve #6656

PR #1449, opened 2024-03-21
[MNG-8082] - Exceptions of proxied SessionScoped components are not working correctly

jonasrutishauser:
Simple fix to handle the exception case correctly.

### DIFF
--- a/maven-core/src/main/java/org/apache/maven/session/scope/internal/SessionScope.java
+++ b/maven-core/src/main/java/org/apache/maven/session/scope/internal/SessionScope.java
@@ -107,7 +107,11 @@ public class SessionScope implements Scope {
     private <T> T createProxy(Key<T> key, Provider<T> unscoped) {
         InvocationHandler dispatcher = (proxy, method, args) -> {
             method.setAccessible(true);
-            return method.invoke(getScopeState().scope(key, unscoped).get(), args);
+            try {
+                return method.invoke(getScopeState().scope(key, unscoped).get(), args);
+            } catch (InvocationTargetException e) {
+                throw e.getCause();
+            }
         };
         Class<T> superType = (Class<T>) key.getTypeLiteral().getRawType();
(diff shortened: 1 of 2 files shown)


cstamas (approved):

rmannibucau (approved):

maven-core/src/main/java/org/apache/maven/session/scope/internal/SessionScope.java
```diff
@@ -107,7 +107,11 @@ public <T> Provider<T> scope(final Key<T> key, final Provider<T> unscoped) {
     private <T> T createProxy(Key<T> key, Provider<T> unscoped) {
         InvocationHandler dispatcher = (proxy, method, args) -> {
             method.setAccessible(true);
-            return method.invoke(getScopeState().scope(key, unscoped).get(), args);
+            try {
+                return method.invoke(getScopeState().scope(key, unscoped).get(), args);
+            } catch (InvocationTargetException e) {
+                throw e.getCause();
```
CrazyHZM, line 113: If the target Exception is null, it will throw NPE, and this could affect exception tracing.

CrazyHZM (changes requested):

maven-core/src/main/java/org/apache/maven/session/scope/internal/SessionScope.java
jonasrutishauser: the variable `e` can never be `null`.

The statement `e.getCause()` can be `null` in theory, but in practice I don't see any case which could give this.

CrazyHZM: `e.getCause()` return `target`. I mean e.g Cause() is null, which can affect line numbers

gnodet (approved):

jira-importer:
Resolve #10795

PR #1455, opened 2024-03-25
Fix split package in org.apache.maven.api.di

gnodet:
@hboutemy


cstamas (approved):

hboutemy (approved):
perhaps we should enable JPMS on the new Maven 4 API to avoid mistakes in the future

PR #1457, opened 2024-03-27
[MNG-8084] Move ModelBuilder and resolver provider to v4 api

gnodet:
This provides a re-implementation of `maven-model-builder` and `maven-resolver-provider` on top of the Maven 4 api.
As a consequence, apart from projects, plugins and actual build, all the v4 api and resolution should be available outside of maven with just the api, `maven-di`, `maven-api-impl` and the resolver.

Note that the builder and resolver provider are not yet used by maven.


cstamas:
Maybe "v3 mode" should use old stuff, and "v4 mode" should use this new stuff? As we also have "v3 scopes vs v4" scopes, and v4 scopes should not be used with v3 modes...

gnodet:
> Maybe "v3 mode" should use old stuff, and "v4 mode" should use this new stuff? As we also have "v3 scopes vs v4" scopes, and v4 scopes should not be used with v3 modes...

How would you decide which one is used ?
The `Maven3ScopeManagerConfiguration` is not used atm afaik.  I just copied those to accommodate the recent changes.

I'm working on rewriting the ProjectBuilder on top of the new ModelBuilder, to at least fix any regression using unit tests and integration tests.

Once we have full support, we could think of having two different builders at the same time if we need to configure them differently.  What kind of changes would you see ?

cstamas:
By project model version?

There is an important distinction:
[mvn3](https://github.com/apache/maven/blob/master/maven-resolver-provider/src/main/java/org/apache/maven/repository/internal/scopes/Maven3ScopeManagerConfiguration.java#L137-L142) vs [mvn4](https://github.com/apache/maven/blob/master/maven-resolver-provider/src/main/java/org/apache/maven/repository/internal/scopes/Maven4ScopeManagerConfiguration.java#L165-L170) RUNTIME resolution scopes, former suffer from https://issues.apache.org/jira/browse/MNG-8041 while latter has it fixed.

gnodet:
> By project model version?
> 
> There is an important distinction: [mvn3](https://github.com/apache/maven/blob/master/maven-resolver-provider/src/main/java/org/apache/maven/repository/internal/scopes/Maven3ScopeManagerConfiguration.java#L137-L142) vs [mvn4](https://github.com/apache/maven/blob/master/maven-resolver-provider/src/main/java/org/apache/maven/repository/internal/scopes/Maven4ScopeManagerConfiguration.java#L165-L170) RUNTIME resolution scopes, former suffer from https://issues.apache.org/jira/browse/MNG-8041 while latter has it fixed.

Right now, mvn3 scopes are not used, so we need those at all ?

cstamas:
I think you are right: mvn4 scopes are "superset" of those of mvn3, plus, if you compare their ordering, they remain in same place (ordering by "width"), this is the two config "dump" output (they have main method that produces this):

Maven3
```
Maven3 defined dependency scopes:
compile (width=3500)
  Query : [ALL]
  Presence: [main-compile, test, main-runtime]
  Main project scope: main-compile
system (width=2500)
  Query : [ALL]
  Presence: [main-compile, test, main-runtime]
  Main project scope: main-compile
runtime (width=2500)
  Query : [BY_BUILD_PATH(runtime)]
  Presence: [test, main-runtime]
  Main project scope: main-runtime
provided (width=1500)
  Query : [BY_BUILD_PATH(compile), SELECT(test, runtime)]
  Presence: [main-compile, test]
  Main project scope: main-compile
test (width=500)
  Query : [BY_PROJECT_PATH(test)]
  Presence: [test]
  Main project scope: test
```

Maven4
```
Maven4 defined dependency scopes:
compile (width=4000)
  Query : [ALL]
  Presence: [main-compile, main-runtime, test-compile, test-runtime]
  Main project scope: main-compile
system (width=3000)
  Query : [ALL]
  Presence: [main-compile, main-runtime, test-compile, test-runtime]
  Main project scope: main-compile
runtime (width=2500)
  Query : [BY_BUILD_PATH(runtime)]
  Presence: [main-runtime, test-runtime]
  Main project scope: main-runtime
provided (width=2000)
  Query : [BY_BUILD_PATH(compile), SELECT(test, runtime)]
  Presence: [main-compile, test-compile, test-runtime]
  Main project scope: main-compile
compile-only (width=1000)
  Query : [SINGLETON(main, compile)]
  Presence: [main-compile]
  Main project scope: main-compile
test (width=1000)
  Query : [BY_PROJECT_PATH(test)]
  Presence: [test-compile, test-runtime]
  Main project scope: test-compile
test-only (width=500)
  Query : [SINGLETON(test, compile)]
  Presence: [test-compile]
  Main project scope: test-compile
test-runtime (width=500)
  Query : [SINGLETON(test, runtime)]
  Presence: [test-runtime]
  Main project scope: test-runtime
none (width=0)
  Query : []
  Presence: []
  Main project scope: null
```

api/maven-api-core/src/main/java/org/apache/maven/api/services/ModelBuilderResult.java
```diff
@@ -54,22 +54,28 @@ public interface ModelBuilderResult {
     Model getFileModel();
 
     /**
-     * Gets the assembled model.
-     *
-     * @return The assembled model, never {@code null}.
+     * Returns the file model + profile injection.
+     * This
```
cstamas, line 58: This?

cstamas (approved):

cstamas (approved):

desruisseaux:
A few more changes seem to be necessary for allowing `maven-plugin-tools` and `maven-plugin-testing` to compile. I created a commit there, but no pull request yet because I'm not sure if I got the intent right: https://github.com/Geomatys/maven/commit/60a60503dda6e352a680774d7e3ab3b71c2b5c11

PR #1459, opened 2024-03-27
[MNG-8089] Introduce fat-JAR types

cstamas:
These are dependency types to be used when dependency is "self contained", or is wanted be used as such: handle if like it has no dependencies, if any.

---

https://issues.apache.org/jira/browse/MNG-8089

### DIFF
--- a/api/maven-api-core/src/main/java/org/apache/maven/api/Type.java
+++ b/api/maven-api-core/src/main/java/org/apache/maven/api/Type.java
@@ -61,10 +61,9 @@ public interface Type extends ExtensibleEnum {
     String JAR = "jar";
 
     /**
-     * Artifact type name for a fat-JAR file that can be placed either on the class-path or on the module-path.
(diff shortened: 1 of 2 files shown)


cstamas:
@desruisseaux ping

desruisseaux:
I suggest to provide only the `fatjar` type, without the `classpath-fatjar` and `modular-fatjar` for now, and modify the Javadoc of `fatjar` for saying that it goes on the classpath only. The reason is that a JAR file can contain at most one `module-info.class` file. So either the `module-info` or all JARs had to be removed, or either they were merged in some way. But the latter would be tricky, and the result would still not be the same as the separated JARs. So I think it is probably safer to said that a fat JAR is never modular.

cstamas:
This makes perfect sense, will do the changes soon

cstamas:
Updated PR

gnodet (approved):

jira-importer:
Resolve #10793

PR #1460, opened 2024-03-29
[MNG-8091] Update Resolver to 2.0.0-alpha-10

cstamas:
~~Currently is under vote, so using staging repository, that will be removed once/if vote passes.~~ Vote passed, but Maven Central sync is busted.

Changes:
* update Resolver to 2.0.0-alpha-10
* Maven must not depend on supplier, it should be other way around.

Once this is merged (and released), Resolver gets these changes:
https://github.com/apache/maven-resolver/pull/454

---

https://issues.apache.org/jira/browse/MNG-8091

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -652,7 +652,8 @@ under the License.
         <enabled>false</enabled>
       </snapshots>
       <id>maven-2080</id>
-      <url>https://repository.apache.org/content/repositories/maven-2080/</url>
+      <!-- Maven Central sync is busted: is released but not published yet -->
+      <url>https://repository.apache.org/content/repositories/releases/</url>
     </repository>
   </repositories>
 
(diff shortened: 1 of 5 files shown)


pom.xml
```diff
@@ -641,6 +641,22 @@ under the License.
   </dependencies>
   <!--bootstrap-end-comment-->
 
+  <repositories>
```
gnodet, line 644: Should we get rid of this now ? 
```
➜  maven git:(model-buider-v4-b) ✗ mima list org.apache.maven.resolver:maven-resolver-connector-basic | grep 2.0.0-alpha-10
org.apache.maven.resolver:maven-resolver-connector-basic:2.0.0-alpha-10
```

cstamas:
```
[cstamas@angeleyes ~]$ jbang mima@maveniverse exists org.apache.maven.resolver:maven-resolver-api:2.0.0-alpha-10
Artifact org.apache.maven.resolver:maven-resolver-api:jar:2.0.0-alpha-10 NOT EXISTS

Checked TOTAL of 1 (existing: 0 not existing: 1)
[cstamas@angeleyes ~]$ 
```

cstamas: List GA uses _maven metadata_, that is on other synced somehow...

gnodet (approved):

PR #1461, opened 2024-03-31
[MNG-8084] add di & api-impl

hboutemy:
also drop model-transform deleted in MNG-7877

published to https://maven.apache.org/ref/4-LATEST/


hboutemy:
notice that there is a chicken and egg problem between resolver and Maven: api-impl depends on resolver supplier, which depends on resolver-provider

cstamas:
Supplier is to be moved to api-impl, as it is not meant to be used in Maven (only in projects that "run maven without maven", ie. resolver integrations w/ maven models). The plan is to move supplier to api-impl, and have resolver supplier depends on api-impl instead. @gnodet

cstamas:
See https://github.com/apache/maven/pull/1460

gnodet:
Fwiw, this will be outdated with https://github.com/apache/maven/pull/1465 and later https://github.com/apache/maven/pull/1457

gnodet:
> Fwiw, this will be outdated with #1465 and later #1457

Also maven-compat is not used anymore (but only included in the distribution for compatibility).

gnodet:
maven-4 0 0-alpha-14-SNAPSHOT

It's really readable unfortunately, but it started from generated graph, so it's correct with current master.

gnodet:
> maven-4 0 0-alpha-14-SNAPSHOT
> 
> It's not readable unfortunately, but it started from generated graph, so it's correct with current master.

Note that the ultimate goal would be, imho, to have a maven-compat3 with all the things from maven 3, and have `maven-core` not depend on the v3 artifacts / api anymore.

hboutemy:
PR updated https://maven.apache.org/ref/4-LATEST/
objective of this PR is to show what we have today, that will help us define targets: schema will evolve when useful

gnodet (approved):
Looks good enough to me, even though it's not perfectly inline with current status.

hboutemy:
I'm happy to update it if you tell me what remains to do: it's just a small update of an OpenOffice document, save as .svg and run the update `prepare-svg.sh` script

gnodet:
> I'm happy to update it if you tell me what remains to do: it's just a small update of an OpenOffice document, save as .svg and run the update `prepare-svg.sh` script
> 
> > It's not readable unfortunately, but it started from generated graph, so it's correct with current master.
> 
> BTW, I'd be interested to be able to create this type of automatic drawing as a reference

`mvn org.fusesource.mvnplugins:maven-graph-plugin:reactor` does render something automatically...
reactor-graph

jira-importer:
Resolve #10798

PR #1463, opened 2024-04-01
[MNG-8043] document new default types, replacing artifact handlers

hboutemy:


maven-resolver-provider/src/site/apt/dependency-types.apt
```diff
@@ -0,0 +1,74 @@
...
+*-----------------------+---------------+------------+-----------+-----------------------+-----------------------+
+| <<<javadoc>>>         | <<<javadoc>>> | <<<jar>>>  | java      | classes               |                       |
+*-----------------------+---------------+------------+-----------+-----------------------+-----------------------+
+| <<<test-jar>>>        | <<<tests>>>   | <<<jar>>>  | java      | classes, patch module |                       |
+*-----------------------+---------------+------------+-----------+-----------------------+-----------------------+
+| <<<modular-jar>>> *   | <<<tests>>>   | <<<jar>>>  | java      | modules               |                       |
+*-----------------------+---------------+------------+-----------+-----------------------+-----------------------+
+| <<<classpath-jar>>> * | <<<tests>>>   | <<<jar>>>  | java      | classes               |                       |
+*-----------------------+---------------+------------+-----------+-----------------------+-----------------------+
+| <<<fatjar>>> *        | <<<tests>>>   | <<<jar>>>  | java      | classes               |                       |
```
cstamas, line 57: fatjar == jar, but:
* can be on classes only (never on modules path)
* stops resolution (just like any existing self sufficient type, i.e. war)

hboutemy: @cstamas what does this mean in term of this PR?
is there something wrong
is there something to add?

jira-importer:
Resolve #10864

PR #1464, opened 2024-04-02
Improve maven startup time

gnodet:
This saves 15-20% on the startup time on the maven project for example.
On bigger projects, most of time is spent in loading projects, so this becomes negligible.


cstamas (approved):

rmannibucau (approved):
OK, ideally if we need more init we can use a global shared forkjoin pool (instead of `ExecutorService executor = Executors.newSingleThreadExecutor();` in jline module) but good enough for now

PR #1471, opened 2024-04-19
[MNG-8031] Backport concurrent friendly transport listener

cstamas:
Backport of simplex transfer listener.

---

https://issues.apache.org/jira/browse/MNG-8031

### DIFF
--- a/maven-embedder/src/main/java/org/apache/maven/cli/transfer/SimplexTransferListener.java
+++ b/maven-embedder/src/main/java/org/apache/maven/cli/transfer/SimplexTransferListener.java
@@ -1,0 +1,231 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.maven.cli.transfer;
+
+import java.io.File;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.concurrent.ArrayBlockingQueue;
+import java.util.concurrent.ConcurrentHashMap;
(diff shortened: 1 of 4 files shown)


slawekjaranowski (approved):

maven-embedder/src/main/java/org/apache/maven/cli/transfer/SimplexTransferListener.java
```diff
@@ -0,0 +1,231 @@
...
+
+import static java.util.Objects.requireNonNull;
+
+/**
+ * A {@link TransferListener} implementation that wraps another delegate {@link TransferListener} but makes it run
+ * on single thread, keeping the listener logic simple. This listener also blocks on last transfer event to allow
+ * output to perform possible cleanup. It spawns a daemon thread to consume queued events that may fall in even
+ * concurrently.
+ *
+ * @since 4.0.0
```
CrazyHZM, line 44: If we support it in 3.9.7, should it be changed to 3.9.7?

CrazyHZM (changes requested):

maven-embedder/src/main/java/org/apache/maven/cli/transfer/SimplexTransferListener.java
cstamas, line 44:
```suggestion
 * @since 3.9.7
```

CrazyHZM (approved):

maven-embedder/src/main/java/org/apache/maven/cli/transfer/SimplexTransferListener.java
```diff
@@ -0,0 +1,231 @@
...
+        } catch (InterruptedException e) {
+            throw new RuntimeException(e);
+        }
+    }
+
+    private final ConcurrentHashMap<File, Boolean> ongoing = new ConcurrentHashMap<>();
+
+    @Override
+    public void transferInitiated(TransferEvent event) {
+        ongoing.putIfAbsent(event.getResource().getFile(), Boolean.TRUE);
```
pshevche, line 157: @cstamas , `TransferResource` supports `file` being `null` ([ref](https://github.com/apache/maven-resolver/blob/master/maven-resolver-api/src/main/java/org/eclipse/aether/transfer/TransferResource.java#L166)). Althought the field is deprecated, the replacement is also nullable ([ref](https://github.com/apache/maven-resolver/blob/master/maven-resolver-api/src/main/java/org/eclipse/aether/transfer/TransferResource.java#L178)). However, this listener implementation will break if the resource has no `file` set. Was it a conscious decision?

cstamas: No, this is a probably oversight... care to provide a PR for this? And thanks!

pshevche: Yeah, I can do that. I'll look into it in https://issues.apache.org/jira/browse/MNG-8150

PR #1477, opened 2024-04-21
[MNG-8081] Profile property activation interpolation

cstamas:
Property profile activation should interpolate.

---

https://issues.apache.org/jira/browse/MNG-8081

### DIFF
--- a/maven-model-builder/src/main/java/org/apache/maven/model/profile/DefaultProfileActivationContext.java
+++ b/maven-model-builder/src/main/java/org/apache/maven/model/profile/DefaultProfileActivationContext.java
@@ -24,6 +24,10 @@ import java.util.List;
 import java.util.Map;
 import java.util.Properties;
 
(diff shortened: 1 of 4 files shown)


slawekjaranowski (commented):
looks ok

mbenson:
I notice that this implementation _only_ interpolates properties in property activations, whereas my submitted implementations for both v3.9.x and v4 attempt to satisfy the "least surprise" principle by applying interpolations to _all_ property activations. IMHO that's the best way forward but I am satisfied if this PR becomes the 3.9.x implementation.

cstamas:
@mbenson can you clarify what you mean by "only [...] property activations" vs "all property activations" as I think don't understand it fully...

mbenson:
Sorry @cstamas, was working late and apparently not 100%. Edited my comment with changed text in bold.

gnodet:
Similar to #1447

PR #1480, opened 2024-04-24
[3.9.x][MNG-8106] Fix metadata merge

cstamas:
As currently if given metadata serves multiple roles (G, A or V level), data loss occurs.

---

https://issues.apache.org/jira/browse/MNG-8106

PR #1481, opened 2024-04-24
[MNG-8106] Fix metadata merge

cstamas:
As currently if given metadata serves multiple roles (G, A or V level), data loss occurs.

---

https://issues.apache.org/jira/browse/MNG-8106

### DIFF
--- a/maven-api-impl/src/main/java/org/apache/maven/internal/impl/resolver/LocalSnapshotMetadata.java
+++ b/maven-api-impl/src/main/java/org/apache/maven/internal/impl/resolver/LocalSnapshotMetadata.java
@@ -113,6 +113,11 @@ final class LocalSnapshotMetadata extends MavenMetadata {
 
         metadata = metadata.withVersioning(metadata.getVersioning().withSnapshotVersions(versions.values()));
 
+        // just carry-on as-is
+        if (recessive.getPlugins() != null && !recessive.getPlugins().isEmpty()) {
+            metadata = metadata.withPlugins(new ArrayList<>(recessive.getPlugins()));
+        }
+
         artifacts.clear();
     }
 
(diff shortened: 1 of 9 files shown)


cstamas:
Note: new IT did not run (it did on 3.9.x PR https://github.com/apache/maven/pull/1480) as master is still alpha-14...

cstamas:
IT https://github.com/apache/maven-integration-testing/pull/336

maven-api-impl/src/main/java/org/apache/maven/internal/impl/resolver/LocalSnapshotMetadata.java
```diff
@@ -113,6 +113,11 @@ protected void merge(Metadata recessive) {
 
         metadata = metadata.withVersioning(metadata.getVersioning().withSnapshotVersions(versions.values()));
 
+        // just carry-on as-is
+        if (recessive.getPlugins() != null && !recessive.getPlugins().isEmpty()) {
+            metadata = metadata.withPlugins(new ArrayList<>(recessive.getPlugins()));
```
slawekjaranowski, line 118: should we also check if `metadata` already contains a `plugins`?

gnodet:
Rebased on master (after switching to beta-1)...

cstamas:
new IT triggered and passed OK
```
[INFO] Running org.apache.maven.it.MavenITmng8106OverlappingDirectoryRolesTest
[INFO] Tests run: 1, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 2.886 s - in org.apache.maven.it.MavenITmng8106OverlappingDirectoryRolesTest
```

maven-api-impl/src/main/java/org/apache/maven/internal/impl/resolver/LocalSnapshotMetadata.java
cstamas: The `metadata` cannot contain plugins, see how it is created:
* in ctor
* only "related" stuff filled in (Versions)
* point is that "not my stuff" should be just carried over as is

slawekjaranowski (approved):

jira-importer:
Resolve #10521

PR #1482, opened 2024-04-24
Clean warnings

gnodet:
- **Cleanup dependencies**
- **Deprecate test classes in maven-compat which cause lots of warnings**


cstamas (approved):
LGTM

PR #1495, opened 2024-05-02
Resolver updates

gnodet:
- **Resolver updates**
- **Remove a few other references**

### DIFF
--- a/maven-resolver-provider/src/main/java/org/apache/maven/repository/internal/RelocatedArtifact.java
+++ b/maven-resolver-provider/src/main/java/org/apache/maven/repository/internal/RelocatedArtifact.java
@@ -116,17 +116,6 @@ public final class RelocatedArtifact extends AbstractArtifact {
         return new RelocatedArtifact(artifact, groupId, artifactId, classifier, extension, version, message);
     }
 
-    @Deprecated
-    @Override
-    public Artifact setFile(File file) {
-        File current = getFile();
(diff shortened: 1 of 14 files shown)


maven-resolver-provider/src/main/java/org/apache/maven/repository/internal/RelocatedArtifact.java
```diff
@@ -116,17 +116,6 @@ public Artifact setVersion(String version) {
         return new RelocatedArtifact(artifact, groupId, artifactId, classifier, extension, version, message);
     }
 
-    @Deprecated
-    @Override
-    public Artifact setFile(File file) {
```
cstamas, line 121: This will break compat, no?

cstamas: maven-resolver-provider RelocatedArtifact is present in mvn3 as well, I mean "binary compat"

gnodet: Mmh, not sure, as a default implementation is provided by `AbstractArtifact`... ?

gnodet: I've revert the removal of the two `setFile()` methods to be safe.

cstamas (approved):

PR #1496, opened 2024-05-02
MNG-8115] Upgrade dependencies

slachiewicz:
Plexus Utils 4
Guice 6
Commons


gnodet:
Isn't upgrading to guice 6 a bit risky for a micro release ?

cstamas:
@slachiewicz I'd pass on this (in fact in whole 3.9.x line), would not budge plexus-utils, as Guillaume mention neither guice... for patch release this is simply put "too much" IMHO

cstamas:
Hm, maybe a small subset of updates in pom.xml only...

slachiewicz:
We are using Guice 6 in tests and it just works. Same with Maven 4. Agree that for patch release this may be too much

slachiewicz:
So plexus-utils 4 we also shouldn't have?

cstamas:
Yup, definitely nothing p-u 4. Let me comment in PR what I think is OK

pom.xml
```diff
@@ -125,31 +125,25 @@ under the License.
...
-    <plexusVersion>2.1.0</plexusVersion>
-    <plexusInterpolationVersion>1.26</plexusInterpolationVersion>
-    <!-- Blocked by 3.4.0+ changes, see MNG-7710 -->
-    <plexusUtilsVersion>3.5.1</plexusUtilsVersion>
-    <guiceVersion>5.1.0</guiceVersion>
-    <guavaVersion>32.0.1-jre</guavaVersion>
-    <guavafailureaccessVersion>1.0.1</guavafailureaccessVersion>
+    <plexusVersion>2.2.0</plexusVersion>
+    <plexusInterpolationVersion>1.27</plexusInterpolationVersion>
+    <guiceVersion>6.0.0</guiceVersion>
```
cstamas, line 137: Nope

pom.xml
```diff
@@ -125,31 +125,25 @@ under the License.
...
+    <guavaVersion>33.2.0-jre</guavaVersion>
```
cstamas, line 138: But, this one should be ok (Maven does not use, it is runtime dep of guice) -- just to get rid of red flags

pom.xml
```diff
@@ -125,31 +125,25 @@ under the License.
...
+    <guavafailureaccessVersion>1.0.2</guavafailureaccessVersion>
```
cstamas, line 139: Goes with guava, so OK

pom.xml
```diff
@@ -352,7 +340,7 @@ under the License.
       <dependency>
         <groupId>org.fusesource.jansi</groupId>
         <artifactId>jansi</artifactId>
-        <version>2.4.0</version>
+        <version>2.4.1</version>
```
cstamas, line 343: +1

pom.xml
```diff
@@ -340,7 +328,7 @@ under the License.
       <dependency>
         <groupId>org.apache.maven.shared</groupId>
         <artifactId>maven-shared-utils</artifactId>
-        <version>3.3.4</version>
+        <version>3.4.2</version>
```
cstamas, line 331: +1

pom.xml
```diff
@@ -125,31 +125,25 @@ under the License.
...
+    <guavaVersion>33.2.0-jre</guavaVersion>
+    <guavafailureaccessVersion>1.0.2</guavafailureaccessVersion>
     <wagonVersion>3.5.3</wagonVersion>
     <securityDispatcherVersion>2.0</securityDispatcherVersion>
     <cipherVersion>2.0</cipherVersion>
     <jxpathVersion>1.3</jxpathVersion>
     <resolverVersion>1.9.20</resolverVersion>
     <slf4jVersion>1.7.36</slf4jVersion>
-    <xmlunitVersion>2.9.1</xmlunitVersion>
+    <xmlunitVersion>2.10.0</xmlunitVersion>
```
cstamas, line 146: only if does not need code change and  builds ok

pom.xml
```diff
@@ -125,31 +125,25 @@ under the License.
...
-    <!-- plexus 2.1.1 is broken, no new release exists yet (nor will exist) -->
-    <plexusVersion>2.1.0</plexusVersion>
-    <plexusInterpolationVersion>1.26</plexusInterpolationVersion>
-    <!-- Blocked by 3.4.0+ changes, see MNG-7710 -->
-    <plexusUtilsVersion>3.5.1</plexusUtilsVersion>
-    <guiceVersion>5.1.0</guiceVersion>
-    <guavaVersion>32.0.1-jre</guavaVersion>
-    <guavafailureaccessVersion>1.0.1</guavafailureaccessVersion>
+    <plexusVersion>2.2.0</plexusVersion>
+    <plexusInterpolationVersion>1.27</plexusInterpolationVersion>
```
cstamas, line 136: +1

pom.xml
```diff
@@ -125,31 +125,25 @@ under the License.
...
     <mockitoVersion>4.11.0</mockitoVersion>
-    <!-- plexus 2.1.1 is broken, no new release exists yet (nor will exist) -->
-    <plexusVersion>2.1.0</plexusVersion>
-    <plexusInterpolationVersion>1.26</plexusInterpolationVersion>
-    <!-- Blocked by 3.4.0+ changes, see MNG-7710 -->
-    <plexusUtilsVersion>3.5.1</plexusUtilsVersion>
-    <guiceVersion>5.1.0</guiceVersion>
-    <guavaVersion>32.0.1-jre</guavaVersion>
-    <guavafailureaccessVersion>1.0.1</guavafailureaccessVersion>
+    <plexusVersion>2.2.0</plexusVersion>
```
cstamas, line 135: unsure, where is this used at all? I'd leave it as there was some issue with qdox, that am unsure was fixed or not

slachiewicz: I've updated qdox for 2.2.0 release. Also, all Plexus annotations are marked as Deprecated - ie more visible.
Not sure if 3.9 would be beneficial as looks like we don't have plans to do any Plexus -> jsr330 migration.

pom.xml
slachiewicz: I'll drop Guava update, not sure about the impact.

slachiewicz:
I've prepared a smaller set of updates, mainly Apache Commons and Classword.
Plexus Utils - leave version as is, only used variable from parent.
No Guava, Guice or plexus annotations updates.

cstamas (approved):

gnodet (approved):
LGTM

jira-importer:
Resolve #9987

PR #1501, opened 2024-05-04
[MNG-8015] Adjustments in new API related to PathType

desruisseaux:
This pull request makes the following changes in API related to `PathType`. Those changes are for supporting the development of the Maven compiler plugin:

* Add new artifact types for annotation processor paths:
  * `PROCESSOR`
  * `MODULAR_PROCESSOR`
  * `CLASSPATH_PROCESSOR`
* Establish a relationship between `org.apache.maven.api.JavaPathType` and `javax.tools.StandardLocation`:
  * New `location()` public method returning `Optional<JavaFileManager.Location>`.
  * New `valueOf(JavaFileManager.Location location)` public static method (converse of the above).
  * New _"Relationship with Java compiler standard location"_ section in class Javadoc.
* Modify the return type of `option(Iterable<? extends Path> paths)` from `String` to `String[]`.
* Delete `DependencyResolverResult.option(PathType)` because it appeared to not be needed.
* Add `DependencyResolverResult.warningForFilenameBasedAutomodules()` as a helper for managing the _"Filename-based automodules detected on the module-path: Please don't publish this project to a public artifact repository"_ warning. This warning already exists in Maven 3.

All methods removed or changed were new in Maven 4-alpha, so there is no compatibility break with stable releases.

### DIFF
--- a/api/maven-api-core/src/main/java/org/apache/maven/api/JavaPathType.java
+++ b/api/maven-api-core/src/main/java/org/apache/maven/api/JavaPathType.java
@@ -46,9 +46,9 @@ import org.apache.maven.api.annotations.Nonnull;
  *
  * <h2>Relationship with Java compiler standard location</h2>
  * This enumeration is closely related to the {@link JavaFileManager.Location} enumerations.
- * A difference is that the latter enumerate input and output files, while {@code JavaPathType}
+ * A difference is that the latter enumerates input and output files, while {@code JavaPathType}
  * enumerates only input dependencies. Another difference is that {@code JavaPathType} contains
- * some enumeration values used only at runtime and therefor not available in {@code javax.tool},
+ * some enumeration values used only at runtime and therefore not available in {@code javax.tool},
  * such as agent paths.
  *
  * @see org.apache.maven.api.services.DependencyResolverResult#getDispatchedPaths()
--- a/api/maven-api-core/src/main/java/org/apache/maven/api/JavaPathType.java
+++ b/api/maven-api-core/src/main/java/org/apache/maven/api/JavaPathType.java
@@ -18,6 +18,10 @@
  */
 package org.apache.maven.api;
 
+import javax.tools.DocumentationTool;
+import javax.tools.JavaFileManager;
+import javax.tools.StandardLocation;
+
 import java.io.File;
 import java.nio.file.Path;
 import java.util.Objects;
@@ -40,6 +44,13 @@ import org.apache.maven.api.annotations.Nonnull;
  * <p>Path types are often exclusive. For example, a dependency should not be both on the Java class-path
  * and on the Java module-path.</p>
  *
+ * <h2>Relationship with Java compiler standard location</h2>
+ * This enumeration is closely related to the {@link JavaFileManager.Location} enumerations.
+ * A difference is that the latter enumerate input and output files, while {@code JavaPathType}
+ * enumerates only input dependencies. Another difference is that {@code JavaPathType} contains
+ * some enumeration values used only at runtime and therefor not available in {@code javax.tool},
+ * such as agent paths.
+ *
  * @see org.apache.maven.api.services.DependencyResolverResult#getDispatchedPaths()
  *
  * @since 4.0.0
@@ -49,11 +60,12 @@ public enum JavaPathType implements PathType {
     /**
      * The path identified by the Java {@code --class-path} option.
      * Used for compilation, execution and Javadoc among others.
+     * The Java tools location is {@link StandardLocation#CLASS_PATH}.
      *
-     * <p><b>Context-sensitive interpretation:</b>
+     * <h4>Context-sensitive interpretation</h4>
      * A dependency with this path type will not necessarily be placed on the class-path.
      * There are two circumstances where the dependency may nevertheless be placed somewhere else:
-     * </p>
+     *
      * <ul>
      *   <li>If {@link #MODULES} path type is also set, then the dependency can be placed either on the
      *       class-path or on the module-path, but only one of those. The choice is up to the plugin,
@@ -63,16 +75,17 @@ public enum JavaPathType implements PathType {
      *       class-path.</li>
      * </ul>
      */
-    CLASSES("--class-path"),
+    CLASSES(StandardLocation.CLASS_PATH, "--class-path"),
 
     /**
      * The path identified by the Java {@code --module-path} option.
      * Used for compilation, execution and Javadoc among others.
+     * The Java tools location is {@link StandardLocation#MODULE_PATH}.
      *
-     * <p><b>Context-sensitive interpretation:</b>
+     * <h4>Context-sensitive interpretation</h4>
      * A dependency with this flag will not necessarily be placed on the module-path.
      * There are two circumstances where the dependency may nevertheless be placed somewhere else:
-     * </p>
+     *
      * <ul>
      *   <li>If {@link #CLASSES} path type is also set, then the dependency <em>should</em> be placed on the
      *       module-path, but is also compatible with placement on the class-path. Compatibility can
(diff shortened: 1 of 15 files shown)


api/maven-api-core/src/main/java/org/apache/maven/api/JavaPathType.java
```diff
@@ -18,6 +18,10 @@
  */
 package org.apache.maven.api;
 
+import javax.tools.DocumentationTool;
```
rmannibucau, line 21: api shouldn't depend on that since it is a generic path setup and there we make it depending on a single consumer so look fishy and not very promishing to be as generic as intended

desruisseaux: The generic API is the `PathType` interface. This dependency appears in th `JavaPathType` enumeration, which is the specialization for the Java tools.

rmannibucau: Ok for me if the name reflects it so something like JavaxToolPathType - or whatever makes it explicit. JavaPathType is way more generic IMHO.

api/maven-api-core/src/main/java/org/apache/maven/api/JavaPathType.java
```diff
@@ -40,6 +44,13 @@
  * <p>Path types are often exclusive. For example, a dependency should not be both on the Java class-path
  * and on the Java module-path.</p>
  *
+ * <h2>Relationship with Java compiler standard location</h2>
+ * This enumeration is closely related to the {@link JavaFileManager.Location} enumerations.
+ * A difference is that the latter enumerate input and output files, while {@code JavaPathType}
```
gnodet, line 49: enumerate -> enumerates

api/maven-api-core/src/main/java/org/apache/maven/api/JavaPathType.java
```diff
@@ -40,6 +44,13 @@
...
+ * enumerates only input dependencies. Another difference is that {@code JavaPathType} contains
+ * some enumeration values used only at runtime and therefor not available in {@code javax.tool},
```
gnodet, line 51: therefor -> therefore

api/maven-api-core/src/main/java/org/apache/maven/api/JavaPathType.java
```diff
@@ -63,16 +75,17 @@ public enum JavaPathType implements PathType {
      *       class-path.</li>
      * </ul>
      */
-    CLASSES("--class-path"),
+    CLASSES(StandardLocation.CLASS_PATH, "--class-path"),
```
gnodet, line 78: Is the `option` string still useful ? That looks really redundant and specific to tool implementations.  Can we only keep the `StandardLocation` ?

api/maven-api-core/src/main/java/org/apache/maven/api/JavaPathType.java
```diff
@@ -187,31 +247,38 @@ public Optional<String> option() {
...
-    public String option(Iterable<? extends Path> paths) {
+    public String[] option(Iterable<? extends Path> paths) {
         return format(null, paths);
     }
 
     /**
      * Implementation shared with {@link Modular}.
      */
-    String format(String moduleName, Iterable<? extends Path> paths) {
+    final String[] format(String moduleName, Iterable<? extends Path> paths) {
```
gnodet, line 262: I wonder if this method should be moved out of the API.

api/maven-api-core/src/main/java/org/apache/maven/api/JavaPathType.java
```diff
@@ -187,31 +247,38 @@ public Optional<String> option() {
...
      * @param paths the path to format as a tool option
      * @return the option associated to this path type followed by the given path elements,
-     *         or an empty string if there is no path element
+     *         or an empty array if there is no path element
      * @throws IllegalStateException if no option is associated to this path type
      */
     @Nonnull
     @Override
-    public String option(Iterable<? extends Path> paths) {
+    public String[] option(Iterable<? extends Path> paths) {
```
gnodet, line 255: I wonder if this method should be moved out of the API.

api/maven-api-core/src/main/java/org/apache/maven/api/PathType.java
```diff
@@ -94,23 +94,24 @@ public String option(Iterable<? extends Path> paths) {
...
      * </ul>
      *
      * @param paths the path to format as a string
      * @return the option associated to this path type followed by the given path elements,
-     *         or an empty string if there is no path element.
+     *         or an empty array if there is no path element.
      */
     @Nonnull
-    String option(Iterable<? extends Path> paths);
+    String[] option(Iterable<? extends Path> paths);
```
gnodet, line 114: I wonder if this method should be moved out of the API.

gnodet (changes requested):
The `option()` methods and field looks very tied to the implementation and I doubt they are part of the compiler API.  Should they be moved to the compiler plugin instead ?

api/maven-api-core/src/main/java/org/apache/maven/api/PathType.java
desruisseaux: The method is proposed here for two reasons: code reuse, and because the method behaviour depends on the option rather than the plugin using it.

### Code reuse
The use of this method would be shared by many plugins. For example, the same `--class-path` option is used by `java`, `javac`, `javadoc`, `javap`, `jdeprscan`, `jdeps`, `jmod` and `jshell`. In current Maven architecture, those tools are handled by independent plugins. This method reduces the need to repeat the same code in all plugins.

**Note:** this strategy relies on Oracle using consistently the same option names for all tools. It seems to be the case for all tools except `serialver`, provided that we use the modern form (e.g. `--class-path` instead of `-cp`).

### Who defines the behaviour
The behaviour of this method depends on the option rather than the plugin. For example, `JavaPathType.CLASSES` formats the given list of files as `--class-path <the files>`. But `JavaPathType.patchModule("org.my.module")` formats the same list of files as `--patch-module org.my.module=<the files>`. This rule is applied consistently in at least `java`, `javac` and `javadoc`. As we can see, the actual work done by this method is determined by the option, not by the plugin, which is why the method appears in `PathType`: for allowing the implementation to be option-dependent.

api/maven-api-core/src/main/java/org/apache/maven/api/JavaPathType.java
desruisseaux: The presence or not of this method depends on the [comment below](https://github.com/apache/maven/pull/1501#discussion_r1590432486).

api/maven-api-core/src/main/java/org/apache/maven/api/JavaPathType.java
desruisseaux: The presence or not of this method depends on the [comment below](https://github.com/apache/maven/pull/1501#discussion_r1590432486).

api/maven-api-core/src/main/java/org/apache/maven/api/JavaPathType.java
desruisseaux: The presence or not of this string depends on the [comment below](https://github.com/apache/maven/pull/1501#discussion_r1590432486).

api/maven-api-core/src/main/java/org/apache/maven/api/JavaPathType.java
desruisseaux: "Java" here refers to standard Java as defined by Oracle (other entities must use other names, e.g. "Jakarta"). The use of standard `javax.tools` API should not block non-standard Java tools to be added to this `JavaPathType` enumeration if desired, because `Location` is an interface with 2 non-default methods easy to implement (`getName()` and `isOutputLocation()`), and the proposed `JavaPathType.location()` property is `Optional` anyway.

The close connection to `javax.tools` API may actually be desirable because contrarily to other API, `javax.tools` is updated immediately when new Java features requires compiler changes. For example, `javax.tools` has been updated in Java 9 with the addition of new methods specific to modules management. By contrast, Maven and the Plexus compiler did not really reflected that evolution as far as I can see.

api/maven-api-core/src/main/java/org/apache/maven/api/Type.java
```diff
@@ -81,6 +81,24 @@ public interface Type extends ExtensibleEnum {
      */
     String MODULAR_JAR = "modular-jar";
 
+    /**
+     * Artifact type name for a JAR file that can be placed either on the annotation processor class-path
+     * or module-path. The path (classes or modules) is chosen by the plugin, possibly using heuristic rules.
+     */
+    String PROCESSOR = "proc";
```
gnodet, line 88: Why not `"processor"` ?

api/maven-api-core/src/main/java/org/apache/maven/api/Type.java
```diff
@@ -81,6 +81,24 @@ public interface Type extends ExtensibleEnum {
...
+
+    /**
+     * Artifact type name for a JAR file to unconditionally place on the annotation processor class-path.
+     * If the JAR is modular, its module information are ignored.
+     */
+    String CLASSPATH_PROCESSOR = "classpath-proc";
```
gnodet, line 94: `"classpath-processor"` ?

api/maven-api-core/src/main/java/org/apache/maven/api/Type.java
```diff
@@ -81,6 +81,24 @@ public interface Type extends ExtensibleEnum {
...
+
+    /**
+     * Artifact type name for a JAR file to unconditionally place on the annotation processor module-path.
+     * If the JAR is not modular, then it is loaded by Java as an unnamed module.
+     */
+    String MODULAR_PROCESSOR = "modular-proc";
```
gnodet, line 100: `"modular-processor"` ?

api/maven-api-core/src/main/java/org/apache/maven/api/Type.java
desruisseaux: I agree. `proc` was used because other types have short names (e.g. "jar") and `proc` is the name of the command-line option. But I usually like better explicit names like "processor".

gnodet (approved):

cstamas (approved):

jira-importer:
Resolve #9864

PR #1505, opened 2024-05-08
[3.9.x][MNG-8117] Backport to Maven 3.9.x

cstamas:
Backport same fix, as very same issue stands there, user cannot really tell what and why is happening.

Maven 3 has only Maven prerequisite, but still same code is present.

This is the backport of https://github.com/apache/maven/pull/1502

---

https://issues.apache.org/jira/browse/MNG-8117


jira-importer:
Resolve #9990

PR #1510, opened 2024-05-17
[MNG-7902] Sort plugins in the validation report

yuehcw:
**Summary**

This pull request addresses the JIRA issue [MNG-7902](https://issues.apache.org/jira/browse/MNG-7902). The goal of this change is to enhance the readability and usability of the validation report by sorting the plugins in a case-insensitive manner. Currently, the report outputs the content of a set of plugins without sorting them, which can result in a random order, making it difficult for users to locate specific entries or identify patterns.

**Changes Made**

Added sorting logic: Implemented a sorting mechanism to order plugins case-insensitively in the validation report.
Updated `reportSessionCollectedValidationIssues` function: Modified this function to include the sorting logic for plugins before they are logged.

**Why This Change Is Necessary**

By sorting the plugins in the validation report, developers can easily scan through the list to find specific entries or identify changes. This enhancement improves the readability and usability of the report, especially in scenarios where users need to review or troubleshoot plugin configurations.

**How the Change Was Implemented**

The sorting logic was added to the `reportSessionCollectedValidationIssues` function in the `DefaultPluginValidationManager` class. The plugins are sorted using `Map.Entry.comparingByKey(String.CASE_INSENSITIVE_ORDER)` before they are logged.

       pulling in other changes.

### DIFF
--- a/maven-core/src/main/java/org/apache/maven/plugin/internal/DefaultPluginValidationManager.java
+++ b/maven-core/src/main/java/org/apache/maven/plugin/internal/DefaultPluginValidationManager.java
@@ -23,7 +23,18 @@ import javax.inject.Singleton;
 
 import java.nio.file.Path;
 import java.nio.file.Paths;
-import java.util.*;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Collection;
+import java.util.Collections;
+import java.util.EnumSet;
+import java.util.HashMap;
+import java.util.LinkedHashMap;
+import java.util.LinkedHashSet;
+import java.util.List;
+import java.util.Locale;
+import java.util.Map;
+import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.stream.Collectors;
 
@@ -213,7 +224,7 @@ public final class DefaultPluginValidationManager extends AbstractEventSpy imple
             logger.warn("Plugin {} validation issues were detected in following plugin(s)", issueLocalitiesToReport);
             logger.warn("");
 
-            // Sorting the plugins (Fix the open issue)
+            // Sorting the plugins
             List<Map.Entry<String, PluginValidationIssues>> sortedEntries = new ArrayList<>(issuesMap.entrySet());
             sortedEntries.sort(Map.Entry.comparingByKey(String.CASE_INSENSITIVE_ORDER));
 
--- a/maven-core/src/main/java/org/apache/maven/plugin/internal/DefaultPluginValidationManager.java
+++ b/maven-core/src/main/java/org/apache/maven/plugin/internal/DefaultPluginValidationManager.java
@@ -23,17 +23,7 @@ import javax.inject.Singleton;
 
 import java.nio.file.Path;
 import java.nio.file.Paths;
-import java.util.Arrays;
-import java.util.Collection;
-import java.util.Collections;
-import java.util.EnumSet;
-import java.util.HashMap;
-import java.util.LinkedHashMap;
-import java.util.LinkedHashSet;
-import java.util.List;
-import java.util.Locale;
-import java.util.Map;
-import java.util.Set;
+import java.util.*;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.stream.Collectors;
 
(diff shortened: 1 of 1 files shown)


maven-core/src/main/java/org/apache/maven/plugin/internal/DefaultPluginValidationManager.java
```diff
@@ -222,7 +212,12 @@ private void reportSessionCollectedValidationIssues(MavenSession mavenSession) {
             logger.warn("");
             logger.warn("Plugin {} validation issues were detected in following plugin(s)", issueLocalitiesToReport);
             logger.warn("");
-            for (Map.Entry<String, PluginValidationIssues> entry : issuesMap.entrySet()) {
+
+            // Sorting the plugins (Fix the open issue)
```
gnodet, line 216: Remove the `(Fix the open issue)` part

gnodet (changes requested):

maven-core/src/main/java/org/apache/maven/plugin/internal/DefaultPluginValidationManager.java
```diff
@@ -23,17 +23,7 @@
...
-import java.util.Collections;
-import java.util.EnumSet;
-import java.util.HashMap;
-import java.util.LinkedHashMap;
-import java.util.LinkedHashSet;
-import java.util.List;
-import java.util.Locale;
-import java.util.Map;
-import java.util.Set;
+import java.util.*;
```
slawekjaranowski, line 26: asterixs should not be used in imports

gnodet: IIRC, they are automatically added if importing more than 10 (?) classes from the same package

gnodet: spotless and impsort plugins can't fix star -> individual imports because that would require access to the full class path and they really only operate on a single file.
That also mean the opposite (so I was wrong), i.e. they won't refactor multiple imports into a single one I think.  It must have been the IDE.  
Which means the PR should not contain the star. 

@slawekjaranowski it seems the AvoidStarImport is [no more enforced](https://github.com/apache/maven-shared-resources/blob/maven-shared-resources-5/src/main/resources/config/maven_checks_nocodestyle.xml).  I think we removed it based on the assumption that it was handled by spotless/impsort, but that does not seem to be the case

@yuehcw can you remove the star import ?

slawekjaranowski: issue added https://issues.apache.org/jira/browse/MSHARED-1401

yuehcw: Hello,

I have made the requested changes and pushed the new commit to this pull request. Please review the updates at your earliest convenience. Thank you for your feedback and guidance!

maven-core/src/main/java/org/apache/maven/plugin/internal/DefaultPluginValidationManager.java
yuehcw: Fixed!

maven-core/src/main/java/org/apache/maven/plugin/internal/DefaultPluginValidationManager.java
```diff
@@ -203,7 +204,7 @@ public void reportPluginMojoValidationIssue(
         mayReportInline(mavenSession.getRepositorySession(), locality, issue);
     }
 
-    private void reportSessionCollectedValidationIssues(MavenSession mavenSession) {
+    public void reportSessionCollectedValidationIssues(MavenSession mavenSession) {
```
slawekjaranowski, line 207: why change to `public`?

slawekjaranowski (approved):
LGTM

gnodet (approved):

slawekjaranowski:
I hope we can cherry-pick it to maven-3.x

PR #1512, opened 2024-05-17
Just a 'bulk update' of dependencies

cstamas:
To trigger CI job, but these should go in one-by-one w/ JIRA issues.


gnodet (changes requested):
I'd rather have the individual PR from dependabot.
Coupled with https://github.com/apache/maven/pull/1525, it should make our life easier.

cstamas:
Agreed

PR #1524, opened 2024-05-17
[MNG-8123] Fix Lifecycle in v3

gnodet:
IT PR: https://github.com/apache/maven-integration-testing/pull/342

### DIFF
--- a/maven-core/src/main/java/org/apache/maven/internal/impl/DefaultLifecycleRegistry.java
+++ b/maven-core/src/main/java/org/apache/maven/internal/impl/DefaultLifecycleRegistry.java
@@ -31,6 +31,7 @@ import java.util.Iterator;
 import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.Objects;
 import java.util.Optional;
 import java.util.Set;
 import java.util.stream.Collectors;
@@ -93,7 +94,7 @@ public class DefaultLifecycleRegistry implements LifecycleRegistry {
 
     @Override
     public Optional<Lifecycle> lookup(String id) {
-        return Optional.empty();
+        return stream().filter(lf -> Objects.equals(id, lf.id())).findAny();
     }
 
     @Named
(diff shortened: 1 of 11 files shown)


maven-core/src/main/java/org/apache/maven/internal/impl/DefaultLifecycleRegistry.java
```diff
@@ -78,37 +83,44 @@ public DefaultLifecycleRegistry(
...
         }
 
         @Override
         public Collection<Lifecycle> provides() {
-            return lifecycles.values().stream().map(this::wrap).collect(Collectors.toList());
+            try {
+                Map<String, org.apache.maven.lifecycle.Lifecycle> all =
+                        container.lookupMap(org.apache.maven.lifecycle.Lifecycle.class);
+                return all.keySet().stream()
+                        .filter(id -> !Lifecycle.CLEAN.equals(id)
```
cstamas, line 115: Why this filter? How are these 4 lifecycles published? Or in other words, why distinguish these 4?

gnodet: We have a kind of chicken/egg situation.  Lifecycle can be published by maven 3 extensions, but lifecycles must also be provided through the Maven 3 API.  We must avoid rewrapping in a kind of loop the default lifecycles exposed using the maven 4 API because it leads to a lookup error (we end-up with a bean creation cycle).

I can see if there's a better way, such as adding a new sub-interface `WrappedLifecycle` to better detect where the lifecycle come from (Maven 3 or 4 api and only wrap if for the other side).

gnodet: Actually, I can't use the instance to detect, as getting the instance will cause the lookup error: the map is lazily managed by sisu, and the instance is created when requested only, but that's what causes the lookup error.

cstamas: Isn't this somewhat similar to Type / ArtifactHandler story? Is same problem present there as well?

cstamas (approved):

PR #1539, opened 2024-05-28
Add test case for DefaultArtifact

chenchc6:
### Description
This pull request adds a new unit test for the `DefaultArtifact` class in the Apache Maven project. The test verifies the correctness of the `DefaultArtifact` class's constructor and its methods by asserting the expected values.

### Changes Made
- Added `DefaultArtifactTest.java` under `maven-artifact/src/test/java/org/apache/maven/artifact/`
- The test case validates:
  - Artifact is not null
  - Group ID, Artifact ID, Version, Scope, Type are correctly assigned
  - Classifier is null
  - Artifact handler is correctly assigned
  
### Purpose
The purpose of this test is to ensure that the `DefaultArtifact` class correctly initializes its fields and that its getter methods return the expected values. This helps in maintaining the robustness and reliability of the Apache Maven project.

### Contribution
- Verifying the correctness of the `DefaultArtifact` class's behavior.
- Enhancing the test coverage of the Apache Maven codebase.
- Providing a reference for future tests and improvements.

### DIFF
--- a/maven-artifact/pom.xml
+++ b/maven-artifact/pom.xml
@@ -36,6 +36,12 @@ under the License.
       <artifactId>junit-jupiter-api</artifactId>
       <scope>test</scope>
     </dependency>
+
+    <dependency>
+      <groupId>org.apache.maven</groupId>
+      <artifactId>maven-core</artifactId>
+      <version>3.6.1</version>
+    </dependency>
   </dependencies>
 
   <build>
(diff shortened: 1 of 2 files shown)


maven-artifact/pom.xml
```diff
@@ -36,6 +36,12 @@ under the License.
       <artifactId>junit-jupiter-api</artifactId>
       <scope>test</scope>
     </dependency>
+
+    <dependency>
+      <groupId>org.apache.maven</groupId>
+      <artifactId>maven-core</artifactId>
+      <version>3.6.1</version>
```
hiufung-kwok, line 43: Should be this be read from the Parent pom, instead of hard-coding it?

hiufung-kwok:
Also, would you mind to update the title to include the associated Jira ticket?

gnodet:
The `maven-artifact` will be deprecated and is not part of the upcoming Maven 4 api.

PR #1546, opened 2024-05-31
[MNG-8136] Update Eclipse Sisu to 0.9.0.M3

cstamas:
Update Eclipse Sisu to 0.9.0.M3

---

https://issues.apache.org/jira/browse/MNG-8136

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -42,7 +42,6 @@ under the License.
     process.</description>
   <url>https://maven.apache.org/ref/${project.version}/</url>
   <inceptionYear>2001</inceptionYear>
-
   <contributors>
     <contributor>
       <name>Stuart McCulloch</name>
@@ -184,7 +183,7 @@ under the License.
     <plexusXmlVersion>4.0.4</plexusXmlVersion>
     <resolverVersion>2.0.0-alpha-11</resolverVersion>
     <securityDispatcherVersion>2.0</securityDispatcherVersion>
-    <sisuVersion>0.9.0.M2</sisuVersion>
+    <sisuVersion>0.9.0-20240531.165402-20</sisuVersion>
     <slf4jVersion>2.0.13</slf4jVersion>
     <stax2ApiVersion>4.2.1</stax2ApiVersion>
     <wagonVersion>3.5.3</wagonVersion>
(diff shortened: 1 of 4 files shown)


cstamas:
@kwin this locally failed on me (it's maybe me), just heads up

kwin (commented):
Let us enable strict parsing at least for Maven 4.0

cstamas:
Will do once we sort out ITs, as something is fishy here: it fails on macOS only (and error seems m-p-p related)

pom.xml
```diff
@@ -416,6 +416,11 @@ under the License.
         <artifactId>javax.annotation-api</artifactId>
         <version>${javaxAnnotationApiVersion}</version>
       </dependency>
+      <dependency>
+        <groupId>org.eclipse.sisu</groupId>
+        <artifactId>org.eclipse.sisu.inject</artifactId>
+        <version>${sisuVersion}</version>
```
kwin, line 422: do we really need both now: `no-asm` and regular?

pom.xml
```diff
@@ -635,6 +640,19 @@ under the License.
...
+  <repositories>
+    <repository>
+      <releases>
+        <enabled>false</enabled>
+      </releases>
+      <snapshots>
+        <enabled>true</enabled>
+      </snapshots>
+      <id>sonatype-snapshots</id>
+      <url>https://oss.sonatype.org/content/repositories/snapshots/</url>
```
kwin, line 653: I am pretty sure this causes the IT failures, because some ITs seem to rely on dynamic versions and check on the wrong repo now (which delivers an outdated result), however I see no artifact being resolved from `sonatype-snapshots` in https://github.com/apache/maven/actions/runs/9322027451/job/25662372946?pr=1546.

kwin: Seems the IT failure was fixed yesterday in https://github.com/apache/maven-integration-testing/commit/4d1c9987e04b2581027f8c1c6645212c7e1f24b8.

pom.xml
cstamas: Yes, in modules we use "plain" while the distro uses no_asm + asm dep.

kwin: Sorry for the noise, missed this is within depMgmt.

kwin (approved):

kwin:
We can get rid of the PathConverter in maven-core/src/main/java/org/apache/maven/configuration/internal/DefaultBeanConfigurator.java now!

jira-importer:
Resolve #9708

PR #1550, opened 2024-06-01
Use Maven Wrapper to build

slawekjaranowski:
We should control Maven version used for builds


slawekjaranowski:
This one is needed https://github.com/apache/maven-integration-testing/pull/345

cstamas (approved):

PR #1551, opened 2024-06-02
Sisu update w strict cp scanning

cstamas:
Same as https://github.com/apache/maven/pull/1546 but enables new feature: strict classpath scanning


cstamas:
This experiment is done, change will be done in other PR

PR #1556, opened 2024-06-05
[MNG-8141] Model builder should report problems it finds during build

cstamas:
And not rely that model was validated, which is not true in some cases. Model builder can still easily detect issues with models while building them.

Provides "escape hatch" for projects stuck on invalid models in form of user property that can be enabled with `-Dmaven.modelBuilder.failOnInvalidModel=false`, this reverts to _old_ behaviour of maven, and the JavaFX reproducer goes back to error "unable to resolve" errors with uninterpolated `${javafx.platform}` property as classifier.

---

https://issues.apache.org/jira/browse/MNG-8141


cstamas:
This PR results in these FATAL issues being reported in JavaFX issue:
https://gist.github.com/cstamas/288a773459ef026e1500699b7a109451

slawekjaranowski (approved):

cstamas:
This change is wrong, am about to roll it back and do alternative fix

cstamas:
Reverted

PR #1558, opened 2024-06-05
[MNG-8066] Default exception handler does not handle recursion

cstamas:
If there is a recursion in throwable causes, Maven will hang forever, instead to return.

---

https://issues.apache.org/jira/browse/MNG-8066


slawekjaranowski (commented):
can we ad a unit test for it?

slawekjaranowski (approved):

PR #1567, opened 2024-06-07
[MNG-8141] Another attempt

cstamas:
WIP


cstamas:
The reproducer now looks like this https://gist.github.com/cstamas/36903dd7701cbae94d5971211e835b05
And is good, goal is to raise awareness to a presence of invalid model in the build. Build still fails due MNG-8147

PR #1571, opened 2024-06-10
[MNG-7194] Change expectation for missing property evaluation

pzygielo:

PR #1574, opened 2024-06-10
[MNG-7194] Change expectation for missing property evaluation

pzygielo:
- Like #1571 but for master


pzygielo:
- The [failure](https://github.com/apache/maven/actions/runs/9448106653/job/26021560449?pr=1574#step:5:2679) - as in #1571 -  is _expected_ with current implementation.
However the current implementation is not expected by me.

jira-importer:
Resolve #8255

PR #1575, opened 2024-06-10
[MNG-8150] Handle absent source/target files in transfer listener

pshevche:
## Summary
The PR address two issues observed in the `SimplexTransferListener` and `ConsoleMavenTransferListener`:
1. [TransferResource#getFile()](https://github.com/apache/maven-resolver/blob/master/maven-resolver-api/src/main/java/org/eclipse/aether/transfer/TransferResource.java#L170) can be null. The current `SimplexTransferListener` will break with an NPE if the `file` is not set on the resource.
2. `TransferResource` is not immutable and does not implement `equals` or `hashCode,` making its usage in collections brittle. Listener consumers are not guaranteed to reuse the same instance across listener invocations. I suggest wrapping it in an immutable identifier.

Let me know if I am overthinking this or if anything else is needed.

Resolves https://issues.apache.org/jira/browse/MNG-8150

### DIFF
--- a/maven-embedder/src/main/java/org/apache/maven/cli/transfer/ConsoleMavenTransferListener.java
+++ b/maven-embedder/src/main/java/org/apache/maven/cli/transfer/ConsoleMavenTransferListener.java
@@ -35,7 +35,7 @@ import org.eclipse.aether.transfer.TransferResource;
  */
 public class ConsoleMavenTransferListener extends AbstractMavenTransferListener {
 
-    private Map<TransferResource, Long> transfers = new LinkedHashMap<>();
+    private Map<TransferResourceIdentifier, TransferResourceAndSize> transfers = new LinkedHashMap<>();
     private FileSizeFormat format = new FileSizeFormat(Locale.ENGLISH); // use in a synchronized fashion
     private StringBuilder buffer = new StringBuilder(128); // use in a synchronized fashion
 
@@ -65,18 +65,19 @@ public class ConsoleMavenTransferListener extends AbstractMavenTransferListener 
     @Override
     public void transferProgressed(TransferEvent event) throws TransferCancelledException {
         TransferResource resource = event.getResource();
-        transfers.put(resource, event.getTransferredBytes());
+        transfers.put(
+                new TransferResourceIdentifier(resource),
+                new TransferResourceAndSize(resource, event.getTransferredBytes()));
 
         buffer.append("Progress (").append(transfers.size()).append("): ");
 
-        Iterator<Map.Entry<TransferResource, Long>> entries =
-                transfers.entrySet().iterator();
+        Iterator<TransferResourceAndSize> entries = transfers.values().iterator();
         while (entries.hasNext()) {
-            Map.Entry<TransferResource, Long> entry = entries.next();
-            long total = entry.getKey().getContentLength();
-            Long complete = entry.getValue();
+            TransferResourceAndSize entry = entries.next();
+            long total = entry.resource.getContentLength();
+            Long complete = entry.transferredBytes;
 
-            String resourceName = entry.getKey().getResourceName();
+            String resourceName = entry.resource.getResourceName();
 
             if (printResourceNames) {
                 int idx = resourceName.lastIndexOf('/');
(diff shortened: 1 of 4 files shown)


cstamas:
Looks nice, but the maven-3.9.x branch (Maven 3.9.x) is Java 8...

pshevche:
> Looks nice, but the maven-3.9.x branch (Maven 3.9.x) is Java 8...

Yeah, I see that now :) Do you want me to keep records on this branch and use a class on `maven-3.9.x` branch or would you rather prefer the same style everywhere?

cstamas:
No, IMO records on master (Maven 4, that is Java 17) is ok, I just warned you when you try to backport this to maven-3.9.x branch...

pshevche:
> No, IMO records on master (Maven 4, that is Java 17) is ok, I just warned you when you try to backport this to maven-3.9.x branch...

Makes sense, thank you. Here is the PR for backport: https://github.com/apache/maven/pull/1576

cstamas (approved):

slawekjaranowski (approved):

maven-embedder/src/main/java/org/apache/maven/cli/transfer/ConsoleMavenTransferListener.java
```diff
@@ -35,7 +35,7 @@
  */
 public class ConsoleMavenTransferListener extends AbstractMavenTransferListener {
 
-    private Map<TransferResource, Long> transfers = new LinkedHashMap<>();
+    private Map<TransferResourceIdentifier, TransferResourceAndSize> transfers = new LinkedHashMap<>();
```
gnodet, line 38: Those 3 fields should be `final`.

maven-embedder/src/main/java/org/apache/maven/cli/transfer/ConsoleMavenTransferListener.java
```diff
@@ -35,7 +35,7 @@
...
     private FileSizeFormat format = new FileSizeFormat(Locale.ENGLISH); // use in a synchronized fashion
```
gnodet, line 39: We need to get rid of the `Locale.ENGLISH` parameter which is unused, though this can be done in a subsequent PR.

maven-embedder/src/main/java/org/apache/maven/cli/transfer/ConsoleMavenTransferListener.java
```diff
@@ -35,7 +35,7 @@
  */
 public class ConsoleMavenTransferListener extends AbstractMavenTransferListener {
```
gnodet, line 36: Can we add a comment specifying that this class is not thread safe and should only be used from a single thread, or wrapped in the `SimplexTransferListener` ?

gnodet (changes requested):
Two minor changes would be nice if included.

maven-embedder/src/main/java/org/apache/maven/cli/transfer/ConsoleMavenTransferListener.java
pshevche: Done in https://github.com/apache/maven/pull/1575/commits/274f1807ab0fcef625efadb26aabcaa2be1fe31b

maven-embedder/src/main/java/org/apache/maven/cli/transfer/ConsoleMavenTransferListener.java
pshevche: Done in https://github.com/apache/maven/pull/1575/commits/274f1807ab0fcef625efadb26aabcaa2be1fe31b

maven-embedder/src/main/java/org/apache/maven/cli/transfer/ConsoleMavenTransferListener.java
pshevche: Addressed in https://github.com/apache/maven/pull/1579

cstamas (approved):

gnodet (approved):

PR #1587, opened 2024-06-17
[MNG-8160] Recreate the transformed artifact if it has been deleted (by the clean goal for example)

gnodet:


cstamas (approved):
But a JIRA for this?

PR #1601, opened 2024-07-08
Fix Maven 4 extensions

gnodet:
- **Add rootDirectory to XmlReaderRequest and fix maven-core exported artifacts**
- **Set the thread context classloader to the container realm to fix class loading from extensions**


gnodet:
Required for https://github.com/apache/maven-xinclude-extension/pull/1

cstamas (approved):

PR #1612, opened 2024-07-11
[MNG-8180] Fail install/deploy if rogue Maven Plugin metadata found

cstamas:
Resolver handles transparently the repository metadata, and in case of plugins it peeks into META-INF/maven/plugin.xml of given artifact JAR to figure out needed metadata bits (prefix, name, etc).

But, this was done "blindly", while it is expected that GA of JAR artifact without classifier (requirement for maven plugins) and GA in embedded plugin metadata must be same.

Decision here is to fail hard, prevent this being installed and deployed, as this is most probably wrong (unsure what maven-indexer or even Sonatype search would do in this case).

---


gnodet (approved):

jira-importer:
Resolve #9818

PR #1613, opened 2024-07-11
[Experiment] Leave out dep from depMgmt

cstamas:
As dep is "stronger" anyway, on conflict resolution will win.


cstamas:
Superseded by https://github.com/apache/maven/pull/1617

PR #1621, opened 2024-07-27
[MNG-8197] Use default classifier when Eclipse Aether specifies none

desruisseaux:
The important change in this commit is in the implementation of `getClassifier()` methods. However, this commit opportunistically refactors the code with a base class shared by `DefaultDependency` and `DefaultDependencyCoordinate` implementations.

## Note on build failure
Trying to compile Maven with itself (4.0.0-beta-4-SNAPSHOT) fails with a `NullPointerException` in the `PluginsMetadataGenerator` class of the _"Maven Compat (deprecated)"_ module. The issue seems more serious than a missing null check, because according the sanity check at line 137, the plugin is operating on the wrong artifact. However, this issue is not related to this commit. Testing with the commit before this one produces the same result. Building with Maven 3 works fine.

---

https://issues.apache.org/jira/browse/MNG-8197

### DIFF
--- a/api/maven-api-core/src/main/java/org/apache/maven/api/services/DependencyResolverRequest.java
+++ b/api/maven-api-core/src/main/java/org/apache/maven/api/services/DependencyResolverRequest.java
@@ -323,7 +323,7 @@ public interface DependencyResolverRequest {
          * @return {@code this} for method call chaining
          */
         @Nonnull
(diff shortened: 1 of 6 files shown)


cstamas (approved):

gnodet (approved):

jira-importer:
Resolve #9672

PR #1625, opened 2024-08-05
[MNG-8195] Add `DependencyResolverResult.getModuleName(Path)` method

desruisseaux:
Clarify the meaning of "module" in documentation, then add two methods for getting Java module information from a `DependencyResolverResult`:

* `getModuleName(Path)` returns a `java.lang.String`
* `getModuleDescription(Path)` returns a `java.lang.module.ModuleDescriptor`

"Get name" may seem redundant with "get description" because the name can be obtained from the description by `ModuleDescriptor.name()`. The difference is that "get name" fallbacks on the `Automatic-Module-Name` attribute of `META-INF/MANIFEST.MF` if the module descriptor is not found.

The rational for providing those methods in the API is because it allows to use the cache managed by the `DefaultDependencyResolverResult` implementation. It avoids decoding many times the `module-info.class` files. Plugins need those information for managing `--add-reads` and similar options.

# Alternative
It would be more natural to have `getModuleName()` and `getModuleDescription()` methods in `Dependency`. But it would imply that dependency contains a `getPath()` method. I'm not sure why it is not already the case.

---

https://issues.apache.org/jira/browse/MNG-8195

### DIFF
--- a/api/maven-api-core/src/main/java/org/apache/maven/api/services/DependencyResolverResult.java
+++ b/api/maven-api-core/src/main/java/org/apache/maven/api/services/DependencyResolverResult.java
@@ -18,12 +18,15 @@
  */
 package org.apache.maven.api.services;
 
+import java.io.IOException;
+import java.lang.module.ModuleDescriptor;
 import java.nio.file.Path;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
 
 import org.apache.maven.api.Dependency;
+import org.apache.maven.api.DependencyScope;
 import org.apache.maven.api.JavaPathType;
 import org.apache.maven.api.Node;
 import org.apache.maven.api.PathType;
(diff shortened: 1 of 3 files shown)


gnodet (approved):
LGTM, I'd try to:
* add an alias in the xml to replace the `<modules>` element
* make this element optional (discover them if not specified explicitely)

cstamas (approved):

desruisseaux:
No additional change envisioned for this pull request if the alternative (methods added in the `Dependency` interface) is not applied.

gnodet:
Could we move the subproject thing into a separate PR ?
I'd rather merge them with the addition of `<subprojects>` in the model.
I've raised https://issues.apache.org/jira/browse/MNG-8210 for that.

desruisseaux:
No problem. I will try to apply this separation and other comments in the next few days.

desruisseaux:
Done with #1649.

gnodet:
> Done with #1649.

Could you remove the commit to rename module -> subproject from this PR ?

desruisseaux:
Done.

jira-importer:
Resolve #9802

PR #1628, opened 2024-08-06
[MNG-8196] Make exception messages match Maven 3 again

oehme:
The boxing of `MojoException` into a `MojoExecutionException` brought two issues:

1. It added another layer to the stacktrace, making it a bit less readable for users and breaking test expectations for Maven extension authors (the latter was my main motivation for this change)
2. It lost the `longMessage`, which meant that this message was no longer shown at the end of the build.

This change fixes both problems by passing through the original `MojoException` and treating this new exception type directly in `DefaultExceptionHandler` and `MojoExecutor`

       pulling in other changes.

---

https://issues.apache.org/jira/browse/MNG-8196

### DIFF
--- a/maven-core/src/main/java/org/apache/maven/lifecycle/internal/MojoExecutor.java
+++ b/maven-core/src/main/java/org/apache/maven/lifecycle/internal/MojoExecutor.java
@@ -37,7 +37,7 @@ import java.util.concurrent.locks.ReentrantLock;
 import java.util.concurrent.locks.ReentrantReadWriteLock;
 
 import org.apache.maven.api.SessionData;
-import org.apache.maven.api.plugin.MojoException;
+import org.apache.maven.api.services.MavenException;
 import org.apache.maven.api.services.MessageBuilderFactory;
 import org.apache.maven.artifact.Artifact;
 import org.apache.maven.artifact.resolver.filter.ArtifactFilter;
@@ -326,7 +326,7 @@ public class MojoExecutor {
                     | PluginManagerException
                     | PluginConfigurationException
                     | MojoExecutionException
-                    | MojoException e) {
+                    | MavenException e) {
                 throw new LifecycleExecutionException(
                         messageBuilderFactory, mojoExecution, session.getCurrentProject(), e);
             }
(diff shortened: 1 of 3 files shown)


cstamas (approved):

maven-core/src/main/java/org/apache/maven/plugin/DefaultBuildPluginManager.java
```diff
@@ -143,7 +143,7 @@ public void executeMojo(MavenSession session, MojoExecution mojoExecution)
                 mojoExecutionListener.beforeMojoExecution(mojoExecutionEvent);
                 mojo.execute();
                 mojoExecutionListener.afterMojoExecutionSuccess(mojoExecutionEvent);
-            } catch (ClassCastException e) {
+            } catch (ClassCastException | MojoException e) {
```
gnodet, line 146: `MojoException` is actually a `RuntimeException`, so the outer catch block needs to be enhanced to rethrow directly instead of calling `mojoExecutionListener.afterExecutionFailure`.  Just to keep the old behaviour.

gnodet: Also, I would extend to any `MavenException`, since those are rather expected conditions.

gnodet (changes requested):

maven-core/src/main/java/org/apache/maven/plugin/DefaultBuildPluginManager.java
oehme: Done

maven-core/src/main/java/org/apache/maven/lifecycle/internal/MojoExecutor.java
```diff
@@ -324,7 +325,8 @@ private void doExecute2(MavenSession session, MojoExecution mojoExecution) throw
             } catch (MojoFailureException
                     | PluginManagerException
                     | PluginConfigurationException
-                    | MojoExecutionException e) {
+                    | MojoExecutionException
+                    | MojoException e) {
```
gnodet, line 329: Should that one be a `MavenException` too given we've broaden the other locations ?

gnodet: Yes, but the intent is that the code either calls `MojoSucceeded` or `MojoFailed`.  Since any `MavenException` can be thrown at this point, they need to be caught imho.

gnodet (approved):

PR #1633, opened 2024-08-09
[MNG-8177] Add contextual info for model warnings

cstamas:
As they really can come from anywhere. In case of this issue even from some eliminated POM that was read while collecting dirty tree, and was later eliminated. So confusing for users.

---

https://issues.apache.org/jira/browse/MNG-8177

### DIFF
--- a/maven-resolver-provider/src/main/java/org/apache/maven/repository/internal/RequestTraceHelper.java
+++ b/maven-resolver-provider/src/main/java/org/apache/maven/repository/internal/RequestTraceHelper.java
@@ -1,0 +1,70 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.maven.repository.internal;
+
+import java.util.stream.Collectors;
+
+import org.apache.maven.model.Plugin;
+import org.eclipse.aether.RequestTrace;
+import org.eclipse.aether.collection.CollectRequest;
+import org.eclipse.aether.collection.CollectStepData;
+import org.eclipse.aether.resolution.ArtifactDescriptorRequest;
+import org.eclipse.aether.resolution.ArtifactRequest;
+import org.eclipse.aether.resolution.DependencyRequest;
+
+/**
+ * Helper class to manage {@link RequestTrace} for better error logging.
+ *
+ * @since 3.9.9
+ */
+public final class RequestTraceHelper {
+
+    /**
+     * Method that creates some informational string based on passed in {@link RequestTrace}. The contents of request
+     * trace can literally be anything, but this class tries to cover "most common" cases that are happening in Maven.
+     */
+    public static String interpretTrace(boolean explain, RequestTrace requestTrace) {
+        while (requestTrace != null) {
+            Object data = requestTrace.getData();
+            if (data instanceof DependencyRequest) {
+                DependencyRequest request = (DependencyRequest) data;
+                return (explain ? "dependency request for " : "") + request;
+            } else if (data instanceof CollectRequest) {
+                CollectRequest request = (CollectRequest) data;
+                return (explain ? "collect request for " : "") + request;
+            } else if (data instanceof CollectStepData) {
+                CollectStepData stepData = (CollectStepData) data;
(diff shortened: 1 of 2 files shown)


maven-resolver-provider/src/main/java/org/apache/maven/repository/internal/RequestTraceHelper.java
```diff
@@ -0,0 +1,70 @@
...
+ *
+ * @since 3.9.9
+ */
+public final class RequestTraceHelper {
+
+    /**
+     * Method that creates some informational string based on passed in {@link RequestTrace}. The contents of request
+     * trace can literally be anything, but this class tries to cover "most common" cases that are happening in Maven.
+     */
+    public static String interpretTrace(boolean explain, RequestTrace requestTrace) {
```
slawekjaranowski, line 42: `explain`  is always `true`

maven-resolver-provider/src/main/java/org/apache/maven/repository/internal/RequestTraceHelper.java
```diff
@@ -0,0 +1,70 @@
...
+            if (data instanceof DependencyRequest) {
+                DependencyRequest request = (DependencyRequest) data;
+                return (explain ? "dependency request for " : "") + request;
+            } else if (data instanceof CollectRequest) {
+                CollectRequest request = (CollectRequest) data;
+                return (explain ? "collect request for " : "") + request;
+            } else if (data instanceof CollectStepData) {
+                CollectStepData stepData = (CollectStepData) data;
+                return (explain ? "collect path " : "")
+                        + stepData.getPath().stream().map(Object::toString).collect(Collectors.joining(" -> "));
```
slawekjaranowski, line 54: It can generate very long one line message ... maybe more readable will be separated by new line

cstamas:
Updated PR: in "normal" invocation, it is one liner:
```
[WARNING] 1 problem was encountered while building the effective model for org.jboss:jboss-dmr:jar:1.3.0.Final during dependency collection step for project
```

When mvn -X used, it explains:
```
[WARNING] 1 problem was encountered while building the effective model for org.jboss:jboss-dmr:jar:1.3.0.Final during dependency collection step for project
Path to offending node from root:
 -> com.acme:example-subsystem:jar:1.0-SNAPSHOT
 -> org.wildfly.core:wildfly-subsystem-test:pom:24.0.0.Final (test)
 -> org.wildfly.core:wildfly-subsystem-test-framework:jar:24.0.0.Final (compile)
 -> org.wildfly.core:wildfly-model-test:jar:24.0.0.Final (compile)
 -> org.wildfly.legacy.test:wildfly-legacy-versions:jar:8.0.1.Final (compile)
[WARNING] 'dependencyManagement.dependencies.dependency.systemPath' for com.sun:tools:jar refers to a non-existing file /home/cstamas/.sdkman/candidates/java/21.0.4-tem/../lib/tools.jar. Please verify that you run Maven using a JDK and not just a JRE. @ org.jboss:jboss-dmr:1.3.0.Final
```

WolfgangHG:
(I am the original reporter of this issue) I preferred the first approach to print full details out of the box. Could you add a hint " (use -X for full details)" to the short version?

cstamas:
@WolfgangHG I think always printing out path would be "too much", Maven is already way way too chatty... but here are few facts to tinker about:
* when descriptor reader reports this error all Maven "knows" that some POM somewhere has some issue (uses system scoped dependency pointing at file that does not exists)
* later on, when collection happened (built the "dirty graph" -> conflicts resolved -> dep tree), just as in this example, the reported POM may not even be present... 
* still, IMO developer _should be aware_ of issues like these, as any dependency change/reshuffle may make the offending POM end up in the build (in reproducer is "just lurking somewhere" but is not entering -- is eliminated)

So IMO one liner is fine, and will add "use -X for full details"

cstamas:
Updated messages:
```
[WARNING] 1 problem was encountered while building the effective model for org.jboss:jboss-dmr:jar:1.3.0.Final during dependency collection step for project (use -X to see details)
```
With `-X`:
```
[WARNING] 1 problem was encountered while building the effective model for org.jboss:jboss-dmr:jar:1.3.0.Final during dependency collection step for project. Path to offending node from root:
 -> com.acme:example-subsystem:jar:1.0-SNAPSHOT
 -> org.wildfly.core:wildfly-subsystem-test:pom:24.0.0.Final (test)
 -> org.wildfly.core:wildfly-subsystem-test-framework:jar:24.0.0.Final (compile)
 -> org.wildfly.core:wildfly-model-test:jar:24.0.0.Final (compile)
 -> org.wildfly.legacy.test:wildfly-legacy-versions:jar:8.0.1.Final (compile)
[WARNING] Problems encountered:
[WARNING] 'dependencyManagement.dependencies.dependency.systemPath' for com.sun:tools:jar refers to a non-existing file /home/cstamas/.sdkman/candidates/java/21.0.4-tem/../lib/tools.jar. Please verify that you run Maven using a JDK and not just a JRE. @ org.jboss:jboss-dmr:1.3.0.Final
```

gnodet (approved):

WolfgangHG:
One more thought: you wrote that the problematic dependency might be replaced later anyway for the effective dependency tree (as happened in my sample). Would it make sense to add another hint like "(problem might be resolved when building effective dependency tree)"?
I probably would still be confused by the new and improved warning message when seeing it the first time.

cstamas:
IMO adding that info may be superfluous (and would lengthen the message much much more). I just assume users are aware what "verbose" tree is and will be able to track it down, after all, the path is present now in message (in -X mode).

PR #1634, opened 2024-08-09
[MNG-8188] Profile properties are not interpolated

cstamas:
Restore of uninterpolated things did "too much", we need to restore activations only.

Maven4 is not affected, manually checked using reproducer.

---

https://issues.apache.org/jira/browse/MNG-8188


slawekjaranowski (approved):
As we have a reproduce we can add also a test

cstamas:
UT added, validated against maven-3.9.x (w/o patch) it fails, it passes in this PR

gnodet (approved):

jira-importer:
Resolve #9809

PR #1635, opened 2024-08-09
[MNG-8182] Resolved errors were created based on collect exceptions

cstamas:
Hence, they missed resolution errors.

---

https://issues.apache.org/jira/browse/MNG-8182

### DIFF
--- a/maven-core/src/main/java/org/apache/maven/plugin/internal/DefaultPluginDependenciesResolver.java
+++ b/maven-core/src/main/java/org/apache/maven/plugin/internal/DefaultPluginDependenciesResolver.java
@@ -25,7 +25,9 @@ import javax.inject.Singleton;
 import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.stream.Collectors;
 
+import com.google.common.collect.Streams;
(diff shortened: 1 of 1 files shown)


maven-core/src/main/java/org/apache/maven/plugin/internal/DefaultPluginDependenciesResolver.java
```diff
@@ -234,8 +236,13 @@ private DependencyResult resolveInternal(
             throw new PluginResolutionException(
                     plugin, e.getResult().getExceptions(), logger.isDebugEnabled() ? e : null);
         } catch (DependencyResolutionException e) {
-            throw new PluginResolutionException(
-                    plugin, e.getResult().getCollectExceptions(), logger.isDebugEnabled() ? e : null);
+            List<Exception> exceptions = Streams.concat(
```
slawekjaranowski, line 239: JDK `Stream.concat` can be used

cstamas: Good catch.

slawekjaranowski (approved):

jira-importer:
Resolve #9815

PR #1636, opened 2024-08-09
[MNG-8177] Add contextual info for model warnings

cstamas:
But also aling Maven4 bits in resolver-provider and impl.

There is a TODO as well!

---

https://issues.apache.org/jira/browse/MNG-8177

### DIFF
--- a/maven-api-impl/src/main/java/org/apache/maven/internal/impl/resolver/RequestTraceHelper.java
+++ b/maven-api-impl/src/main/java/org/apache/maven/internal/impl/resolver/RequestTraceHelper.java
@@ -1,0 +1,62 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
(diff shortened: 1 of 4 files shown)


maven-api-impl/src/main/java/org/apache/maven/internal/impl/resolver/RequestTraceHelper.java
```diff
@@ -0,0 +1,62 @@
...
+/**
+ * Helper class to manage {@link RequestTrace} for better error logging.
+ */
+public final class RequestTraceHelper {
+
+    /**
+     * Method that creates some informational string based on passed in {@link RequestTrace}. The contents of request
+     * trace can literally be anything, but this class tries to cover "most common" cases that are happening in Maven.
+     */
+    public static String interpretTrace(boolean explain, RequestTrace requestTrace) {
```
slawekjaranowski, line 39: `explain` is always `true`

gnodet (approved):

PR #1638, opened 2024-08-10
[MPLUGIN-530] Deprecate requirements in plugin descriptor 1.1.0

kwin:
It is already removed for 2.0.0

       pulling in other changes.

### DIFF
--- a/api/maven-api-plugin/src/main/mdo/plugin.mdo
+++ b/api/maven-api-plugin/src/main/mdo/plugin.mdo
@@ -409,7 +409,7 @@ under the License.
         <field xdoc.separator="blank">
           <name>requirements</name>
           <version>1.0.0/1.1.0</version>
-          <description></description>
+          <description>Use Maven 4 Dependency Injection (for v4 plugins) or JSR 330 annotations (for v3 plugins) to inject dependencies instead.</description>
           <association>
             <type>Requirement</type>
             <multiplicity>*</multiplicity>
@@ -535,20 +535,17 @@ under the License.
         <field>
           <name>role</name>
           <required>true</required>
-          <version>1.0.0+</version>
           <type>String</type>
           <description></description>
         </field>
         <field xml.tagName="role-hint">
           <name>roleHint</name>
-          <version>1.0.0+</version>
           <type>String</type>
           <description></description>
         </field>
         <field xml.tagName="field-name">
           <name>fieldName</name>
           <required>true</required>
-          <version>1.0.0+</version>
           <type>String</type>
           <description>The field name which has this requirement.</description>
         </field>


kwin:
@gnodet Have you deliberately removed that from the plugin descriptor 2.0.0?

gnodet:
> @gnodet Have you deliberately removed that from the plugin descriptor 2.0.0? It is still written in https://github.com/apache/maven-plugin-tools/blob/9332b09acc08d2e063f12a8158a57ca8ec56fadd/maven-plugin-tools-generators/src/main/java/org/apache/maven/tools/plugin/generator/PluginDescriptorFilesGenerator.java#L593 (even for V4/2.0.0)

The injection mechanism is different for V4 mojo as it uses the new DI API / implementation.

kwin:
Does that mean the plugin descriptors for v4 should no longer contain those?

gnodet:
> Does that mean the plugin descriptors for v4 should no longer contain those?

It doesn't afaik.
https://github.com/apache/maven/blob/master/api/maven-api-plugin/src/main/mdo/plugin.mdo#L411

kwin:
v4 is using plugin descriptor 2.0. Having `requirements` only for version 1.0 and 1.1 makes it disappear in 2.0 (just look at the generated doc from modello)

gnodet:
So if you're targeting master, I'm not sure this PR is really needed. @kwin wdyt ?

cstamas:
Same question here: IF this goes to master, the deprecation message is plain wrong (mvn4 API has DI annotation, and not JSR330 is to be used). Am unsure is this PR really needed.

kwin:
> So if you're targeting master, I'm not sure this PR is really needed.

We are somehow going in cycles with this PR:
@gnodet I think we agreed that even for Plugin Descriptor v2 the `requirements` is still evaluated! That means it should be marked as deprecated in both 1.1 and 2.0! I am gonna fix the version in this PR as well for `requirements`, but you should go over all other elements as well!

Regarding

> the deprecation message is plain wrong (mvn4 API has DI annotation, and not JSR330 is to be used)

I am gonna clarify that in the deprecation that this is not necessarily JSR 330 annotation but alternatively some other annotation evaluated at run time.

> Am unsure is this PR really needed.

For me the plugin descriptor is the primary contract between a plugin and Maven, and therefore it needs to be formally deprecated there before we can ultimately get rid of it in the future!

gnodet:
> > So if you're targeting master, I'm not sure this PR is really needed.
> 
> We are somehow going in cycles with this PR: @gnodet I think we agreed that even for Plugin Descriptor v2 the `requirements` is still evaluated! That means it should be marked as deprecated in both 1.1 and 2.0! I am gonna fix the version in this PR as well for `requirements`, but you should go over all other elements as well!

As you've found out, it's not in the generated xml schema for v2.  

I don't see how it could be supported with the v2 schema without removing the DI support and restoring plexus.

> Regarding
> 
> > the deprecation message is plain wrong (mvn4 API has DI annotation, and not JSR330 is to be used)
> 
> I am gonna clarify that in the deprecation that this is not necessarily JSR 330 annotation but alternatively some other annotation evaluated at run time.

If you consider the XML the API, you can't bind to the implementation.  Why are we talking about JSR330 annotations or `@Component` if that's the case ?

> 
> > Am unsure is this PR really needed.
> 
> For me the plugin descriptor is the primary contract between a plugin and Maven, and therefore it needs to be formally deprecated there before we can ultimately get rid of it in the future!

It's fine to add the deprecation to the 1.1 model imho.  Both 1.1 and 2.0 are supported by Maven 4.

gnodet (commented):
LGTM.  However, this won't appear anymore, because we don't generate the 1.1.0 schema.

kwin:
I plan to backport to 3.x and publish it, because there were also some other changes being supported meanwhile in Maven 3.x. I don't think this causes any harm here. Later on we can remove all outdated elements from the MDO in master.

gnodet:
> I plan to backport to 3.x and publish it, because there were also some other changes being supported meanwhile in Maven 3.x. I don't think this causes any harm here. Later on we can remove all outdated elements from the MDO in master.

They are part of modello support for history.  We do actually use the same MDO in two different versions to generate the v3 and v4 model classes.  I agree this is not used for plugins, but I don't see the point of removing those.

PR #1639, opened 2024-08-10
[MNG-8180] Handle NPE due non-existent tags

cstamas:
There was an NPE possibility when plugin.xml had no expected tags present.

Also: maven-compat has plugin.xml (!) w/o "name" tag, it NPEd and failed build. This was NOT picked up by CI as "rebuild itself" step does not install (just verify).

---

### DIFF
--- a/maven-api-impl/src/main/java/org/apache/maven/internal/impl/resolver/PluginsMetadata.java
+++ b/maven-api-impl/src/main/java/org/apache/maven/internal/impl/resolver/PluginsMetadata.java
@@ -24,6 +24,8 @@ import java.util.Date;
 import java.util.LinkedHashMap;
 import java.util.List;
 
+import org.apache.maven.api.annotations.Nonnull;
+import org.apache.maven.api.annotations.Nullable;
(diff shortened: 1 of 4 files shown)


gnodet (approved):

slawekjaranowski (approved):

gnodet:
@cstamas should we remove that plugin.xml at the same time ?
The mojos referenced in the plugin.xml do not exist in maven-compat, so I think it's useless.

cstamas:
@gnodet yup, we could remove plugin.xml, and am unsure about the rest, as I see in history, they belong together. Also, they are being mention here https://cwiki.apache.org/confluence/display/MAVENOLD/Deterministic+Lifecycle+Planning and to me it seems like those are not even m3 -> m2 backward compat, but in fact m2.1 -> m2.0 (!!!) backward compat thing.

cstamas:
but let's do it in separate JIRA IMO

cstamas:
Created https://issues.apache.org/jira/browse/MNG-8206

PR #1640, opened 2024-08-10
Renaming of `ArtifactCoordinate.getVersion()` + documentation

desruisseaux:
The pull request contains two changes:

* Documentation on `Artifact`, `ArtifactCoordinate`, `Dependency` and `DependencyCoordinate`.
* Renaming of `ArtifactCoordinate.getVersion()` as `getVersionConstraint()`.

This pull request is located after #1621 for avoiding merge conflict. It would be easier to merge (if accepted) #1621 before to merge this one.

### DIFF
--- a/api/maven-api-core/src/main/java/org/apache/maven/api/Artifact.java
+++ b/api/maven-api-core/src/main/java/org/apache/maven/api/Artifact.java
@@ -24,7 +24,8 @@ import org.apache.maven.api.annotations.Nonnull;
 
 /**
  * Pointer to a resolved resource such as a <abbr>JAR</abbr> file or <abbr>WAE</abbr> application.
- * {@code Artifact} instances are created when <dfn>resolving</dfn> {@link Artifact} instances.
+ * Each {@code Artifact} instance is basically a pointer to a file in the Maven repository.
+ * {@code Artifact} instances are created when <dfn>resolving</dfn> {@link ArtifactCoordinate} instances.
  * Resolving is the process that selects a {@linkplain #getVersion() particular version}
  * and downloads the artifact in the local repository.
  * The download may be deferred to the first time that the file is needed.
@@ -72,6 +73,8 @@ public interface Artifact {
     /**
      * {@return the version of the artifact}. Contrarily to {@link ArtifactCoordinate},
      * each {@code Artifact} is associated to a specific version instead of a range of versions.
+     * If the {@linkplain #getBaseVersion() base version} contains a meta-version such as {@code LATEST},
+     * {@code RELEASE} or {@code SNAPSHOT}, those keywords are replaced by, for example, the actual timestamp.
      *
      * @see ArtifactCoordinate#getVersionConstraint()
      */
@@ -79,8 +82,10 @@ public interface Artifact {
     Version getVersion();
 
     /**
-     * {@return the base version of the artifact}.
-     * TODO: this javadoc is not helpful.
+     * {@return the version or meta-version of the artifact}.
+     * A meta-version is a version suffixed with {@code LATEST}, {@code RELEASE} or {@code SNAPSHOT} keyword.
+     * Meta-versions are represented in a base version by their symbols (e.g., {@code SNAPSHOT}),
+     * while they are replaced by, for example, the actual timestamp in the {@linkplain #getVersion() version}.
      */
     @Nonnull
     Version getBaseVersion();
--- a/api/maven-api-core/src/main/java/org/apache/maven/api/Artifact.java
+++ b/api/maven-api-core/src/main/java/org/apache/maven/api/Artifact.java
@@ -73,8 +73,8 @@ public interface Artifact {
     /**
      * {@return the version of the artifact}. Contrarily to {@link ArtifactCoordinate},
      * each {@code Artifact} is associated to a specific version instead of a range of versions.
-     * If the {@linkplain #getBaseVersion() base version} contains a meta-version such as {@code LATEST},
-     * {@code RELEASE} or {@code SNAPSHOT}, those keywords are replaced by, for example, the actual timestamp.
+     * If the {@linkplain #getBaseVersion() base version} contains a meta-version such as {@code SNAPSHOT},
+     * those keywords are replaced by, for example, the actual timestamp.
      *
      * @see ArtifactCoordinate#getVersionConstraint()
      */
@@ -83,7 +83,7 @@ public interface Artifact {
 
     /**
      * {@return the version or meta-version of the artifact}.
-     * A meta-version is a version suffixed with {@code LATEST}, {@code RELEASE} or {@code SNAPSHOT} keyword.
+     * A meta-version is a version suffixed with the {@code SNAPSHOT} keyword.
      * Meta-versions are represented in a base version by their symbols (e.g., {@code SNAPSHOT}),
      * while they are replaced by, for example, the actual timestamp in the {@linkplain #getVersion() version}.
      */
--- a/api/maven-api-core/src/main/java/org/apache/maven/api/Artifact.java
+++ b/api/maven-api-core/src/main/java/org/apache/maven/api/Artifact.java
@@ -23,76 +23,83 @@ import org.apache.maven.api.annotations.Immutable;
 import org.apache.maven.api.annotations.Nonnull;
 
 /**
- * An artifact points to a resource such as a jar file or war application.
+ * Pointer to a resolved resource such as a <abbr>JAR</abbr> file or <abbr>WAE</abbr> application.
+ * {@code Artifact} instances are created when <dfn>resolving</dfn> {@link Artifact} instances.
+ * Resolving is the process that selects a {@linkplain #getVersion() particular version}
+ * and downloads the artifact in the local repository.
+ * The download may be deferred to the first time that the file is needed.
  *
  * @since 4.0.0
  */
 @Experimental
 @Immutable
 public interface Artifact {
-
     /**
-     * Returns a unique identifier for this artifact.
+     * {@return a unique identifier for this artifact}.
      * The identifier is composed of groupId, artifactId, extension, classifier, and version.
      *
-     * @return the unique identifier
+     * @see ArtifactCoordinate#getId()
      */
     @Nonnull
     default String key() {
+        String c = getClassifier();
         return getGroupId()
                 + ':'
                 + getArtifactId()
                 + ':'
                 + getExtension()
-                + (getClassifier().isEmpty() ? "" : ":" + getClassifier())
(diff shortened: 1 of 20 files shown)


api/maven-api-core/src/main/java/org/apache/maven/api/Artifact.java
```diff
@@ -23,76 +23,83 @@
 import org.apache.maven.api.annotations.Nonnull;
 
 /**
- * An artifact points to a resource such as a jar file or war application.
+ * Pointer to a resolved resource such as a <abbr>JAR</abbr> file or <abbr>WAE</abbr> application.
+ * {@code Artifact} instances are created when <dfn>resolving</dfn> {@link Artifact} instances.
+ * Resolving is the process that selects a {@linkplain #getVersion() particular version}
+ * and downloads the artifact in the local repository.
```
gnodet, line 29: So things are a bit blurry:
 * artifact resolution is:  resolve version (see `VersionResolver#resolveVersion()`), and then resolve (i.e. download) the file
 * dependency resolution is: collect dependencies, flatten the graph, resolve (i.e. download) the file

During dependency collection, versions are resolved (@cstamas right?). The flattening phase removes branches of the graph so that one artifact per GA is present. The resolution phase will then download the result artifacts.

We don't have any way to download an artifact for which the version has been resolved though (that may be a missing bit).  So if we introduce `ResolvedArtifact` and `ResolvedDependency`, it may help a bit:
  * version resolution: ArtifactCoordinate -> Artifact
  * artifact resolution: Artifact -> ResolvedArtifact
  * dependency collection -> Node + Dependency
  * dependency resolution: list of Dependency -> list of ResolvedDependency

api/maven-api-core/src/main/java/org/apache/maven/api/Artifact.java
```diff
@@ -23,76 +23,83 @@
...
-     * @return the artifactId
+     * @see ArtifactCoordinate#getArtifactId()
      */
     @Nonnull
     String getArtifactId();
 
     /**
-     * The version of the artifact.
+     * {@return the version of the artifact}. Contrarily to {@link ArtifactCoordinate},
+     * each {@code Artifact} is associated to a specific version instead of a range of versions.
```
gnodet, line 74: range or meta version (`LATEST`, `RELEASE`, `SNAPSHOT`)

api/maven-api-core/src/main/java/org/apache/maven/api/Artifact.java
```diff
@@ -23,76 +23,83 @@
...
      */
     @Nonnull
     Version getVersion();
 
     /**
-     * The base version of the artifact.
-     *
-     * @return the version
+     * {@return the base version of the artifact}.
+     * TODO: this javadoc is not helpful.
```
gnodet, line 83: Agreed !
So the base version is the resolved range version or the meta version (i.e. it can be `4.0.0-SNAPSHOT`) while the `getVersion()` should return `4.0.0-20240809.104932-36`).
I'm actually not quite sure if the original `VersionConstraint` was a range... @cstamas  ?

api/maven-api-core/src/main/java/org/apache/maven/api/ArtifactCoordinate.java
```diff
@@ -23,64 +23,68 @@
 import org.apache.maven.api.annotations.Nonnull;
 
 /**
- * The {@code Coordinate} object is used to point to an {@link Artifact}
- * but the version may be specified as a range instead of an exact version.
+ * Point to an {@link Artifact} but with the version specified as a range instead of an exact version.
+ * Each {@code ArtifactCoordinate} instance is basically a pointer to a file in the Maven repository,
+ * except that the version may not be defined precisely.
```
gnodet, line 28: Not sure I like the wording, since the actual pointer is the `Artifact`. `ArtifactCoordinate` is used to create those pointers by resolving the version.

I think relocations do not come into play for artifacts, but only for dependencies, at least that's what the [resolver ArtifactResolver's doc seems to indicate](https://github.com/apache/maven-resolver/blob/97dfd1c2b9deb15734d5e401807e55cd0498332a/maven-resolver-impl/src/main/java/org/eclipse/aether/impl/ArtifactResolver.java#L41-L43)

api/maven-api-core/src/main/java/org/apache/maven/api/ArtifactCoordinate.java
```diff
@@ -23,64 +23,68 @@
...
      * @return the classifier or an empty string if none, never {@code null}
      */
     @Nonnull
     String getClassifier();
 
     /**
-     * The version of the artifact.
-     *
-     * @return the version
+     * {@return the range of versions of the artifact}.
```
gnodet, line 56: Specific version, range, or meta version.

api/maven-api-core/src/main/java/org/apache/maven/api/Dependency.java
```diff
@@ -23,30 +23,45 @@
 import org.apache.maven.api.annotations.Nonnull;
 
 /**
+ * Result of resolving a {@code DependencyCoordinate}.
+ * Resolving is the process that clarifies the obligation (optional or mandatory status),
+ * selects a particular version and downloads the artifact in the local repository.
```
gnodet, line 28: It's a bit more complicated as explained earlier.  `Dependency` is the output of the `collection` process, which builds the graph of dependencies. It then needs to be `flattened` and `resolved`.  The version selection is done for each dependency during the collection phase, the flatten phase will keep only a single version per GA (groupId+artifactId), the resolution will actually download the dependencies (or artifacts) that have been computed.

api/maven-api-core/src/main/java/org/apache/maven/api/Dependency.java
```diff
@@ -23,30 +23,45 @@
...
+     * If may be, for example, at compile time only, at run time or at test time.
+     *
+     * @see DependencyCoordinate#getScope()
+     */
     @Nonnull
     DependencyScope getScope();
 
+    /**
+     * Returns whether the dependency is optional or mandatory.
+     * Contrarily to {@link DependencyCoordinate}, the obligation of a dependency cannot be unspecified.
```
gnodet, line 56: It's not specified at all, it's computed.  The collection phase will compute the scope and optionality of each dependency.

api/maven-api-core/src/main/java/org/apache/maven/api/DependencyScope.java
```diff
@@ -28,9 +28,9 @@
 import org.apache.maven.api.annotations.Nonnull;
 
 /**
- * Dependency scope.
- * This represents at which time the dependency will be used, for example, at compile time only,
- * at run time or at test time.  For a given dependency, the scope is directly derived from the
+ * Represents at which time the dependency will be used.
+ * If may be, for example, at compile time only, at run time or at test time.
```
gnodet, line 32: `Indicates when the dependency will be used. For example, it may be at compile time only, at runtime, or at test time.`

gnodet (changes requested):
I'm fine with the code changes.  The javadoc can be refined. @cstamas, I think your input would be valuable here...

gnodet:
@desruisseaux One possibility would be to split this PR.  Keep this one for the doc, and create another one for the getVersionConstraint() method if that allows you to go further.

api/maven-api-core/src/main/java/org/apache/maven/api/Artifact.java
```diff
@@ -23,76 +23,83 @@
 import org.apache.maven.api.annotations.Nonnull;
 
 /**
- * An artifact points to a resource such as a jar file or war application.
+ * Pointer to a resolved resource such as a <abbr>JAR</abbr> file or <abbr>WAE</abbr> application.
```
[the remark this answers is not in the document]
desruisseaux: I thought that "resolved" means that all properties were fixed to their final values, especially the version and obligation. Before resolution, an `ArtifactCoordinate` refers to an ensemble of files. After resolution, an `Artifact` refers to a single file. I was seeing the download as a separated concern. I have no problem in including download in the definition of "resolved", but maybe we should write this definition somewhere, e.g. in the package-info file.

api/maven-api-core/src/main/java/org/apache/maven/api/Artifact.java
desruisseaux: Added a "definitions of terms" section in `package-info.java` with the above explanation about artifact resolution and dependency resolution. Since those definitions include the download part, they answer my [previous comment](https://github.com/apache/maven/pull/1640#discussion_r1714881957).

api/maven-api-core/src/main/java/org/apache/maven/api/Artifact.java
desruisseaux: Then the `Version` interface should said something about those meta-versions. I will put a first attempt, but it should be completed with an explanation of `LATEST`, `RELEASE` and `SNAPSHOT` semantic.

api/maven-api-core/src/main/java/org/apache/maven/api/Artifact.java
desruisseaux: So does it means that the discussion about meta-versions in [above comment](https://github.com/apache/maven/pull/1640#discussion_r1713059218) actually applies here?

api/maven-api-core/src/main/java/org/apache/maven/api/ArtifactCoordinate.java
desruisseaux: Replaced by "Identification of an ensemble of {@link Artifact}s in a range of versions."

desruisseaux:
Added a commit trying to resolve the comments to the best of my understanding. Did à `git push --force` for resolving conflicts.

api/maven-api-core/src/main/java/org/apache/maven/api/Artifact.java
gnodet: I forgot that `LATEST` and `RELEASE` are actually deprecated.  So we should not document them I think.

desruisseaux: Done the removal of `LATEST` and `RELEASE` from the documentation.

gnodet:
@desruisseaux I've added follow-on commits at https://github.com/gnodet/maven/tree/refactor/getVersionConstraint, containing a proposal to add `ResolvedArtifact`, `ResolvedDependency` and `ProvidedArtifact` interfaces and another one with more javadoc.

desruisseaux:
Thanks, it looks good to me. We may need to revisit #1625 after this pull request is merged.

PR #1641, opened 2024-08-11
[MNG-8180] Handle NPE due non-existent tags

cstamas:
Original commit made Maven fail build itself.

---


jira-importer:
Resolve #9818

PR #1645, opened 2024-08-12
[MNG-8180] Back out from failing the build

cstamas:
Just warn the user for now.

Port of accepted change from 3.9.x: https://github.com/apache/maven/pull/1642

---


jira-importer:
Resolve #9818

PR #1648, opened 2024-08-12
[MNG-8165] Align mvn.sh script with mvn.cmd

cstamas:
As one does check for .mvn directory in FS root while other does not, stops one level before.

---

https://issues.apache.org/jira/browse/MNG-8165


gnodet (approved):

jira-importer:
Resolve #9687

PR #1649, opened 2024-08-13
[MNG-8210] Replace Maven "module" term by "subproject" in documentation

desruisseaux:
Replace Maven "module" term by "subproject" in documentation for avoiding confusion with JPMS modules. This commit was initially part of #1625 and moved to a separated PR for allowing inclusion with [MNG-8210](https://issues.apache.org/jira/browse/MNG-8210) instead of [MNG-8195](https://issues.apache.org/jira/browse/MNG-8195).

PR #1651, opened 2024-08-13
[MNG-8210] Replace Maven "module" term by "subproject"

gnodet:
Based on @desruisseaux initial commit to rename subproject #1649.
This PR adds `subprojects` to 4.1.0 POM, deprecate `modules`, and add a discovery mechanism which kicks in when:
  * pomVersion == 4.1.0
  * packaging == pom
  * no modules or subprojects are specified

### DIFF
--- a/api/maven-api-core/src/main/java/org/apache/maven/api/plugin/annotations/Mojo.java
+++ b/api/maven-api-core/src/main/java/org/apache/maven/api/plugin/annotations/Mojo.java
@@ -76,8 +76,8 @@ public @interface Mojo {
     boolean projectRequired() default true;
 
     /**
-     * if the Mojo uses the Maven project and its child modules.
-     * @return uses the Maven project and its child modules
+     * if the Mojo uses the Maven project and its subprojects.
+     * @return uses the Maven project and its subprojectss
      */
     boolean aggregator() default false;
 
--- a/api/maven-api-core/src/main/java/org/apache/maven/api/plugin/annotations/Mojo.java
+++ b/api/maven-api-core/src/main/java/org/apache/maven/api/plugin/annotations/Mojo.java
@@ -77,7 +77,7 @@ public @interface Mojo {
 
     /**
      * if the Mojo uses the Maven project and its subprojects.
-     * @return uses the Maven project and its subprojectss
+     * @return uses the Maven project and its subprojects
      */
     boolean aggregator() default false;
 
--- a/api/maven-api-core/src/main/java/org/apache/maven/api/services/ModelBuilder.java
+++ b/api/maven-api-core/src/main/java/org/apache/maven/api/services/ModelBuilder.java
@@ -25,7 +25,11 @@ import org.apache.maven.api.model.Model;
 
 public interface ModelBuilder extends Service {
 
-    List<String> VALID_MODEL_VERSIONS = List.of("4.0.0", "4.1.0");
+    String MODEL_VERSION_4_0_0 = "4.0.0";
+
+    String MODEL_VERSION_4_1_0 = "4.0.0";
+
+    List<String> VALID_MODEL_VERSIONS = List.of(MODEL_VERSION_4_0_0, MODEL_VERSION_4_1_0);
 
     ModelBuilderResult build(ModelBuilderRequest request) throws ModelBuilderException;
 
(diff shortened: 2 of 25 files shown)


maven-core/src/main/java/org/apache/maven/project/MavenProject.java
```diff
@@ -708,6 +708,9 @@ public List<Plugin> getBuildPlugins() {
     }
 
     public List<String> getModules() {
```
cstamas, line 710: Deprecate it? Or maybe does not matter, as in API 4 plugins have no reach to this class?

api/maven-api-core/src/main/java/org/apache/maven/api/services/ModelBuilder.java
```diff
@@ -25,7 +25,11 @@
 
 public interface ModelBuilder extends Service {
 
-    List<String> VALID_MODEL_VERSIONS = List.of("4.0.0", "4.1.0");
+    String MODEL_VERSION_4_0_0 = "4.0.0";
+
+    String MODEL_VERSION_4_1_0 = "4.0.0";
```
desruisseaux, line 30: Typo? Shouldn't the value be "4.1.0"?

mthmulders:
Shouldn't the subprojects be transformed to modules in the consumer pom?

maven-core/src/main/java/org/apache/maven/project/MavenProject.java
gnodet: Yes, I was assuming this change targets 4.x api and plugins, so I would not care about the v3 api.

gnodet:
> Shouldn't the subprojects be transformed to modules in the consumer pom?

Actually, the modules/subprojects are build information, so not available in the consumer POM.

api/maven-api-core/src/main/java/org/apache/maven/api/plugin/annotations/Mojo.java
```diff
@@ -76,8 +76,8 @@
     boolean projectRequired() default true;
 
     /**
-     * if the Mojo uses the Maven project and its child modules.
-     * @return uses the Maven project and its child modules
+     * if the Mojo uses the Maven project and its subprojects.
+     * @return uses the Maven project and its subprojectss
```
Bukama, line 80: `subprojectss` -> `subprojects`

maven-api-impl/src/main/java/org/apache/maven/internal/impl/model/DefaultModelValidator.java
```diff
@@ -360,6 +360,61 @@ public void validateFileModel(Model m, ModelBuilderRequest request, ModelProblem
...
+                                Severity.WARNING,
+                                Version.V41,
+                                "modules",
+                                null,
+                                "deprecated modules element, use subprojects instead",
+                                m.getLocation("modules"));
+                    } else {
+                        addViolation(
+                                problems,
+                                Severity.WARNING,
```
Bukama, line 409: Why not error?

maven-core/src/main/java/org/apache/maven/project/DefaultProjectBuilder.java
```diff
@@ -121,6 +121,7 @@
 public class DefaultProjectBuilder implements ProjectBuilder {
     public static final String BUILDER_PARALLELISM = "maven.projectBuilder.parallelism";
     public static final int DEFAULT_BUILDER_PARALLELISM = Runtime.getRuntime().availableProcessors() / 2 + 1;
+    private static final String POM_4_0_0 = "4.0.0";
```
Bukama, line 124: Why not `MODEL_VERSION_4.0.0` as above?

maven-api-impl/src/main/java/org/apache/maven/internal/impl/model/DefaultModelValidator.java
gnodet: Good catch !

maven-core/src/main/java/org/apache/maven/project/DefaultProjectBuilder.java
gnodet: That one isn't used anymore, I'll remove it.

cstamas (approved):

jira-importer:
Resolve #9793

PR #1656, opened 2024-08-15
[MNG-8211] Fail the build if CI Friendly revision used without value

cstamas:
As this is almost always source of confusion. If feature is used and there is no proper value set, fail the build, as users for sure does not plan to deploy artifacts with version `${revision}` (or any expression in project.version).

Still, to not be disruptive, the old behaviour can be achieved by setting `maven.build.allowExpressionInEffectiveProjectVersion`=true project property.

---

https://issues.apache.org/jira/browse/MNG-8211

### DIFF
--- a/maven-model-builder/src/main/java/org/apache/maven/model/interpolation/AbstractStringBasedModelInterpolator.java
+++ b/maven-model-builder/src/main/java/org/apache/maven/model/interpolation/AbstractStringBasedModelInterpolator.java
@@ -30,7 +30,9 @@ import java.util.Properties;
 
 import org.apache.maven.model.Model;
 import org.apache.maven.model.building.ModelBuildingRequest;
+import org.apache.maven.model.building.ModelProblem;
 import org.apache.maven.model.building.ModelProblemCollector;
+import org.apache.maven.model.building.ModelProblemCollectorRequest;
 import org.apache.maven.model.path.PathTranslator;
 import org.apache.maven.model.path.UrlNormalizer;
 import org.codehaus.plexus.interpolation.AbstractValueSource;
@@ -158,6 +160,16 @@ public abstract class AbstractStringBasedModelInterpolator implements ModelInter
         // Overwrite existing values in model properties. Otherwise it's not possible
         // to define them via command line e.g.: mvn -Drevision=6.5.7 ...
         versionProcessor.overwriteModelProperties(modelProperties, config);
+        String version = model.getVersion();
+        if (version != null && version.startsWith("${") && version.endsWith("}")) {
+            String property = version.substring(2, version.length() - 1);
+            if (!modelProperties.containsKey(property)) {
+                problems.add(new ModelProblemCollectorRequest(ModelProblem.Severity.FATAL, ModelProblem.Version.BASE)
+                        .setMessage("Model '" + model.getId() + "' uses CI Friendly Versions expression '${" + property
+                                + "}' without value being set")
+                        .setLocation(model.getLocation("")));
+            }
+        }
         valueSources.add(new MapBasedValueSource(modelProperties));
 
         valueSources.add(new MapBasedValueSource(config.getSystemProperties()));
--- a/maven-model-builder/src/main/java/org/apache/maven/model/interpolation/AbstractStringBasedModelInterpolator.java
+++ b/maven-model-builder/src/main/java/org/apache/maven/model/interpolation/AbstractStringBasedModelInterpolator.java
@@ -30,9 +30,7 @@ import java.util.Properties;
 
 import org.apache.maven.model.Model;
 import org.apache.maven.model.building.ModelBuildingRequest;
-import org.apache.maven.model.building.ModelProblem;
 import org.apache.maven.model.building.ModelProblemCollector;
-import org.apache.maven.model.building.ModelProblemCollectorRequest;
 import org.apache.maven.model.path.PathTranslator;
 import org.apache.maven.model.path.UrlNormalizer;
 import org.codehaus.plexus.interpolation.AbstractValueSource;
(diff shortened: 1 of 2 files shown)


cstamas:
Example failure: https://gist.github.com/cstamas/9dbb79be59ba22b6bc5c823d95593e50

maven-model-builder/src/main/java/org/apache/maven/model/interpolation/AbstractStringBasedModelInterpolator.java
```diff
@@ -158,6 +160,16 @@ public Object getValue(String expression) {
         // Overwrite existing values in model properties. Otherwise it's not possible
         // to define them via command line e.g.: mvn -Drevision=6.5.7 ...
         versionProcessor.overwriteModelProperties(modelProperties, config);
+        String version = model.getVersion();
+        if (version != null && version.startsWith("${") && version.endsWith("}")) {
+            String property = version.substring(2, version.length() - 1);
+            if (!modelProperties.containsKey(property)) {
+                problems.add(new ModelProblemCollectorRequest(ModelProblem.Severity.FATAL, ModelProblem.Version.BASE)
+                        .setMessage("Model '" + model.getId() + "' uses CI Friendly Versions expression '${" + property
+                                + "}' without value being set")
```
laeubi, line 169:
```suggestion
                        .setMessage("Project '" + model.getId() + "' uses versions expression '" + version
                                + "' but a value is not defined")
```

laeubi (changes requested):
It seems currently only one value is assumed but one can combine values e.g. 

`<version>${revision}${sha1}${changelist}</version>`

see https://github.com/apache/maven/blob/76ff89f50750424de8b4a4cba8e7bc0cbb02991f/maven-model-builder/src/main/java/org/apache/maven/model/validation/DefaultModelValidator.java#L1231-L1245

cstamas:
This PR does it in wrong place:
* this should happen in validator when validating effective model?

cstamas:
Example of failure: https://gist.github.com/cstamas/01b1b6152aca56f1383148ef74245583

cstamas:
This PR would really help to reduce user confusion with CI Friendly version.
Combined with Nisse https://github.com/maveniverse/nisse they can have full experience.

gnodet (approved):

cstamas:
Last example of failure but also the "escape hatch" in effect: https://gist.github.com/cstamas/1e880b2381bf1c543db052a1298e5074

dalbani:
Thanks for keeping the old behaviour with `maven.build.allowExpressionInEffectiveProjectVersion`, as I'm using a setup based on `versions-maven-plugin` to dynamically replace `${revision}`.
As in:
```xml
<project xmlns="http://maven.apache.org/POM/4.0.0"
         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
    <groupId>com.example.app</groupId>
    <artifactId>app</artifactId>
    <version>${revision}</version>

    <properties>
        <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
        <project.reporting.outputEncoding>UTF-8</project.reporting.outputEncoding>
        <java.version>21</java.version>
        <maven.build.allowExpressionInEffectiveProjectVersion>true</maven.build.allowExpressionInEffectiveProjectVersion>
        <maven.compiler.source>${java.version}</maven.compiler.source>
        <maven.compiler.target>${java.version}</maven.compiler.target>
    </properties>

    <build>
        <finalName>${project.artifactId}</finalName>

        <plugins>
            <plugin>
                <groupId>io.github.git-commit-id</groupId>
                <artifactId>git-commit-id-maven-plugin</artifactId>
                <version>9.0.2</version>
                <executions>
                    <execution>
                        <goals>
                            <goal>revision</goal>
                        </goals>
                        <phase>initialize</phase>
                    </execution>
                </executions>
                <configuration>
                    <abbrevLength>8</abbrevLength>
                    <commitIdGenerationMode>full</commitIdGenerationMode>
                    <generateGitPropertiesFile>true</generateGitPropertiesFile>
                </configuration>
            </plugin>
        </plugins>
    </build>

    <profiles>
        <profile>
            <id>set-version</id>

            <build>
                <plugins>
                    <plugin>
                        <groupId>org.codehaus.mojo</groupId>
                        <artifactId>versions-maven-plugin</artifactId>
                        <version>2.18.0</version>
                        <executions>
                            <execution>
                                <goals>
                                    <goal>set</goal>
                                </goals>
                                <phase>initialize</phase>
                            </execution>
                        </executions>
                        <configuration>
                            <generateBackupPoms>false</generateBackupPoms>
                        </configuration>
                    </plugin>
                </plugins>
            </build>
        </profile>
    </profiles>
</project>
```

Which I can then use as:
```sh
if [ $CI_COMMIT_BRANCH == $CI_DEFAULT_BRANCH ]; then
  NEW_VERSION='${git.total.commit.count}'
else
  NEW_VERSION=${CI_COMMIT_REF_SLUG}-SNAPSHOT
fi
./mvnw --activate-profiles set-version --define newVersion=${NEW_VERSION} initialize
./mvnw source:jar deploy
```

That way, I can let `git-commit-id-maven-plugin` calculate the version without any `git` command in my CI environment.

Hathoute:
This change broke our pipeline today (we had `${revision}` while using the flatten plugin, but we migrated to `versions:set` and left the version with the revision).

We are using eclipse-temurin image to build maven projects, and they just pushed the 3.9.10 images yesterday (which replaced the `3-eclipse-temurin-xx` docker images).

PR #1660, opened 2024-08-16
[MNG-8214] Improve model velocity template to support subclasses

kwin:
This makes the constructor protected and takes the Builder as argument

       pulling in other changes.

### DIFF
--- a/src/mdo/model.vm
+++ b/src/mdo/model.vm
@@ -136,9 +136,9 @@ public class ${class.name}
     final ${type} $field.name;
     #end
     #if ( $locationTracking )
-      #if ( ! $class.superClass )
-    /** Locations */
+    /** Locations (this potentially hides the same name field from the super class) */
     final Map<Object, InputLocation> locations;
+      #if ( ! $class.superClass )
     /** Location tracking */
     final InputLocation importedFrom;
       #end
@@ -171,13 +171,16 @@ public class ${class.name}
         Map<Object, InputLocation> newlocs = builder.locations != null ? builder.locations : Collections.emptyMap();
         Map<Object, InputLocation> oldlocs = builder.base != null && builder.base.locations != null ? builder.base.locations : Collections.emptyMap();
       #if ( ! $class.superClass )
-        this.locations = new HashMap<>();
+        Map<Object, InputLocation> mutableLocations = new HashMap<>();
         this.importedFrom = builder.importedFrom;
-        this.locations.put("", newlocs.containsKey("") ? newlocs.get("") : oldlocs.get(""));
+        mutableLocations.put("", newlocs.containsKey("") ? newlocs.get("") : oldlocs.get(""));
+      #else
+        Map<Object, InputLocation> mutableLocations = new HashMap<>(super.locations);
       #end
         #foreach ( $field in $class.getFields($version) )
-        this.locations.put("${field.name}", newlocs.containsKey("${field.name}") ? newlocs.get("${field.name}") : oldlocs.get("${field.name}"));
+        mutableLocations.put("${field.name}", newlocs.containsKey("${field.name}") ? newlocs.get("${field.name}") : oldlocs.get("${field.name}"));
         #end
+        this.locations = Collections.unmodifiableMap(mutableLocations);
     #end
     }
 
@@ -234,7 +237,7 @@ public class ${class.name}
     }
 
     #end
-    #if ( $locationTracking && !$class.superClass )
+    #if ( $locationTracking )
     /**
      * Gets the location of the specified field in the input source.
      */
@@ -249,6 +252,7 @@ public class ${class.name}
         return locations != null ? locations.keySet() : null;
     }
 
+    #if ( !$class.superClass )
     /**
      * Gets the input location that caused this model to be read.
      */
@@ -257,6 +261,7 @@ public class ${class.name}
         return importedFrom;
     }
 
+    #end
     #end
     /**
      * Creates a new builder with this object as the basis.
--- a/src/mdo/model.vm
+++ b/src/mdo/model.vm
@@ -482,7 +482,7 @@ public class ${class.name}
     #end
         @Nonnull
         public ${class.name} build() {
-            // this method should not contain any logic other than creating the object in order to ease subclassing
+            // this method should not contain any logic other than creating (or reusing) an object in order to ease subclassing
             if (base != null
     #foreach ( $field in $allFields )
                     && (${field.name} == null || ${field.name} == base.${field.name})
--- a/src/mdo/model.vm
+++ b/src/mdo/model.vm
@@ -145,55 +145,39 @@ public class ${class.name}
     #end
 
     /**
-      * Constructor for this class, package protected.
+      * Constructor for this class, to be called from its subclasses and {@link Builder}.
       * @see Builder#build()
       */
-    ${class.name}(
-    #if ( $class == $root )
-        String namespaceUri,
-        String modelEncoding,
-    #end
-    #foreach ( $field in $allFields )
-      #set ( $sep = "#if(${locationTracking}||$field!=${allFields[${allFields.size()} - 1]}),#end" )
-      #set ( $type = ${types.getOrDefault($field,${types.getOrDefault($field.type,$field.type)})} )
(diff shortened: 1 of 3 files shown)


gnodet:
I don't really see what's your use case for that ? We already have inheritance such as `Build` extends `BuildBase` and it works correctly.  Is that just a cleanup ?

kwin:
This is not about inheritance within MDO models but inheritance of classes generated from MDO, like for https://github.com/apache/maven-plugin-tools/blob/9332b09acc08d2e063f12a8158a57ca8ec56fadd/maven-plugin-tools-api/src/main/java/org/apache/maven/tools/plugin/ExtendedMojoDescriptor.java#L31. This is currently not possible due to the restricted visibility of the constructors and the lack of a constructor taking a builder as argument.

src/mdo/model.vm
```diff
@@ -142,46 +142,33 @@ public class ${class.name}
...
+            builder${sep}
       #if ( $locationTracking )
             locations
       #end
         );
     #end
     #if ( $class == $root )
-        this.namespaceUri = namespaceUri;
-        this.modelEncoding = modelEncoding;
+        this.namespaceUri = builder.namespaceUri;
```
kwin, line 160: This requires the base inheritance logic as well

gnodet:
> This is not about inheritance within MDO models but inheritance of classes generated from MDO, like for https://github.com/apache/maven-plugin-tools/blob/9332b09acc08d2e063f12a8158a57ca8ec56fadd/maven-plugin-tools-api/src/main/java/org/apache/maven/tools/plugin/ExtendedMojoDescriptor.java#L31. This is currently not possible due to the restricted visibility of the constructors and the lack of a constructor taking a builder as argument.

Fair point.  However the use case of ExtendedMojoDescriptor is definitely a bad one.  Those were needed because the models did not really evolved, but we allowed small additions without properly versioning them, and then rely on such hacks.

kwin:
> However the use case of ExtendedMojoDescriptor is definitely a bad one.

I only partially agree, extending the MDO for this case is not useful as the additional attribute really only is useful in the plugin-tools context and therefore this class is only used internally. Although one could probably do things differently, the effort when not extending the original Maven classes is considerably higher (as one somehow needs to establish the connection between the raw Maven object and the extended plugin tools one).

src/mdo/model.vm
```diff
@@ -142,46 +142,33 @@ public class ${class.name}
...
-      #if ( $type.startsWith("List<") )
-        #set ( $type = ${type.replace('List<','Collection<')} )
-      #end
-        $type $field.name${sep}
-    #end
-    #if ( $locationTracking )
-        Map<Object, InputLocation> locations
-    #end
-    ) {
+    #set ( $additionalArguments = "#if(${locationTracking}), Map<Object, InputLocation> locations#end" )
```
gnodet, line 148: The `$additionalArguments` need to be removed, they are also fields on the `builder` argument.

src/mdo/model.vm
```diff
@@ -480,23 +467,8 @@ public class ${class.name}
...
-                ${field.name} != null ? ${field.name} : (base != null ? base.${field.name} : ${field.defaultValue})${sep}
-      #else
-                ${field.name} != null ? ${field.name} : (base != null ? base.${field.name} : null)${sep}
-      #end
-    #end
-    #if ( $locationTracking )
-                locations
-    #end
-            );
+    #set ( $additionalArguments = "#if(${locationTracking}), locations#end" )
```
gnodet, line 470: Same here

gnodet (changes requested):
The `$additionalArguments` needs to be removed

src/mdo/model.vm
```diff
@@ -145,55 +145,39 @@ public class ${class.name}
...
+        #end
       #end
     #end
     #if ( $locationTracking )
+        Map<Object, InputLocation> newlocs = builder.locations != null ? builder.locations : Collections.emptyMap();
+        Map<Object, InputLocation> oldlocs = builder.base != null && builder.base.locations != null ? builder.base.locations : Collections.emptyMap();
       #if ( ! $class.superClass )
-        this.locations = ImmutableCollections.copy(locations);
-        this.importedFrom = importedFrom;
+        this.locations = new HashMap<>();
```
gnodet, line 174: Fields should be immutable.  Maybe the generated should look like:
```
        this.locations = Map.of(
                "", newlocs.containsKey("") ? newlocs.get("") : oldlocs.get(""),
                "modules", newlocs.containsKey("modules") ? newlocs.get("modules") : oldlocs.get("modules"),
                "distributionManagement", newlocs.containsKey("distributionManagement") ? newlocs.get("distributionManagement") : oldlocs.get("distributionManagement"),
                "properties", newlocs.containsKey("properties") ? newlocs.get("properties") : oldlocs.get("properties"),
                "dependencyManagement", newlocs.containsKey("dependencyManagement") ? newlocs.get("dependencyManagement") : oldlocs.get("dependencyManagement"),
                "dependencies", newlocs.containsKey("dependencies") ? newlocs.get("dependencies") : oldlocs.get("dependencies"),
                "repositories", newlocs.containsKey("repositories") ? newlocs.get("repositories") : oldlocs.get("repositories"),
                "pluginRepositories", newlocs.containsKey("pluginRepositories") ? newlocs.get("pluginRepositories") : oldlocs.get("pluginRepositories"),
                "reporting", newlocs.containsKey("reporting") ? newlocs.get("reporting") : oldlocs.get("reporting"));
```

gnodet (changes requested):
The `locations` field should be immutable.

src/mdo/model.vm
kwin: This is orthogonal to this PR. The same logic from https://github.com/apache/maven/pull/1660/files#diff-ad6fc545933fb055ede5c7cf4ca4f05c63de54967e3253e4244b00adcaa7dcc9L503-L511 just has been moved to the constructor.

gnodet: Well, no.  The constructor was getting a possibly mutable map, but it was doing an immutable copy before assigning it to the field.

gnodet: I really dislike the knowledge that is being spread across the `Builder#build()` method and the actual constructor.  But the shallow-copy would be really a performance loss during the initial transformation of the model.

I think it would be much easier to simply change the visibility of constructors to `protected`, I think this would be enough to allow extension from outside the package, as the generated code already allows extension (as we have `Model` extending `ModelBase` and other similar classes).

kwin: > I think it would be much easier to simply change the visibility of constructors to protected

This would require duplicating all of the functionality from the super-classes `Builder.build()` method! This is exactly what I want to prevent with this approach.

The `locations` map contains `null` values therefore cannot leverage `Map.ofEntries`.

gnodet: > > I think it would be much easier to simply change the visibility of constructors to protected
> 
> This would require duplicating all of the functionality from the super-classes `Builder.build()` method! This is exactly what I want to prevent with this approach.

If you find a better way that keep the current features... but the objects have to be immutable.
Honestly, this is a very rare case, so duplicating a few lines of code may not be so dramatic.
And as I said, the model is now supposed to evolve, rather than be hacked externally (with an outdated XSD + model which do not reflect the reality, etc...)

kwin: Done now in https://github.com/apache/maven/pull/1660/commits/8ce4cdedd2040f9a97b2cc6a586b67b4aae749e6. Please check again.

src/mdo/model.vm
```diff
@@ -493,40 +482,15 @@ public class ${class.name}
     #end
         @Nonnull
         public ${class.name} build() {
+            // this method should not contain any logic other than creating the object in order to ease subclassing
```
gnodet, line 485: Is that a hint that the below code should be removed ?  If so, I disagree, as this is an important optimisation.
If not, maybe the line should be removed or made more explicit...

kwin: No it isn't, this needs to be kept as optimization. I clarified the comment in https://github.com/apache/maven/pull/1660/commits/effa73709e4b49a9509f0891ee3c6225b0b9a02c.

gnodet (approved):

PR #1684, opened 2024-08-27
[MNG-8218] Always normalize Path in PathSource

gnodet:
Possible fix for https://issues.apache.org/jira/browse/MNG-8218
An IT would be interesting to determine if this is the only problem.


cstamas (approved):

jira-importer:
Resolve #9785

PR #1690, opened 2024-08-29
[MNG-8229] CI Friendly versions confuse model pool

cstamas:
Fixing the method (there is one GA inside reactor), but this still does not fix the issue, now it fails elsewhere.

The problem in short is CI friendly version and its half-assed solution as model version "suddenly" change from file-activated models to effective ones.

---

https://issues.apache.org/jira/browse/MNG-8229

### DIFF
--- a/maven-core/src/main/java/org/apache/maven/project/ReactorModelPool.java
+++ b/maven-core/src/main/java/org/apache/maven/project/ReactorModelPool.java
@@ -19,6 +19,7 @@
 package org.apache.maven.project;
 
 import java.nio.file.Path;
+import java.util.Collection;
 import java.util.Collections;
 import java.util.HashSet;
 import java.util.Map;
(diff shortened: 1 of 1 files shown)


maven-core/src/main/java/org/apache/maven/project/ReactorModelPool.java
```diff
@@ -48,13 +49,20 @@ class ReactorModelPool {
...
+            return models.iterator().next();
+        } else {
+            return models.stream()
+                    .filter(m -> version == null || version.equals(getVersion(m)))
+                    .reduce((a, b) -> {
+                        throw new IllegalStateException(
+                                "Multiple modules with key " + a.getGroupId() + ':' + a.getArtifactId());
+                    })
+                    .orElse(null);
+        }
```
gnodet, line 65: This does not change the behaviour, right ? What's the point ?

gnodet: Oh, it does change the behaviour, as if there's a single version, it will always be used, even if there's a version mismatch.  I think that's wrong.
Multiple versions of the same GA inside a reactor is a supported (albeit rare) use case.

jira-importer:
Resolve #9982

PR #1696, opened 2024-08-30
[MNG-7255] Infer groupId for intra-reactor dependencies

gnodet:
JIRA issue: [MNG-7255](https://issues.apache.org/jira/browse/MNG-7255)
IT PR: https://github.com/apache/maven-integration-testing/pull/356

### DIFF
--- a/maven-api-impl/src/main/java/org/apache/maven/internal/impl/model/BuildModelTransformer.java
+++ b/maven-api-impl/src/main/java/org/apache/maven/internal/impl/model/BuildModelTransformer.java
@@ -28,6 +28,7 @@ import java.util.Optional;
 import org.apache.maven.api.di.Named;
 import org.apache.maven.api.di.Singleton;
 import org.apache.maven.api.model.Dependency;
+import org.apache.maven.api.model.InputLocation;
 import org.apache.maven.api.model.Model;
 import org.apache.maven.api.model.Parent;
 import org.apache.maven.api.services.ModelTransformer;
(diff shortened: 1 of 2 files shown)


cstamas (approved):
Simple as that.

But a "minority report": I personally disagree with this, as it will encourage using A only for dependencies (as G and V are inferred), thus, will make depMgt less compelling. Not to mention how will this plus BOMs behave?

gnodet:
> Simple as that.
> 
> But a "minority report": I personally disagree with this, as it will encourage using A only for dependencies (as G and V are inferred), thus, will make depMgt less compelling. Not to mention how will this plus BOMs behave?

Not sure to understand.  We now warn when using BOMs inside a reactor for dependency management and the only point in using dependency management for dependencies coming from the reactor is exactly to avoid writing the version. I don't really see how this is different.  Note that this only applies to dependencies that are part of the reactor.  Projects that have multiple subprojects usually define in their dependency management section the list of artifacts actually produced during the build to avoid having to specify the version everywhere.  That's what we have in [maven](https://github.com/apache/maven/blob/master/pom.xml#L201-L353), [maven-resolver](https://github.com/jline/jline3/blob/4ccf282c91daa0fc96ed6030e053dd28a4aba342/builtins/src/test/java/org/jline/example/Example.java#L328-L335) and almost all projects I've seen.  I don't see any benefit in having to maintain this section manually, since there's absolutely no value there.  What am I missing ? @cstamas

gnodet:
> > Simple as that.
> > But a "minority report": I personally disagree with this, as it will encourage using A only for dependencies (as G and V are inferred), thus, will make depMgt less compelling. Not to mention how will this plus BOMs behave?
> 
> Not sure to understand. We now warn when using BOMs inside a reactor for dependency management and the only point in using dependency management for dependencies coming from the reactor is exactly to avoid writing the version. I don't really see how this is different. Note that this only applies to dependencies that are part of the reactor. Projects that have multiple subprojects usually define in their dependency management section the list of artifacts actually produced during the build to avoid having to specify the version everywhere. That's what we have in [maven](https://github.com/apache/maven/blob/master/pom.xml#L201-L353), [maven-resolver](https://github.com/jline/jline3/blob/4ccf282c91daa0fc96ed6030e053dd28a4aba342/builtins/src/test/java/org/jline/example/Example.java#L328-L335) and almost all projects I've seen. I don't see any benefit in having to maintain this section manually, since there's absolutely no value there. What am I missing ? @cstamas

I've just added location tracking. Would you like to add more checks ?  Instead of using the current pom `groupId` to infer the missing value, we could check if there's a single matching `artifactId` in the reactor and use that one instead, just like we do for the `version`.

PR #1700, opened 2024-09-04
[MNG-8120] Refactor ModelBuilder and ProjectBuilder

gnodet:
Refactor `ModelBuilder` and `ProjectBuilder`.
With the introduction of the build pom and raw -> build pom transformation, the construction of the effective poms in two steps become very problematic.  Over the time, multiple caches have been added to the ProjectBuilder and ModelBuilder related classes which are often redundant.

This PR thus changes things and move the recursive construction of the models fully into the `ModelBuilder` in a single call.  When building build poms, a first step is done by parsing the file models from the root, then building all needed effective models from those. All the inference can be cleanly done because the builder has all the file models ready.  The result will be used by the `ProjectBuilder` to build the projects.

IT PR: https://github.com/apache/maven-integration-testing/pull/357


maven-core/src/main/java/org/apache/maven/internal/impl/InternalMavenSession.java
```diff
@@ -35,6 +36,10 @@ static InternalMavenSession from(Session session) {
 
     List<Project> getProjects(List<org.apache.maven.project.MavenProject> projects);
 
+    /**
+     * May return null if the input projcet is null or is not part of the reactor.
```
cstamas, line 40: typo: project

maven-core/src/main/java/org/apache/maven/internal/transformation/impl/DefaultConsumerPomBuilder.java
```diff
@@ -79,71 +78,74 @@ class DefaultConsumerPomBuilder implements ConsumerPomBuilder {
...
+    private final ModelNormalizer modelNormalizer;
+    private final ModelPathTranslator modelPathTranslator;
+    private final ModelProcessor modelProcessor;
+    private final ModelUrlNormalizer modelUrlNormalizer;
+    private final ModelValidator modelValidator;
+    private final PluginConfigurationExpander pluginConfigurationExpander;
+    private final PluginManagementInjector pluginManagementInjector;
+    private final SuperPomProvider superPomProvider;
+    private final ModelVersionParser versionParser;
+    // To break circular dependency    private final Provider<RepositorySystem> repositorySystem;
```
cstamas, line 96: drop this?

cstamas (approved):

PR #1701, opened 2024-09-04
[MNG-8233] Client should handle modelIds as "opaque"

cstamas:
Javadoc states wrong things (about format of modelId, they are paths in reality), but also these all are "implementation details". Crafting modelIds to "walk" lineage is wrong, there is a dedicated method that provide "opaque" key Strings in defined order to do that. And client should not care (nor assume) anything about format of the key String.

As per MNG-8233, "modelIds" in Maven4 are NOT `g:a:v`s as Javadoc wrongly states, but paths.

---

https://issues.apache.org/jira/browse/MNG-8233


gnodet (approved):
I'm not sure we should keep them after MNG-8120, but I'm fine with those changes for now.

PR #1710, opened 2024-09-09
[MNG-8230] Rewrite CI friendly versions

gnodet:
CI Friendly versions are now fully opened, so not limited anymore to the 3 predefined properties which were `sha1`, `changelist` and `revision`.
So expression on version and parent version are interpolated with their proper values, taken from the session user properties and from the session's root project properties (uninterpolated).
An additional validation check after computing this file model to ensure that no expression is used in version at this point. This ensures that during the next step (raw model build), all the full GAV are known, as they will be used to infer missing dependency information.  It allows remove possible problems with plugins, not being fully aware of those versions.

--------
JIRA issue: https://issues.apache.org/jira/browse/MNG-8230
IT PR: https://github.com/apache/maven-integration-testing/pull/366


cstamas (commented):
This PR seems mixed with DI?

gnodet:
> This PR seems mixed with DI?

Yes, it requires a session scope which is not well supported by the sisu/di bridge.  I'll rebase once #1722 is merged.

cstamas:
IMO, the proper way of working for CFV is:
* any placeholder can be used (OK)
* as long it has value defined, but this part is NOT OK, at least I see no code that makes sure version is not left placeholder?

Or I miss something?

gnodet:
> IMO, the proper way of working for CFV is:
> 
> * any placeholder can be used (OK)
> * as long it has value defined, but this part is NOT OK, at least I see no code that makes sure version is not left placeholder?
> 
> Or I miss something?

This role is supposed to be hold by the `ModelValidator` ...

cstamas (approved):

jira-importer:
Resolve #9777

PR #1719, opened 2024-09-11
[MNG-8261] Remove jline-terminal-ffm direct dependency

gnodet:
JIRA issue: https://issues.apache.org/jira/browse/MNG-8261

### DIFF
--- a/apache-maven/pom.xml
+++ b/apache-maven/pom.xml
@@ -99,7 +99,15 @@ under the License.
     </dependency>
     <dependency>
       <groupId>org.jline</groupId>
-      <artifactId>jline</artifactId>
+      <artifactId>jline-reader</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>org.jline</groupId>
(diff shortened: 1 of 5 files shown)


cstamas (commented):
Could not we do this simpler? Maven really needs jansi-core, so everything else can be added in apache-maven "bundle" only, no?

gnodet:
> Could not we do this simpler? Maven really needs jansi-core, so everything else can be added in apache-maven "bundle" only, no?

Well, that's what I did, i.e. make sure `maven-embedded` only depends on `janis-core`.  However the `maven-jline` subproject actually depends on JLine (it provides plexus interactivity services based on JLine).
Btw, those services were broken, so I restored them, this can be tested easily by using `mvn versions:set` and using left/right arrow keys (compare Maven 3.x with this PR).

cstamas (approved):

jira-importer:
Resolve #9994

PR #1721, opened 2024-09-11
[MNG-8252] Fully infer the parent coordinates if the location points to a valid model (#1706)

gnodet:
Fix things
The original PR was passing UT and IT, but to sure how it worked
Anyway, this aims at fixing the computation. Using the `BuildModelTransformer` is not a good option: when the full reactor is loaded, we parse file models and expect them to have valid groupId/artifactId, so the processing has to be done during file model loading.


jira-importer:
Resolve #9766

PR #1723, opened 2024-09-12
[MNG-8256] FilteredProjectDependencyGraph fix for non-transitive case

cstamas:
The `FilteredProjectDependencyGraph` class fix for non-transitive case and an UT. Also contains some improvement in classes, making them immutable as intended.

---

https://issues.apache.org/jira/browse/MNG-8256


gnodet (approved):

jira-importer:
Resolve #9606

PR #1725, opened 2024-09-16
DI cleanup

gnodet:
- **Implement DI scopes**
- **Use DI OutOfScopeException in session's scope proxies**
- **Make sure we don't discover the same components multiple times**
- **Use proper DI api and move discovery into a module**
- **Add session scope support in ApiRunner**
- **Use injector's discovery**


gnodet:
Incorporated in #1722

PR #1739, opened 2024-09-23
[MNG-8264] Deprecate maven 3 specific modules

gnodet:
- **[MNG-8264] Deprecate maven-toolchain-builder**
- **[MNG-8264] Deprecate maven-settings-builder**
- **[MNG-8264] Deprecate maven-resolver-provider**
- **[MNG-8264] Deprecate maven-model-builder**

### DIFF
--- a/maven-model-builder/pom.xml
+++ b/maven-model-builder/pom.xml
@@ -28,7 +28,7 @@ under the License.
 
   <artifactId>maven-model-builder</artifactId>
 
-  <name>Maven Model Builder</name>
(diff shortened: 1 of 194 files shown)


cstamas (approved):

cstamas:
maven-artifact?

gnodet:
> maven-artifact?

This would be part of a second deprecation imho.  This round is mostly about deprecating subprojects which are not used anymore when booting Maven 4.  So any extension that was extending those parts will definitely not work.

PR #1743, opened 2024-09-24
Add license header to m2.conf

XenoAmess:
...just found no reason not adding it, as all other scripts/shell have added, and this file format actually can accept \# comments...

### DIFF
--- a/apache-maven/pom.xml
+++ b/apache-maven/pom.xml
@@ -151,7 +151,6 @@ under the License.
           <artifactId>apache-rat-plugin</artifactId>
           <configuration>
             <excludes combine.children="append">
-              <exclude>src/assembly/maven/bin/m2.conf</exclude>
               <!-- these are partial scripts, resulting in mvn scripts -->
               <exclude>src/assembly/shared/init</exclude>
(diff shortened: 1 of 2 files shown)


cstamas:
Perso I'd rather fix spotless on BOM, as we still want BOM formatted (instead as in this PR, skipping it) @gnodet ?

maven-bom/pom.xml
```diff
@@ -204,6 +204,14 @@ under the License.
           </execution>
         </executions>
       </plugin>
+      <plugin>
+        <groupId>com.diffplug.spotless</groupId>
+        <artifactId>spotless-maven-plugin</artifactId>
```
cstamas, line 209: This is a no go

cstamas (changes requested):

maven-bom/pom.xml
XenoAmess: @cstamas removed. hope you find another better way for resolving this.

gnodet:
> Perso I'd rather fix spotless on BOM, as we still want BOM formatted (instead as in this PR, skipping it) @gnodet ?

Which BOM is broken ?

gnodet (approved):

PR #1772, opened 2024-10-03
Maven CLIng smaller bugfixed and importments

cstamas:
And now the UTs are running only when in Maven4

### DIFF
--- a/maven-cli/src/test/java/org/apache/maven/cling/invoker/forked/DefaultForkedMavenInvokerTest.java
+++ b/maven-cli/src/test/java/org/apache/maven/cling/invoker/forked/DefaultForkedMavenInvokerTest.java
@@ -20,28 +20,34 @@ package org.apache.maven.cling.invoker.forked;
 
 import java.nio.file.Files;
 import java.nio.file.Path;
+import java.nio.file.Paths;
 
+import org.apache.maven.api.cli.ParserRequest;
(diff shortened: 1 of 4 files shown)


maven-cli/src/test/java/org/apache/maven/cling/invoker/forked/DefaultForkedMavenInvokerTest.java
```diff
@@ -20,28 +20,34 @@
...
 import org.junit.jupiter.api.Test;
 import org.junit.jupiter.api.io.TempDir;
 
-@Disabled("Works ONLY with Maven4!")
 public class DefaultForkedMavenInvokerTest {
     @Test
     void smoke(@TempDir Path tempDir) throws Exception {
+        // works only in recent Maven4
+        Assumptions.assumeTrue(Files.isRegularFile(
+                Paths.get(System.getProperty("maven.home")).resolve("conf").resolve("maven.properties")));
```
gnodet, line 39: Does this mean they should be ITs ?

cstamas: Um, no, they just require this system property to be set (surefire in POM does it) AND that conf/maven.propeties exists that was introduced on latest Maven4

gnodet (approved):

jira-importer:
Resolve #9978

PR #1775, opened 2024-10-04
[MNG-8283] More mvnd related changes

cstamas:
Changes:
* there is only one parser for "just maven", no need for 3
* aligned scopes (public) of local context for simplicity, we can fix visibility later
* allow custom guice modules (unused, may undo this)
* split logging setup in two steps: config and activate

---

https://issues.apache.org/jira/browse/MNG-8283


gnodet (approved):

jira-importer:
Resolve #9978

PR #1782, opened 2024-10-07
[MNG-8288] "path cannot be null"

cstamas:
It seems Maven is eager to get POM from rootDirectory to support CI Friendly versions. This change makes it "best effort", as in if there is no POM, just walk away.

---

https://issues.apache.org/jira/browse/MNG-8288


jira-importer:
Resolve #9750

PR #1787, opened 2024-10-08
Make rootDirectory mandatory

cstamas:
Make rootDirectory mandatory. Expose locator (and detectors) via ServiceLoaders, as with polyglot more detector may be needed (ie polyglot extension will need to register own "detector").


gnodet:
I don't really see why we would default `rootDirectory` to `${maven.multimoduleDirectory}`. 
The `mvnd` client does not have to set the `rootDirectory`, only the `cwd` and arguments.  It also set `${maven.multimoduleDirectory}` for backward compatibility.

gnodet (approved):
Could you have a look at my changes?
I'm fine otherwise.

cstamas:
Looks good +1

PR #1798, opened 2024-10-15
Cleanup after last commit

cstamas:
LookupInvoker does terminal stuff now, remove them from "early" entry classes, also adopt mvnenc


maven-cli/src/main/java/org/apache/maven/cling/invoker/LookupInvoker.java
```diff
@@ -413,6 +413,7 @@ protected void helpOrVersionAndMayExit(C context) throws Exception {
             } else {
                 context.terminal.writer().println(CLIReportingUtils.showVersion());
             }
+            context.terminal.writer().flush();
```
gnodet, line 416: I would add a flush after displaying the help, just so that we don't rely on `displayHelp` to call `flush`...

gnodet (approved):

PR #1801, opened 2024-10-15
[MNG-8311] empty but existing <localRepository/> in settings.xml

danthe1st:
In https://issues.apache.org/jira/browse/MNG-8311, I reported the issue that an empty `<localRepository/>` element in the settings.xml uses the current working directory or home directory for the local repository.

This PR fixes that issue by adding a check for the empty `String`.


danthe1st:
Note that I think I cannot self-assign myself to the issue or close it even though it was created by me. It might be necessary for someone else with the necessary permissions to do that.

cstamas (approved):

PR #1809, opened 2024-10-16
[MNG-8296] Support creating unknown Language on the fly

gnodet:
JIRA issue: [MNG-8296](https://issues.apache.org/jira/browse/MNG-8296)


cstamas:
Will this help to the use case in JIRA?

gnodet:
> Will this help to the use case in JIRA?

Yes, that's a fix. Maven 3 does not have this `Language`, it's just a `String` on the ArtifactHandler, so we need to create those on the fly.

cstamas (approved):

jira-importer:
Resolve #9566

PR #1810, opened 2024-10-16
[MNG-8302] Warn when appropriate

cstamas:
First, `rootDirectory` is nullable, CLIng code was not assuming this, fixed.
Second, emit the "no root found" warning ONLY when appropriate (when we have a POM in picture).

---

https://issues.apache.org/jira/browse/MNG-8302

### DIFF
--- a/maven-cli/src/main/java/org/apache/maven/cling/invoker/Utils.java
+++ b/maven-cli/src/main/java/org/apache/maven/cling/invoker/Utils.java
@@ -31,7 +31,6 @@ import java.util.function.Function;
 
 import org.apache.maven.api.annotations.Nonnull;
 import org.apache.maven.api.annotations.Nullable;
-import org.apache.maven.api.cli.ParserException;
 import org.apache.maven.api.services.model.RootLocator;
 import org.apache.maven.cli.logging.Slf4jConfiguration;
 import org.apache.maven.execution.MavenExecutionRequest;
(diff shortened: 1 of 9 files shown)


maven-cli/src/main/java/org/apache/maven/cling/invoker/Utils.java
```diff
@@ -144,4 +150,37 @@ public static int toPlexusLoggingLevel(Slf4jConfiguration.Level level) {
...
+    }
+
+    @Nonnull
+    public static Path findMandatoryRoot(Path topDirectory) throws ParserException {
+        requireNonNull(topDirectory, "topDirectory");
+        Path rootDirectory;
+        for (RootLocator rootLocator : ServiceLoader.load(RootLocator.class).stream()
+                .map(ServiceLoader.Provider::get)
+                .toList()) {
+            rootDirectory = rootLocator.findMandatoryRoot(topDirectory);
```
gnodet, line 179: In this case, we'll take only the first `RootLocator`, because calling `findMandatoryRoot` should throw an exception if that `RootLocator` cannot find the root.  That's not really in line with the `findRoot` (not mandatory above).
Should that be:
```
return Optional.ofNullable(findMandatory(topDirectory)).orElseThrow();
```

gnodet:
Should we create an IT to check the various use cases ?

gnodet (approved):

cstamas:
> Should we create an IT to check the various use cases ?

We could, but this is really minimal impact and more "user experience", nothing biggie IMO.

PR #1821, opened 2024-10-19
[MNG-8331] Sometimes versioned dependencies disappear

jjkester:

       pulling in other changes.

### DIFF
--- a/maven-api-impl/src/main/java/org/apache/maven/internal/impl/model/DefaultModelBuilder.java
+++ b/maven-api-impl/src/main/java/org/apache/maven/internal/impl/model/DefaultModelBuilder.java
@@ -564,8 +564,7 @@ public class DefaultModelBuilder implements ModelBuilder {
         // Infer inner reactor dependencies version
         //
         Model transformFileToRaw(Model model) {
-            List<Dependency> newDeps = new ArrayList<>();
-            boolean modified = false;
+            List<Dependency> newDeps = new ArrayList<>(model.getDependencies().size());
             for (Dependency dep : model.getDependencies()) {
                 if (dep.getVersion() == null) {
                     Dependency.Builder depBuilder = null;
@@ -591,13 +590,14 @@ public class DefaultModelBuilder implements ModelBuilder {
                     }
                     if (depBuilder != null) {
                         newDeps.add(depBuilder.build());
-                        modified = true;
                     } else {
                         newDeps.add(dep);
                     }
+                } else {
+                    newDeps.add(dep);
                 }
             }
-            return modified ? model.withDependencies(newDeps) : model;
+            return model.withDependencies(newDeps);
         }
 
         String replaceCiFriendlyVersion(Map<String, String> properties, String version) {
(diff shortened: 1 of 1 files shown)


mthmulders:
Corresponding IT: https://github.com/apache/maven-integration-testing/pull/393.

cstamas:
There is a more trivial PR https://github.com/apache/maven/pull/1822

cstamas:
Another alternative mixing the two solutions: https://github.com/apache/maven/pull/1823

cstamas:
@jjkester could you please add the "empty check" to the beginning of the method (so avoid any work and allocation if there is nothing to be done)? like in #1823

jjkester:
> @jjkester could you please add the "empty check" to the beginning of the method (so avoid any work and allocation if there is nothing to be done)? like in #1823

I did not add one because in practice I do not expect many projects to have zero dependencies. If a project does have zero dependencies, the only "waste" is the allocation of an unnecessary `ArrayList` of size 0, and unnecessarily creating a new `Model` instance with that list. Personally I don't think that optimalization (given the assumed rarity) is worth the additional complexity in the method, which is already quite complex.

If my assumption and/or preference is not in line with the general way in which Maven is coded, I'm happy to add the optimalization, but I'd like to check the necessity first.

cstamas:
For example, all the parents, aggregators, BOMs etc will not have any of those. This is about all the "subprojects" within a project (just to clarify "I do not expect project with zero dependencies" -- they may still have subprojects with zero deps even if project as a whole does have deps) and while I cannot tell the _usual ratio_ of things like parent vs non-parent POMs in a project, they are there.

Just my 5cts

cstamas:
Ugh, reformatting the file will fail the build I guess

jjkester:
Reverted the accidental reformat, force-of-habit hotkey in IntelliJ...

maven-api-impl/src/main/java/org/apache/maven/internal/impl/model/DefaultModelBuilder.java
```diff
@@ -564,20 +564,21 @@ public void mergeRepositories(List<Repository> toAdd, boolean replace) {
...
-            List<Dependency> newDeps = new ArrayList<>();
-            boolean modified = false;
+            if (model.getDependencies().isEmpty()) {
+                return model;
+            }
+            List<Dependency> newDeps = new ArrayList<>(model.getDependencies().size());
             for (Dependency dep : model.getDependencies()) {
                 if (dep.getVersion() == null) {
-                    Dependency.Builder depBuilder = null;
                     Model depModel = getRawModel(model.getPomFile(), dep.getGroupId(), dep.getArtifactId());
```
jjkester, line 573: @gnodet I think this part of the code (starting here or below `if (depModel != null)`) is a good candidate for extraction to improve readability. It seems to just apply the new logic to retrieve the version from the aggregated build. However, I do not have a clear name for that method since I'm not familiar with the common terminology. Do you have a suggestion?

gnodet: Good point, I've extracted it in https://github.com/apache/maven/pull/1822

mthmulders (approved):
The changes make sense to me. I'm happy to see the added IT's have passed, proving the bug is fixed.

gnodet:
> The changes make sense to me. I'm happy to see the added IT's have passed, proving the bug is fixed.

For the IT to run, a branch in the maven-integration-testing fork with the same name than this branch and in the same org must exists.

mthmulders:
> For the IT to run, a branch in the maven-integration-testing fork with the same name than this branch and in the same org must exists.

This is exactly what is happening: look for `MavenITmng8331VersionedAndUnversionedDependenciesTest` in [this pipeline](https://github.com/apache/maven/actions/runs/11417228901/job/31769368782?pr=1821).

PR #1822, opened 2024-10-19
[MNG-8331] Fix disappearing dependencies

gnodet:
JIRA issue: [MNG-8331](https://issues.apache.org/jira/browse/MNG-8331)
IT PR: https://github.com/apache/maven-integration-testing/pull/393

### DIFF
--- a/maven-api-impl/src/main/java/org/apache/maven/internal/impl/model/DefaultModelBuilder.java
+++ b/maven-api-impl/src/main/java/org/apache/maven/internal/impl/model/DefaultModelBuilder.java
@@ -595,6 +595,8 @@ public class DefaultModelBuilder implements ModelBuilder {
                     } else {
                         newDeps.add(dep);
                     }
+                } else {
+                    newDeps.add(dep);
                 }
             }
             return modified ? model.withDependencies(newDeps) : model;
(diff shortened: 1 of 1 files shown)


maven-api-impl/src/main/java/org/apache/maven/internal/impl/model/DefaultModelBuilder.java
```diff
@@ -595,6 +595,8 @@ Model transformFileToRaw(Model model) {
                     } else {
                         newDeps.add(dep);
                     }
+                } else {
+                    newDeps.add(dep);
```
mthmulders, line 599: Because of this, the size of `newDeps` will always be equal to the number of objects in `model.getDependencies()`. We might as well pre-allocate the `ArrayList` with the correct size, so we avoid resizing `newDeps` while looping through `model.getDependencies()`.

maven-api-impl/src/main/java/org/apache/maven/internal/impl/model/DefaultModelBuilder.java
```diff
@@ -595,6 +595,8 @@ Model transformFileToRaw(Model model) {
...
                 }
             }
             return modified ? model.withDependencies(newDeps) : model;
```
mthmulders, line 602: I feel that remembering `modified` no longer makes sense and obfuscates what this method is doing. What purpose of that variable am I missing?

cstamas:
Another alternative mixing the two solutions: https://github.com/apache/maven/pull/1823

maven-api-impl/src/main/java/org/apache/maven/internal/impl/model/DefaultModelBuilder.java
gnodet: The point is to not recreate a new `Model` object is there's no need to.  We could go further and delay the creation of the list until it's actually needed, but that would mean store the index, etc... I think the code was less readable.

gnodet:
Let's go with #1821

jira-importer:
Resolve #9522

PR #1823, opened 2024-10-19
[MNG-8331] Mix of the two fixes

cstamas:
Don't do anything if not needed.

---

https://issues.apache.org/jira/browse/MNG-8331


maven-api-impl/src/main/java/org/apache/maven/internal/impl/model/DefaultModelBuilder.java
```diff
@@ -591,13 +593,14 @@ Model transformFileToRaw(Model model) {
...
-                        modified = true;
                     } else {
                         newDeps.add(dep);
                     }
+                } else {
+                    newDeps.add(dep);
                 }
             }
-            return modified ? model.withDependencies(newDeps) : model;
+            return model.withDependencies(newDeps);
```
gnodet, line 603: `model.withDependencies()` will always create a new object if not modified.
I don't see why this would be needed.

cstamas:
Closing my "mix" draft to favor one of existing PRs.

PR #1824, opened 2024-10-20
[MNG-8334] Returns redirect to log

cstamas:
It got lost when improving logging infrastructure

---

https://issues.apache.org/jira/browse/MNG-8334

PR #1826, opened 2024-10-20
[MNG-8334] Fix output redirection

gnodet:
JIRA issue: [MNG-8334](https://issues.apache.org/jira/browse/MNG-8334)

- **Fix empty output when using -l or redirecting output**
- **Fix incomplete log in file**

Alternative PR to #1824 @cstamas, and it should fix both type of redirections (`-l` and shell stdout `>` redirection)

This PR is more in line with what I had in mind.
In particular, this would allow stuff like the multiplexed output from mvnd
where we display information to the terminal, but not the actual log output.

### DIFF
--- a/api/maven-api-cli/src/main/java/org/apache/maven/api/cli/Options.java
+++ b/api/maven-api-cli/src/main/java/org/apache/maven/api/cli/Options.java
@@ -18,10 +18,10 @@
  */
 package org.apache.maven.api.cli;
 
-import java.io.PrintWriter;
 import java.util.Collection;
 import java.util.Map;
 import java.util.Optional;
+import java.util.function.Consumer;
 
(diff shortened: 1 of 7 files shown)


cstamas:
Looks good, the only strange thing is that `mvn -v -l log.txt` does not redirect (emit version to console). Also command `mvn -v > log.txt` sends ansi escapes to log file.

cstamas (approved):
Almost all issues fixed, is good. Am also for enhancing logging (like multiplexing).

One issue (or non issue): `mvn -v` is NOT (yet) colored, as I see it exits so fast that Jansi (or jline) is not yet inited.

cstamas (approved):
Last bits fixed, good to go!

jira-importer:
Resolve #9519

PR #1832, opened 2024-10-22
[MNG-8344] Support multiple operators in variable expansion

gnodet:
JIRA issue: [MNG-8344](https://issues.apache.org/jira/browse/MNG-8344)

----

The goal is to enhance the supported syntax for operators in variable expansion, so that we can use those as a ternary operator:
```
maven.user.config = ${env.MAVEN_XDG:+${env.XDG_CONFIG_HOME:-${user.home}/.config}/maven}:-${user.home}/.m2}
```


jira-importer:
Resolve #9510

PR #1843, opened 2024-10-24
[MNG-8341] Fix possible deadlock in ModelBuilder

gnodet:
The deadlock can happen when the root pom inherit from a parent which is a subproject, the subproject inheriting from an external parent.

JIRA issue: [MNG-8341](https://issues.apache.org/jira/browse/MNG-8341)
IT PR: https://github.com/apache/maven-integration-testing/pull/394

The reproducer project is https://github.com/apache/pdfbox.git master, but an IT would be a good idea.


cstamas (approved):

jira-importer:
Resolve #9512

PR #1857, opened 2024-10-26
[MNG-8340] Resolve parent according to the exact model location

gnodet:
JIRA issue: [MNG-8340](https://issues.apache.org/jira/browse/MNG-8340)

We need to double check one last time that it works, or better, add an IT.

### DIFF
--- a/impl/maven-core/src/main/java/org/apache/maven/project/DefaultProjectBuilder.java
+++ b/impl/maven-core/src/main/java/org/apache/maven/project/DefaultProjectBuilder.java
@@ -351,7 +351,7 @@ public class DefaultProjectBuilder implements ProjectBuilder {
                     ModelBuilderRequest.RequestType type = pomFile != null
                                     && this.request.isProcessPlugins()
                                     && this.request.getValidationLevel() == ModelBuildingRequest.VALIDATION_LEVEL_STRICT
-                            ? ModelBuilderRequest.RequestType.BUILD_PROJECT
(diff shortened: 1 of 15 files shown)


cstamas:
Just a plain project with one module shade + dep reduced POM would work as an IT?

cstamas (approved):

gnodet:
> Just a plain project with one module shade + dep reduced POM would work as an IT?

It needs child subproject, and the shade plugin should generate the Pom in the target folder (instead of in the same directory than the Pom).

cstamas:
Something is still not right, as this PR w/ shade in IT https://github.com/apache/maven-integration-testing/pull/400 now causes NPE

gnodet:
It seems this PR also fixes [MNG-8336](https://issues.apache.org/jira/browse/MNG-8336)

impl/maven-core/src/main/java/org/apache/maven/project/DefaultProjectBuilder.java
```diff
@@ -913,7 +913,7 @@ private Model injectLifecycleBindings(
         }
         project.setPluginArtifactRepositories(pluginRepositories);
 
-        if (request.getRequestType() == ModelBuilderRequest.RequestType.BUILD_POM) {
+        if (request.getRequestType() == ModelBuilderRequest.RequestType.BUILD_PROJECT) {
```
cstamas, line 916: One correct change from my "guesswork" :smile:

cstamas (approved):

jira-importer:
Resolve #9515

PR #1861, opened 2024-10-30
[MNG-8362] Adding some missing config properties

cstamas:
To have them listed in generated docs.

---

https://issues.apache.org/jira/browse/MNG-8362


api/maven-api-core/src/main/java/org/apache/maven/api/Constants.java
```diff
@@ -386,5 +386,26 @@ public final class Constants {
     @Config(type = "java.lang.Boolean", defaultValue = "true")
     public static final String MAVEN_CONSUMER_POM = "maven.consumer.pom";
 
+    /**
+     * User property for disabling version resolver cache.
+     *
+     * @since 3.0.0
+     */
+    @Config(type = "java.lang.Boolean", defaultValue = "false")
+    public static final String MAVEN_VERSION_RESOLVER_NO_CACHE = "aether.versionResolver.noCache";
```
gnodet, line 395: Maybe we can make it start with `maven.` instead of `aether.` to be consistent with all other properties ?

cstamas: Agreed. 
This component is in maven, but implements Resolver API (as Resolver is "incomplete"), but yes, I agree, this is a Maven implementation detail, not resolver.

cstamas: Fixed

gnodet (approved):

gnodet (approved):

jira-importer:
Resolve #9491

PR #1880, opened 2024-11-06
CI: fix caching

cstamas:
Fix and introduce caching for all jobs.


cstamas:
This is put on hold: we want to sort out CI to not pollute local repo with built artifacts.

PR #1881, opened 2024-11-06
[MNG-8370] Add maven.repo.local.head

cstamas:
As counterpart to maven.repo.local.tail, but this one "prepends" while other "appends". This means one can do like this

```
$ mvn install -Prun-its -Dmaven.repo.local.head=~/.m2/repository-it
```

And have stuff installed into dedicated IT local repo (as IT bits must be installed), instead to pollute own local repo.

---

https://issues.apache.org/jira/browse/MNG-8370


gnodet (approved):

slawekjaranowski:
@cstamas, @gnodet what do you think to add it to 3.9.x?

cstamas:
Yes plase

slawekjaranowski:
Ok, I will do

jira-importer:
Resolve #9483

PR #1882, opened 2024-11-06
[MNG-8332] Detach CLIng from deprecated Embedder

cstamas:
WIP

For start, just copy pasta and some cleanup.

---

https://issues.apache.org/jira/browse/MNG-8332


cstamas:
Now that embedder depends on cli, the extension.xml generation should be fully removed from embedder (not just xsd generation). Make it all in cli and embedder should pull what it needed.

gnodet (approved):
LGMT

PR #1887, opened 2024-11-07
IT fix: Drop ancient reporting, use latest

cstamas:
Attempt to stop ITs pulling in ancient (and vulnerable) deps.


cstamas:
Supersedes https://github.com/apache/maven/pull/1874

michael-o (approved):
Reasonable change.

PR #1910, opened 2024-11-14
Exception usage cleanup

cstamas:
This PR merely cleans up exception usage across Maven.


impl/maven-impl/src/main/java/org/apache/maven/internal/impl/model/DefaultModelBuilder.java
```diff
@@ -1733,12 +1734,50 @@ ModelSource resolveReactorModel(String groupId, String artifactId, String versio
...
+                }
+            }
+        }
+
+        private <T, E extends Exception> T cache(
+                String groupId, String artifactId, String version, String tag, ThrowingSupplier<T, E> supplier)
+                throws E {
+            return ThrowingSupplierWrapper.process(
+                    supplier, s -> cache.computeIfAbsent(groupId, artifactId, version, tag, s));
         }
```
gnodet, line 1777: This is exactly the kind of boiler plate code I wanted to avoid using runtime exceptions.

cstamas: most of this undone, we need to deal with one single exception only

cstamas: IMO, the latest push is quite sensible. Also, have to add "boilerplate code wanted to avoid" vs all the changes revealed just by the compiler as exception when from unchecked to checked... which one is worth more for you?

gnodet:
I really don't think that works.  What if the ModelBuilder calls internally a service which throws an unchecked exception ? If the caller catches ModelBuilderException (which he would have to because that one is checked), then it may miss the unchecked exceptions.  I don't really see a good justification for having only a few ones being unchecked or checked.  That could work if there were real differences, such as transient issues (as in downloading and a network error happens), but most exceptions here are not of this kind.

cstamas:
Ok but @gnodet we need to pick up at least the bugfix from this PR, and consider maybe some others (you are catching ModelBuildingException while in fact ModelBuilderException is being thrown, etc).

gnodet (approved):

PR #1913, opened 2024-11-15
[MNG-8379] Decrypt all of settings on building it

cstamas:
The decryption should happen transparently, also, resolver should not decrypt for itself only, whole maven should have access to all.

This PR also disables Maven 3.0 IT MNG-4459 as Maven4 stops with this "security through obscurity" (keep encrypted pw in memory kinda for "security reasons" but in reality any mojo or extension can decrypt anything they want).

---

https://issues.apache.org/jira/browse/MNG-8379


cstamas:
LGTM! Thanks!

gnodet (approved):

PR #1917, opened 2024-11-17
Refactor/replace plexus abstract log enabled with slf4j

timtebeek:
As seen on https://cwiki.apache.org/confluence/plugins/servlet/mobile?contentId=181305684#MavenEcosystemCleanup-DropPlexusContainerforJSR-330+SisuGuiceextension

-Leverages https://github.com/openrewrite/rewrite-apache/pull/41

This was the last one of this series; wasn't too sure about removal of `extends AbstractLogEnabled` from the deprecated classes; applied now to the classes that were not explicitly backwards compatibility tests, and retained in the two cases that were.

@elharo could you also look over this last one? That would mostly conclude the removal except where not yet possible.

### DIFF
--- a/compat/maven-compat/src/main/java/org/apache/maven/artifact/deployer/DefaultArtifactDeployer.java
+++ b/compat/maven-compat/src/main/java/org/apache/maven/artifact/deployer/DefaultArtifactDeployer.java
@@ -45,8 +45,6 @@ import org.eclipse.aether.deployment.DeploymentException;
 import org.eclipse.aether.metadata.MergeableMetadata;
 import org.eclipse.aether.repository.RemoteRepository;
 import org.eclipse.aether.util.artifact.SubArtifact;
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
 
 /**
  * DefaultArtifactDeployer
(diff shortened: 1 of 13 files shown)


elharo (approved):

timtebeek:
Thanks for the quick review! Looks to have been a one off failure in CI:
> Caused by: java.net.ConnectException: HTTP connect timed out

gnodet:
What's the benefits of refactoring deprecated classes ?

timtebeek:
Hmm; you're right. I only just now noticed _all_ of the classes here are in `compat`, as opposed to just the few I'd reverted when tests broke on the changed backwards compatibility. Guess that means there's little sense in changing this now; luckily the changes were mostly automated. Thanks for pointing it out!

I did already have my eye on completely removing `maven-compat` and was considering automation in that area next. Is there any set of steps on what would be required? Would this better fit a Slack or mailling list conversation?

gnodet:
> I did already have my eye on completely removing `maven-compat` and was considering automation in that area next. Is there any set of steps on what would be required? Would this better fit a Slack or mailling list conversation?

`maven-compat` is not really used anymore in Maven core project anymore, the problem rely more in other plugins.
See https://github.com/apache/maven/pull/1487

PR #1919, opened 2024-11-18
[MNG-8378] Upgrade to Resolver 2.0.4

cstamas:
Use latest Resolver with bugfixes and enhancements. Also, downgrades default JDK Transport to HTTP/1.1.
https://issues.apache.org/jira/secure/ReleaseNote.jspa?projectId=12320628&version=12355229

---

https://issues.apache.org/jira/browse/MNG-8378


jira-importer:
Resolve #9474

PR #1925, opened 2024-11-19
Enable GitHub issues

gnodet:


kwin:
I think this deserves some official vote on the mailing list and a JIRA issue outlining details:
Release Notes
Parallel usage of JIRA or migration?

CrazyHZM (approved):

olamy:
can we have this one read only first https://issues.apache.org/jira/projects/MNG/ 
and with some messages to redirect to correct location to create issues now.

gnodet:
@olamy 
Imho it's a bit premature to just make the jira project read only.  We can definitely add some warning and maybe disallow the creation of new issues (if possible).  But we still have issues there and I certainly don't want to duplicate issues.  I know Apache SSHD has gone through the same kind of migration some time ago.  New issues were created on GH, but existing ones have not been migrated.  We ended identifying with `[GH-xxx]` for the GitHub ones and `[SSHD-xxx]` for the JIRA issues.  For some time, the release notes contained both kind of issues, but over time, the migration is done painlessly imho.

olamy:
My only concern is to have new issues created in Jira.
Is there any way to at least prevent creating new issue in Jira?

gnodet:
> My only concern is to have new issues created in Jira. 

Yes, agreed.

> Is there any way to at least prevent creating new issue in Jira?

I think it should be doable from
https://issues.apache.org/jira/plugins/servlet/project-config/MNG/permissions
We need to investigate further and the profile seems shared between all maven projects for now.
Or maybe we switch all repos to GH issues at the same time.

olamy:
> > My only concern is to have new issues created in Jira.
> 
> Yes, agreed.
> 
> > Is there any way to at least prevent creating new issue in Jira?
> 
> I think it should be doable from https://issues.apache.org/jira/plugins/servlet/project-config/MNG/permissions We need to investigate further and the profile seems shared between all maven projects for now. Or maybe we switch all repos to GH issues at the same time.

Yes. all at the same time looks good to me.

maybe we can have some tools to copy one specific jira issue to GH? (when working on it)

CrazyHZM:
Enabling the GitHub issue gives Maven more user feedback, so it looks good.

slachiewicz (approved):

slawekjaranowski:
> We should also be prepared to move issues reported here to the correct place like plugins, shared lib

We can move issues between github projects - there is an option transfer issue ....

slawekjaranowski (approved):
I'm for enabled issues here and disable possibility to creating new issues in JIra

slawekjaranowski:
We also should first finish - https://github.com/apache/maven-site/pull/588 to have some guide how to works with GitHub issues in all our project in the same way.

slachiewicz (approved):

hboutemy:
do we all agree that even if this issue is marked approved by many, it's not yet ready to be merged?

.asf.yaml
```diff
@@ -14,6 +14,9 @@ github:
     rebase: true
   autolink_jira:
     - MNG
+  del_branch_on_merge: true
```
hboutemy, line 17: this line can be split and merge quickly on a separate PR

.asf.yaml
```diff
@@ -14,6 +14,9 @@ github:
...
+  features:
+    issues: true
```
hboutemy, line 19: this activation is not ready for merge to me: https://lists.apache.org/thread/56mhgp2kynylo5ctpdk5m1qczjzj6wln

hboutemy (changes requested):

.asf.yaml
bmarwell: Are we done now? @hboutemy @slachiewicz I think we should enable GH issues.

Bukama:
As this repository is one of the last ones, we need to update this PR (at least rebase and solve conflicts). I suggest that we bring the PR and repository in line with the other enabling ones

* Leave `del_branch_on_merge: true` in this PR
* Add ISSUE_TEMPLATE folder
* Add release_drafter config/workflow
* Update PR template to the one without `clean`
* Update pom.xml

Further work:
* Add add required files to 3.9 branch.
* Update maven-site with changed link(s)

Note: releases and milestones already exist

For details / links to examples see the confluence page: https://cwiki.apache.org/confluence/display/MAVEN/JIRA+to+GitHub+Issues+switching


@gnodet @slawekjaranowski @hboutemy

Bukama:
In consultation with @gnodet I close this PR and recreate new ones with everything left.

PR #1929, opened 2024-11-20
[MNG-8385] Introduce proto session

cstamas:
And make use of it in CLIng. Also, move from "late" Resolver session factory to "early" CLIng invocation of `PropertyContributor` SPI and make contribution visible across whole Maven, not only Resolver.

---

https://issues.apache.org/jira/browse/MNG-8385


its/core-it-suite/src/test/resources/mng-8385/spi-extension/src/main/resources/META-INF/maven/org.apache.maven.api.di.Inject
```diff
@@ -0,0 +1 @@
+org.apache.maven.its.mng8385.CustomPropertyContributor
```
gnodet, line 1: That one should be automatically generated if you compile with enabled annotation processors.
See https://github.com/apache/maven/blob/master/api/maven-api-di/src/main/java/org/apache/maven/di/tool/DiIndexProcessor.java

its/core-it-suite/src/test/resources/mng-8385/spi-extension/.mvn/placeholder.txt
gnodet, line 1: It should be unneeded, since you added `root="true"` on the pom.

api/maven-api-spi/src/main/java/org/apache/maven/api/spi/PropertyContributor.java
```diff
@@ -38,6 +40,19 @@ public interface PropertyContributor extends SpiService {
      * Invoked just before session is created with a mutable map that carries collected user properties so far.
      *
      * @param userProperties The mutable user properties, never {@code null}.
+     * @see #contribute(ProtoSession)
      */
-    void contribute(Map<String, String> userProperties);
+    default void contribute(Map<String, String> userProperties) {}
```
gnodet, line 45: Glad we keep compatibility somehow !

impl/maven-cli/src/main/java/org/apache/maven/cling/invoker/spi/PropertyContributorsHolder.java
```diff
@@ -0,0 +1,42 @@
...
+package org.apache.maven.cling.invoker.spi;
+
+import java.util.Map;
+
+import org.apache.maven.api.di.Inject;
+import org.apache.maven.api.di.Named;
+import org.apache.maven.api.spi.PropertyContributor;
+
+/**
+ * Hack: to circumvent sisu/plexus/di stuff.
```
gnodet, line 28: I really need to investigate what's the problem here.

gnodet (approved):
LGTM, a few minor enhancements are possible to remove two files from the IT.

cstamas:
Dropped two files, locally is ok, let see ITs and merge

jira-importer:
Resolve #9468

PR #1936, opened 2024-11-25
[MNG-8395] Add a <Source> element in the model.

desruisseaux:
JIRA issue: [MNG-8395](https://issues.apache.org/jira/browse/MNG-8395)

Add a `<Source>` element in the model. Properties are:

- `directory` (inherited from `FileSet`)
- `includes`  (inherited from `PatternSet`)
- `excludes`  (inherited from `PatternSet`)
- `scope`
- `lang`
- `module`
- `targetVersion`
- `targetPath` (taken from `<Resource>`)
- `filtering` (taken from `<Resource>`)
- `enabled`

This commit also renames `source` parameter value in `reader-stax.vm` for avoiding name collision with the new `Source` model element.

### DIFF
--- a/api/maven-api-model/src/main/mdo/maven.mdo
+++ b/api/maven-api-model/src/main/mdo/maven.mdo
@@ -775,6 +775,7 @@
           <type>String</type>
         </field>
         <field>
+          <!-- TODO: replaced by <Source> with "resource" scope. -->
           <name>resources</name>
           <version>3.0.0+</version>
           <description>
@@ -789,6 +790,7 @@
           </association>
         </field>
         <field>
+          <!-- TODO: replaced by <Source> with "testResource" scope. -->
           <name>testResources</name>
           <version>4.0.0+</version>
           <description>
@@ -857,6 +859,22 @@
       </description>
       <fields>
         <field>
+          <name>sources</name>
+          <version>4.1.0+</version>
+          <description>
+            All the sources to compile and resources files to copy for a project or it's unit tests.
+            The sources can be Java source files, generated source files, scripts or resources for examples.
+            Each source is specified by a mandatory {@code directory} element, which is relative to the POM.
+            The kind of sources (codes to compile or resources to copy) and their usage (for the main code
+            or for the tests) is specified by the {@code scope} element together with each source directory.
+          </description>
+          <association>
+            <type>Source</type>
+            <multiplicity>*</multiplicity>
+          </association>
+        </field>
+        <field>
+          <!-- TODO: replaced by <Source> with "main" scope. -->
           <name>sourceDirectory</name>
           <version>3.0.0+</version>
           <required>true</required>
@@ -869,6 +887,7 @@
           <type>String</type>
         </field>
         <field>
+          <!-- TODO: replaced by <Source> with "script" scope. -->
           <name>scriptSourceDirectory</name>
           <version>4.0.0+</version>
           <required>true</required>
@@ -882,6 +901,7 @@
           <type>String</type>
         </field>
         <field>
+          <!-- TODO: replaced by <Source> with "test" scope. -->
           <name>testSourceDirectory</name>
           <version>4.0.0+</version>
           <required>true</required>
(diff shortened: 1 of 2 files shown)


api/maven-api-model/src/main/mdo/maven.mdo
```diff
@@ -1945,6 +1965,178 @@
...
+          <description>
+            <![CDATA[
+            Name of the Java module (or other language-specific module) which is built by the sources.
+            This element can be specified in a Maven project containing multiple Java modules.
+            It is generally not needed for non-modular projects, or for modular projects having only one module.
+
+            <p>If a module name is specified for resources or script files,
+            then this value modifies the directory where the files will be copied.
+            For example, if a Java module name is "foo.biz", then the {@code foo/bar.properties}
+            resource file will be copied as {@code foo.biz/foo/bar.properties}.</p>
```
gnodet, line 2049: Is that a Maven convention, a Java convention, or something imposed by the JVM ?

api/maven-api-model/src/main/mdo/maven.mdo
```diff
@@ -1945,6 +1965,178 @@
...
+          <description>
+            <![CDATA[
+            The version of the platform where the code will be executed.
+            In a Java environment, this is the value of the {@code --release} compiler option.
+            If a Java project contains multiple main sources with different target versions,
+            then a multi-version <abbr>JAR</abbr> file will be created.
+            If this element is omitted, then the default target version is the compiler default value,
+            which is usually the version of the Java environment running Maven.
+
+            <p>If a target version is specified for resources or script files,
```
gnodet, line 2069: Should that be `if a target version, different from the default target, version is specified` ?

gnodet:
I like it a lot, this would solve a number of problems.
I think we'd need to deprecate the previous `resources`, `sourceDirectory`, `scriptSourceDirectory` and `testSourceDirectory` in the 4.x model.

api/maven-api-model/src/main/mdo/maven.mdo
desruisseaux: It is a Java compiler convention. In a multi-modules project, `javac` automatically inserts module names at the root of class directories. For example, if a module named `my.foo.module` contains a `foo.bar.biz.Bla` class, then in single-module project (as today), `javac` generates the following file in the Maven `target/classes/` directory:

* `foo/bar/biz/Bla.class`

But in a multi-modules project (would be optional and not the default), `javac` generates the same file at the following location instead:

* `my.foo.module/foo/bar/biz/Bla.class`

Other Java tools such as `java` understand this directory hierarchy. I.e. the `--module-path` can be a directory of modules, with no need to enumerate on the path each module found in the directory. I'm not sure if it is a requirement that the directory names match module names, but this is what `javac` does at least by default.

Above documentation of `module` element tries to match what `javac` does. It could be rephrased as _"During Java compilation, we let `javac` do whatever it wants. For resources, we try to copy them in the same directory as `javac`"_. The current wording tries to be more specific about what happens, but the goal is to match `javac` and the documentation would be updated if we find mismatches.

api/maven-api-model/src/main/mdo/maven.mdo
```diff
@@ -1945,6 +1965,178 @@
...
+        <field>
+          <name>filtering</name>
+          <version>4.1.0+</version>
+          <description>
+            <![CDATA[
+            Whether resources are filtered to replace tokens with parameterized values.
+            The values are taken from the {@code properties} element and from the properties
+            in the files listed in the {@code filters} element.
+
+            <p>This filtering should not be confused with the filtering of paths done by the
```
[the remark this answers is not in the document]
desruisseaux: Right. Maybe `stringFiltering` or `stringFilteringEnabled`?

 [JEP 465](https://openjdk.org/jeps/465) talked about "string template" or "string interpolation", but I think that what Maven is doing is simpler (e.g., I don't think that `${x + y}` computes the sum?)

api/maven-api-model/src/main/mdo/maven.mdo
```diff
@@ -1945,6 +1965,178 @@
...
+          </description>
+          <type>boolean</type>
+          <defaultValue>false</defaultValue>
+        </field>
+        <field>
+          <name>enabled</name>
+          <version>4.1.0+</version>
+          <description>
+            <![CDATA[
+            Whether the directory described by this source element should be included in the build.
```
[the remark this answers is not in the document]
desruisseaux: I guess so. I wasn't aware of that. How should we express that?

api/maven-api-model/src/main/mdo/maven.mdo
gnodet: Agreed. So specifying a target version for the `source` is not sufficient to create a multi-release jar. That's why I think the sentence:
> If a target version is specified for resources or script files, then this value modifies the directory where the files will be copied.

is not completely accurate.  This is only the case if the version is not the base one.

desruisseaux: Ah, understood. You are right. I will fix tomorrow.

cstamas (approved):

gnodet:
I wonder if we should include in next RC. @cstamas ?

cstamas:
> I wonder if we should include in next RC. @cstamas ? Though it would require a bit more work to fix the deprecation and migration stuff. We've done a similar thing for `module` -> `subproject`.

I'd say go for it. We have many other things to sort out anyway.

gnodet (approved):

jira-importer:
Resolve #9458

PR #1937, opened 2024-11-25
[MNG-8394] Event bridge and properties fix

cstamas:
JIRA issue: [MNG-8394](https://issues.apache.org/jira/browse/MNG-8394)

Changes:
* The event type conversion is wrong (mvn3-mvn4 event bridge).
* The user properties were pushed too late for logging configuration. Now we push user properties two times, immediately (only those keys that does not exists), and 2nd time after property contributor we "refresh" all we pushed.


jira-importer:
Resolve #9459

PR #1940, opened 2024-11-27
[MNG-8403] Maven ITs use maven-executor

cstamas:
The goal of this PR is manifold, but major one is to be able to use in ITs new options introduced in Maven4. Currently the "embedded" mode supports only Maven3 options, as Maven4 got new CLI entry point (CLIng), while verifier uses old MavenCli entry point, that is also deprecated. Finally, a full cleanup of (black) magic happened as well, keep ITs simple and clean.

Changes:
* dropped from ITs classpath maven-shared-util
* dropped from ITs classpath maven-verifier, copied last master Verifier to maven-it-helper and modified
* enhancements to new maven-executor to make it fully replace maven-verifier
* ITs are now using new infra and are using new CLIng "entry point" as well (so far ITs used deprecated maven-embedder/MavenCLI class).

---

https://issues.apache.org/jira/browse/MNG-8403


its/core-it-suite/pom.xml
```diff
@@ -526,7 +515,9 @@ under the License.
           <forkCount>0</forkCount>
           <reuseForks>true</reuseForks>
           <skip>true</skip>
+          <promoteUserPropertiesToSystemProperties>false</promoteUserPropertiesToSystemProperties>
           <systemPropertyVariables>
+            <project.build.directory>${project.build.directory}</project.build.directory>
```
cstamas, line 520: not needed, this was meant as place for IT own repo

cstamas:
@gnodet no, let's do it other way: remove the "shared" verifier?

gnodet (approved):

jira-importer:
Resolve #9451

PR #1949, opened 2024-12-05
[MNG-8403] Collapse IT utils and helpers

cstamas:
Changes:
* move all into same package
* stop direct use of ResourceExtractor, redirect all ITs to superclass method
* drop unneeded ResourceExtractor, it never "extracts" anything
* un-fork ITs where applicable (ie. due new CLI param or so that was not supported before)

---

https://issues.apache.org/jira/browse/MNG-8403


gnodet (approved):

jira-importer:
Resolve #9451

PR #1957, opened 2024-12-08
[MNG-8410] API cleanup

gnodet:
* Rename maven-api-meta to maven-api-annotations

The `org.apache.maven.api.plugin` package can't be moved to `maven-api-plugin` because that would cause a circularity between m-api-core and m-api-plugin.

I'm still on the edge wrt to `maven-api-core` which defines `org.apache.maven.api` and `org.apache.maven.api.services`...   Should we rename those to `maven-api-session` and `org.apache.maven.api.session` ? Or let then the way they are...

So maybe just renaming `maven-api-meta` is enough ?


cstamas (approved):

jira-importer:
Resolve #9444

PR #1965, opened 2024-12-11
[MNG-5729] Use monotonic time measurements

gnodet:
This PR introduces a `MonotonicClock` in the Maven API.
It provides accurate duration computation, and relative date/time support, leveraging`System.nanoTime()`.
The clock is UTC-based and provided `Instant` need to be zoned if needed for correct display.

---

https://issues.apache.org/jira/browse/MNG-5729


cstamas (approved):

cstamas:
Is this for rc2 or 4.1?

gnodet:
> Is this for rc2 or 4.1?

I'm on the edge.  The updated PR is more lightweight and does not define a new `Temporal` implementation, but simply reuses the well known `Instant` instead, but provides a `Clock` implementation which uses the monotonic system clock.  This makes the code a lot simpler and safer.

cstamas:
This PR looks quite non-invasive, so I'd just merge it, make it rc2

cstamas (approved):

jira-importer:
Resolve #6563

PR #1968, opened 2024-12-12
[MNG-8406] Proper IT isolation

cstamas:
Implement proper IT isolation, in a way, that user running ITs should not have it's own env "mixed in" into IT runs. In essence, use "alternate" `user.home` for ITs. Also use proper means to set user home. This will also stop fork 4 ITs (as they were forked due presence of env variable).

Note: yes, this causes longer IT execution time (as can be seen), as the ITs user home (core-it-suite/target/user-home) starts as empty (tail is set to "outer" local repo), so ITs do download stuff.

---

https://issues.apache.org/jira/browse/MNG-8406


jira-importer:
Resolve #9448

PR #1973, opened 2024-12-12
[MNG-8244] Using before:all / all / after:all is not triggered

gnodet:
- **[MNG-5729] Fix possible NPE with introduction of mononic clock**
- **[MNG-8244] Using before:all / all / after:all is not triggered**


jira-importer:
Resolve #9621

PR #1979, opened 2024-12-13
[MNG-3282] Docgen: remove property numbering, they are misleading and properties are "floating" anyway (alphabetically)

gnodet:
Requires #1975 (Upgrade to resolver 2.0.5)


jira-importer:
Resolve #4656

PR #1981, opened 2024-12-13
[MNG-8414] The consumer pom should warn if not able to downgrade the model version to 4.0.0

gnodet:
JIRA issue: [MNG-8414](https://issues.apache.org/jira/browse/MNG-8414)

Warn if the consumer POM cannot be downgraded to 4.0.0.
The warning can be removed by adding `preserve.model.version='true'` to the `<project>` element.


jira-importer:
Resolve #9440

PR #1982, opened 2024-12-13
[MNG-8437] mvnsh

cstamas:
The beginning

https://asciinema.org/a/KxFoz2NUyI0joa4kM4ZRRCJgD

Changes:
* (unrelated) contains fix for property handling in embedded executor and in lookup invoker
* pulled `-o` (offline) option to "generic" Options from MavenOptions
* introduce mvnsh (scripts, options, CLIng main class)
* simplified invokers (only one context needed for maven now)
* invokers made "reentrant", it all depends HOW context is created

Related PRs:
* mvnd changes https://github.com/apache/maven-mvnd/pull/1228

---

https://issues.apache.org/jira/browse/MNG-8437


gnodet (approved):
Awesome ! Really nice work !

impl/maven-cli/src/main/java/org/apache/maven/cling/invoker/LookupInvoker.java
```diff
@@ -553,13 +579,19 @@ protected void settings(C context, boolean emitSettingsWarnings, SettingsBuilder
                 .build();
 
         customizeSettingsRequest(context, settingsRequest);
+        if (context.eventSpyDispatcher != null) {
```
alextu, line 582: Our regression tests fail because the `SettingsBuilderRequest` event no longer fires after this PR has been merged. Would that be because `eventSpyDispatcher` is now somehow null at this point?

cstamas: This method is called twice, once (very early) when loading up extensions, then spy is null. The 2nd invocation is when "maven boots" and then spy should not be null...

alextu: Here's a reproducer https://github.com/alextu/maven4-reproducer

PR #1986, opened 2024-12-16
Maven 4 Compatibility Testing

gnodet:
A starting point for testing Maven 4 compatibility with a set of projects.
This currently scans all ASF projects for the ones that have a pom.xml ...
We may want to add some additional inclusions / exclusions...

### DIFF
--- a/.github/workflows/test-projects.yml
+++ b/.github/workflows/test-projects.yml
@@ -1,0 +1,147 @@
+name: Maven 4 Compatibility Testing
+
+on:
+  workflow_dispatch:
+    inputs:
+      maven_version:
+        description: 'Maven 4 version to test (e.g., 4.0.0-rc-1)'
+        required: true
+        default: '4.0.0-rc-1'
+      maven_url:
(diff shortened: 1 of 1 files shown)


slawekjaranowski:
We can also enable Maven4 by default in our builds - https://github.com/apache/maven-gh-actions-shared?tab=readme-ov-file#testing-with-maven-4

gnodet:
> We can also enable Maven4 by default in our builds - https://github.com/apache/maven-gh-actions-shared?tab=readme-ov-file#testing-with-maven-4

I'm currently mostly interested in finding issues in Maven 4 rather than fixing our own builds, so I don't want to focus on Maven projects only.
I've created https://github.com/gnodet/maven4-testing whith in-progress testing results at  https://github.com/gnodet/maven4-testing/issues/133

cstamas:
We may want to add tool based on your findings, like:
* up enforcer (if under some version) to the version that works with mvn4 (not to "latest")
* same for exec-m-p
* and so on...

All input on this are welcome!

PR #1987, opened 2024-12-16
[MNG-8438] maven-jline: Migrate to Maven DI (off javax.inject)

cstamas:
Move to Maven DI

---

https://issues.apache.org/jira/browse/MNG-8438

### DIFF
--- a/apache-maven/pom.xml
+++ b/apache-maven/pom.xml
@@ -110,6 +110,16 @@ under the License.
       <artifactId>jline-terminal-ffm</artifactId>
     </dependency>
 
+    <!-- (legacy) DI annotations -->
+    <dependency>
+      <groupId>javax.inject</groupId>
+      <artifactId>javax.inject</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>javax.annotation</groupId>
+      <artifactId>javax.annotation-api</artifactId>
+    </dependency>
+
     <!-- DI Runtime -->
     <dependency>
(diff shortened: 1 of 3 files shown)


slachiewicz:
How all this would work - we have many `external` components that in theory should work outside of Maven world, based on javax.inject. 
If we say that's only theory - should components migrate to Maven DI?

cstamas:
Maven4 "core" (the new one) is  already Maven DI, while "old core" (impl/maven-core) is still sisu DI. Still, there is a "bridge" that makes Maven DI components visible to Sisu, hence "everything work" even if "new core" is already Maven DI.

For outside of Maven (this repo), Maven4 stuff can already migrate to Maven DI, while stuff requiring Maven 3 AND 4 support should remain on Sisu as that is the common denominator of two.

gnodet:
> Maven4 "core" (the new one, impl/maven-impl) is already Maven DI, while "old core" (impl/maven-core) is still sisu DI. Still, there is a "bridge" that makes Maven DI components visible to Sisu, hence "everything work" even if "new core" is already Maven DI.
> 
> For outside of Maven (this repo), Maven4 stuff can already migrate to Maven DI, while stuff requiring Maven 3 AND 4 support should remain on Sisu as that is the common denominator of two.

Right, so in short, when migrating components to the new Maven 4 API, they should be migrated to the `org.apache.maven.api.di` package too.

jira-importer:
Resolve #9416

PR #1994, opened 2024-12-17
[MNG-8447] Lossy ProblemCollector

cstamas:
Historically (from Maven3) "problem collection" (in various scenarios, like building effective settings and toolchains, to building models) were done by passing around `List<Problem>` structure. This proved quite ineffective, as in case of (really) huge projects with quite big count of problems reported choked Maven (but also produced unusable output). We are aware of projects producing 3 million warnings! Dumping all of them onto console/log makes really no sense.

This PR changes Maven that by default it reports "problems" (settings building, toolchains building, or model building) as one liner warnings: how much and where were problems collected. This produces much more less overwhelming output than happens on master. User can control "build errors" using `-e`, so `mvn -e` will dump errors on screen/log.

The new `org.apache.maven.api.services.ProblemCollector<P>` class, that is somewhat "drop in" replacement for `List`, but with huge difference: it is "lossy", in a way, it maintains counters precisely, but actual problem instances are simply dropped after threshold `maven.builder.maxProblems` is surpassed (default 100). See `org.apache.maven.api.Constants#MAVEN_BUILDER_MAX_PROBLEMS`.

Rules:
* on problem addition counters are always updated
* if below threshold, store the problem (and will be reported later to the user)
* if above threshold, try to drop one problem with severity lower than currently reported problem, if succeeded, store the current problem, otherwise drop it

---

https://issues.apache.org/jira/browse/MNG-8447


impl/maven-core/src/main/java/org/apache/maven/project/DefaultProjectBuilder.java
```diff
@@ -498,18 +499,14 @@ private List<ProjectBuildingResult> build(File pomFile, boolean recursive) {
...
             }
 
             List<ProjectBuildingResult> results = new ArrayList<>();
             List<ModelBuilderResult> allModels = results(result).toList();
             for (ModelBuilderResult r : allModels) {
-                List<ModelProblem> problems = new ArrayList<>(r.getProblems());
-                results(r)
-                        .filter(c -> c != r)
-                        .flatMap(c -> c.getProblems().stream())
-                        .forEach(problems::remove);
```
cstamas, line 512: Sus right here.

cstamas: No need for this anymore

gnodet (approved):

jira-importer:
Resolve #9407

PR #2000, opened 2024-12-19
[MNG-8451] Core inject depMgt for plugin resolution

cstamas:
Maven Core injects "own provided" artifacts when resolving plugins.

---

https://issues.apache.org/jira/browse/MNG-8451

### DIFF
--- a/impl/maven-core/src/main/java/org/apache/maven/plugin/internal/DefaultPluginDependenciesResolver.java
+++ b/impl/maven-core/src/main/java/org/apache/maven/plugin/internal/DefaultPluginDependenciesResolver.java
@@ -22,14 +22,23 @@ import javax.inject.Inject;
 import javax.inject.Named;
 import javax.inject.Singleton;
 
+import java.io.IOException;
+import java.io.InputStream;
+import java.util.ArrayList;
+import java.util.Collections;
 import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.Properties;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ConcurrentMap;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
 import org.apache.maven.RepositoryUtils;
 import org.apache.maven.api.DependencyScope;
+import org.apache.maven.api.annotations.Nullable;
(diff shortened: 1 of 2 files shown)


cstamas:
This works cool, but breaks https://issues.apache.org/jira/browse/MNG-5783
We need to implement the proposed fix in refd issue.

impl/maven-core/src/main/java/org/apache/maven/plugin/internal/DefaultPluginDependenciesResolver.java
```diff
@@ -208,12 +221,29 @@ private DependencyResult resolveInternal(
...
+                    // align version if needed
+                    if (!Objects.equals(
+                            pluginDep.getArtifact().getVersion(),
+                            managedDep.getArtifact().getVersion())) {
+                        pluginDep = pluginDep.setArtifact(pluginDep
+                                .getArtifact()
+                                .setVersion(managedDep.getArtifact().getVersion()));
+                    }
+                    // align scope
+                    pluginDep = pluginDep.setScope(managedDep.getScope());
```
gnodet, line 245: Should we flag them as `system` scope ?

gnodet: ... to avoid downloading the jars ?

cstamas: Yes, but there is more to this PR, it breaks https://issues.apache.org/jira/browse/MNG-5783 We need to implement the proposed fix in refd issue (real classpath vs `${plugin.artifacts}`).

cstamas: Former contains plugin stuff + maven core (imported), while latter should contain all

gnodet:
> This works cool, but breaks https://issues.apache.org/jira/browse/MNG-5783 We need to implement the proposed fix in refd issue.

The fix was implemented in https://github.com/apache/maven/commit/b7088a34edd8cd34ab3b376d9baf90bdb6f8dc00 AFAIK.  This may be something slightly different.

PR #2004, opened 2024-12-20
[MNG-8461] Initial settings method must restore context state

cstamas:
Effective settings are (should be) created twice, once for "early boot" of Plexus when extensions are loaded up, and then again when Maven "boots".

Bug was that early call "corrupted" (inited settings) in context causing that 2nd required call (due spy) was omitted. This resulted in lack of settings related spy events firing (as we do have IT for spy but it does not test settings events).

---

https://issues.apache.org/jira/browse/MNG-8461

Related: https://github.com/alextu/maven4-reproducer/issues/1

### DIFF
--- a/impl/maven-cli/src/main/java/org/apache/maven/cling/invoker/LookupInvoker.java
+++ b/impl/maven-cli/src/main/java/org/apache/maven/cling/invoker/LookupInvoker.java
@@ -501,8 +501,12 @@ public abstract class LookupInvoker<C extends LookupContext> implements Invoker 
      * If there are Maven3 passwords presents in settings, this results in doubled warnings emitted. So Plexus DI
      * creation call keeps "emitSettingsWarnings" false. If there are fatal issues, it will anyway "die" at that
      * spot before warnings would be emitted.
+     * <p>
+     * The method returns a "cleaner" runnable, as during extension loading the context needs to be "cleaned", restored
+     * to previous state (as it was before extension loading).
      */
-    protected void settings(C context, boolean emitSettingsWarnings, SettingsBuilder settingsBuilder) throws Exception {
+    protected Runnable settings(C context, boolean emitSettingsWarnings, SettingsBuilder settingsBuilder)
+            throws Exception {
         Options mavenOptions = context.invokerRequest.options();
 
         Path userSettingsFile = null;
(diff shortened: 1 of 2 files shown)


cstamas:
I think we can... but am now looking at MavenITmng4936EventSpyTest and it seems it does test for this thing but did not fail?

cstamas:
So, we need to dig in, why reproducer fails, and why IT succeeds. Locally running the related IT I have this (mng-4936/target/spy.log):

fixed: Apache Maven 4.0.0-rc-3-SNAPSHOT (a9f6f3b71d21542869f635c197dc8076583937f1)
```
init
event: org.apache.maven.api.services.SettingsBuilderRequest$SettingsBuilderRequestBuilder$DefaultSettingsBuilderRequest
event: org.apache.maven.internal.impl.DefaultSettingsBuilder$DefaultSettingsBuilderResult
event: org.apache.maven.api.services.ToolchainsBuilderRequest$ToolchainsBuilderRequestBuilder$DefaultToolchainsBuilderRequest
event: org.apache.maven.internal.impl.DefaultToolchainsBuilder$DefaultToolchainsBuilderResult
event: org.apache.maven.execution.DefaultMavenExecutionRequest
event: org.apache.maven.lifecycle.internal.DefaultExecutionEvent
event: org.apache.maven.lifecycle.internal.DefaultExecutionEvent
event: org.apache.maven.lifecycle.internal.DefaultExecutionEvent
event: org.apache.maven.lifecycle.internal.DefaultExecutionEvent
event: org.apache.maven.lifecycle.internal.DefaultExecutionEvent
event: org.apache.maven.execution.DefaultMavenExecutionResult
close
```

master: Apache Maven 4.0.0-rc-3-SNAPSHOT (575ad3719c79b061c78d531ec9d29793f159c2ec)
```
init
event: org.apache.maven.api.services.SettingsBuilderRequest$SettingsBuilderRequestBuilder$DefaultSettingsBuilderRequest
event: org.apache.maven.internal.impl.DefaultSettingsBuilder$DefaultSettingsBuilderResult
event: org.apache.maven.api.services.ToolchainsBuilderRequest$ToolchainsBuilderRequestBuilder$DefaultToolchainsBuilderRequest
event: org.apache.maven.internal.impl.DefaultToolchainsBuilder$DefaultToolchainsBuilderResult
event: org.apache.maven.execution.DefaultMavenExecutionRequest
event: org.apache.maven.lifecycle.internal.DefaultExecutionEvent
event: org.apache.maven.lifecycle.internal.DefaultExecutionEvent
event: org.apache.maven.lifecycle.internal.DefaultExecutionEvent
event: org.apache.maven.lifecycle.internal.DefaultExecutionEvent
event: org.apache.maven.lifecycle.internal.DefaultExecutionEvent
event: org.apache.maven.execution.DefaultMavenExecutionResult
close
```

Basically they are same.

cstamas:
Still cannot figure, and I starting to doubt in our ITs... The only difference I found so far:
* reproducer uses `.mvn/extensions.xml` to load up spy listener
* IT uses `-Dmaven.ext.class.path=spy-0.1.jar` so it is much earlier already present in system

Still, as the code change shows, the event should not be emitted, but it seems it was? Checked also, the IT is forked, and forked executor is used, so it cannot be some "leftover"...

jira-importer:
Resolve #9394

PR #2006, opened 2024-12-21
[MNG-8460] Implement flush()

cstamas:
LoggingOutputStream used by default as sysout and syserr were NOT flushing (def method is empty), and flush happened only when EOL was printed.

Still, prompts and help:evaluate does print out unterminated strings.

---

https://issues.apache.org/jira/browse/MNG-8460


gnodet (approved):

jira-importer:
Resolve #9393

PR #2009, opened 2024-12-23
[MNG-8461] Integrate reproducer as IT

cstamas:
Integrate the reproducer into ITs.

---

https://issues.apache.org/jira/browse/MNG-8461


jira-importer:
Resolve #9394

PR #2010, opened 2024-12-23
[MNG-8473] Site fixes

cstamas:
Maven4 site cleanup: make all new modules generating models also document them, expose them all and redirect all links to new documentation.

Lifecycle is the only thing missing, as it is generated on the fly, there is no more source to "point at"...

---

https://issues.apache.org/jira/browse/MNG-8473


gnodet (commented):
Why not pointing at:
* https://github.com/apache/maven/blob/751d3f18c97663c5f435ef615a24dd639a0f2611/impl/maven-core/src/main/java/org/apache/maven/internal/impl/DefaultLifecycleRegistry.java#L383-L412
* https://github.com/apache/maven/blob/751d3f18c97663c5f435ef615a24dd639a0f2611/impl/maven-core/src/main/java/org/apache/maven/internal/impl/DefaultLifecycleRegistry.java#L472-L476
* https://github.com/apache/maven/blob/751d3f18c97663c5f435ef615a24dd639a0f2611/impl/maven-core/src/main/java/org/apache/maven/internal/impl/DefaultLifecycleRegistry.java#L361-L365

They are not much readable for non developers, but maybe our target would know enough ?


The changes to the modello generation leads to unneeded processing if not building the site. That was done on purpose to simplify the config ?

cstamas:
re "unneeded processing": exactly, I don't care if XSD is being built but unused, what I do care about is simple POM.
re "pointing": i agree, why not. Before the class was literally "listing" phases...

gnodet (approved):

PR #2011, opened 2024-12-23
[MNG-8469] Fix interpolation precedence order

cstamas:
Restore precedence as it was in Maven3:
- basedir
- buildTime
- prefixed model
- user props
- model props
- system props
- env
- prefixless model

---

https://issues.apache.org/jira/browse/MNG-8469


cstamas:
Fixes all https://github.com/gnodet/maven4-testing/issues?q=is%3Aissue+is%3Aopen+%22recursive+variable+reference%22

jira-importer:
Resolve #9385

PR #2014, opened 2025-01-01
[MNG-8475] In the loop scenario, StringBuilder is used instead of concatenation

CrazyHZM:
Under loops, string concatenation using StringBuilder performs better.


michael-o (approved):

gnodet (approved):

jira-importer:
Resolve #9956

PR #2021, opened 2025-01-04
[MNG-8487] Completely isolate UTs

cstamas:
UTs were using real user home, and in case user had user-wide extensions.xml, it resulted in failure. Goal: make sure all UTs are properly protected and isolated from user env, by providing "fake" user home (to not have user-wide extensions in real user home affect test outcome).

---

https://issues.apache.org/jira/browse/MNG-8487


elharo (approved):

slawekjaranowski (approved):

jira-importer:
Resolve #9947

PR #2022, opened 2025-01-04
Delete unused mode constants

elharo:
We can always add these later if we need them, but right now their existence implies things about the code that aren't true.

### DIFF
--- a/api/maven-api-xml/src/main/java/org/apache/maven/api/xml/XmlNode.java
+++ b/api/maven-api-xml/src/main/java/org/apache/maven/api/xml/XmlNode.java
@@ -28,7 +28,7 @@ import org.apache.maven.api.annotations.Nullable;
 import org.apache.maven.api.annotations.ThreadSafe;
 
 /**
- * An immutable xml node.
+ * An immutable XML node.
  *
  * @since 4.0.0
  */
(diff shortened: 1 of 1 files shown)


elharo:
some unrelated flakage in CI:

Error:  	Failed to read artifact descriptor for org.sonatype.sisu:sisu-guice:jar:no_aop:3.1.0: Failed to collect dependencies at org.apache.maven.plugins:maven-dependency-plugin:jar:3.3.0 -> org.apache.maven.shared:maven-common-artifact-filters:jar:3.2.0 -> org.eclipse.sisu:org.eclipse.sisu.plexus:jar:0.0.0.M5 -> org.sonatype.sisu:sisu-guice:jar:no_aop:3.1.0: The following artifacts could not be resolved: org.sonatype.sisu:sisu-guice:pom:3.1.0 (absent): Could not transfer artifact org.sonatype.sisu:sisu-guice:pom:3.1.0 from/to central (https://repo.maven.apache.org/maven2): HTTP connect timed out
Error:  -> [Help 1]

gnodet (changes requested):
It's the default mode, so obviously not usually explicitly defined, though I could find a couple of code that does use it:
https://github.com/search?q=combine.self%3D%22merge%22+&type=code

Removing the default value would not allow specifying the default behaviour anymore, which may break some use cases (not 100% sure).

Also, this removes the only place where those values are documented and leave the code with no documentation at all.

So not sure what we are the benefits here.

gnodet:
> > It's the default mode, so obviously not usually explicitly defined, though I could find a couple of code that does use it: https://github.com/search?q=combine.self%3D%22merge%22+&type=code
> > Removing the default value would not allow specifying the default behaviour anymore, which may break some use cases (not 100% sure).
> 
> This doesn't remove the mode. It removes the unused constant. 

Right.

> If the constant is unused, should it be used somewhere?

Maybe the merge method could check if the attributes have known values (i.e. `override` or `merge`) and log something if that's not the case ?

elharo:
Or throw an exception?

gnodet:
> Or throw an exception?

Throwing an exception would indeed be easier.  We'd need to assess that not too many POMs using invalid values...

gnodet:
Superseded by #2038

PR #2024, opened 2025-01-04
[MNGSITE-393] Remove references to Maven 2.x

elharo:


api/maven-api-settings/src/main/mdo/settings.mdo
```diff
@@ -542,9 +542,9 @@
           <type>String</type>
           <description>
             <![CDATA[
-            A repository id or (since Maven 2.0.9) an expression matching one or many repository ids to mirror, e.g.,
+            A repository id or an expression matching one or many repository ids to mirror, e.g.,
             <code>central</code> or <code>*,!repo1</code>.
-            <code>*</code> (since Maven 2.0.5), <code>external:*</code> (since Maven 2.0.9) and <code>external:http:*</code> (since Maven 3.8.0) have
+            <code>*</code>, <code>external:*</code> and <code>external:http:*</code> (since Maven 3.8.0) have
             a special meaning: see <a href="/guides/mini/guide-mirror-settings.html">Mirror Settings</a> guide.
```
gnodet, line 548: Would it make sense to move the documentation in the code rather than keeping it external ?
It's easier to maintain as using a single PR allows updating both the code and the doc.

gnodet (commented):
I've seen a few other "Maven 2" references in the main model.
Would it make sense to change them as well ?
https://github.com/apache/maven/blob/master/api/maven-api-model/src/main/mdo/maven.mdo

elharo:
Yes to remove other references, but that's for another PR. This bug has already seen dozens of PRs and may require dozens more before it;'s done.

api/maven-api-settings/src/main/mdo/settings.mdo
elharo: I'm not sure what you're proposing here.

gnodet: My suggestion was to inline the content of https://maven.apache.org/guides/mini/guide-mirror-settings.html into the model, so that the generated javadoc for the [Mirror class](https://maven.apache.org/ref/4.0.0-rc-2/api/apidocs/org/apache/maven/api/settings/Mirror.html) will contain the documentation.

elharo: I wouldn't prefer that. guide-mirror-settings.html is written for people who simply use Maven for their own projects and might never look at the API docs.

gnodet: Agreed.  My point was to move the documentation inside the code base along the source code.
It could be put in https://github.com/apache/maven/tree/master/src/site rather than in an unversioned area from a different git repository.

elharo: Maybe then. I'm not sure how maven site is published and what parts come from where.

gnodet: Since we remove references to old version, what's the point of keeping the ref for Maven 3.8.0 ?

gnodet (approved):

PR #2025, opened 2025-01-05
[MNG-8493] 'serialVersionUID' can be annotated with '@Serial' annotation

Stellar1999:


elharo (commented):
Honestly this feels like noise to me. I don't think it's important enough to justify the screen real estate it takes

impl/maven-logging/src/main/java/org/apache/maven/slf4j/MavenBaseLogger.java
```diff
@@ -146,6 +147,7 @@
  */
 public class MavenBaseLogger extends LegacyAbstractLogger {
 
+    @Serial
```
gnodet, line 150: Why does `org.slf4j.helpers.AbstractLogger` implements `Serializable` in the first place ? 
What's the use case for serialising those ?

Stellar1999: Maybe only the SLF4J team knows the exact use case. My guess is that it might involve transmitting logger objects over a network. However, if AbstractClass implements Serializable, I think we should also define a serialVersionUID.

elharo: Honestly there is none. It's just something developers put in 20 years ago because maybe it would help someone somewhere. Java object serialization isn't much used these days, and Brian Goetz uses it as a canonical example of bad software design and development:

https://www.youtube.com/watch?v=n6K_8s3Sx4s
https://www.youtube.com/watch?v=dOgfWXw9VrI&t=1957s

What can I say? It was 1997 and none of us really knew what we were doing yet, Sun included. 

Interestingly, some of the initialization issues that make Java serialization dangerous are the same reasons we should move away from field injection with Plexus to constructor injection with Guice.

gnodet: I think we can safely remove the `serialVersionUID` from `MavenBaseLogger`

Stellar1999: I think so.I have done it.

gnodet (approved):

jira-importer:
Resolve #9944

PR #2026, opened 2025-01-05
Fix comparisons against null in PathSource.equals

elharo:


api/maven-api-core/src/test/java/org/apache/maven/api/services/PathSourceTest.java
```diff
@@ -0,0 +1,34 @@
...
+ */
+package org.apache.maven.api.services;
+
+import java.nio.file.Paths;
+
+import org.junit.jupiter.api.Test;
+
+import static org.junit.jupiter.api.Assertions.assertFalse;
+
+public class PathSourceTest {
```
gnodet, line 27: Test classes should not be public.

api/maven-api-core/src/test/java/org/apache/maven/api/services/PathSourceTest.java
```diff
@@ -0,0 +1,34 @@
...
+
+    @Test
+    public void testEquals() {
```
gnodet, line 30: Test methods don't need to be public either.

gnodet (changes requested):
The best practices lead to using package private classes / methods.

api/maven-api-core/src/test/java/org/apache/maven/api/services/PathSourceTest.java
elharo: Interesting. They used to have to be. Did that change in JUnit 5?

api/maven-api-core/src/test/java/org/apache/maven/api/services/PathSourceTest.java
elharo: done

gnodet (approved):

PR #2029, opened 2025-01-06
[MNG-8490] Combine XmlNode and XmlNodeImpl.

Stellar1999:


Stellar1999:
We need to synchronize the implementation in plexus-xml, so we will keep it as a draft for now.

gnodet (changes requested):
I'm -1 on this.
The role of an API is to decouple the contract from the implementation.
All interfaces in Maven have almost a single implementation, except a couple of interfaces.
The Single Interface Single Implementation is only an anti-pattern in a closed world, but the API is designed to be consumed by all plugins, extensions, etc...

gnodet:
And this implementation class changes quite often with bug fixes, so I would not put it in the API.

Stellar1999:
I’ve almost completed most of the coding work and linked it to a plexus-xml PR. However, this PR seems to have a broader impact than expected. Could you please help evaluate it?

Stellar1999:
> And this implementation class changes quite often with bug fixes, so I would not put it in the API.

I’ve placed most of the implementation in the Util class to avoid impacting the API layer. Could you please take another look?

Stellar1999:
There are still some comment-related tasks remaining, and I will complete them later.

api/maven-api-xml/src/main/java/org/apache/maven/api/xml/XmlNode.java
```diff
@@ -18,121 +18,244 @@
...
 
-    @Nonnull
-    Map<String, String> getAttributes();
+    // ----------------------------------------------------------------------
+    // Attribute handling
+    // ----------------------------------------------------------------------
 
-    @Nullable
-    String getAttribute(@Nonnull String name);
+    public Map<String, String> getAttributes() {
```
elharo, line 151: needs javadoc, particularly about immutability

api/maven-api-xml/src/main/java/org/apache/maven/api/xml/XmlNode.java
```diff
@@ -18,121 +18,244 @@
...
+    public String toStringXml() throws XMLStreamException {
+        StringWriter writer = new StringWriter();
+        XmlNodeWriter.write(writer, this);
+        return writer.toString();
+    }
+
+    public String toStringObject() {
+        StringBuilder sb = new StringBuilder();
+        sb.append("XmlNode[");
+        boolean w = false;
```
elharo, line 238: rename variable

api/maven-api-xml/src/main/java/org/apache/maven/api/xml/XmlNode.java
```diff
@@ -18,121 +18,244 @@
...
+        w = addToStringField(sb, prefix, o -> !o.isEmpty(), "prefix", w);
+        w = addToStringField(sb, namespaceUri, o -> !o.isEmpty(), "namespaceUri", w);
```
elharo, line 240: why is this variable reassigned?

api/maven-api-xml/src/main/java/org/apache/maven/api/xml/XmlNode.java
```diff
@@ -18,121 +18,244 @@
...
+        w = addToStringField(sb, name, o -> !o.isEmpty(), "name", w);
+        w = addToStringField(sb, value, o -> !o.isEmpty(), "value", w);
+        w = addToStringField(sb, attributes, o -> !o.isEmpty(), "attributes", w);
+        w = addToStringField(sb, children, o -> !o.isEmpty(), "children", w);
+        w = addToStringField(sb, location, Objects::nonNull, "location", w);
+        sb.append("]");
+        return sb.toString();
+    }
+
+    private static <T> boolean addToStringField(StringBuilder sb, T o, Function<T, Boolean> p, String n, boolean w) {
```
elharo, line 250: rename w variable and other variables; in general avoid single letter names

apache-maven/pom.xml
```diff
@@ -125,6 +125,10 @@ under the License.
       <groupId>org.eclipse.sisu</groupId>
       <artifactId>org.eclipse.sisu.plexus</artifactId>
     </dependency>
+    <dependency>
+      <groupId>org.codehaus.plexus</groupId>
+      <artifactId>plexus-xml</artifactId>
```
gnodet, line 130: We tried hard to get rid of that dependency, don't add it back.

api/maven-api-xml/pom.xml
```diff
@@ -35,6 +35,14 @@
       <groupId>org.apache.maven</groupId>
       <artifactId>maven-api-annotations</artifactId>
     </dependency>
+    <dependency>
+      <groupId>com.fasterxml.woodstox</groupId>
```
gnodet, line 39: No external dependencies in the API.

api/maven-api-xml/src/main/java/org/apache/maven/api/xml/XmlNode.java
```diff
@@ -18,121 +18,244 @@
...
- * An immutable xml node.
- *
- * @since 4.0.0
+ *  NOTE: remove all the util code in here when separated, this class should be pure data.
  */
-@Experimental
-@ThreadSafe
-@Immutable
-public interface XmlNode {
+public class XmlNode implements Serializable {
```
gnodet, line 34: Use an interface and a builder API, similar to requests / models in other parts of the API.

api/maven-api-xml/src/main/java/org/apache/maven/api/xml/XmlNode.java
```diff
@@ -18,121 +18,244 @@
...
+        return children;
+    }
 
-    default XmlNode merge(@Nullable XmlNode source) {
-        return merge(source, (Boolean) null);
+    public int getChildCount() {
+        return children.size();
     }
 
-    XmlNode merge(@Nullable XmlNode source, @Nullable Boolean childMergeOverride);
```
gnodet, line 111: This removes one important piece from the whole API.  It suddenly becomes completely unavailable.  A service needs to be defined.  However, services are currently only available from the `Session`, but it cannot be used here.
A service locator could be used, but that's really overkill.  No better idea at this point though (and don't put the `merge` code in the API, this really needs to be decoupled).

gnodet (changes requested):
A few points need to be addressed:
* no third party dependency in the API
* no dependency on plexus-xml, this would re-create a cyclic dependency
* use an interface + builder api, similar to other data classes in the API
* the merge code needs to be abstracted and decoupled using a service interface so that it can be used from plugins (they are only provided with the API jars)

For the merge service, maybe using a factory using `ServiceLoader` if we want to make that generally available.  However, if we only target plugins, a service accessible from the `Session` could actually be used.
But we do have plugins that use `Xpp3Dom.mergeXpp3Dom` so that feature cannot be removed from the API

Stellar1999:
> [MNG-8490] Combine XmlNode and XmlNodeImpl. codehaus-plexus/plexus-xml#54



> A few points need to be addressed:
> 
> * no third party dependency in the API
> * no dependency on plexus-xml, this would re-create a cyclic dependency
> * use an interface + builder api, similar to other data classes in the API
> * the merge code needs to be abstracted and decoupled using a service interface so that it can be used from plugins (they are only provided with the API jars)
> 
> For the merge service, maybe using a factory using `ServiceLoader` if we want to make that generally available. However, if we only target plugins, a service accessible from the `Session` could actually be used. But we do have plugins that use `Xpp3Dom.mergeXpp3Dom` so that feature cannot be removed from the API

First of all, thank you for your code review. After reflecting on your comments and considering how to implement them, I suddenly realized that if I use the ServiceLoader approach to abstract the merge functionality and expose it, it doesn't seem significantly different from not combining the implementations of XmlNode and XmlNodeImpl. Would it be worth considering not combining them?

api/maven-api-xml/src/main/java/org/apache/maven/api/xml/XmlNode.java
```diff
@@ -18,121 +18,244 @@
...
     /**
      * This default mode for combining a DOM node during merge means that where element names match, the process will
      * try to merge the element attributes and values, rather than overriding the recessive element completely with the
      * dominant one. This means that wherever the dominant element doesn't provide the value or a particular attribute,
      * that value or attribute will be set from the recessive DOM node.
      */
-    String DEFAULT_SELF_COMBINATION_MODE = SELF_COMBINATION_MERGE;
+    public static final String DEFAULT_SELF_COMBINATION_MODE = SELF_COMBINATION_MERGE;
+
+    protected final String prefix;
```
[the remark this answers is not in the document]
gnodet: This isn't necessary if using an interface + builder and non public class.

api/maven-api-xml/src/main/java/org/apache/maven/api/xml/XmlNode.java
elharo: This is exactly what I don't what to do. That just makes the API more complex than it needs to be.

api/maven-api-xml/src/main/java/org/apache/maven/api/xml/XmlNode.java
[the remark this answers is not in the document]
elharo: Yes it is. They'd just be in a different class.

api/maven-api-xml/src/main/java/org/apache/maven/api/xml/XmlNode.java
```diff
@@ -18,121 +18,244 @@
...
+        this.value = value;
+        this.attributes = ImmutableCollections.copy(attributes);
+        this.children = ImmutableCollections.copy(children);
+        this.location = location;
+    }
+    // ----------------------------------------------------------------------
+    // Name handling
+    // ----------------------------------------------------------------------
+
+    public String getPrefix() {
```
elharo, line 127: Javadoc on all public methods, and thi9s should be more than simply "Returns prefix". It should say what a prefix is.

impl/maven-xml/src/main/java/org/apache/maven/internal/xml/XmlNodeUtil.java
```diff
@@ -18,160 +18,28 @@
...
+import static org.apache.maven.api.xml.XmlNode.SELF_COMBINATION_REMOVE;
 
-    /**
-     * @since 3.2.0
-     * @return input location
-     */
-    public Object getInputLocation() {
-        return location;
-    }
+public class XmlNodeUtil {
```
elharo, line 42: I prefer the Guava convention of using plurals for utility classes; e.g. instead of XmlNodeUtil consider XmlNodes

PR #2030, opened 2025-01-07
[EXPERIMENT] Make CI use split repo

cstamas:
And then reuse Central cache across jobs, while the rest (locally installed) are throwaway.

Changes:
* rework CI to use MAVEN_ARGS to make all Maven invocations use split repo, CI caches now only central
* EmbeddedExecutor is now equivalent to forked: it also observes MAVEN_ARGS env variable (forked was always, as it is `mvn` script invoked).
* Helper test made "resilient" to split repo: it now asserts "ends with" as split repo may add prefixes to local repository paths.
* Verifier methods made "repository aware" (null = installed, non-null = remote repositoryId)
* Tests doing artifact LRM ops now need to be aware of their origin

Result of these changes should be drastically lower pulls from Central, ideally zero in fact (only when a dependency changes).


cstamas:
Giving up this work, we still end up w/ dirty central caches as:
* split repo is only about "repoId"
* but many ITs redefine "central" as local file repo -> those would end up polluting cache
* some ITs redefine "centra" with bringing up Jetty -> same thing as above, but goes over resolver

PR #2031, opened 2025-01-07
[MNG-8494] Restore Maven 3 compatibility

gnodet:


compat/maven-model-builder/src/main/java/org/apache/maven/model/building/ModelCache.java
```diff
@@ -18,23 +18,37 @@
...
  * @deprecated use {@link org.apache.maven.api.services.ModelBuilder} instead
  */
 @Deprecated(since = "4.0.0")
 public interface ModelCache {
 
-    <T> T computeIfAbsent(String groupId, String artifactId, String version, String tag, Supplier<T> data);
+    /**
+     * Puts the specified data into the cache.
+     *
+     * @param groupId The group id of the cache record, must not be {@code null}.
```
elharo, line 36: Oracle javadoc guidelines are no initial capitalization on Javadoc comments and period only after or before a complete sentence

compat/maven-model-builder/src/main/java/org/apache/maven/model/building/ModelCacheTag.java
```diff
@@ -46,7 +46,25 @@ interface ModelCacheTag<T> {
...
+     * cache is populated but the state of the cache must not change so we need to make a copy.
+     *
+     * @param data The data to store in the cache, must not be {@code null}.
+     * @return The data being stored in the cache, never {@code null}.
+     */
+    T intoCache(T data);
+
+    /**
+     * Creates a copy of the data suitable for retrieval from the cache. The retrieved data can be mutated after the
+     * cache is queried but the state of the cache must not change so we need to make a copy.
```
elharo, line 59: queried,

gnodet:
99% of the code here is copy/pasted from maven-3.9.x branch to restore compatibility... so I'm fine with any change you suggest, but I really don't expect you to closely review all the changed files.

compat/maven-embedder/src/main/java/org/apache/maven/cli/transfer/AbstractMavenTransferListener.java
```diff
@@ -33,31 +34,177 @@
...
+                return formattedProgressedSize + "/" + formattedSize;
+            } else {
+                return format(progressedSize);
+            }
+        }
     }
 
-    protected AbstractMavenTransferListener(MessageBuilderFactory messageBuilderFactory, PrintWriter out) {
-        this.messageBuilderFactory = messageBuilderFactory;
+    protected PrintStream out;
```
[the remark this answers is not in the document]
gnodet: I fully agree, however the code has already been cleaned up in the non deprecated version:
https://github.com/apache/maven/blob/698614e3cc71c38f2979e5e165172aeb911bc8be/impl/maven-cli/src/main/java/org/apache/maven/cling/transfer/AbstractMavenTransferListener.java#L36
That class is purely for compatibility, so no `protected` stuff should be changed.

The maven-embedder is here for compatibility and not used when launching Maven from CLI directly.

cstamas (approved):

cstamas:
IMHO last commit should be undone: 
* Maven2 "way" was to use SecDispatcher directlly
* since Maven3 the "official way" was provided in Maven3 API (and SecDispatcher is not even published, see https://issues.apache.org/jira/browse/MNG-4384 Sadly this _workaround_ got picked up despite there is official Maven 3 API for that
* IMHO in Maven4 we _should not_ support Maven2 "way", only Maven 3 API

PR #2032, opened 2025-01-08
[MNG-8502] Embedded executor should obey MAVEN_ARGS env variable

cstamas:
Salvaged from the "split repo for ITs" experiment. The experiment failed but this improvement is good to have, as makes the embedded executor behave more closely like forked executor.

Embedded so far did not obey MAVEN_ARGS env variable while forked did.

---

https://issues.apache.org/jira/browse/MNG-8502


gnodet (approved):

jira-importer:
Resolve #9936

PR #2018, opened 2025-01-03
[MNG-8482] Use instanceof assignments to get rid of casting expressions

breun:
JIRA issue: [MNG-8482](https://issues.apache.org/jira/browse/MNG-8482)

Following this checklist to help us incorporate your
contribution quickly and easily:

 - [x] Make sure there is a [JIRA issue](https://issues.apache.org/jira/browse/MNG) filed
       for the change (usually before you start working on it).  Trivial changes like typos do not
       require a JIRA issue. Your pull request should address just this issue, without
       pulling in other changes.
 - [x] Each commit in the pull request should have a meaningful subject line and body.
 - [x] Format the pull request title like `[MNG-XXX] SUMMARY`,
       where you replace `MNG-XXX` and `SUMMARY` with the appropriate JIRA issue.
 - [x] Also format the first line of the commit message like `[MNG-XXX] SUMMARY`.
       Best practice is to use the JIRA issue title in both the pull request title and in the first line of the commit message.
 - [x] Write a pull request description that is detailed enough to understand what the pull request does, how, and why.
 - [x] Run `mvn clean verify` to make sure basic checks pass. A more thorough check will
       be performed on your pull request automatically.
 - [x] You have run the [Core IT][core-its] successfully.

If your pull request is about ~20 lines of code you don't need to sign an
[Individual Contributor License Agreement](https://www.apache.org/licenses/icla.pdf) if you are unsure
please ask on the developers list.

To make clear that you license your contribution under
the [Apache License Version 2.0, January 2004](http://www.apache.org/licenses/LICENSE-2.0)
you have to acknowledge this by using the following check-box.

 - [x] I hereby declare this contribution to be licenced under the [Apache License Version 2.0, January 2004](http://www.apache.org/licenses/LICENSE-2.0)

 - [ ] In any other case, please file an [Apache Individual Contributor License Agreement](https://www.apache.org/licenses/icla.pdf).

[core-its]: https://maven.apache.org/core-its/core-it-suite/

PR #2033, opened 2025-01-08
Just an IT run for PR 2018

cstamas:
Fixes https://github.com/apache/maven/pull/2018

Changes to pick:
* 94cb54a45372c5a92d2ade42c7c1ad1fca05b128
* 0c58851fec68211679766bd965b4dab451ea4992


cstamas:
@breun @gnodet applied some fixes (and there was also a conflict) to Nils PR, let's see IT. IF OK, Nils you can pick changes from here.

cstamas:
The ITs are NOT part of current reactor (the "maven build") and they are still building projects mostly with Java=8, so those classes and resources should not be modified.

PR #2037, opened 2025-01-10
[MNG-8484] Replace BaseRequest#unmodifiable with List.copyOf()

gnodet:
`List.copyOf()` is exactly what we need to make a defensive copy of the input collection.  It also has the benefit of not doing anything if the input collection is already an immutable list.

### DIFF
--- a/api/maven-api-core/src/main/java/org/apache/maven/api/services/BaseRequest.java
+++ b/api/maven-api-core/src/main/java/org/apache/maven/api/services/BaseRequest.java
@@ -18,10 +18,6 @@
  */
 package org.apache.maven.api.services;
 
-import java.util.ArrayList;
-import java.util.Collection;
-import java.util.Collections;
-
 import org.apache.maven.api.ProtoSession;
 import org.apache.maven.api.annotations.Experimental;
 import org.apache.maven.api.annotations.Nonnull;
@@ -45,11 +41,5 @@ abstract class BaseRequest<S extends ProtoSession> {
     @Nonnull
     public S getSession() {
         return session;
-    }
-
-    protected static <T> Collection<T> unmodifiable(Collection<T> obj) {
-        return obj != null && !obj.isEmpty()
-                ? Collections.unmodifiableCollection(new ArrayList<>(obj))
-                : Collections.emptyList();
     }
 }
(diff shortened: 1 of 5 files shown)


api/maven-api-core/src/main/java/org/apache/maven/api/services/BaseRequest.java
```diff
@@ -52,9 +51,16 @@ public static <T> T nonNull(T obj, String message) {
...
-                ? Collections.unmodifiableCollection(new ArrayList<>(obj))
-                : Collections.emptyList();
+    /**
+     * Creates a defensive immutable copy of a collection or returns an empty immutable list if the input is null or empty.
+     * This method is useful when creating immutable objects to ensure collection fields cannot be modified by external code.
+     *
+     * @param source the source collection to create an immutable copy from
+     * @param <T> the type of elements in the collection
+     * @return an unmodifiable view of the collection, or an empty list if the input is null or empty
+     * @throws NullPointerException if the collection contains null elements
```
elharo, line 61: Why do we do this? In general, lists with null elements are OK, though they might be wrong in a specific case. However if null elements are wrong in the specific case I'd expect the source list to already have checked that.

gnodet: In general, non null elements in list are a burden. Unless there's a specific semantic associated with a null value, I'd rather forbid those.  For example if you want to resolve a list of artifacts, passing null values makes absolutely no sense to me.

api/maven-api-core/src/main/java/org/apache/maven/api/services/BaseRequest.java
```diff
@@ -52,9 +51,16 @@ public static <T> T nonNull(T obj, String message) {
...
+     */
+    protected static <T> List<T> copyToImmutable(Collection<T> source) {
+        return source != null && !source.isEmpty() ? List.copyOf(source) : List.of();
     }
```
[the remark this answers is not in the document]
gnodet: You're right.  But then, I'll go further and just remove the method and use `List.copyOf(xxx)` in the 5 places where this method is actually used.

cstamas (approved):

jira-importer:
Resolve #9380

PR #2040, opened 2025-01-10
[MNG-8510] Remove Utils class from maven-di

gnodet:


impl/maven-di/src/main/java/org/apache/maven/di/impl/ReflectionUtils.java
```diff
@@ -382,4 +367,49 @@ public static <T> Binding<T> bindingFromConstructor(Key<T> key, Constructor<T> c
 
         return binding.withKey(key);
     }
+
+    public static void getDisplayString(StringBuilder sb, Object object) {
```
elharo, line 371: These two methods can now be private.

jira-importer:
Resolve #9926

PR #2042, opened 2025-01-12
[MNG-8512] Fix org.slf4j.simpleLogger.defaultLogLevel Configuration conflict

CrazyHZM:
JIRA issue: [MNG-8512](https://issues.apache.org/jira/browse/MNG-8512)

### DIFF
--- a/impl/maven-cli/src/main/java/org/apache/maven/cling/logging/impl/MavenSimpleConfiguration.java
+++ b/impl/maven-cli/src/main/java/org/apache/maven/cling/logging/impl/MavenSimpleConfiguration.java
@@ -20,7 +20,9 @@ package org.apache.maven.cling.logging.impl;
 
 import org.apache.maven.cling.logging.BaseSlf4jConfiguration;
 import org.apache.maven.slf4j.MavenLoggerFactory;
+import org.apache.maven.slf4j.MavenSimpleLogger;
 import org.slf4j.ILoggerFactory;
+import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 /**
@@ -29,6 +31,8 @@ import org.slf4j.LoggerFactory;
  * @since 3.1.0
  */
 public class MavenSimpleConfiguration extends BaseSlf4jConfiguration {
+    private static final Logger LOGGER = LoggerFactory.getLogger(MavenSimpleConfiguration.class);
+
     @Override
     public void setRootLoggerLevel(Level level) {
         String value =
@@ -37,7 +41,14 @@ public class MavenSimpleConfiguration extends BaseSlf4jConfiguration {
                     case INFO -> "info";
                     default -> "error";
                 };
-        System.setProperty("org.slf4j.simpleLogger.defaultLogLevel", value);
+
+        String current = System.setProperty(MavenSimpleLogger.DEFAULT_LOG_LEVEL_KEY, value);
+        if (current != null && !value.equalsIgnoreCase(current)) {
+            LOGGER.info(
+                    "System property '" + MavenSimpleLogger.DEFAULT_LOG_LEVEL_KEY + "' is already set to '" + current
+                            + "' - ignoring system property and get log level from -X/-e/-q options, log level will be set to"
+                            + value);
+        }
     }
 
     @Override
--- a/impl/maven-logging/src/main/java/org/apache/maven/slf4j/MavenLoggerFactory.java
+++ b/impl/maven-logging/src/main/java/org/apache/maven/slf4j/MavenLoggerFactory.java
@@ -21,13 +21,14 @@ package org.apache.maven.slf4j;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ConcurrentMap;
 
+import org.apache.maven.logging.api.LogLevelRecorder;
 import org.slf4j.ILoggerFactory;
 import org.slf4j.Logger;
 
 /**
  * LogFactory for Maven which can create a simple logger or one which, if set, fails the build on a severity threshold.
  */
-public class MavenLoggerFactory implements org.apache.maven.logging.api.LogLevelRecorder, ILoggerFactory {
+public class MavenLoggerFactory implements LogLevelRecorder, ILoggerFactory {
     final DefaultLogLevelRecorder logLevelRecorder = new DefaultLogLevelRecorder();
     final ConcurrentMap<String, Logger> loggerMap = new ConcurrentHashMap<>();
 
--- a/impl/maven-logging/src/main/java/org/apache/maven/slf4j/MavenSimpleLogger.java
+++ b/impl/maven-logging/src/main/java/org/apache/maven/slf4j/MavenSimpleLogger.java
@@ -41,6 +41,8 @@ public class MavenSimpleLogger extends MavenBaseLogger {
 
     static Consumer<String> logSink;
 
+    public static final String DEFAULT_LOG_LEVEL_KEY = "org.slf4j.simpleLogger.defaultLogLevel";
+
     public static void setLogSink(Consumer<String> logSink) {
         MavenSimpleLogger.logSink = logSink;
     }


impl/maven-cli/src/main/java/org/apache/maven/cling/logging/impl/MavenSimpleConfiguration.java
```diff
@@ -37,7 +40,16 @@ public void setRootLoggerLevel(Level level) {
                     case INFO -> "info";
                     default -> "error";
                 };
-        System.setProperty("org.slf4j.simpleLogger.defaultLogLevel", value);
+
+        String defaultLogLevelKey = "org.slf4j.simpleLogger.defaultLogLevel";
+        String current = System.getProperty(defaultLogLevelKey);
+        if (current == null) {
+            System.setProperty(defaultLogLevelKey, value);
+        } else {
```
elharo, line 48: what if the levels are the same? Then there shouldn't be a warning.

elharo (commented):
I know it's tricky with system properties, but anyway this can be tested? Maybe with an IT?

Also, rerun the failing CI suite. It's known flaky.

impl/maven-cli/src/main/java/org/apache/maven/cling/logging/impl/MavenSimpleConfiguration.java
```diff
@@ -37,7 +40,16 @@ public void setRootLoggerLevel(Level level) {
...
+            LOGGER.warn(
+                    "org.slf4j.simpleLogger.defaultLogLevel already set to: {},ignore -X/-e/-q command line options",
```
gnodet, line 50: Maybe rephrase as:
```
LOGGER.warn(
        "System property '{}' is already set to '{}' - ignoring log level from -X/-e/-q options",
        defaultLogLevelKey,
        current);
```

impl/maven-cli/src/main/java/org/apache/maven/cling/logging/impl/MavenSimpleConfiguration.java
CrazyHZM: When the value is the same, It doesn't cause `-Dorg slf4j.SimpleLogger.DefaultLogLevel` configuration affects the maven command line configuration failure.
@elharo

impl/maven-cli/src/main/java/org/apache/maven/cling/logging/impl/MavenSimpleConfiguration.java
CrazyHZM: It does look better.

impl/maven-cli/src/main/java/org/apache/maven/cling/logging/impl/MavenSimpleConfiguration.java
elharo: I'm not sure I follow. Can you rephrase?

impl/maven-cli/src/main/java/org/apache/maven/cling/logging/impl/MavenSimpleConfiguration.java
```diff
@@ -37,7 +40,17 @@ public void setRootLoggerLevel(Level level) {
...
                 };
-        System.setProperty("org.slf4j.simpleLogger.defaultLogLevel", value);
+
+        String defaultLogLevelKey = "org.slf4j.simpleLogger.defaultLogLevel";
+        String current = System.getProperty(defaultLogLevelKey);
+        if (current == null) {
+            System.setProperty(defaultLogLevelKey, value);
+        } else {
+            LOGGER.warn(
+                    "System property '{}' is already set to '{}' - ignoring log level from -X/-e/-q options",
```
elharo, line 50: should we specify what the log level that's being ignored is? that is, add a third parameter?

Also, it's really easy to mix parameters up. Simple string concatenation is preferable.

impl/maven-cli/src/main/java/org/apache/maven/cling/logging/impl/MavenSimpleConfiguration.java
CrazyHZM: For example, we run the command  `mvn clean install -Dorg slf4j.SimpleLogger.DefaultLogLevel -X`, it will lead to effective `-X`, and `-Dorgslf4j.SimpleLogger.DefaultLogLevel` failure (I think I said against the above)

elharo: You're probably correct, but I do not understand what you are saying.

gnodet: The `-X` option will ultimately call `MavenSimpleConfiguration#setRootLoggerLevel` which will set the `org.slf4j.simpleLogger.defaultLogLevel` system property to the value, see https://github.com/apache/maven/blob/99777ace0226034f82299898e598a7f977d525ef/impl/maven-cli/src/main/java/org/apache/maven/cling/logging/impl/MavenSimpleConfiguration.java#L40

impl/maven-cli/src/main/java/org/apache/maven/cling/logging/impl/MavenSimpleConfiguration.java
```diff
@@ -37,7 +40,17 @@ public void setRootLoggerLevel(Level level) {
                     case INFO -> "info";
                     default -> "error";
                 };
-        System.setProperty("org.slf4j.simpleLogger.defaultLogLevel", value);
+
+        String defaultLogLevelKey = "org.slf4j.simpleLogger.defaultLogLevel";
```
gnodet, line 44: This should be `MavenSimpleLogger.DEFAULT_LOG_LEVEL_KEY`

impl/maven-cli/src/main/java/org/apache/maven/cling/logging/impl/MavenSimpleConfiguration.java
gnodet:
```
➜  maven git:(feat/MNG-8512) ✗ ~/work/git/maven/apache-maven/target/apache-maven-4.0.0-rc-3-SNAPSHOT/bin/mvn -X -Dorg.slf4j.simpleLogger.defaultLogLevel=debug

[WARNING] System property 'org.slf4j.simpleLogger.defaultLogLevel' is already set to 'debug' - ignoring log level from -X/-e/-q options
Apache Maven 4.0.0-rc-3-SNAPSHOT (cc9dc285f04d254e786b9e727ebea1d368bca400)
...
```

I think what @elharo means is that we don't want to print a warning if we're using `-X` and the system property is already set to the correct value.

gnodet: Maybe we should compare and use the _lowest_ level ? And simply ignore if the two values are the same?

gnodet: Actually, I'm not sure we want user properties to take precedence over CLI options.
We use the same behavior for color for example:
https://github.com/apache/maven/blob/bebc3d4a2e987e59aa0666f71cf3b8688dde6f7a/impl/maven-cli/src/main/java/org/apache/maven/cling/invoker/LookupInvoker.java#L209-L212

impl/maven-cli/src/main/java/org/apache/maven/cling/logging/impl/MavenSimpleConfiguration.java
elharo: Use mvn spotless:apply and it should fix anything IntelliJ does wrong. And you can always edit in vi or whatever plain text editor you prefer.

impl/maven-cli/src/main/java/org/apache/maven/cling/logging/impl/MavenSimpleConfiguration.java
CrazyHZM: We still need to tell the user that the two configurations conflict, but changing to the INFO level makes more sense. What do you think?
cc @gnodet @elharo

gnodet: Yes, you're right.  That could be done in a single call as `System.setProperty` returns the previous value.  So we could check and if non null and different from the new value, log something at info level.

impl/maven-cli/src/main/java/org/apache/maven/cling/logging/impl/MavenSimpleConfiguration.java
```diff
@@ -37,7 +41,14 @@ public void setRootLoggerLevel(Level level) {
                     case INFO -> "info";
                     default -> "error";
                 };
-        System.setProperty("org.slf4j.simpleLogger.defaultLogLevel", value);
+
+        String current = System.getProperty(MavenSimpleLogger.DEFAULT_LOG_LEVEL_KEY);
```
gnodet, line 45:
```suggestion
        String current = System.setProperty(MavenSimpleLogger.DEFAULT_LOG_LEVEL_KEY, value);
        if (current != null && !Objects.equals(value, current)) {
            LOGGER.info("System property '" + MavenSimpleLogger.DEFAULT_LOG_LEVEL_KEY + "' is already set to '"
                    + current + "' - ignoring log level from -X/-e/-q options");
        }
```

gnodet, line 45:
```suggestion
        String current = System.setProperty(MavenSimpleLogger.DEFAULT_LOG_LEVEL_KEY, value);
        if (current != null && !Objects.equals(value, current)) {
            LOGGER.info("System property '" + MavenSimpleLogger.DEFAULT_LOG_LEVEL_KEY + "' is already set to '"
                    + current + "' - ignoring log level from -X/-e/-q options");
        }
```

gnodet, line 45:
```suggestion
        String current = System.setProperty(MavenSimpleLogger.DEFAULT_LOG_LEVEL_KEY, value);
        if (current != null && !Objects.equals(value, current)) {
```

CrazyHZM: Does this logically seem to be maven command options taking precedence over slf4j config?
@gnodet

gnodet: Yes, a CLI option seems more explicit to me that a value from system properties.  CLI options are usually set by the user for a given invocation, while system properties can come from various sources (also from CLI, that's right).
I think this would be in line with the example I pasted a few days ago:
```
        String styleColor = mavenOptions
                .color()
                .orElse(userProperties.getOrDefault(
                        Constants.MAVEN_STYLE_COLOR_PROPERTY, userProperties.getOrDefault("style.color", "auto")));
```
where the color option comes first from the CLI options, then user properties (in that case, we have the new, then the legacy property).

CrazyHZM: The log msg also needs to be adjusted.

gnodet (approved):

gnodet:
@CrazyHZM feel free to squash/merge this PR.

jira-importer:
Resolve #9377

PR #2043, opened 2025-01-13
[MNG-8522] Deprecate maven-builder-support artifact and its classes

gnodet:
JIRA issue: [MNG-8522](https://issues.apache.org/jira/browse/MNG-8522)

The `maven-builder-support` contains classes shared between `maven-model-builder`, `maven-settings-builder` and `maven-toolchain-builder`, all of those subprojects being deprecated as they are part of the Maven 3 API.
This PR thus also deprecates this subproject and its classes.

### DIFF
--- a/compat/maven-builder-support/pom.xml
+++ b/compat/maven-builder-support/pom.xml
@@ -35,6 +35,11 @@ under the License.
 
   <dependencies>
     <dependency>
+      <groupId>org.apache.maven</groupId>
+      <artifactId>maven-api-core</artifactId>
+      <scope>provided</scope>
+    </dependency>
+    <dependency>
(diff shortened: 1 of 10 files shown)


cstamas (approved):

CrazyHZM (approved):

slachiewicz:
It would be best if we could backport these types of changes to 3.x line - so it will be visible in the code, without the need to change plugin/shared code to 4.x line.

cstamas:
We cannot backport this deprecation, as in maven3 there is no replacement to offer. In maven4.0 we deprecated this, and once Maven3 backward compat is gone, this code is gone as well.

PR #2045, opened 2025-01-13
Simplify

gnodet:


gnodet:
Related to https://github.com/apache/maven/pull/2041
I don't think those calls are actually needed.

PR #2048, opened 2025-01-15
[MNG-8503] Configure logging using maven.logger.* properties rather than org.slf4j.simpleLogger.*

gnodet:
JIRA issue: [MNG-8503](https://issues.apache.org/jira/browse/MNG-8503)


gnodet:
Should the constants be moved to https://github.com/apache/maven/blob/master/api/maven-api-core/src/main/java/org/apache/maven/api/Constants.java and annotated with `@Config` so that they appear in https://maven.apache.org/ref/4-LATEST/maven-configuration.html ?

cstamas:
I'd agree w/ config stuff, make them documented.

cstamas (approved):

PR #2049, opened 2025-01-15
[MNG-8524] DefaultInterpolator should be used by injection

CrazyHZM:
JIRA Issue: [[MNG-8524]](https://issues.apache.org/jira/browse/MNG-8524)


impl/maven-cli/src/main/java/org/apache/maven/cling/extensions/ExtensionConfigurationModule.java
```diff
@@ -36,7 +37,9 @@ public class ExtensionConfigurationModule implements Module {
 
     private final CoreExtensionEntry extension;
     private final UnaryOperator<String> callback;
-    private final DefaultInterpolator interpolator = new DefaultInterpolator();
+
+    @Inject
+    private DefaultInterpolator interpolator;
```
cstamas, line 42: This is not a component. Just modify the ctor and make it final, and pass it in where instance is created

gnodet:
I'm not sure we want to change the maven 3 code base too much.  I've tried to revert most changes to get back into a compatible state, so if we add a constructor, it might break something.  And for not much value, as this code is not really used.

cstamas:
Agreed, the compat/maven-embedder changes should be undone.

CrazyHZM:
@gnodet @cstamas I also agree with this and have reverted the compatibility logic of maven3.

gnodet:
There are two references to `new DefaultInterpolator()` that can be easily removed:
* https://github.com/apache/maven/blob/f5a407cf79494d90ad1df4aa6336239c542d98e1/impl/maven-impl/src/main/java/org/apache/maven/internal/impl/DefaultSettingsBuilder.java#L83-L85
* https://github.com/apache/maven/blob/f5a407cf79494d90ad1df4aa6336239c542d98e1/impl/maven-impl/src/main/java/org/apache/maven/internal/impl/DefaultToolchainsBuilder.java#L60-L62

We can simply inline those constructors, as they are only used in a few tests.  This will move the references to the implementation to the tests instead of the main code.

CrazyHZM:
> There are two references to `new DefaultInterpolator()` that can be easily removed:
> 
> * https://github.com/apache/maven/blob/f5a407cf79494d90ad1df4aa6336239c542d98e1/impl/maven-impl/src/main/java/org/apache/maven/internal/impl/DefaultSettingsBuilder.java#L83-L85
> * https://github.com/apache/maven/blob/f5a407cf79494d90ad1df4aa6336239c542d98e1/impl/maven-impl/src/main/java/org/apache/maven/internal/impl/DefaultToolchainsBuilder.java#L60-L62
> 
> We can simply inline those constructors, as they are only used in a few tests. This will move the references to the implementation to the tests instead of the main code.

@gnodet Done.

impl/maven-cli/src/main/java/org/apache/maven/cling/extensions/ExtensionConfigurationModule.java
```diff
@@ -36,6 +36,7 @@ public class ExtensionConfigurationModule implements Module {
 
     private final CoreExtensionEntry extension;
     private final UnaryOperator<String> callback;
+
     private final DefaultInterpolator interpolator = new DefaultInterpolator();
```
cstamas, line 40: What about this one?

gnodet: This component is loaded very early without the full container afaik.  This would require an `Interpolator` to be bound to the `ProtoSession` and `ProtoLookup`.  Currently the only component added is `ClassWorld`.

gnodet (approved):

PR #2056, opened 2025-01-22
[MNG-8536] Use JDK methods to copy Map/List

cstamas:
Drop the ImmutableCollections and use JDK
Map/List copyOf() method instead.

---

https://issues.apache.org/jira/browse/MNG-8536


cstamas:
Seems the dropped `ImmutableCollections.copy(Map)` method uses own `MapN` implementation that did preserve copied map ordering, hence, `Map.copyOf` is not a simple drop-in replacement...

gnodet:
This now causes OutOfMemory errors.

jira-importer:
Resolve #9908

PR #2057, opened 2025-01-22
[MNG-8535] Ability to capture stdout with maven-executor

cstamas:
WIP

Needs https://github.com/jline/jline3/issues/1160

---

https://issues.apache.org/jira/browse/MNG-8535


impl/maven-cli/src/main/java/org/apache/maven/cling/ClingSupport.java
```diff
@@ -60,8 +64,26 @@ private ClingSupport(ClassWorld classWorld, boolean classWorldManaged) {
...
+
+    /**
+     * The main entry point with IO management.
+     */
+    public int run(String[] args, @Nullable InputStream in, @Nullable OutputStream out, @Nullable OutputStream err)
+            throws IOException {
         try (Invoker invoker = createInvoker()) {
-            return invoker.invoke(parseArguments(args));
+            ParserRequest.Builder parserRequestBuilder = parserRequestBuilder(args);
+            if (in != null) {
```
gnodet, line 77: Aren't those checks useless ?
It's definitely no big deal, but
```
ParserRequest parserRequestBuilder = parserRequestBuilder(args)
        .in(in).out(out).err(err)
        .build();
```

impl/maven-executor/src/main/java/org/apache/maven/cling/executor/embedded/EmbeddedMavenExecutor.java
```diff
@@ -296,11 +306,18 @@ protected Context doCreate(Path mavenHome, ExecutorRequest executorRequest) {
                         try {
                             if (r.stdoutConsumer().isPresent()
                                     || r.stderrConsumer().isPresent()) {
+                                System.setProperty("org.jline.terminal.output", "out");
```
gnodet, line 309: We don't have JLine in the classpath, right ? Else `TerminalBuilder.PROP_OUTPUT` would be more appropriate.

impl/maven-jline/src/main/java/org/apache/maven/jline/MessageUtils.java
```diff
@@ -49,8 +49,11 @@ public static void systemInstall() {
     public static void systemInstall(Consumer<TerminalBuilder> builderConsumer, Consumer<Terminal> terminalConsumer) {
         MessageUtils.terminal = new FastTerminal(
                 () -> {
-                    TerminalBuilder builder =
-                            TerminalBuilder.builder().name("Maven").dumb(true);
+                    TerminalBuilder builder = TerminalBuilder.builder()
+                            .jansi(false)
+                            .jna(false)
+                            .jni(true)
+                            .name("Maven");
```
gnodet, line 56: We may still want to have `dumb(true)`.
For example if we are on an unsupported os/architecture combination, we'll end up with a dumb terminal, but with an annoying warning (and nothing the user can do)...

cstamas:
This PR is superseded by several others. Dropping this one.

PR #2058, opened 2025-01-22
[MNG-8527] Re-enable consumer POM, remove unwanted managed dependencies, developers, contributors, mailing lists and issue management

gnodet:
JIRA issue: [MNG-8527](https://issues.apache.org/jira/browse/MNG-8527)

This PR aims at:
* enabling again the consumer POM on Maven 4 for all models, not only 4.1.0 models
* reducing the size of consumer POMs

This last item was really problematic and one of the reason why it was disabled for 4.0.0.
When a project imports huge BOMs (and sometimes cascading), you could end up with tons of managed dependencies that were inlined in the consumer BOM (it has no parent and no imports).
This PR fixes the problem by making sure only the managed dependencies that can actually have an effect are kept in the consumer POM.

It also removes informational data:
* developers
* contributors
* mailing lists
* issue management

I think those informations are not important on the consumer POM and are better related to build POMs.


cstamas (approved):

gnodet:
I need to add an IT which will test dependency management pruning.

PR #2061, opened 2025-01-26
Redirect the <sourceDirectory> element to <Source>

desruisseaux:
Notable changes:

* Add a `SourceRoot` interface for the properties declared in the POM's `<Source>` element.
* In the POM, deprecate the `sourceDirectory`, `testSourceDirectory`, `scriptSourceDirectory`, `resources`, `testResources` elements and the `Resource` structure. They are replaced by `<source>` elements.
* In the `MavenProject` class, replace the `compileSourceRoots`, `testCompileSourceRoots` and `scriptSourceRoots` fields by redirections to the new `sources` field with paths wrapped in `SourceRoot` objects. The resource fields are not completely replaced, information are duplicated.
* The getters and setters methods for above-cited fields are deprecated.
* This commit contains other opportunistic deprecations on methods which were receiving or returning instances of types that were already deprecated.

In the POM file, the above-cited deprecated properties are ignored if a corresponding `<source>` element exists. For example, if there is a source with `<scope>main</scope>` and `<lang>java</lang>`, then `<sourceDirectory>` is ignored. This rule exits because Maven sets default values to those fields, which can interfere with user's setting.

**Note:** the `SourceRoot` API uses method names without the `get` prefix in anticipation for possible use of records in the future.

### DIFF
--- a/api/maven-api-model/src/main/mdo/maven.mdo
+++ b/api/maven-api-model/src/main/mdo/maven.mdo
@@ -2041,7 +2041,7 @@
             ]]>
           </description>
           <type>String</type>
-          <defaultValue>main</defaultValue>
+          <defaultValue>java</defaultValue>
         </field>
         <field>
           <name>module</name>
--- a/api/maven-api-model/src/main/mdo/maven.mdo
+++ b/api/maven-api-model/src/main/mdo/maven.mdo
@@ -775,7 +775,6 @@
           <type>String</type>
         </field>
         <field>
-          <!-- TODO: replaced by <Source> with "main" scope and "resources" language -->
           <name>resources</name>
           <version>3.0.0+</version>
           <description>
@@ -783,6 +782,8 @@
             files associated with a project. These resources are often included in the final
             package.
             The default value is {@code src/main/resources}.
+
+            @deprecated Replaced by &lt;Source&gt; with "main" scope and "resources" language.
           </description>
           <association>
             <type>Resource</type>
@@ -790,13 +791,14 @@
           </association>
         </field>
         <field>
-          <!-- TODO: replaced by <Source> with "test" scope and "resources" language -->
           <name>testResources</name>
           <version>4.0.0+</version>
           <description>
             This element describes all the classpath resources such as properties
             files associated with a project's unit tests.
             The default value is {@code src/test/resources}.
+
+            @deprecated Replaced by &lt;Source&gt; with "test" scope and "resources" language.
           </description>
           <association>
             <type>Resource</type>
@@ -874,7 +876,6 @@
           </association>
         </field>
         <field>
-          <!-- TODO: replaced by <Source> with "main" scope. -->
           <name>sourceDirectory</name>
           <version>3.0.0+</version>
           <required>true</required>
@@ -883,11 +884,12 @@
             generated build system will compile the sources from this directory when the project is
             built. The path given is relative to the project descriptor.
             The default value is {@code src/main/java}.
+
+            @deprecated Replaced by &lt;Source&gt; with "main" scope.
           </description>
           <type>String</type>
         </field>
         <field>
-          <!-- TODO: replaced by <Source> with "script" language -->
           <name>scriptSourceDirectory</name>
           <version>4.0.0+</version>
           <required>true</required>
@@ -897,11 +899,12 @@
             contents will be copied to the output directory in most cases (since scripts are
             interpreted rather than compiled).
             The default value is {@code src/main/scripts}.
+
+            @deprecated Replaced by &lt;Source&gt; with "script" language.
           </description>
           <type>String</type>
         </field>
         <field>
-          <!-- TODO: replaced by <Source> with "test" scope. -->
           <name>testSourceDirectory</name>
           <version>4.0.0+</version>
           <required>true</required>
@@ -910,6 +913,8 @@
             project. The generated build system will compile these directories when the project is
             being tested. The path given is relative to the project descriptor.
             The default value is {@code src/test/java}.
+
+            @deprecated Replaced by &lt;Source&gt; with "test" scope.
           </description>
           <type>String</type>
         </field>
@@ -2143,7 +2148,6 @@
       </fields>
     </class>
     <class>
-      <!-- TODO: Replaced by <Source> with "resources" language. -->
       <name>Resource</name>
       <description>This element describes all of the classpath resources associated with a project
         or unit tests.</description>
@@ -2161,6 +2165,8 @@
             element with this value: {@code org/apache/maven/messages}.
             This is not required if you simply put the resources in that directory
             structure at the source, however.
+
+            @deprecated Replaced by &lt;Source&gt; with "resources" language.
           </description>
           <type>String</type>
         </field>
--- a/impl/maven-core/src/main/java/org/apache/maven/project/MavenProject.java
+++ b/impl/maven-core/src/main/java/org/apache/maven/project/MavenProject.java
@@ -33,8 +33,12 @@ import java.util.Objects;
 import java.util.Properties;
 import java.util.Set;
 import java.util.function.Predicate;
+import java.util.stream.Stream;
 
 import org.apache.maven.RepositoryUtils;
+import org.apache.maven.api.Language;
+import org.apache.maven.api.ProjectScope;
+import org.apache.maven.api.SourceRoot;
 import org.apache.maven.artifact.Artifact;
 import org.apache.maven.artifact.ArtifactUtils;
 import org.apache.maven.artifact.DependencyResolutionRequiredException;
@@ -42,6 +46,7 @@ import org.apache.maven.artifact.factory.ArtifactFactory;
 import org.apache.maven.artifact.handler.ArtifactHandler;
 import org.apache.maven.artifact.repository.ArtifactRepository;
 import org.apache.maven.artifact.resolver.filter.ArtifactFilter;
+import org.apache.maven.impl.DefaultSourceRoot;
 import org.apache.maven.lifecycle.internal.DefaultProjectArtifactFactory;
 import org.apache.maven.model.Build;
 import org.apache.maven.model.CiManagement;
@@ -121,8 +126,10 @@ public class MavenProject implements Cloneable {
 
     private Set<Artifact> pluginArtifacts;
 
+    @Deprecated
     private List<ArtifactRepository> remoteArtifactRepositories;
 
+    @Deprecated
     private List<ArtifactRepository> pluginArtifactRepositories;
 
     private List<RemoteRepository> remoteProjectRepositories;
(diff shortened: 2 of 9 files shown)


api/maven-api-model/src/main/mdo/maven.mdo
```diff
@@ -775,28 +775,30 @@
...
-          <!-- TODO: replaced by <Source> with "main" scope and "resources" language -->
           <name>resources</name>
           <version>3.0.0+</version>
           <description>
             This element describes all the classpath resources such as properties
             files associated with a project. These resources are often included in the final
             package.
             The default value is {@code src/main/resources}.
+
+            @deprecated Replaced by &lt;Source&gt; with "main" scope and "resources" language.
```
gnodet, line 786:
```suggestion
            @deprecated Replaced by {@code &lt;Source&gt;} with "main" scope and "resources" language.
```

api/maven-api-model/src/main/mdo/maven.mdo
```diff
@@ -775,28 +775,30 @@
...
         <field>
-          <!-- TODO: replaced by <Source> with "test" scope and "resources" language -->
           <name>testResources</name>
           <version>4.0.0+</version>
           <description>
             This element describes all the classpath resources such as properties
             files associated with a project's unit tests.
             The default value is {@code src/test/resources}.
+
+            @deprecated Replaced by &lt;Source&gt; with "test" scope and "resources" language.
```
gnodet, line 801:
```suggestion
            @deprecated Replaced by {@code &lt;Source&gt;} with "test" scope and "resources" language.
```

api/maven-api-model/src/main/mdo/maven.mdo
```diff
@@ -883,11 +884,12 @@
             generated build system will compile the sources from this directory when the project is
             built. The path given is relative to the project descriptor.
             The default value is {@code src/main/java}.
+
+            @deprecated Replaced by &lt;Source&gt; with "main" scope.
```
gnodet, line 888:
```suggestion
            @deprecated Replaced by {@code &lt;Source&gt;} with "main" scope.
```

api/maven-api-model/src/main/mdo/maven.mdo
```diff
@@ -897,11 +899,12 @@
             contents will be copied to the output directory in most cases (since scripts are
             interpreted rather than compiled).
             The default value is {@code src/main/scripts}.
+
+            @deprecated Replaced by &lt;Source&gt; with "script" language.
```
gnodet, line 903:
```suggestion
            @deprecated Replaced by {@code &lt;Source&gt;} with "script" language.
```

api/maven-api-model/src/main/mdo/maven.mdo
```diff
@@ -910,6 +913,8 @@
             project. The generated build system will compile these directories when the project is
             being tested. The path given is relative to the project descriptor.
             The default value is {@code src/test/java}.
+
+            @deprecated Replaced by &lt;Source&gt; with "test" scope.
```
gnodet, line 917:
```suggestion
            @deprecated Replaced by {@code &lt;Source&gt;} with "test" scope.
```

api/maven-api-model/src/main/mdo/maven.mdo
```diff
@@ -2161,6 +2165,8 @@
             element with this value: {@code org/apache/maven/messages}.
             This is not required if you simply put the resources in that directory
             structure at the source, however.
+
+            @deprecated Replaced by &lt;Source&gt; with "resources" language.
```
gnodet, line 2169:
```suggestion
            @deprecated Replaced by {@code &lt;Source&gt;} with "resources" language.
```

gnodet (changes requested):
Javadoc fixes

api/maven-api-model/src/main/mdo/maven.mdo
desruisseaux: `{@code]` escapes HTML characters, so the output would really be `&lt;Source&gt;` in the Javadoc. If it was a Java file, I would have written {@code <Source>} (allowing direct use of `<` and `>` are the purpose of `{@code]` and `{@literal]`). But since this is a `.mdo` file, I wasn't sure if `<Source>` wouldn't be interpreted as XML. I thought that escaping the XML characters was the safest, which means that I cannot put them in a`{@code]` element.

desruisseaux: Ah sorry, I jsut saw the CI output. I see that `&lt;` has been replaced by `<` in the generated Java file. Then you are right, I need to add the `{@code}` tags.

api/maven-api-model/src/main/mdo/maven.mdo
```diff
@@ -775,28 +775,30 @@
...
             package.
             The default value is {@code src/main/resources}.
+
+            @deprecated Replaced by {@code &lt;Source&gt;} with {@code main} scope and {@code resources} language.
           </description>
           <association>
             <type>Resource</type>
             <multiplicity>*</multiplicity>
           </association>
         </field>
```
gnodet, line 792: All deprecated elements should be added
```
          <annotations>
            <annotation>@Deprecated</annotation>
          </annotations>
```
This will add the usual `@Deprecated` annotation to the field and getter.

gnodet (changes requested):
Add a `@Deprecated` annotation.

api/maven-api-model/src/main/mdo/maven.mdo
desruisseaux: Should we also replace `<version>3.0.0+</version>` by `<version>4.0.0</version>`? I see that other deprecated elements such as `module` have the latter.

gnodet: No the version here refers to the `modelVersion`, not Maven's version, so we don't want to change them.  New elements are introduced in model 4.1.0 which is fine.

impl/maven-core/src/main/java/org/apache/maven/project/MavenProject.java
```diff
@@ -420,17 +425,29 @@ public void addTestCompileSourceRoot(String path) {
...
      * @param scope the scope of the sources to return, or {@code null} for all scopes
      * @param language the language of the sources to return, or {@code null} for all languages
      *
+     * @see #addSourceRoot(ProjectScope, Language, Path)
+     *
      * @since 4.0.0
      */
-    public Stream<SourceRoot> getSourceRoots(ProjectScope scope, Language language) {
-        Stream<SourceRoot> s = sources.values().stream();
+    public Stream<SourceRoot> getEnabledSourceRoots(ProjectScope scope, Language language) {
```
gnodet, line 449: Don't we need those methods on `org.apache.maven.api.Project` ? That's what will be available to the Maven 4 plugins.

desruisseaux: Looking at it now, opportunistically adding `{@return}` tags in Javadoc of `Project`. In this process, I came across this method (abridged):

```java
    /**
     * Returns the project artifacts as immutable list. Elements are the project POM artifact and the artifact
     * produced by this project build, if applicable. Hence, the returned list may have one or two elements
     * (never less than 1, never more than 2), depending on project packaging.
     */
    List<ProducedArtifact> getArtifacts();
```

If 2 elements means, for example, "path to a JAR file + path to its POM file", would it be easier to add something like `Optional<ProducedArtifact> getPOM()` method in `ProducedArtifact`? Then, the signature of above `getArtifacs()` would stay the same but its documentation changed for saying that the list usually contains only one element. The goal is to allow a project to produce more than 1 "JAR + POM" pair in the future (it may be useful with Java _modules source hierarchy_, if the user choose to organize his project that way).

gnodet: I don't think that works.  We can have multiple POMs or multiple JARs attached, so there's no one-to-one mapping between a jar and a pom.

api/maven-api-core/src/main/java/org/apache/maven/api/Project.java
```diff
@@ -234,4 +239,32 @@ default String getId() {
...
+     * If the given scope is {@code null}, then this method returns the enabled sources for all scopes.
+     * If the given language is {@code null}, then this method returns the enabled sources for all languages.
+     * An arbitrary number of source roots may exist for the same scope and language.
+     * It may be, for example, the case of a multi-versions project.
+     * The iteration order is the order in which the sources are declared in the POM file.
+     *
+     * @param scope the scope of the sources to return, or {@code null} for all scopes
+     * @param language the language of the sources to return, or {@code null} for all languages
+     */
+    Stream<SourceRoot> getEnabledSourceRoots(ProjectScope scope, Language language);
```
gnodet, line 269: The original idea was to have the `Project` completely immutable and have modifications done through the `ProjectManager` service.  Plugins can add sources during the build, so both accessors and setters are currently defined on `ProjectManager`, see https://github.com/apache/maven/blob/dc6d48ce00c3ebc0579079cadc93ae6175ed183b/api/maven-api-core/src/main/java/org/apache/maven/api/services/ProjectManager.java#L120-L158

desruisseaux: The methods to remove would be `Project.getSourceRoots()` and `getEnabledSourceRoots(…)`? I thought that the [previous comment](https://github.com/apache/maven/pull/2061#discussion_r1930340516) was about adding them.

gnodet: Yes, but we need methods to add a `Source` to a given project at runtime I think.
The incoherence is more about the location of these.  The `Project` is supposed to be immutable, that's why they are located inside the `ProjectManager` which purpose is to hold modifications to the project.

desruisseaux: Pushed addition of two `addSourceRoot` method (the second one is a convenience) and removed the 4 above-cited methods.

Note: for interfaces such as `Project` that are intended to be unmodifiable, should be remove the `get` prefix? It would express the intent that those interfaces have no setter methods, and would allow to implement some of them with `record`. The latter may not be applicable to `DefaultProject`, but some other implementations may benefit from that.

gnodet: Yes, it could make sense to align with the `record` way of doing things.

gnodet: I guess the most important ones would be the model classes themselves, but this would be quite a big change.

gnodet:
I think we also need to move the `Project#getSourceRoots()` and `Project#getEnabledSourceRoots(ProjectScope,Language)` to `ProjectManager` as they return values that are not meant to be immutable.  I know from an implementation POV, they are still managed by the underlying `MavenProject` itself, but the idea would be to get rid of the underlying Maven 3 layer in the future.

api/maven-api-core/src/main/java/org/apache/maven/api/Project.java
desruisseaux: Maybe we should start with interfaces that are new in Maven 4. Nevertheless closing this conversation as it would be the topic for a separated pull request.

gnodet:
Thx, a few minor changes in https://github.com/Geomatys/maven/pull/1
At some point, it would be nice to have an IT that would use the new `source` element, as this is supposed to work even with the current set of plugins I think.

gnodet (approved):

desruisseaux:
> At some point, it would be nice to have an IT that would use the new `source` element

Test added in [pull request 320 of compiler plugin](https://github.com/apache/maven-compiler-plugin/pull/320) ([commit](https://github.com/apache/maven-compiler-plugin/pull/320/commits/195a80949b19aca24834b2be22d27c529537d10e)).

jira-importer:
Resolve #9458

PR #2062, opened 2025-01-27
[MNG-8477] Fix problems with file activation being wrongly cached

gnodet:
JIRA issue: [MNG-8477](https://issues.apache.org/jira/browse/MNG-8477)

Disable profile activation interpolation before checking profile activations.
The interpolated activations are cached and lead to wrong evaluations.
Given the activations are interpolated anyway during activation, there's
no need to do that before.


cstamas (approved):
Verified with MNG-8477 reproducer, fixed! Puush

https://gist.github.com/cstamas/817b35264336928d3d6deff046d4677a

jira-importer:
Resolve #9955

PR #2063, opened 2025-01-28
[MNG-8544] Conflicting extensions went unnoticed

cstamas:
Now that Maven4 has several extension sources, we need to be able to detect conflicts. For now, we allow only one GA and fail the bootstrapping, if conflict discovered.

User should fix the situation, with the help of detailed error message enumerating all the conflicts.

---

https://issues.apache.org/jira/browse/MNG-8544


gnodet (approved):
LGTM

PR #2067, opened 2025-01-28
[MNG-8545] More repository handling fixes

cstamas:
Multiple fixes:
* cling does not override repositories properly at bootstrap
* Source problems caused cache misses

---

https://issues.apache.org/jira/browse/MNG-8545


gnodet (approved):

gnodet (approved):
I've changed a few things to introduce a `Sources` class which holds the factory methods and the implementations.

PR #2069, opened 2025-01-29
[MNG-8513] Move xml combination mode validation to validator

cstamas:
As Central has POMs using wrong values, causing even dependency collation to fail.

---

https://issues.apache.org/jira/browse/MNG-8513


gnodet (approved):
LGTM

jira-importer:
Resolve #9376

PR #2076, opened 2025-01-30
[MNG-8551] Improve early error reporting

cstamas:
CLIng changes:
* parser does not throw but signals if "parsing failed" to invoker via new flag. Invoker should "early bail out" in these cases, as it is possible that even args were not parsed.
* logging is now fixed, in a way, that we depend on runtime configured Slf4j backend being in use. At "early stages" logger is just accumulating log entries, and logs are emitted to STDERR (in case of "early failure") or to proper Slf4j backend, once configured.
* Parser now validates options, all those that can down the road cause error.
* dropped ParserException and other unused bits.

In short: "early" messages go to stderr, while once logger is set up, to logger backend.

---

https://issues.apache.org/jira/browse/MNG-8551


gnodet (approved):
LGTM

PR #2078, opened 2025-02-01
[MNG-8554] Fix UT suite for CLI

cstamas:
The whole UT suite was written to rely on maven.properties of the "outer build", hence UTs were not even running when built with Maven 3. Latest changes made it simply fail.

Fix the UT suite, as BasePasrser is depending on maven.properties being loaded, so provide one in a fake maven.home. Now the UT is always running, irrelevant of Maven being built with.

---

https://issues.apache.org/jira/browse/MNG-8554


jira-importer:
Resolve #9892

PR #2085, opened 2025-02-05
[MNG-8561] SourceRoot should be more lenient wrt duplicates

gnodet:
Add includes/excludes to the SourceKey

JIRA issue: [MNG-8561](https://issues.apache.org/jira/browse/MNG-8561)

### DIFF
--- a/impl/maven-core/src/main/java/org/apache/maven/project/MavenProject.java
+++ b/impl/maven-core/src/main/java/org/apache/maven/project/MavenProject.java
@@ -417,14 +417,14 @@ public class MavenProject implements Cloneable {
      * @since 4.0.0
      */
     public Stream<SourceRoot> getEnabledSourceRoots(ProjectScope scope, Language language) {
-        Stream<SourceRoot> s = sources.stream().filter(SourceRoot::enabled);
+        Stream<SourceRoot> stream = sources.stream().filter(SourceRoot::enabled);
         if (scope != null) {
-            s = s.filter((source) -> scope.equals(source.scope()));
+            stream = stream.filter(source -> scope.equals(source.scope()));
         }
         if (language != null) {
-            s = s.filter((source) -> language.equals(source.language()));
+            stream = stream.filter(source -> language.equals(source.language()));
         }
-        return s;
+        return stream;
     }
 
     /**
--- a/impl/maven-core/src/main/java/org/apache/maven/project/MavenProject.java
+++ b/impl/maven-core/src/main/java/org/apache/maven/project/MavenProject.java
@@ -22,6 +22,7 @@ import java.io.File;
 import java.io.IOException;
 import java.io.Writer;
 import java.nio.file.Path;
+import java.util.AbstractList;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Collections;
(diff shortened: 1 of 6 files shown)


impl/maven-impl/src/main/java/org/apache/maven/impl/DefaultSourceRoot.java
```diff
@@ -136,6 +136,31 @@ public DefaultSourceRoot(final ProjectScope scope, final Language language, fina
         enabled = true;
     }
 
+    /**
+     * Creates a new instance for the given directory and scope. The language is assumed Java.
```
elharo, line 140: assumed to be

impl/maven-impl/src/main/java/org/apache/maven/impl/DefaultSourceRoot.java
```diff
@@ -136,6 +136,31 @@ public DefaultSourceRoot(final ProjectScope scope, final Language language, fina
...
+     *
+     * @param scope scope of source code (main or test)
+     * @param language language of the source code
+     * @param directory directory of the source code
+     */
+    public DefaultSourceRoot(
+            final ProjectScope scope,
+            final Language language,
```
elharo, line 148: If it's assumed to be Java why are we specifying it here?

elharo (changes requested):
Needs tests

desruisseaux:
We could also remove this `SourceKey` stuff completely. The intend was to prevent ambiguous situations or potential mistakes such as declaring the same sources for two Java releases. But those checks are superficial anyway.

gnodet:
I've added support for wrapping the `<resources>` and `<testResources>`, remove the `SourceKey` completely to restore previous behavior and added an IT.

impl/maven-core/src/main/java/org/apache/maven/project/MavenProject.java
```diff
@@ -770,23 +738,69 @@ public Build getBuild() {
...
+            @Override
+            public Resource get(int index) {
+                return toResource(getEnabledSourceRoots(ProjectScope.TEST, Language.RESOURCES)
+                        .toList()
+                        .get(index));
+            }
+
+            @Override
+            public int size() {
+                return (int) getEnabledSourceRoots(ProjectScope.TEST, Language.RESOURCES)
```
elharo, line 778: why the cast to int? I would expect this would return an int already. Surely it isn't big enough to require a long?

gnodet: `Stream.count()` actually returns a `long` ...

gnodet: Remember a stream can be generated on the fly using `Stream#generate`, so it does not need to hold all elements in memory.

impl/maven-core/src/main/java/org/apache/maven/project/MavenProject.java
```diff
@@ -449,7 +417,7 @@ public Collection<SourceRoot> getSourceRoots() {
      * @since 4.0.0
      */
     public Stream<SourceRoot> getEnabledSourceRoots(ProjectScope scope, Language language) {
-        Stream<SourceRoot> s = sources.values().stream().filter(SourceRoot::enabled);
+        Stream<SourceRoot> s = sources.stream().filter(SourceRoot::enabled);
```
elharo, line 420: Maybe s --> stream
When I see a single letter s variable, I think string, not stream.

elharo (approved):

PR #2086, opened 2025-02-05
[MNG-8530] Prompter fixes

cstamas:
This change aligns now what we really want:
* maven-core provides Plexus Interactivity API (package export was already present, but NOT artifactExport)
* maven-jline provides Plexus Interactivity API implementation that delegates to new M4 API implementation
* new Maven4 API Prompter implementation provided
* exploded the super overloaded MessageFactory implementation

---

https://issues.apache.org/jira/browse/MNG-8530


gnodet (approved):
Looks good, nice work !

impl/maven-jline/src/main/java/org/apache/maven/jline/LegacyPlexusInteractivity.java
```diff
@@ -0,0 +1,134 @@
...
+import org.eclipse.sisu.Priority;
+
+/**
+ * This class is injected into any legacy component that would want to use legacy "Plexus Interactivity API".
+ * It simply delegates to Maven4 API {@link DefaultPrompter}.
+ */
+@Experimental
+@Named
+@Singleton
+@Priority(10)
```
gnodet, line 43: @cstamas Would it make sense to move that one to `compat/maven-compat` ? The idea would be to avoid introducing any new dependency on sisu in `impl/*` jars ?

jira-importer:
Resolve #9373

PR #2088, opened 2025-02-06
[MNG-8564] Move to maven-compat Plexus legacy

cstamas:
This merely moves a newly introduced legacy support class to maven-compat.

---

https://issues.apache.org/jira/browse/MNG-8564


gnodet (approved):
Looks good !

PR #2091, opened 2025-02-06
[MNG-8563] Improve the object cache memory usage

gnodet:
Improve memory consumption as the current code keeps lots of empty maps in memory.


jira-importer:
Resolve #9884

PR #2092, opened 2025-02-07
[MNG-8575] Replace a list with O(N²) performance by O(N) at least during iteration.

desruisseaux:
JIRA issue: [MNG-8575](https://issues.apache.org/jira/browse/MNG-8575)

The use of `AbstractList` with an implementation that creates the whole list every time that `get(int)` is invoked implies that iterations over that list would have a performance cost of O(N²) where _N_ is the number of elements. Extending `AbstractSequentialList` instead brings back the performance cost to O(N) at least during iterations.

### DIFF
--- a/impl/maven-core/src/main/java/org/apache/maven/project/MavenProject.java
+++ b/impl/maven-core/src/main/java/org/apache/maven/project/MavenProject.java
@@ -35,7 +35,6 @@ import java.util.Map;
 import java.util.Objects;
 import java.util.Properties;
 import java.util.Set;
-import java.util.concurrent.CopyOnWriteArrayList;
 import java.util.function.Predicate;
 import java.util.stream.Stream;
 
@@ -149,7 +148,7 @@ public class MavenProject implements Cloneable {
     /**
      * All sources of this project, in the order they were added.
      */
-    private List<SourceRoot> sources = new CopyOnWriteArrayList<>();
+    private Set<SourceRoot> sources = new LinkedHashSet<>();
 
     @Deprecated
     private ArtifactRepository releaseArtifactRepository;
@@ -327,9 +326,7 @@ public class MavenProject implements Cloneable {
      * @since 4.0.0
      */
     public void addSourceRoot(SourceRoot source) {
-        if (!sources.contains(source)) {
-            sources.add(source);
-        }
+        sources.add(source);
     }
 
     /**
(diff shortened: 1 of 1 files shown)


desruisseaux:
It brings a related question. The following code also has a O(N²) performance cost if a large number _N_ of sources are added through this method:

```java
    private List<SourceRoot> sources = new CopyOnWriteArrayList<>();

    public void addSourceRoot(SourceRoot source) {
        if (!sources.contains(source)) {
            sources.add(source);
        }
    }
```

We could reduce the cost to O(N) by using `LinkedHashSet` instead of a list. However, `CopyOnWriteArrayList` is thread-safe while `LinkedHashSet` is not. Do we really need thread-safety for that particular field while everything else in `MavenProject` is not thread-safe? I agree that we may want thread-safety later, but maybe it would need to a more general strategy (e.g. with immutable records)?

cstamas:
Maybe we want two methods to add sources? And one could be a "batch" (several sources at once) to be used by Project Builder and other adding one by one like for example some plugin may be? Am unsure would the rate of second call justify complicating things....

desruisseaux:
A batch method would not fundamentally change the O(N²) nature of the operation, unless it temporarily copy the list in an hash set, in which case we lost the concurrency characteristics of `CopyOnWriteArrayList`.

I think that the first question still: do we need to care about concurrency or thread safety now, given that the rest of the class is not thread-safe? If no, we can just replace `CopyOnWriteArrayList` by `LinkedHashSet` and we are done.

gnodet:
> A batch method would not fundamentally change the O(N²) nature of the operation, unless it temporarily copy the list in an hash set, in which case we lost the concurrency characteristics of `CopyOnWriteArrayList`.
> 
> I think that the first question still: do we need to care about concurrency or thread safety now, given that the rest of the class is not thread-safe? If no, we can just replace `CopyOnWriteArrayList` by `LinkedHashSet` and we are done.

I don't think we need thread safety here.  Thread safe

> A batch method would not fundamentally change the O(N²) nature of the operation, unless it temporarily copy the list in an hash set, in which case we lost the concurrency characteristics of `CopyOnWriteArrayList`.
> 
> I think that the first question still: do we need to care about concurrency or thread safety now, given that the rest of the class is not thread-safe? If no, we can just replace `CopyOnWriteArrayList` by `LinkedHashSet` and we are done.

I don't think we do.  We can defer that at the time we reverse the MavenProject implementation to be implemented on top of Project.

gnodet (approved):

cstamas (approved):

PR #2095, opened 2025-02-09
Fix links to Settings Descriptor Reference

kwin:
Adjust link for Maven 4 Password Encryption


cstamas (approved):
But this is maven-compat module, so unsure why here?

kwin:
I guess for historical reasons, we can move later on to a more appropriate place.

PR #2099, opened 2025-02-10
[MNG-8573] OS Service

cstamas:
And move the OS static helper to util, is not
profile specific.

---

https://issues.apache.org/jira/browse/MNG-8573

### DIFF
--- a/api/maven-api-core/src/main/java/org/apache/maven/api/services/OsService.java
+++ b/api/maven-api-core/src/main/java/org/apache/maven/api/services/OsService.java
@@ -1,0 +1,64 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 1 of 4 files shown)


api/maven-api-core/src/main/java/org/apache/maven/api/services/OsService.java
```diff
@@ -0,0 +1,60 @@
...
+    String version();
+
+    /**
+     * Returns the OS family.
+     */
+    @Nonnull
+    String family();
+
+    /**
+     * Returns {@code true} if you need to make extra hoops and loops.
```
pzygielo, line 57: I hope this javadoc could be reworded.

cstamas: sure, am just testing our reviewers "eagle eye"

cstamas:
The goal is simple:
* the API is meant for plugins and extensions, while Maven "inner stuff" do have the static helper class
* having Service (instead of static class/methods) we are in greater control and are future proof
* for example (unsure how feasible this would be), we could test plugins on "spoofed OS"es?

gnodet:
I've enhanced the javadoc, but I wonder if it would make more sense to use proper enums, as we don't even define the values.

cstamas:
Yes, I intentionally left them out, as IMO this service would be most used for things like:
* print out some OS related stuff (like `mvn -V` does)
* the "is windows" stuff is hot and will be most used, as I envision

I just wanted to keep this OS stuff small, minimal.

gnodet (approved):
LGTM

jira-importer:
Resolve #9877

PR #2105, opened 2025-02-11
Add PR Automation action

slawekjaranowski:
https://github.com/apache/maven-gh-actions-shared?tab=readme-ov-file#pull-request-automation

### DIFF
--- a/.github/workflows/pr-automation.yml
+++ b/.github/workflows/pr-automation.yml
@@ -1,0 +1,27 @@
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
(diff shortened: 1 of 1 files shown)


slachiewicz (approved):

Bukama (approved):

gnodet:
We currently have more than one milestone opened for master branch: the upcoming rc-3, but we also have 4.0.0-rc-4 , 4.1.0 and 4.x.  What happens in this case ?

slawekjaranowski:
> We currently have more than one milestone opened for master branch: the upcoming rc-3, but we also have 4.0.0-rc-4 , 4.1.0 and 4.x. What happens in this case ?

When we merg PR without milestone assigned automat will search milestone which has in description target branch name.
So we need at least milestones for each branch which we want to accept merges for.

If PR will has milestone assigned nothing will happen.

Brach to milestone assign example in:
https://github.com/apache/maven-clean-plugin/milestones

We can have more milestone which will be not used by automat.

slawekjaranowski:
I also added branch names to maven milestones https://github.com/apache/maven/milestones

PR #2112, opened 2025-02-15
[MNG-8582] Keep stdout/stderr functioning

cstamas:
This call has severe implications (re-ties logger, and in case is invoked multiple times, causes "loop").
Make sure it happens only once, This changes does not affect mvn and mvnd (it installs own stdout/stderr), but does affects mvnsh and resident maven,

---

https://issues.apache.org/jira/browse/MNG-8582


gnodet (approved):

jira-importer:
Resolve #9863

PR #2117, opened 2025-02-19
[MNG-8587] mvnsh navigation

cstamas:
Ability to navigate on disk and new commands:
* `!` execute shell command (executable or script)
* `cd` change cwd
* `pwd` print cwd

---

https://issues.apache.org/jira/browse/MNG-8587


cstamas:
@gnodet help needed, tried to implement the "shell" command as usually done with `!`, but for some reason did not work, `!ls` invoked the `ls` command (the `!` got removed for some reason). Why?

impl/maven-cli/src/main/java/org/apache/maven/cling/invoker/Directory.java
```diff
@@ -0,0 +1,68 @@
...
+    }
+
+    public static class MutableDirectory extends Directory {
+        public MutableDirectory(Path directory) {
+            super(directory);
+        }
+
+        public void changeDirectory(String seg) {
+            requireNonNull(seg, "seg");
+            if ("..".equals(seg)) {
```
gnodet, line 52: I don't the point for this specific test.
What if you `cd ../.`, that would be the same ?
Also `directory.resolve("..")` should work, so I don't get why we'd need this branch.

impl/maven-cli/src/main/java/org/apache/maven/cling/invoker/Directory.java
```diff
@@ -0,0 +1,68 @@
...
+package org.apache.maven.cling.invoker;
+
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.util.function.Function;
+import java.util.function.Supplier;
+
+import static java.util.Objects.requireNonNull;
+
+public class Directory implements Supplier<Path>, Function<String, Path> {
```
gnodet, line 28: That looks overkill to me.  Why do we use `Function<String, Path>` instead of just a `Path` btw ?
And have a non `final` `Path` field for the `cwd` ?

cstamas:
Applied proposed changes. And some explanations:
* removing special handling for ".." and simplified things
* we deal with 3 paths: installationDir, userDir and CWD. Former two are never changing, latter is mutable (but only in mvnsh).
* hence, installation and user made plain Path, but it shows the intent of `Function<String, Path>` now: I just wanted to prevent repeated use of `normalize()` (as user may say `-Dfoo=../../bar.xml` or alike). But given there is a handful access to installation and user, this is fine.
* renamed Directory to CWD to have clear intent (and it is accessed MUST MORE than former two, hence i made here a new method `resolve(seg)` (equivalent of former function). Also, I tried to made clear how CWD is mutable, but allows transitions to existing directories ONLY.

cstamas:
Anyone?

impl/maven-cli/src/main/java/org/apache/maven/cling/invoker/mvnsh/builtin/BuiltinShellCommandRegistryFactory.java
```diff
@@ -130,53 +129,140 @@ public String name() {
...
+            ProcessBuilder builder = new ProcessBuilder();
+            List<String> processArgs = new ArrayList<>();
+            if (Os.IS_WINDOWS) {
+                processArgs.add("cmd.exe");
+                processArgs.add("/c");
+            } else {
+                processArgs.add("sh");
+                processArgs.add("-c");
+            }
+            processArgs.add(String.join(" ", args));
```
slawekjaranowski, line 142: we should add each item as new argument  ... not a one arument with spaces

cstamas: This is shell `-c` and it must be one argument (space separated)

impl/maven-cli/src/main/java/org/apache/maven/cling/invoker/mvnsh/builtin/BuiltinShellCommandRegistryFactory.java
```diff
@@ -75,15 +75,14 @@ private BuiltinShellCommandRegistry(LookupContext shellContext) {
...
-            for (Command c : commands) {
-                commandName.put(c, c.name().toLowerCase());
-            }
-            commandExecute.put(Command.MVN, new CommandMethods(this::mvn, this::mvnCompleter));
-            commandExecute.put(Command.MVNENC, new CommandMethods(this::mvnenc, this::mvnencCompleter));
-            registerCommands(commandName, commandExecute);
+            Map<String, CommandMethods> commandExecute = new HashMap<>();
+            commandExecute.put("!", new CommandMethods(this::shell, this::defaultCompleter));
+            commandExecute.put("cd", new CommandMethods(this::cd, this::cdCompleter));
+            commandExecute.put("ls", new CommandMethods(this::ls, this::defaultCompleter));
```
gnodet, line 81: JLine [defines a bunch of commands](https://github.com/jline/jline3/blob/master/demo/src/main/java/org/apache/felix/gogo/jline/Posix.java) already.  They are currently leveraging Felix Gogo, but it should be fairly easy to abstract from it and make them reusable.  
Would that make sense to reuse them here ?

impl/maven-cli/src/main/java/org/apache/maven/cling/invoker/mvnsh/builtin/BuiltinShellCommandRegistryFactory.java
```diff
@@ -130,53 +129,133 @@ public String name() {
...
+                shellContext.writer.accept(shellContext.cwd.get().toString());
+            } catch (Exception e) {
+                saveException(e);
             }
-            return null;
         }
 
         private void mvn(CommandInput input) {
             try {
                 shellMavenInvoker.invoke(mavenParser.parseInvocation(
```
gnodet, line 206: I don't think this is cause by this commit, but when I hit `Ctrl+C` on the prompt, `mvnsh` prints a new line.  However, when doing the same when running `mvn` command (from inside the `mvnsh`), it completely exits.  I think it should abort the build, but restart the prompt.

impl/maven-cli/src/main/java/org/apache/maven/cling/invoker/mvnsh/builtin/BuiltinShellCommandRegistryFactory.java
```diff
@@ -130,53 +129,133 @@ public String name() {
...
+            return List.of(new ArgumentCompleter(new StringsCompleter(
+                    "clean",
+                    "validate",
+                    "compile",
+                    "test",
+                    "package",
+                    "verify",
+                    "install",
+                    "deploy",
+                    "wrapper:wrapper")));
```
gnodet, line 225: Should that be computed somehow ?
If we can lookup a `LifecycleRegistry`, we could use it with something like:
```
context.lookup(LifecycleRegistry.class).stream().flatMap(Lifecycle:allPhases).map(Phase::name)
```

impl/maven-cli/src/main/java/org/apache/maven/cling/invoker/mvnsh/builtin/BuiltinShellCommandRegistryFactory.java
gnodet: Let me push that change

impl/maven-cli/src/main/java/org/apache/maven/cling/invoker/mvnsh/builtin/BuiltinShellCommandRegistryFactory.java
gnodet: Let me push the change.

impl/maven-cli/src/main/java/org/apache/maven/cling/invoker/mvnsh/builtin/BuiltinShellCommandRegistryFactory.java
cstamas: I was intentionally limiting "command set" to "navigation" only. Keep core simple: can run mvn/mvnenc (like in real shell, provided by distro) and you can navigate your workdir. All the rest is not in maven core anyway, and IMO should be provided like extensions IMO. I think we don't want a super-uber generic script/shell here, just leave the door open for whoever wants it, be able to do it.

impl/maven-cli/src/main/java/org/apache/maven/cling/invoker/mvnsh/builtin/BuiltinShellCommandRegistryFactory.java
cstamas: My intent here was to offer "most frequent" goals plus `wrapper:wrapper` as Maven4 does not have wrapper lifecycle anymore. But yes, make it complete by lookup is ok.

impl/maven-cli/src/main/java/org/apache/maven/cling/invoker/mvnsh/builtin/BuiltinShellCommandRegistryFactory.java
gnodet: Then, I'd remove `ls` and see if we can provide a posix set of commands in a separate extension.
I'd really like a property in `~/.m2/maven.properties` that could be parsed to discover extensions: this could include a list of GAvoids which would contain project-less mojos that could be run as commands, such as toolbox, etc...

cstamas: Removed `ls`, and yes, we may want to provide some standard extensions. And yes, that would be useful feat.

impl/maven-cli/src/main/java/org/apache/maven/cling/invoker/mvnsh/builtin/BuiltinShellCommandRegistryFactory.java
cstamas: This is fixed: the commands should handle InvokerException.ExitExcepion which contract is "do not output anything, I did all, you just exit w/ exitCode". Now, in case of mvnsh this is simply a check exitCode=0 all if fine, or non-zero then report error (but again, all the error conditions are printed out already).

cstamas: This now fixes also the `mvnsh> mvn -s nonexistent` case, where the cause of failure was swallowed:
```
maven mvnsh> mvn -s fofo
[ERROR] The specified user settings file does not exist: /home/cstamas/Worx/apache-maven/maven/fofo
[ERROR] mvn command exited with exit code 1
maven mvnsh> 
```

gnodet: > I don't think this is cause by this commit, but when I hit `Ctrl+C` on the prompt, `mvnsh` prints a new line. However, when doing the same when running `mvn` command (from inside the `mvnsh`), it completely exits. I think it should abort the build, but restart the prompt.

You last commit fixes `mvn --help`.  To fix `Ctrl+C` when running a command, we'd have to register an interrupt handler on the terminal and then interrupt the running maven command.

cstamas: Re ctrl+c case, am really unsure what to do, esp if user uses `-T` command... Adding this single line to mvnsh invoker does NOT work (and even deadlocks in some circumstances):
```
            context.terminal.handle(Terminal.Signal.INT, signal -> executeThread.interrupt());

```

cstamas: This what happens now:
```
maven mvnsh> mvn package
[INFO] Please upgrade to Mimir version 0.3.5-SNAPSHOT (you are using version 0.3.4)
[INFO] Mimir session created
[INFO] --------------------------------------------------------------------------------------------------------------------------
[INFO] Reactor Build Order:
[INFO] 
[INFO] Apache Maven                                                                                                         [pom]
[INFO] Maven 4 API                                                                                                          [pom]
[INFO] Maven 4 API :: Meta annotations                                                                                      [jar]
[INFO] Maven 4 API :: Dependency Injection                                                                                  [jar]
[INFO] Maven 4 API :: XML                                                                                                   [jar]
[INFO] Maven 4 API :: Model                                                                                                 [jar]
[INFO] Maven 4 API :: Plugin                                                                                                [jar]
[INFO] Maven 4 API :: Settings                                                                                              [jar]
[INFO] Maven 4 API :: Toolchain                                                                                             [jar]
[INFO] Maven 4 API :: Repository Metadata                                                                                   [jar]
[INFO] Maven 4 API :: Core                                                                                                  [jar]
[INFO] Maven 4 API :: SPI                                                                                                   [jar]
[INFO] Maven 4 API :: CLI                                                                                                   [jar]
[INFO] Maven 4 Implementation Modules                                                                                       [pom]
[INFO] Maven 4 Dependency Injection                                                                                         [jar]
[INFO] Maven 4 XML Implementation                                                                                           [jar]
[INFO] Maven 4 API Implementation                                                                                           [jar]
[INFO] Maven 4 JLine integration                                                                                            [jar]
[INFO] Maven 4 Logging                                                                                                      [jar]
[INFO] Maven Compatibility Modules                                                                                          [pom]
[INFO] Maven Artifact                                                                                                       [jar]
[INFO] Maven Model                                                                                                          [jar]
[INFO] Maven Builder Support (deprecated)                                                                                   [jar]
[INFO] Maven Model Builder (deprecated)                                                                                     [jar]
[INFO] Maven 3 Plugin API                                                                                                   [jar]
[INFO] Maven Repository Metadata Model                                                                                      [jar]
[INFO] Maven Settings                                                                                                       [jar]
[INFO] Maven Toolchain Model                                                                                                [jar]
[INFO] Maven Toolchain Builder (deprecated)                                                                                 [jar]
[INFO] Maven 4 Core                                                                                                         [jar]
[INFO] Maven Settings Builder (deprecated)                                                                                  [jar]
[INFO] Maven Artifact Resolver Provider (deprecated)                                                                        [jar]
[INFO] Maven Compat (deprecated)                                                                                            [jar]
[INFO] Maven 4 CLI                                                                                                          [jar]
[INFO] Maven Embedder (deprecated)                                                                                          [jar]
[INFO] Apache Maven Distribution                                                                                            [pom]
[INFO] Maven 4 Executor                                                                                                     [jar]
[INFO] 
[INFO] ------------------------------------------------< org.apache.maven:maven >------------------------------------------------
[INFO] Building Apache Maven 4.0.0-rc-3-SNAPSHOT                                                                           [1/37]
[INFO]   from pom.xml
[INFO] ---------------------------------------------------------[ pom ]----------------------------------------------------------
[INFO] 
[INFO] --- enforcer:3.5.0:enforce (enforce-bytecode-version) @ maven ---
[INFO] Rule 0: org.apache.maven.enforcer.rules.version.RequireJavaVersion passed
[INFO] Rule 1: org.codehaus.mojo.extraenforcer.dependencies.EnforceBytecodeVersion passed
[INFO] 
[INFO] --- enforcer:3.5.0:enforce (enforce-maven-version) @ maven ---
[INFO] Rule 0: org.apache.maven.enforcer.rules.version.RequireMavenVersion passed
[INFO] 
[INFO] --- enforcer:3.5.0:enforce (enforce-java-version) @ maven ---
[INFO] Rule 0: org.apache.maven.enforcer.rules.version.RequireJavaVersion passed
[INFO] 
[INFO] --- enforcer:3.5.0:enforce (ensure-no-sonatype-cipher-and-sec-dispatcher) @ maven ---
[INFO] Rule 0: org.apache.maven.enforcer.rules.dependency.BannedDependencies passed
[INFO] 
[INFO] --- dependency:3.7.1:properties (prepare-for-mockito-agent) @ maven ---
[INFO] 
[INFO] --- checkstyle:3.4.0:check (checkstyle-check) @ maven ---
^C[INFO] You have 0 Checkstyle violations.
[INFO] 
[INFO] --- spotless:2.43.0:check (default) @ maven ---
[INFO] --------------------------------------------------------------------------------------------------------------------------
[INFO] Reactor Summary for Apache Maven 4.0.0-rc-3-SNAPSHOT:
[INFO] 
[INFO] Apache Maven ......................................................................................... FAILURE [  0.493 s]
[INFO] Maven 4 API .......................................................................................... SKIPPED
[INFO] Maven 4 API :: Meta annotations ...................................................................... SKIPPED
[INFO] Maven 4 API :: Dependency Injection .................................................................. SKIPPED
[INFO] Maven 4 API :: XML ................................................................................... SKIPPED
[INFO] Maven 4 API :: Model ................................................................................. SKIPPED
[INFO] Maven 4 API :: Plugin ................................................................................ SKIPPED
[INFO] Maven 4 API :: Settings .............................................................................. SKIPPED
[INFO] Maven 4 API :: Toolchain ............................................................................. SKIPPED
[INFO] Maven 4 API :: Repository Metadata ................................................................... SKIPPED
[INFO] Maven 4 API :: Core .................................................................................. SKIPPED
[INFO] Maven 4 API :: SPI ................................................................................... SKIPPED
[INFO] Maven 4 API :: CLI ................................................................................... SKIPPED
[INFO] Maven 4 Implementation Modules ....................................................................... SKIPPED
[INFO] Maven 4 Dependency Injection ......................................................................... SKIPPED
[INFO] Maven 4 XML Implementation ........................................................................... SKIPPED
[INFO] Maven 4 API Implementation ........................................................................... SKIPPED
[INFO] Maven 4 JLine integration ............................................................................ SKIPPED
[INFO] Maven 4 Logging ...................................................................................... SKIPPED
[INFO] Maven Compatibility Modules .......................................................................... SKIPPED
[INFO] Maven Artifact ....................................................................................... SKIPPED
[INFO] Maven Model .......................................................................................... SKIPPED
[INFO] Maven Builder Support (deprecated) ................................................................... SKIPPED
[INFO] Maven Model Builder (deprecated) ..................................................................... SKIPPED
[INFO] Maven 3 Plugin API ................................................................................... SKIPPED
[INFO] Maven Repository Metadata Model ...................................................................... SKIPPED
[INFO] Maven Settings ....................................................................................... SKIPPED
[INFO] Maven Toolchain Model ................................................................................ SKIPPED
[INFO] Maven Toolchain Builder (deprecated) ................................................................. SKIPPED
[INFO] Maven 4 Core ......................................................................................... SKIPPED
[INFO] Maven Settings Builder (deprecated) .................................................................. SKIPPED
[INFO] Maven Artifact Resolver Provider (deprecated) ........................................................ SKIPPED
[INFO] Maven Compat (deprecated) ............................................................................ SKIPPED
[INFO] Maven 4 CLI .......................................................................................... SKIPPED
[INFO] Maven Embedder (deprecated) .......................................................................... SKIPPED
[INFO] Apache Maven Distribution ............................................................................ SKIPPED
[INFO] Maven 4 Executor ..................................................................................... SKIPPED
[INFO] --------------------------------------------------------------------------------------------------------------------------
[INFO] BUILD FAILURE
[INFO] --------------------------------------------------------------------------------------------------------------------------
[INFO] Total time:  0.560 s
[INFO] Finished at: 2025-02-24T12:40:07+01:00
[INFO] --------------------------------------------------------------------------------------------------------------------------
[INFO] Mimir session closed (LOCATED=0/0 TRANSFER=0/0 STORED=0/0)
[ERROR] java.lang.InterruptedException
[ERROR] 
maven mvnsh> 
```

impl/maven-cli/src/main/java/org/apache/maven/cling/invoker/mvnenc/EncryptInvoker.java
```diff
@@ -78,8 +78,8 @@ protected int execute(EncryptContext context) throws Exception {
             context.addInHeader("This tool is part of Apache Maven 4 distribution.");
             context.addInHeader("");
 
-            Thread executeThread = Thread.currentThread();
-            context.terminal.handle(Terminal.Signal.INT, signal -> executeThread.interrupt());
+            context.terminal.handle(
+                    Terminal.Signal.INT, signal -> Thread.currentThread().interrupt());
```
gnodet, line 82: That sounds weird.  The current thread cannot be the Maven execution thread, unless maybe it actually reads from stdin.  It should be a terminal's pump thread.

impl/maven-cli/src/main/java/org/apache/maven/cling/invoker/mvnsh/ShellInvoker.java
```diff
@@ -132,16 +134,28 @@ public boolean isCommandOrScript(String command) {
...
+                            systemRegistry.execute(reader.readLine(
+                                    context.cwd.get().getFileName().toString() + " mvnsh> ",
+                                    null,
+                                    (MaskingCallback) null,
+                                    null));
+                        } catch (Exception e) {
+                            failure.set(e);
+                        }
+                    });
+                    context.terminal.handle(Terminal.Signal.INT, signal -> commandThread.interrupt());
```
gnodet, line 153: Ah, yes, that one is correct :-)

impl/maven-cli/src/main/java/org/apache/maven/cling/invoker/mvnenc/EncryptInvoker.java
cstamas: mvnenc command is always single threaded, and did install this handler from beginning to handle ctrl+c while inside of ConsolePrompt (ESC is back, Ctrl+C is "abort").

gnodet (approved):

impl/maven-cli/src/main/java/org/apache/maven/cling/invoker/mvnsh/builtin/BuiltinShellCommandRegistryFactory.java
```diff
@@ -130,53 +130,133 @@ public String name() {
...
+            }
+        }
+
+        private List<Completer> cdCompleter(String name) {
+            return List.of(new ArgumentCompleter(new Completers.DirectoriesCompleter(shellContext.cwd)));
+        }
+
+        private void pwd(CommandInput input) {
+            try {
+                shellContext.writer.accept(shellContext.cwd.get().toString());
```
cstamas, line 183: Am inconsistent here: sometimes am using context.writer and sometimes context.logger... let's settle on one of these

jira-importer:
Resolve #9366

PR #2119, opened 2025-02-21
[MNG-8588] Add activated profiles to Project interface

Giovds:

 - [x] Make sure there is a [MNG-8588](https://issues.apache.org/jira/browse/MNG-8588) filed
       pulling in other changes.

### DIFF
--- a/api/maven-api-core/src/main/java/org/apache/maven/api/Project.java
+++ b/api/maven-api-core/src/main/java/org/apache/maven/api/Project.java
@@ -20,6 +20,7 @@ package org.apache.maven.api;
 
 import java.nio.file.Path;
 import java.util.List;
+import java.util.Map;
 import java.util.Optional;
 
 import org.apache.maven.api.annotations.Experimental;
@@ -237,4 +238,17 @@ public interface Project {
      */
     @Nonnull
     Optional<Project> getParent();
+
+    /**
+     * Gets the identifiers of all profiles that contributed to this project's effective model. This includes active
+     * profiles from the project's POM and all its parent POMs as well as from external sources like the
+     * {@code settings.xml}. The profile identifiers are grouped by the identifier of their source, e.g.
+     * {@code <groupId>:<artifactId>:<version>} for a POM profile or {@code external} for profiles from the
+     * {@code settings.xml}.
+     *
+     * @return The identifiers of all activated profiles, indexed by the source from which the profiles originated, never
+     *         {@code null}.
+     */
+    @Nonnull
+    Map<String, List<String>> getActivatedProfileIdsBySource();
 }
--- a/impl/maven-core/src/main/java/org/apache/maven/internal/impl/DefaultProject.java
+++ b/impl/maven-core/src/main/java/org/apache/maven/internal/impl/DefaultProject.java
@@ -23,6 +23,7 @@ import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.List;
+import java.util.Map;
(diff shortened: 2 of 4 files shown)


Giovds:
I'm trying to move the maven-help-plugin forward into a 4.0.0 release. It currently (3.x) uses a method to get the activated profiles from the project. This method is not yet available on the new interface. I think it is necessary to introduce this method in order to properly get the activated profiles from the build without needing to rebuild the project again. 

I've decided to keep the interface the same for now, but I can imagine that we would want a better representation such as a new class or making use of the Profile class.

gnodet:
@Giovds I see the tests are using `getActiveProfiles()` instead of `getInjectedProfileIds()`.
Is that because you wanted to use that method instead ?

gnodet:
> @gnodet I think it is time for weekend for me. I was adding missing tests to the `getActiveProfiles()` and ended up forgetting to add the tests for the `getInjectedProfileIds()` 🙈. I'll add those in the next commit.
> 
> Thinking out loud / alternative:
> 
> While testing though, it might be better to replace the `getInjectedProfileIds()` with `getActiveProfiles()` in the `DefaultProject` and utilise the `Locations`. We could reference the actual file used for the external profiles instead of hardcoding "external". We could also reference the line number of the active profile as well.
> 
> image
> **maven-help-plugin 3.5.1** image
> 
> Instead of ` - staged-releases (source: external)`
> 
> we could do something like ` - staged-releases (source: settings.xml)` and fall back to "external" if not present.
> 
> and ` - run-its (source: org.apache.maven.plugins:maven-help-plugin:maven-plugin:4.0.0-SNAPSHOT line 224)` if "pom".
> 
> I'd have to do some more testing to see if this would work.

If that works, that would be even better and allow a leaner API imho.

gnodet:
> > @gnodet I think it is time for weekend for me. I was adding missing tests to the `getActiveProfiles()` and ended up forgetting to add the tests for the `getInjectedProfileIds()` 🙈. I'll add those in the next commit.
> > Thinking out loud / alternative:
> > While testing though, it might be better to replace the `getInjectedProfileIds()` with `getActiveProfiles()` in the `DefaultProject` and utilise the `Locations`. We could reference the actual file used for the external profiles instead of hardcoding "external". We could also reference the line number of the active profile as well.
> > image
> > **maven-help-plugin 3.5.1** image
> > Instead of ` - staged-releases (source: external)`
> > we could do something like ` - staged-releases (source: settings.xml)` and fall back to "external" if not present.
> > and ` - run-its (source: org.apache.maven.plugins:maven-help-plugin:maven-plugin:4.0.0-SNAPSHOT line 224)` if "pom".
> > I'd have to do some more testing to see if this would work.
> 
> If that works, that would be even better and allow a leaner API imho.

We should definitely try it instead.  Adding `getActiveProfiles()` to `Project` should be sufficient to cover both mojos usage I think.

gnodet:
Btw, are you working on top of https://github.com/apache/maven-help-plugin/pull/121 or did you start from scratch ?

Giovds:
> Btw, are you working on top of https://github.com/apache/maven-help-plugin/pull/121 or did you start from scratch ?

Yes, I took the mvn4 branch as baseline. Did have to do some changes after bumping to beta-5, and I’m not sure if it is going to be comparable with the rc releases. 

> We should definitely try it instead. Adding getActiveProfiles() to Project should be sufficient to cover both mojos usage I think.

I believe so too!

jira-importer:
Resolve #9367

PR #2110, opened 2025-02-14
[MNG-8583] Missing SISU bean definitions from maven-resolver-provider

gnodet:
JIRA issue: [MNG-8583](https://issues.apache.org/jira/browse/MNG-8583)

The beans are not recognized by Sisu as the needed META-INF file is missing.

PR #2123, opened 2025-02-24
[MNG-8590] Make sure maven-resolver-provider does not depend on maven-impl

gnodet:
JIRA issue: [MNG-8590](https://issues.apache.org/jira/browse/MNG-8590)

In order to be able to use the maven-resolver-provider as it used to be (see [MNG-8583](https://issues.apache.org/jira/browse/MNG-8583)), we need to break the dependency from `maven-resolver-provider` to `maven-impl` so that we don't have duplicate bean definitions.

Related to #2110


cstamas (approved):

jira-importer:
Resolve #9855

PR #2125, opened 2025-02-25
[MNG-8592] Package Sisu enabled JAR separately

cstamas:
We create additional JAR that is very same to main JAR but has sisu index present.

---

https://issues.apache.org/jira/browse/MNG-8592


gnodet (approved):
LGTM

jira-importer:
Resolve #9856

PR #2126, opened 2025-02-25
Minor Javadoc cleanup

elharo:


cstamas:
You are aware you are editing `compat` subproject? You know, a subproject that should be left as is (ie. "frozen") and is there only for backward compatibility reasons?

elharo:
Also worth noting: ComparableVersion is public API used by code that is not in Maven and not part of a Maven plugin. We should not just abandon it.

cstamas:
Right, we demoted it to "compat" (with Maven 3) libraries, probably deprecated it as well (or will soon). Noting in Maven uses these classes anymore.

OTOH. by not "just leaving" these sources, you may raise by some bystander the (wrong) assumption that we do "develop" these libs, and just increase the already chaotic situation and confusion (as it was in Maven 3, multiple Version implementations and so on).

gnodet:
> One more time: ComparableVersion is used by third party, non-Maven code and has direct usages as well (i.e. not as a library but as a program in its own right). Furthermore, there's at least one committer (me) who's actively maintaining it. It is not deprecated and should not be deprecated. If you don't want to work on it, that's fine. No one has time to work on everything. But please don't block work on useful parts of the project.

Sure, the point is that, if it's not used in Maven anymore (and that's the case for all classes in compat/* jars), and that they need to be maintained, they'd be better moved into some maven-shared jar instead.

compat/maven-artifact/src/main/java/org/apache/maven/artifact/versioning/ComparableVersion.java
```diff
@@ -36,25 +36,25 @@
...
+ * <li>Version components in the text can be digits or strings,</li>
+ * <li>Strings are checked for well-known qualifiers, and the qualifier ordering is used for version ordering.
  *     Well-known qualifiers (case-insensitive) are:<ul>
  *     <li><code>alpha</code> or <code>a</code></li>
  *     <li><code>beta</code> or <code>b</code></li>
  *     <li><code>milestone</code> or <code>m</code></li>
  *     <li><code>rc</code> or <code>cr</code></li>
  *     <li><code>snapshot</code></li>
- *     <li><code>(the empty string)</code> or <code>ga</code> or <code>final</code></li>
+ *     <li><code>ga</code> or <code>final</code></li>
```
gnodet, line 51: Why removing the empty string ? That sounds like useful info wrt ordering.

gnodet: This indicates that when sorting, no qualifier will be sorted at the same level than `ga` or `final`.  I would not remove this bit.

elharo: The empty string is not a qualifier. It's an absence of a qualifier. Not having a qualifier sorts the same as having a qualifier of "ga" or "final" buit I have trouble including it in a list of qualifiers.

elharo: done

gnodet (approved):

PR #2131, opened 2025-02-26
[MNG-8594] Add atFile option

cstamas:
Where user can create ad-hoc command line parms. The difference between .mvn/maven.conf and this file is that this file allows goals as well. The CLI and atFile are merged, in this order, hence, goals in atFile will come AFTER goals specified on CLI, if CLI has them.

Source order (first wins for options, while goals are collected/aggregated from sources):
* CLI
* atFile (if specified, may have goals that are appended)
* maven.config (if present, may not have goals)

Note: option is called `af` or long `at-file` for similarity with `javac @file`, but commons CLI does not support options without leading hyphen.

---

https://issues.apache.org/jira/browse/MNG-8594


gnodet:
Looks good, that can be handy !
Any way to add a small UT or IT maybe ?

cstamas:
@gnodet Added IT, that revealed precedence issue in case of User Properties (`-D`) that was present for maven.config as well, fixed.

gnodet (commented):
LGTM

jira-importer:
Resolve #9852

PR #2132, opened 2025-02-26
[MNG-8588] Add activated profiles to Project interface

gnodet:
- **[MNG-8588] Add activated profiles to Project interface**
- **[MNG-8588] Add tests for getInjectedProfileIds()**
- **Define getActiveProfiles() instead**
- **Add getProfiles methods and fix location tracking on settings**

### DIFF
--- a/api/maven-api-core/src/main/java/org/apache/maven/api/Project.java
+++ b/api/maven-api-core/src/main/java/org/apache/maven/api/Project.java
@@ -20,13 +20,13 @@ package org.apache.maven.api;
 
 import java.nio.file.Path;
 import java.util.List;
-import java.util.Map;
 import java.util.Optional;
 
 import org.apache.maven.api.annotations.Experimental;
 import org.apache.maven.api.annotations.Nonnull;
 import org.apache.maven.api.model.Build;
 import org.apache.maven.api.model.Model;
+import org.apache.maven.api.model.Profile;
 
 /**
  * Interface representing a Maven project which can be created using the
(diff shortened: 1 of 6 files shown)


gnodet:
@Giovds I went ahead and added the `get[All][Active]Profiles()` methods on `Project`.
I've locally updated `m-help-p` to use those and had to fix location tracking in settings for the real source to be displayed (instead of just `pom`, `external` or `settings.xml`).
Next step would be to actually remove the `source` field from `Profile`...

impl/maven-core/src/main/java/org/apache/maven/internal/impl/DefaultProject.java
```diff
@@ -165,6 +166,46 @@ public Optional<Project> getParent() {
...
+    @Nonnull
+    public List<Profile> getAllProfiles() {
+        List<org.apache.maven.model.Profile> activeProfiles = new ArrayList<>();
+        for (MavenProject project = this.project; project != null; project = project.getParent()) {
+            activeProfiles.addAll(project.getModel().getProfiles());
+        }
+        return activeProfiles.stream()
+                .map(org.apache.maven.model.Profile::getDelegate)
+                .toList();
+    }
```
Giovds, line 185:
```suggestion
    public List<Profile> getAllProfiles() {
        List<org.apache.maven.model.Profile> profiles = new ArrayList<>();
        for (MavenProject project = this.project; project != null; project = project.getParent()) {
            profiles.addAll(project.getModel().getProfiles());
        }
        return profiles.stream()
                .map(org.apache.maven.model.Profile::getDelegate)
                .toList();
    }
```

api/maven-api-core/src/main/java/org/apache/maven/api/Project.java
```diff
@@ -237,4 +238,68 @@ default String getId() {
...
+     * This method traverses the parent hierarchy and includes profiles defined in parent POMs.
+     * The returned list contains profiles from the current project and all of its ancestors in
+     * the project inheritance chain.
+     *
+     * @return a non-null, possibly empty list of all profiles from this project and its parents
+     * @see Profile
+     * @see #getProfiles()
+     */
+    @Nonnull
+    List<Profile> getAllProfiles();
```
Giovds, line 267: I'm trying to think of a different name for the `getAll(Active)Profiles()` methods, as it might be interpreted the same as `getProfiles()`. I was thinking something alone the lines of `getAll(Active)ProfilesIncludingInherited()`, or rename `get(Active)Profiles()` to something like `getCurrent(Active)Profiles()`

impl/maven-core/src/main/java/org/apache/maven/internal/impl/DefaultProject.java
```diff
@@ -165,6 +166,46 @@ public Optional<Project> getParent() {
...
+    public List<Profile> getAllActiveProfiles() {
+        List<org.apache.maven.model.Profile> activeProfiles = new ArrayList<>();
+        for (MavenProject project = this.project; project != null; project = project.getParent()) {
+            activeProfiles.addAll(project.getActiveProfiles());
+        }
+        return activeProfiles.stream()
+                .map(org.apache.maven.model.Profile::getDelegate)
+                .toList();
+    }
+
```
Giovds, line 208: You could make the methods a one liner, if you'd like with something like this I think:

```java
    return Stream.iterate(this.project, Objects::nonNull, MavenProject::getParent)
                 .flatMap(project -> project.getActiveProfiles().stream())
                 .map(org.apache.maven.model.Profile::getDelegate)
                 .toList();
```

api/maven-api-core/src/main/java/org/apache/maven/api/Project.java
gnodet: What about:
* `getDeclaredProfiles()` / `getEffectiveProfiles()`
* `getDeclaredActiveProfiles()` / `getEffectiveActiveProfiles()`
The `declared` would make reference to only profiles on the current project, while `effective` would make reference to profiles from the project and it's hierarchy, similar to the `effective` model which is computed with the parents merged into it.

cstamas (approved):

jira-importer:
Resolve #9367

PR #2136, opened 2025-03-01
[MNG-8601] Add type test-java-source type

cstamas:
Similar to java-source type, but for tests sources.

---

https://issues.apache.org/jira/browse/MNG-8601


gnodet (approved):

jira-importer:
Resolve #9845

PR #2139, opened 2025-03-03
[MNG-8602] Fix BOM consumer pom

gnodet:
JIRA issue: [MNG-8602](https://issues.apache.org/jira/browse/MNG-8602)

### DIFF
--- a/impl/maven-core/src/main/java/org/apache/maven/internal/transformation/impl/DefaultConsumerPomBuilder.java
+++ b/impl/maven-core/src/main/java/org/apache/maven/internal/transformation/impl/DefaultConsumerPomBuilder.java
@@ -22,7 +22,6 @@ import javax.inject.Inject;
 import javax.inject.Named;
 
 import java.nio.file.Path;
-import java.util.ArrayList;
 import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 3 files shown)


impl/maven-core/src/main/java/org/apache/maven/internal/transformation/impl/DefaultConsumerPomBuilder.java
```diff
@@ -193,72 +202,72 @@ private ModelBuilderResult buildModel(RepositorySystemSession session, Path src)
...
+
+    static Model transformBom(Model model, MavenProject project) {
+        boolean preserveModelVersion = model.isPreserveModelVersion();
+
+        Model.Builder builder = prune(
+                Model.newBuilder(model, true)
+                        .preserveModelVersion(false)
+                        .root(false)
+                        .parent(null)
+                        // .dependencyManagement(dependencyManagement.withDependencies(dependencies))
```
cstamas, line 244: why was this added at all? Was it envisioned that BOM may contain only reactor deps? As then we may want to enforce that

gnodet: See https://issues.apache.org/jira/browse/MNG-7920

gnodet: The problem is that the consumer POM for the BOM was originally created like a POM, i.e. from the raw model, and it was missing some versions for managed dependencies.  The fix was overwriting the dependency version with the model version.  This is now correctly implemented by using the effective model instead of the raw model.

PR #2144, opened 2025-03-04
[experiment] Mimir

cstamas:
Use Maveniverse Mimir to cache central artifacts. This makes caching way simpler as well. Update Maven used to build to Maven 4.0.0-rc-3 as well.

### DIFF
--- a/.github/ci-extensions.xml
+++ b/.github/ci-extensions.xml
@@ -21,6 +21,6 @@ under the License.
   <extension>
     <groupId>eu.maveniverse.maven.mimir</groupId>
     <artifactId>extension3</artifactId>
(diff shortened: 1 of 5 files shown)


cstamas:
> bit unfortunate name https://grafana.com/oss/mimir/

Am sure this name is (re)usable :smile: https://en.wikipedia.org/wiki/M%C3%ADmir

cstamas:
The reason for this experiment is this:
https://github.com/orgs/community/discussions/145314

We regularly see macOS runners (only!!!) failing to get to central, unsure why

gnodet (approved):
LGTM

PR #2145, opened 2025-03-04
[MNG-8599] Fix configuration for maven-source-plugin

slawekjaranowski:
There is no sources parameter for maven-source-plugin

added in #2133

sources.jar still contains generated code 😄


cstamas (commented):
oh, dang

gnodet (approved):

PR #2160, opened 2025-03-18
Replace a `NullPointerException` by a more explicit error message

desruisseaux:
`Session.requireDependencyScope(…)` is annotated as `@Nonnull` but its implementation can still return null, causing `NullPointerException` later. This pull request replace the null value by an `IllegalArgumentException` that describes the problem.


desruisseaux:
Build failures on MacOS are the usual "Could not transfer artifact org.apache.maven.surefire:maven-surefire-common:pom:3.5.2 from/to central (https://repo.maven.apache.org/maven2): HTTP connect timed out".

gnodet (approved):

PR #2154, opened 2025-03-14
[IT] Related fixes

cstamas:
Changes:
* untangle overly complex helper/tool
* make sure tool cannot return "" (empty string), add "output validation"
* make ITs use mimir (IT settings redefined Central in a way Mimir did not wanted to handle it)

PR #2164, opened 2025-03-19
Fix ITs referencing rc-3-SNAPSHOT

gnodet:
This should fix ITs being broken in PRs


cstamas (approved):
FYI is included in https://github.com/apache/maven/pull/2154 as well

PR #2176, opened 2025-03-21
[MNG-8646] Make IT tests use "outer" Mimir caching as well

cstamas:
As it turned out, ITs were not using Mimir as they redefined Maven Central to allow snapshots, and that made Mimir not handle the IT used Maven Central (while those did not contain snapshots, it was still pointing at canonical Maven Central URL).

Fixing this revealed several other (related or unrelated) issues:
* MavenProject was NPE prone regarding remote repositories handling
* MappedList was NPE prone regarding null values (emitted by MavenProject remote repositories)
several ProjectModelBuilder (detailed below) had no remote repositories set anymore (as central is not in super POM anymore)
* some ITs revaled ReactorReader issues as well

The `org.apache.maven.project.DefaultProjectBuilder issues` fixed:
* model request type CONSUMER_PARENT is used for non-locally built projects (ie. plugins invoked with direct mojo invocation), this seems strange. This skips CI friendly replacement for example as shows in Maven ITs
* remote repositories are injected ONLY when injecting lifecycle plugins (processPlugins=true). This was ok when root pom was defining default repository, but that's not the case anymore, and this results in projects built that has 0 remote reposes.

At the end of the day, all issues were fixed but one IT that we declared "invalid" was disabled (w/ remark about why's).

---

https://issues.apache.org/jira/browse/MNG-8646

### DIFF
--- a/api/maven-api-core/src/main/java/org/apache/maven/api/services/ModelBuilderRequest.java
+++ b/api/maven-api-core/src/main/java/org/apache/maven/api/services/ModelBuilderRequest.java
@@ -59,7 +59,7 @@ public interface ModelBuilderRequest extends Request<Session> {
         BUILD_EFFECTIVE,
         /**
          * The request is used specifically to parse the POM used as a basis for creating the consumer POM.
-         * This POM will not ungergo any profile activation.
+         * This POM will not undergo any profile activation.
          */
         BUILD_CONSUMER,
         /**
--- a/impl/maven-core/src/main/java/org/apache/maven/ReactorReader.java
+++ b/impl/maven-core/src/main/java/org/apache/maven/ReactorReader.java
@@ -98,7 +98,7 @@ class ReactorReader implements MavenWorkspaceReader {
     }
 
     public File findArtifact(Artifact artifact) {
-        MavenProject project = getProject(artifact, true);
+        MavenProject project = getProject(artifact);
 
         if (project != null) {
             File file = findArtifact(project, artifact, true);
@@ -140,7 +140,7 @@ class ReactorReader implements MavenWorkspaceReader {
 
     @Override
     public Model findModel(Artifact artifact) {
-        MavenProject project = getProject(artifact, false);
+        MavenProject project = getProject(artifact);
         return project == null ? null : project.getModel().getDelegate();
     }
 
@@ -470,8 +470,8 @@ class ReactorReader implements MavenWorkspaceReader {
         return projectLocalRepository;
     }
 
-    private MavenProject getProject(Artifact artifact, boolean includeAllProjects) {
-        return (includeAllProjects ? getAllProjects() : getProjects())
+    private MavenProject getProject(Artifact artifact) {
+        return getProjects()
                 .getOrDefault(artifact.getGroupId(), Collections.emptyMap())
                 .getOrDefault(artifact.getArtifactId(), Collections.emptyMap())
                 .getOrDefault(artifact.getBaseVersion(), null);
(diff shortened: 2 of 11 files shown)


cstamas:
Heuristic:
```
                    ModelBuilderRequest.RequestType type = pomFile != null
                                    && this.request.isProcessPlugins()
                                    && this.request.getValidationLevel() == ModelBuildingRequest.VALIDATION_LEVEL_STRICT
                            ? ModelBuilderRequest.RequestType.BUILD_EFFECTIVE
                            : ModelBuilderRequest.RequestType.CONSUMER_PARENT;
```

Am unsure why "process plugins" matters here, I'd rather go: `pomFile != null` (so is checkout, so to say "current project") then assume build effective, and if no file present, then assume is "remote/external".

cstamas:
Formatting. But the change is too "big" IMO.
Maybe just "if pom being built is from reactor, do not build consumer?". Still, that means the process plugins value is ignored?

gnodet:
> Formatting. But the change is too "big" IMO. Maybe just "if pom being built is from reactor, do not build consumer?". Still, that means the process plugins value is ignored?

So m-assembly-p does [the following](https://github.com/apache/maven-assembly-plugin/blob/5c324065f129e4803196bdfedb6102c730a703b5/src/main/java/org/apache/maven/plugins/assembly/archive/task/AddDependencySetsTask.java#L169-L171)
```java
        return new DefaultProjectBuildingRequest(configSource.getMavenSession().getProjectBuildingRequest())
                .setValidationLevel(ModelBuildingRequest.VALIDATION_LEVEL_MINIMAL)
                .setProcessPlugins(false);
```
so we need to remove both conditions, else it will have no effect.

The actual problem is that there's no different made by m-assembly-p between loading a pom from the repository (as a consumer) and loading from the projects being built.  Maybe that's the one that needs to be fixed ?

gnodet:
> > Formatting. But the change is too "big" IMO. Maybe just "if pom being built is from reactor, do not build consumer?". Still, that means the process plugins value is ignored?
> 
> So m-assembly-p does [the following](https://github.com/apache/maven-assembly-plugin/blob/5c324065f129e4803196bdfedb6102c730a703b5/src/main/java/org/apache/maven/plugins/assembly/archive/task/AddDependencySetsTask.java#L169-L171)
> 
> ```java
>         return new DefaultProjectBuildingRequest(configSource.getMavenSession().getProjectBuildingRequest())
>                 .setValidationLevel(ModelBuildingRequest.VALIDATION_LEVEL_MINIMAL)
>                 .setProcessPlugins(false);
> ```
> 
> so we need to remove both conditions, else it will have no effect.
> 
> The actual problem is that there's no different made by m-assembly-p between loading a pom from the repository (as a consumer) and loading from the projects being built. Maybe that's the one that needs to be fixed ?

I'm going to experiment with https://github.com/apache/maven-assembly-plugin/pull/217

gnodet:
> > > Formatting. But the change is too "big" IMO. Maybe just "if pom being built is from reactor, do not build consumer?". Still, that means the process plugins value is ignored?
> > 
> > 
> > So m-assembly-p does [the following](https://github.com/apache/maven-assembly-plugin/blob/5c324065f129e4803196bdfedb6102c730a703b5/src/main/java/org/apache/maven/plugins/assembly/archive/task/AddDependencySetsTask.java#L169-L171)
> > ```java
> >         return new DefaultProjectBuildingRequest(configSource.getMavenSession().getProjectBuildingRequest())
> >                 .setValidationLevel(ModelBuildingRequest.VALIDATION_LEVEL_MINIMAL)
> >                 .setProcessPlugins(false);
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > so we need to remove both conditions, else it will have no effect.
> > The actual problem is that there's no different made by m-assembly-p between loading a pom from the repository (as a consumer) and loading from the projects being built. Maybe that's the one that needs to be fixed ?
> 
> I'm going to experiment with [apache/maven-assembly-plugin#217](https://github.com/apache/maven-assembly-plugin/pull/217)

It seems to work locally for me.  @cstamas could you double check when you have some time ?

cstamas:
I dislike it. It means we need to modify _legacy_ code base to make it work with Maven 4? Are we sure we cannot make it work by improving the heuristic instead?

gnodet (approved):

jira-importer:
Resolve #9359

PR #2180, opened 2025-03-24
[MNG-8647] Set the default source directory to `src/${scope}/${lang}` as per documentation

desruisseaux:
JIRA issue: [MNG-8647](https://issues.apache.org/jira/browse/MNG-8647)

The `maven.mdo` file for the `<Source>` element said that the default directories are `src/${scope}/${lang}`. but this is effective only if no source is specified at all. It would be helpful to have this default value when some `<Source>` are specified. The intend is to allow users to write:

```xml
  <build>
    <sources>
      <source>
        <targetVersion>21</targetVersion>
      </source>
      <source>
        <scope>test</scope>
      </source>
    </sources>
</build>
```

instead of:

```xml
  <build>
    <sources>
      <source>
        <directory>src/main/java</directory>
        <targetVersion>21</targetVersion>
      </source>
      <source>
        <scope>test</scope>
        <directory>src/test/java</directory>
      </source>
    </sources>
</build>
```

### DIFF
--- a/api/maven-api-core/src/main/java/org/apache/maven/api/SourceRoot.java
+++ b/api/maven-api-core/src/main/java/org/apache/maven/api/SourceRoot.java
@@ -28,8 +28,9 @@ import java.util.Optional;
  * The sources may be Java main classes, test classes, resources or anything else identified by the scope.
  *
  * <h2>Default values</h2>
- * The default implementation of all methods in this interface match the default values
- * documented in {@code maven.mdo} (the <abbr>POM</abbr> model).
+ * The properties in this interface are defined in the {@code <Source>} element of the
+ * {@linkplain org.apache.maven.api.model.Model Maven project descriptor}.
+ * For each property, the default value is either empty or documented in the project descriptor.
  */
 public interface SourceRoot {
     /**
@@ -64,7 +65,10 @@ public interface SourceRoot {
 
     /**
      * {@return in which context the source files will be used}.
-     * The default value is {@link ProjectScope#MAIN}.
+     * Not to be confused with dependency scope.
+     * The default value is {@code "main"}.
+     *
+     * @see ProjectScope#MAIN
      */
     default ProjectScope scope() {
         return ProjectScope.MAIN;
@@ -72,7 +76,9 @@ public interface SourceRoot {
 
     /**
      * {@return the language of the source files}.
-     * The default value is {@link Language#JAVA_FAMILY}.
+     * The default value is {@code "java"}.
+     *
+     * @see Language#JAVA_FAMILY
      */
     default Language language() {
         return Language.JAVA_FAMILY;
--- a/api/maven-api-core/src/main/java/org/apache/maven/api/SourceRoot.java
+++ b/api/maven-api-core/src/main/java/org/apache/maven/api/SourceRoot.java
@@ -26,13 +26,23 @@ import java.util.Optional;
 /**
  * A root directory of source files.
  * The sources may be Java main classes, test classes, resources or anything else identified by the scope.
+ *
+ * <h2>Default values</h2>
+ * The default implementation of all methods in this interface match the default values
+ * documented in {@code maven.mdo} (the <abbr>POM</abbr> model).
  */
 public interface SourceRoot {
     /**
      * {@return the root directory where the sources are stored}.
      * The path is relative to the <abbr>POM</abbr> file.
+     *
+     * <h4>Default implementation</h4>
+     * The default value is <code>src/{@linkplain #scope() scope}/{@linkplain #language() language}</code>
+     * as a relative path. Implementation classes may override this default with an absolute path instead.
      */
-    Path directory();
+    default Path directory() {
+        return Path.of("src", scope().id(), language().id());
+    }
 
     /**
      * {@return the list of pattern matchers for the files to include}.
@@ -62,8 +72,11 @@ public interface SourceRoot {
 
     /**
      * {@return the language of the source files}.
+     * The default value is {@link Language#JAVA_FAMILY}.
      */
-    Language language();
+    default Language language() {
+        return Language.JAVA_FAMILY;
+    }
 
     /**
      * {@return the name of the Java module (or other language-specific module) which is built by the sources}.
--- a/api/maven-api-model/src/main/mdo/maven.mdo
+++ b/api/maven-api-model/src/main/mdo/maven.mdo
@@ -875,7 +875,7 @@
           <description>
             All the sources to compile and resources files to copy for a project or it's unit tests.
             The sources can be Java source files, generated source files, scripts or resources for examples.
-            Each source is specified by a mandatory {@code directory} element, which is relative to the POM.
+            Each source is specified by a {@code directory} element, which is relative to the POM.
             The kind of sources (codes to compile or resources to copy) and their usage (for the main code
             or for the tests) is specified by the {@code scope} element together with each source directory.
           </description>
@@ -2005,7 +2005,7 @@
         <![CDATA[
         Description of the sources associated with a project main code or unit tests.
         The sources can be Java source files, generated source files, scripts or resources for examples.
-        A source is specified by a mandatory {@code directory} element, which is relative to the POM.
+        A source is specified by a {@code directory} element, which is relative to the POM.
         The directory content can optionally by reduced to a subset with the {@code includes} and
         {@code excludes} elements. The kind of sources (codes, resources, <i>etc.</i>) and their
         usage (main code, tests, <i>etc.</i>) is specified by the {@code scope} element.
--- a/api/maven-api-model/src/main/mdo/maven.mdo
+++ b/api/maven-api-model/src/main/mdo/maven.mdo
@@ -874,9 +874,9 @@
           <version>4.1.0+</version>
           <description>
             All the sources to compile and resources files to copy for a project or it's unit tests.
-            The sources can be Java source files, generated source files, scripts or resources for examples.
+            The sources can be Java source files, generated source files, scripts, or resources for examples.
             Each source is specified by a {@code directory} element, which is relative to the POM.
-            The kind of sources (codes to compile or resources to copy) and their usage (for the main code
+            The kind of sources (source files to compile or resources to copy) and their usage (for the main code
             or for the tests) is specified by the {@code scope} element together with each source directory.
           </description>
           <association>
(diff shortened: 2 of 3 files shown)


api/maven-api-core/src/main/java/org/apache/maven/api/SourceRoot.java
```diff
@@ -26,13 +26,23 @@
 /**
  * A root directory of source files.
  * The sources may be Java main classes, test classes, resources or anything else identified by the scope.
+ *
+ * <h2>Default values</h2>
+ * The default implementation of all methods in this interface match the default values
+ * documented in {@code maven.mdo} (the <abbr>POM</abbr> model).
```
elharo, line 32: I don't quite follow this, might need a bit of rewrite to clarify

api/maven-api-core/src/main/java/org/apache/maven/api/SourceRoot.java
```diff
@@ -26,13 +26,23 @@
...
  */
 public interface SourceRoot {
     /**
      * {@return the root directory where the sources are stored}.
      * The path is relative to the <abbr>POM</abbr> file.
+     *
+     * <h4>Default implementation</h4>
+     * The default value is <code>src/{@linkplain #scope() scope}/{@linkplain #language() language}</code>
```
elharo, line 40: main is not a scope. Is that OK?

api/maven-api-core/src/main/java/org/apache/maven/api/SourceRoot.java
```diff
@@ -62,8 +72,11 @@ default ProjectScope scope() {
 
     /**
      * {@return the language of the source files}.
+     * The default value is {@link Language#JAVA_FAMILY}.
```
elharo, line 75: Just use the constant value here. No need to hide this behind a variable in the docs.

api/maven-api-model/src/main/mdo/maven.mdo
```diff
@@ -875,7 +875,7 @@
           <description>
             All the sources to compile and resources files to copy for a project or it's unit tests.
             The sources can be Java source files, generated source files, scripts or resources for examples.
-            Each source is specified by a mandatory {@code directory} element, which is relative to the POM.
+            Each source is specified by a {@code directory} element, which is relative to the POM.
             The kind of sources (codes to compile or resources to copy) and their usage (for the main code
```
elharo, line 879: codes --> code or perhaps source files

api/maven-api-model/src/main/mdo/maven.mdo
```diff
@@ -2005,7 +2005,7 @@
         <![CDATA[
         Description of the sources associated with a project main code or unit tests.
         The sources can be Java source files, generated source files, scripts or resources for examples.
```
elharo, line 2007: scripts,

api/maven-api-model/src/main/mdo/maven.mdo
```diff
@@ -2005,7 +2005,7 @@
...
-        A source is specified by a mandatory {@code directory} element, which is relative to the POM.
+        A source is specified by a {@code directory} element, which is relative to the POM.
```
elharo, line 2008: This seems confusingly similar to source files.

api/maven-api-model/src/main/mdo/maven.mdo
```diff
@@ -2005,7 +2005,7 @@
...
         The directory content can optionally by reduced to a subset with the {@code includes} and
```
elharo, line 2009: by --> be

impl/maven-impl/src/main/java/org/apache/maven/impl/DefaultSourceRoot.java
```diff
@@ -65,24 +65,26 @@ public final class DefaultSourceRoot implements SourceRoot {
...
         excludes = matchers(fs, source.getExcludes());
         stringFiltering = source.isStringFiltering();
         enabled = source.isEnabled();
         moduleName = nonBlank(source.getModule());
 
-        value = nonBlank(source.getScope());
+        String value = nonBlank(source.getScope());
         scope = (value != null) ? session.requireProjectScope(value) : ProjectScope.MAIN;
 
         value = nonBlank(source.getLang());
```
elharo, line 78: We should not be reusing these local variables for different things

elharo (changes requested):

api/maven-api-core/src/main/java/org/apache/maven/api/SourceRoot.java
desruisseaux: Is the following better?

> The properties in this interface are defined in the `<Source>` element of the [Maven project descriptor 4.1](https://maven.apache.org/xsd/maven-4.1.0.xsd). For each property, the default value is either empty or documented in the project descriptor. The default implementations of the methods provided in this interface are consistent with that specification.

elharo: Can we link to something easier to read than the XSD?
Consider deleting "The default implementations of the methods provided in this interface are consistent with that specification." as that's implied by the previous sentence

api/maven-api-core/src/main/java/org/apache/maven/api/SourceRoot.java
desruisseaux: Currently, this is called "scope" with two recognized values: `main` or `test`. I agree that there is a risk of confusion with dependency scopes. However, the "scope" term is already used in different contexts. Maven core currently has:

* `ProjectScope` with `main` and `test` (the scopes in question above).
* `DependencyScope` with `compile`, `runtime`, `provided`, `test`, etc.
* `PathScope` with `main-compile`, `main-runtime`, `test-compile` and `test-runtime`.
*  `Scope` for lifecycle and visibility of objects in the dependency injection system.
*  `SessionScope` (internal)
* `MojoExecutionScope` (internal)

I agree that users would see only dependency scopes and project scopes. We could rename the latter. However, we could also said that a scope is like `name` or a `id` properties: they are used for different kinds of objects, and their meaning depend on the object where they appear.

api/maven-api-core/src/main/java/org/apache/maven/api/SourceRoot.java
desruisseaux: I will put a link to https://maven.apache.org/ref/current/maven-model/maven.html on the assumption that `current` will redirect to version 4.1 some day.

impl/maven-impl/src/main/java/org/apache/maven/impl/DefaultSourceRoot.java
desruisseaux: This is often true. But in this particular case, I have some difficulties to find a nice alternative. Multiplying the number of variables also increases the risk that we use the wrong ones. In the current code, the `value` variable is used each time for initializing a final field, then immediately assigned the next value, with each block separated by a blank line. There is no much room for confusion.

elharo: It's not new in this PR so it's not a blocker, but I don't agree this is OK and I wouldn't have approved it if this were new code. Each could simply be named based on the source field it's derived from

elharo (approved):

api/maven-api-core/src/main/java/org/apache/maven/api/SourceRoot.java
gnodet: What about linking to the model directly ? Everything in the xsd should be in the javadoc too ?
So maybe `{@link org.apache.maven.api.model.Model} ?

impl/maven-impl/src/main/java/org/apache/maven/impl/DefaultSourceRoot.java
gnodet: Or we need to relax the fact that we disallow inner blocks.  Those provide a scoping for local variables, so I don't think we should forbid both inner scopes _and_ reusing variables.

gnodet (approved):

impl/maven-impl/src/main/java/org/apache/maven/impl/DefaultSourceRoot.java
desruisseaux: > Or we need to relax the fact that we disallow inner blocks.

Yes exactly. I thought about relaxing that rule just for this method with a `@SuppressWarnings("checkstyle:...")`, but thought the maybe the community wanted strict adherence to the checkstyle rules.

api/maven-api-core/src/main/java/org/apache/maven/api/SourceRoot.java
desruisseaux: True, I will do that.

Actually, I didn't realized that these generated classes where in public API. Since they are, shouldn't we go ahead with the use of `record` before GA, or is it too late in 4.0 release candidates for that?

impl/maven-impl/src/main/java/org/apache/maven/impl/DefaultSourceRoot.java
gnodet: I'm fine with local relaxation.

api/maven-api-core/src/main/java/org/apache/maven/api/SourceRoot.java
gnodet: I don't think that's possible, the records cannot inherit from another record.

desruisseaux:
I rebased the pull request on the latest commit of the main branch. Just realized that it caused GitHub to said that there is a lot of changes, but actually the only changes were in the link to the model and in the addition of inner blocks.

impl/maven-impl/src/main/java/org/apache/maven/impl/DefaultSourceRoot.java
```diff
@@ -64,30 +64,38 @@ public final class DefaultSourceRoot implements SourceRoot {
...
-
-        value = nonBlank(source.getLang());
-        language = (value != null) ? session.requireLanguage(value) : Language.JAVA_FAMILY;
-
-        value = nonBlank(source.getTargetVersion());
-        targetVersion = (value != null) ? session.parseVersion(value) : null;
-
-        value = nonBlank(source.getTargetPath());
-        targetPath = (value != null) ? baseDir.resolve(value) : null;
+        {
```
elharo, line 75: No. I am going to -1 this. This is worse than what we had before. Just rename all the variables and don't use the word value, which was a non-descriptive name in the first place, even if it were only used in one place.

elharo (changes requested):

impl/maven-impl/src/main/java/org/apache/maven/impl/DefaultSourceRoot.java
desruisseaux: Reverted to the previous version.

api/maven-api-core/src/main/java/org/apache/maven/api/SourceRoot.java
gnodet: However, we should change to use record-style getters (so without the `get` prefix).  But I would generate both for now and deprecate the `getXxx()` for removal in 4.1.0.

elharo (approved):

PR #2181, opened 2025-03-24
requireNonNull

elharo:
Let's see if this breaks anything....


gnodet:
Why not using a static import ? `requireNonNull` is quite unambiguous...

elharo:
Only because the IntelliJ inline refactoring didn't use a static import. If the tests pass I can do that and a couple of other things to clean this up.

desruisseaux (approved):

PR #2183, opened 2025-03-25
[MNG-8645] Fix dependency management section of consumer POM

gnodet:
JIRA issue: [MNG-8645](https://issues.apache.org/jira/browse/MNG-8645)

The implicit call to Type#toString() instead of Type#id() causes the wrong string id to generated, thus causing the code to not find matches nodes and discarding all managed dependencies.

I'll try to add an IT later.


slawekjaranowski (approved):
tested on m-jar-p - looks ok

jira-importer:
Resolve #9663

PR #2184, opened 2025-03-25
[MNG-8075] Deprecate project.baseUri in favor of project.basedir.uri

gnodet:
JIRA issue: [MNG-8075](https://issues.apache.org/jira/browse/MNG-8075)

The original issue was adding support for `${project.baseUri}`, however, it has been inadvertently dropped when refactoring the Maven 4 model builder.  Overall, instead of providing a new property, I think using the newly supported syntax of `${project.basedir.uri}` is a better idea.

This PR isn't really complete as it only deprecates the expression in repository URIs for now.


cstamas (approved):

jira-importer:
Resolve #9801

PR #2188, opened 2025-03-26
Remove redundant code

elharo:


compat/maven-compat/src/main/java/org/apache/maven/repository/metadata/MetadataGraph.java
```diff
@@ -128,9 +128,7 @@ private void processTreeNodes(MetadataGraphVertex parentVertex, MetadataTreeNode
         }
 
         MetadataGraphVertex vertex = new MetadataGraphVertex(node.md, versionedVertices, scopedVertices);
-        if (!vertices.contains(vertex)) {
-            vertices.add(vertex);
-        }
+        vertices.add(vertex);
```
gnodet, line 131: THat's a non trivial change, is this semantic change on purpose ?

elharo:
I had to look twice at that one too. Vertices is a set. Adding the same
thing twice is a noop.

Elliotte Rusty Harold
***@***.***

gnodet (approved):

PR #2195, opened 2025-03-27
[MNG-5668] Execute after:* phases when build fails

gnodet:
JIRA issue: [MNG-5668](https://issues.apache.org/jira/browse/MNG-5668)

When a build step fails, Maven should still execute its corresponding after:* phases to ensure proper cleanup. This fix modifies the BuildPlanExecutor to:

- Execute after:* phases when their corresponding before:* phase has been executed
- Maintain proper phase ordering during failure handling
- Handle cleanup phase failures gracefully without affecting the original error
- Preserve concurrent build capabilities

This ensures cleanup tasks (like resource cleanup or test environment teardown) are properly executed even when the build fails.


cstamas (approved):

jira-importer:
Resolve #6656

PR #2197, opened 2025-03-28
[MNG-8548] - Add cycle detection to dependency injection

gnodet:
JIRA issue: [MNG-8548](https://issues.apache.org/jira/browse/MNG-8548)

Adds cycle detection during instance creation in the DI container to prevent StackOverflowError when circular dependencies are present. The error message now clearly shows the dependency chain that caused the cycle.


elharo (approved):
nice!

jira-importer:
Resolve #9899

PR #2199, opened 2025-03-28
minor cleanups

elharo:
Another batch of small cleanups IntelliJ noticed.


gnodet:
I'm quite reluctant at modifying the compat/* code.  Its aim is to provide as much compatibility as possible with Maven 3 when required.  Changes could lead to minor incompatibilities for no real benefit, as they are deprecated and will be phased out anyway.  So I don't see the benefits in fixing that code.

That definitely does not apply to code in api/* or impl/* of course.

elharo:
This is all stuff detected by IntelliJ static analysis, so relatively safe. If it's in the repo then we need to read it, understand it, and maintain it so impriovements along these lines are valuable, compat or otherwise.

gnodet:
> This is all stuff detected by IntelliJ static analysis, so relatively safe. If it's in the repo then we need to read it, understand it, and maintain it so impriovements along these lines are valuable, compat or otherwise.

I don't think we need to maintain it at all.  I wonder if we could reuse the 3.x jars instead actually...

PR #2200, opened 2025-03-28
Remove a few thousand lines of commented and obsolete code

elharo:
Mostly in tests


compat/maven-compat/src/main/java/org/apache/maven/toolchain/java/JavaToolchain.java
```diff
@@ -27,38 +27,4 @@
  * @deprecated Use {@link org.apache.maven.api.JavaToolchain} instead.
  */
 @Deprecated(since = "4.0.0")
```
gnodet, line 29: -1
no binary incompatibility whatsoever in those compat classes please

gnodet: Sorry, misread, I thought you removed the interface

gnodet:
Same comment as in #2199 , I'd rather only fixing actual bugs in this code...  I think it's wasted time, but...

elharo:
S

> Same comment as in #2199 , I'd rather only fixing actual bugs in this code... I think it's wasted time, but...

ans same response. Either delete it or maintain it. There is no third option.

desruisseaux:
I also think that removing unused codes is helpful. When I'm considering to make a change, I sometime do a search for some keywords in a large code base for having a sense of whether it is used a lot. Dead codes or commented-out codes create false positives which make the analysis more tedious.

Related to above comment but not directly related to this PR, reducing the visibility (i.e. not `public` unless we really intend to allow the usage of a class or method outside its package) is also helpful.

gnodet:
> I also think that removing unused codes is helpful. When I'm considering to make a change, I sometime do a search for some keywords in a large code base for having a sense of whether it is used a lot. Dead codes or commented-out codes create false positives which make the analysis more tedious.

I do agree.  Though I would rather focus on trying to kill the (not completely dead yet) code asap, rather than spending too much effort on it.

gnodet (approved):

PR #2218, opened 2025-04-02
Add missing Javadoc to API subprojects

gnodet:
This PR adds missing Javadoc to various classes and interfaces in the API subprojects.

## Changes:
- Add class-level and method-level Javadoc to various classes in api/ subprojects
- Add missing package-info.java for org.apache.maven.di.tool package
- Enhance existing package-info.java files with more detailed documentation
- Ensure consistent Javadoc style across the API

These changes improve the API documentation, making it more comprehensive and consistent, which will help developers using the Maven API.

### DIFF
--- a/api/maven-api-annotations/src/main/java/org/apache/maven/api/annotations/Config.java
+++ b/api/maven-api-annotations/src/main/java/org/apache/maven/api/annotations/Config.java
@@ -24,34 +24,79 @@ import java.lang.annotation.Retention;
 import java.lang.annotation.RetentionPolicy;
 import java.lang.annotation.Target;
 
+/**
+ * Annotation used to mark fields that represent configuration properties.
+ * This annotation provides metadata about how the configuration property
+ * should be handled, including its source, type, default value, and whether it's read-only.
+ *
+ * @since 4.0.0
+ */
 @Experimental
 @Documented
 @Retention(RetentionPolicy.CLASS)
 @Target(ElementType.FIELD)
 public @interface Config {
 
+    /**
+     * Specifies the source of the configuration property, which determines when and where the property
+     * will be read from or set for consumption in the Maven build lifecycle.
+     *
+     * The source indicates whether the property is:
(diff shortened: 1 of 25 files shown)


api/maven-api-annotations/src/main/java/org/apache/maven/api/annotations/Config.java
```diff
@@ -24,18 +24,45 @@
...
+ * @since 4.0.0
+ */
 @Experimental
 @Documented
 @Retention(RetentionPolicy.CLASS)
 @Target(ElementType.FIELD)
 public @interface Config {
 
+    /**
+     * Specifies the source of the configuration property.
```
elharo, line 41: What is a "source" here? Anything we can say beyond the class name?

api/maven-api-core/src/main/java/org/apache/maven/api/ExtensibleEnums.java
```diff
@@ -24,28 +24,69 @@
 import java.util.Objects;
 import java.util.Set;
 
+/**
```
elharo, line 27: Not for this PR, but I wonder if we should have a different name here to avoid confusion with Java enums, which these aren't

api/maven-api-core/src/main/java/org/apache/maven/api/ExtensibleEnums.java
```diff
@@ -66,37 +107,73 @@ public String toString() {
...
     }
 
+    /**
+     * Implementation of the PathScope interface.
+     */
     private static class DefaultPathScope extends DefaultExtensibleEnum implements PathScope {
         private final ProjectScope projectScope;
         private final Set<DependencyScope> dependencyScopes;
 
+        /**
```
elharo, line 117: again not for this PR, but why is this all in one class? It feels like these should be split out to different classes

elharo (approved):

slawekjaranowski:
source branch deleted

PR #2219, opened 2025-04-02
[MNG-8696] PathModularizationCache needs to set cache

elharo:
…o a public constructor in DefaultDependencyResolverResult. It only hasn't failed yet because we're only passing null for this argument but there's a TODO to change that. 

Alternatively, we could remove PathModularizationCache from the DefaultDependencyResolverResult.

### DIFF
--- a/impl/maven-impl/src/main/java/org/apache/maven/impl/DefaultDependencyResolverResult.java
+++ b/impl/maven-impl/src/main/java/org/apache/maven/impl/DefaultDependencyResolverResult.java
@@ -108,8 +108,7 @@ public class DefaultDependencyResolverResult implements DependencyResolverResult
      */
     public DefaultDependencyResolverResult(
             DependencyResolverRequest request, List<Exception> exceptions, Node root, int count) {
-        // TODO pass a new cache instead of null?
-        this(request, null, exceptions, root, count);
+        this(request, new PathModularizationCache(), exceptions, root, count);
     }
 
     DefaultDependencyResolverResult(
@@ -119,7 +118,11 @@ public class DefaultDependencyResolverResult implements DependencyResolverResult
             Node root,
             int count) {
         this.request = request;
-        this.cache = cache;
+        if (cache == null) {
+            this.cache = new PathModularizationCache();
+        } else {
+            this.cache = cache;
+        }
         this.exceptions = exceptions;
         this.root = root;
         nodes = new ArrayList<>(count);


gnodet:
The `DefaultDependencyResolverResult` does not really respect the `@Immutable` contract which is on `Result<REQ>`.
It may be better refactored in a different class that would be package protected, contains the `addDependency` and `addOutputDirectory`, while the `DefaultDependencyResolverResult` would be kept immutable.

elharo (commented):
PTAL. Took @desruisseaux 's suggestion to make the constructor package-private and add another constructor without the PathModularizationCache argument, which delegates to the package-private constructor with a null argument.

impl/maven-impl/src/main/java/org/apache/maven/impl/DefaultDependencyResolverResult.java
```diff
@@ -102,19 +102,27 @@ public class DefaultDependencyResolverResult implements DependencyResolverResult
...
+
+    DefaultDependencyResolverResult(
             DependencyResolverRequest request,
             PathModularizationCache cache,
             List<Exception> exceptions,
             Node root,
             int count) {
         this.request = request;
-        this.cache = cache;
+        if (cache == null) {
```
desruisseaux, line 121: This `if` … `else` block could be replaced by `this.cache = Objects.requireNonNull(cache)`. For consistency, we may do the same for other arguments too except `root` which is nullable.

elharo: I don't think that's the same. cache is nullable. We need to create a cache if the argument is null, not throw an exception. Null is passed for this argument and we don't want to break those existing usages.

desruisseaux: The new constructor should replace all the calls to the previous constructor with a null cache. We are sure that the new constructor is invoked outside the package since the old constructor became package-private. So we only need to ensure that the calls inside the package pass a non-null cache, and I think that it is already the case.

elharo: It's risky and brittle to allow null caches given the multiple locations that invoke methods on the cache and the lack of test coverage on this class. It is unsafe to allow this field to be null, even if the constructor argument is. Even if it doesn't cause an uncaught NPE now it will in the future.

desruisseaux: Yes I agree, which is why my proposal above was to use `Objects.requireNonNull`.

desruisseaux: The rational was that a cache created here is of limited use. The intend was to have a session-wide cache, which is why this method was requesting the cache in argument. The creation of a cache in the constructor was only a temporary workaround for the fact that we have not investigated where a session-wide cache should be stored.

elharo: Whatever the intent was, this field is dereferenced without null checks in multiple places. That's dangerous.

Pankraz76, line 121:
```suggestion
        if (cache != null) {
```

impl/maven-impl/src/main/java/org/apache/maven/impl/DefaultDependencyResolverResult.java
```diff
@@ -102,19 +102,27 @@ public class DefaultDependencyResolverResult implements DependencyResolverResult
...
+            this.cache = new PathModularizationCache();
```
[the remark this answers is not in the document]
Pankraz76:
```suggestion
            this.cache = cache;
```
might consider init on declaration to give cohesion; default - making first things first.

impl/maven-impl/src/main/java/org/apache/maven/impl/DefaultDependencyResolverResult.java
```diff
@@ -102,19 +102,27 @@ public class DefaultDependencyResolverResult implements DependencyResolverResult
...
+        } else {
+            this.cache = cache;
+        }
```
Pankraz76, line 125:
```suggestion
        }
```
one concern.

impl/maven-impl/src/main/java/org/apache/maven/impl/DefaultDependencyResolverResult.java
```diff
@@ -102,19 +102,27 @@ public class DefaultDependencyResolverResult implements DependencyResolverResult
      * to {@link #addDependency(Node, Dependency, Predicate, Path)}.
      *
      * @param request the corresponding request
-     * @param cache cache of module information about each dependency
```
Pankraz76, line 105: `private final PathModularizationCache cache = new PathModularizationCache();` L:98

- https://www.reddit.com/r/learnjava/comments/ozgtxm/is_initializing_variables_at_their_declaration/
- https://softwareengineering.stackexchange.com/questions/379039/what-is-the-best-practice-for-initializing-variables-outside-or-inside-of-the-c
-

Pankraz76 (changes requested):
might consider init on declaration to give cohesion; default - making first things first.

impl/maven-impl/src/main/java/org/apache/maven/impl/DefaultDependencyResolverResult.java
desruisseaux: In the particular case of this issue, creating the cache at the location where the field is defined is the opposite of the intend, because it makes the `PathModularizationCache` lifetime strongly coupled with the `DefaultDependencyResolverResult` lifetime. Such coupling makes `PathModularizationCache` close to useless. It was done that way in the previous version of this class only as a workaround for the fact that I didn't knew where to store a session-wide cache, but that workaround was intended to be temporary.

More details in a next comment.

desruisseaux:
I think that the `DefaultDependencyResolverResult` constructor is creating some confusion. I created a new pull request that I propose as an alternative to this one: #2347. For seeing the purpose, there is two design constraints to consider:

* In order to be useful, `PathModularizationCache` needs to live longer than `DefaultDependencyResolverResult`. Any change that makes the lifetime of those two objects strongly coupled make the cache close to useless.
* At the time I wrote these classes, I didn't know how session-wide caches should be managed in Maven (I still haven't investigated). Therefore, `PathModularizationCache` was instantiated together with `DefaultDependencyResolverResult` *as a temporary workaround, until we learn more about how session-wide caches are managed in Maven.*

The pull #2347 tries to address the same issue as this pull request, but in a way that makes more apparent that `DefaultDependencyResolverResult` and `PathModularizationCache` lifetimes are distinct in principle, and that the current coupling may be temporary.

impl/maven-impl/src/main/java/org/apache/maven/impl/DefaultDependencyResolverResult.java
Pankraz76: my suggest would not compile. im sorry.

jira-importer:
Resolve #9351

PR #2226, opened 2025-04-03
[MNG-8637] Pull out Standalone API from UTs

gnodet:
JIRA issue: [MNG-8637](https://issues.apache.org/jira/browse/MNG-8637)

This commit moves the Standalone API implementation from the test directory to the main source directory, making it available for reuse outside of tests.

The Standalone API provides a way to run Maven API in a standalone mode (without the full Maven environment), primarily for testing and specialized execution scenarios. This is now the proper way to utilize Maven 4 in MIMA-like scenarios.

Key changes:
- Moved ApiRunner and RepositorySystemSupplier from test to main source
- Added UnsupportedInStandaloneModeException for operations not supported in standalone mode
- Added package-info.java with documentation for the standalone package


cstamas:
@gnodet sure!

elharo (approved):

impl/maven-impl/pom.xml
```diff
@@ -112,6 +112,22 @@ under the License.
       <groupId>org.apache.maven.resolver</groupId>
       <artifactId>maven-resolver-impl</artifactId>
     </dependency>
+    <dependency>
```
cstamas, line 115: am unsure are these deps really needed here, maybe only for test?

As if you make them required, you make apache transport or basic connector "mandatory". I'd rather put back these into test scope of maven-impl and let "mima-like use case" user pick deps he needs instead?

cstamas:
Not to mention: now if this pulled out, what happens to maven-testing supplier? Should they be the same?

gnodet:
> Not to mention: now if this pulled out, what happens to maven-testing supplier? Should they be the same?

Maven testing uses mocks, so not really affected.

impl/maven-impl/pom.xml
gnodet: I've moved out the transports and kept named locks and connectors, as they are really the only implementation.  If needed, we can revisit later.

cstamas (approved):

jira-importer:
Resolve #9673

PR #2228, opened 2025-04-03
[MRESOLVER-569] Pass on exceptions

cstamas:
Pass on exceptions that are "per repo" for better
error reporting.

---

https://issues.apache.org/jira/browse/MRESOLVER-569


cstamas:
This is wrong, dropping this.

PR #2232, opened 2025-04-04
[MNG-8673] SourceRoot includes and excludes should be String

gnodet:
JIRA issue: [MNG-8673](https://issues.apache.org/jira/browse/MNG-8673)

### DIFF
--- a/api/maven-api-core/src/main/java/org/apache/maven/api/SourceRoot.java
+++ b/api/maven-api-core/src/main/java/org/apache/maven/api/SourceRoot.java
@@ -19,7 +19,6 @@
 package org.apache.maven.api;
 
 import java.nio.file.Path;
-import java.nio.file.PathMatcher;
 import java.util.List;
 import java.util.Optional;
 
@@ -50,7 +49,7 @@ public interface SourceRoot {
      * The default implementation returns an empty list, which means to apply a language-dependent pattern.
      * For example, for the Java language, the pattern includes all files with the {@code .java} suffix.
      */
-    default List<PathMatcher> includes() {
+    default List<String> includes() {
         return List.of();
     }
 
@@ -59,7 +58,7 @@ public interface SourceRoot {
      * The exclusions are applied after the inclusions.
      * The default implementation returns an empty list.
      */
-    default List<PathMatcher> excludes() {
+    default List<String> excludes() {
         return List.of();
     }
 
--- a/api/maven-api-core/src/main/java/org/apache/maven/api/SourceRoot.java
+++ b/api/maven-api-core/src/main/java/org/apache/maven/api/SourceRoot.java
@@ -46,6 +46,11 @@ public interface SourceRoot {
 
     /**
      * {@return the list of pattern matchers for the files to include}.
+     * The path separator is {@code /} on all platforms, including Windows.
+     * If the pattern does not start with {@code regex:} or {@code glob:}, the default variation of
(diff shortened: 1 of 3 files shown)


api/maven-api-core/src/main/java/org/apache/maven/api/SourceRoot.java
```diff
@@ -50,7 +49,7 @@ default Path directory() {
      * The default implementation returns an empty list, which means to apply a language-dependent pattern.
```
desruisseaux, line 49: Proposed new block of comment:

```
    /**
     * {@return the list of patterns for the files to include}.
     * The path separator is {@code /} on all platforms, including Windows.
     * The prefix before the {@code :} character, if present, is the syntax.
     * If no syntax is specified, the default is a Maven-specific variation
     * of the {@code "glob"} pattern.
     *
     * <p>The default implementation returns an empty list, which means to apply a language-dependent pattern.
     * For example, for the Java language, the default pattern is {@code "*.java"}.</p>
     */
```

api/maven-api-core/src/main/java/org/apache/maven/api/SourceRoot.java
```diff
@@ -59,7 +58,7 @@ default List<PathMatcher> includes() {
      * The exclusions are applied after the inclusions.
```
desruisseaux, line 58: Heading comment would become:

```
     * {@return the list of patterns for the files to exclude}.
```

impl/maven-impl/src/main/java/org/apache/maven/impl/DefaultSourceRoot.java
```diff
@@ -149,8 +145,8 @@ public DefaultSourceRoot(
             final ProjectScope scope,
```
desruisseaux, line 145: We can take this opportunity for adding the following missing comments:

```
     * @param includes list of patterns for the files to include, or {@code null} if unspecified
     * @param excludes list of patterns for the files to exclude, or {@code null} if unspecified
```

api/maven-api-core/src/main/java/org/apache/maven/api/SourceRoot.java
gnodet: What if on windows we have a drive specified ? We need to differentiate:
```
C:/foo/**
```
So maybe rephrase as: `If the pattern does not start with {@code regex:} or {@code glob:}, the default variation of the {@code glob} pattern will be used.`

desruisseaux: Just to complete: "glob:" and "regex:" are the two syntaxes that all `FileSystem` shall support according NIO Javadoc, but a file system may support others syntaxes as well. It may be nice to avoid a formulation that restrict the possibilities to those two syntaxes.

desruisseaux (approved):

api/maven-api-core/src/main/java/org/apache/maven/api/SourceRoot.java
desruisseaux: According Stackoverflow, Windows drives are restricted to 1 letter. So I suggest to replace _"If the pattern does not start with regex: or glob:"_ by _"The prefix before the : character, if present and longer than one character, is the syntax"_.

desruisseaux:
Added pull request #2236 as a follow-up of this one.

PR #2233, opened 2025-04-04
[MNG-8674] Deprecate mergeId in the Maven 3 model and remove it in the new model

gnodet:
JIRA issue: [MNG-8674](https://issues.apache.org/jira/browse/MNG-8674)

### DIFF
--- a/api/maven-api-model/src/main/mdo/maven.mdo
+++ b/api/maven-api-model/src/main/mdo/maven.mdo
@@ -2208,7 +2208,7 @@
         </field>
         <field xml.transient="true">
           <name>mergeId</name>
-          <version>4.0.0+</version>
+          <version>4.0.0</version>
           <description>
             FOR INTERNAL USE ONLY. This is a unique identifier assigned to each
             resource to allow Maven to merge changes to this resource that take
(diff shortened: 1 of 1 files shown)


api/maven-api-model/src/main/mdo/maven.mdo
```diff
@@ -2220,6 +2220,21 @@
         </field>
       </fields>
       <codeSegments>
+        <codeSegment>
+            <version>4.0.0+</version>
+            <code>
+            <![CDATA[
+    @Deprecated
```
elharo, line 2227: So these are new methods being added in a deprecated state? That's unusual.

There should be a Javadoc comment about them of some kind, perhaps just indicating that they should not be used at all.

And if they're deprecated because they're only for internal use, that's not really what deprecated is typically used for.

elharo (changes requested):
This one confuses me.

api/maven-api-model/src/main/mdo/maven.mdo
gnodet: I've added a bit of javadoc.
Those methods are not new, they are not explicitly added, because the switch above from `<version>4.0.0+</version>` to `<version>4.0.0</version>` effectively does not generate them anymore.  So I added them for backward compatibility as no-op.

elharo (approved):

PR #2236, opened 2025-04-05
[MNG-8686] Add `SourceRoot.matcher(boolean)` method

desruisseaux:
Following the revert of includes/excludes filters from `PathMatcher` to `String` in #2232, this pull request adds a new `SourceRoot.matcher(boolean)` method which provides the `PathMatcher`. Contrarily to the previous methods, this new method provides a single matcher combining the work of all includes and all excludes. The default implementation is ported from https://github.com/apache/maven-clean-plugin/pull/243.

### DIFF
--- a/api/maven-api-core/src/main/java/org/apache/maven/api/SourceRoot.java
+++ b/api/maven-api-core/src/main/java/org/apache/maven/api/SourceRoot.java
@@ -19,6 +19,8 @@
 package org.apache.maven.api;
 
 import java.nio.file.Path;
+import java.nio.file.PathMatcher;
+import java.util.Collection;
 import java.util.List;
 import java.util.Optional;
 
@@ -45,29 +47,46 @@ public interface SourceRoot {
     }
 
     /**
-     * {@return the list of pattern matchers for the files to include}.
+     * {@return the list of patterns for the files to include}.
      * The path separator is {@code /} on all platforms, including Windows.
-     * The patterns are used to match paths relative to the {@code directory}.
-     * The prefix before the {@code :} character, if present, is the syntax.
-     * If no syntax is specified, the default is a Maven-specific variation
-     * of the {@code "glob"} pattern.
+     * The prefix before the {@code :} character, if present and longer than 1 character, is the syntax.
+     * If no syntax is specified, or if its length is 1 character (interpreted as a Windows drive),
+     * the default is a Maven-specific variation of the {@code "glob"} pattern.
      *
      * <p>
      * The default implementation returns an empty list, which means to apply a language-dependent pattern.
      * For example, for the Java language, the pattern includes all files with the {@code .java} suffix.
+     *
+     * @see java.nio.file.FileSystem#getPathMatcher(String)
      */
     default List<String> includes() {
         return List.of();
     }
 
     /**
-     * {@return the list of pattern matchers for the files to exclude}.
+     * {@return the list of patterns for the files to exclude}.
      * The exclusions are applied after the inclusions.
      * The default implementation returns an empty list.
      */
     default List<String> excludes() {
         return List.of();
     }
+
+    /**
+     * {@return a matcher combining the include and exclude patterns}.
+     * If the user did not specify any includes, the given {@code defaultIncludes} are used.
+     * These defaults depend on the plugin.
+     * For example, the default include of the Java compiler plugin is <code>"**&sol;*.java"</code>.
+     *
+     * <p>If the user did not specify any excludes, the default is often files generated
+     * by Source Code Management (<abbr>SCM</abbr>) software or by the operating system.
+     * Examples: <code>"**&sol;.gitignore"</code>, <code>"**&sol;.DS_Store"</code>.</p>
+     *
+     * @param defaultIncludes the default includes if unspecified by the user
+     * @param useDefaultExcludes whether to add the default set of patterns to exclude,
+     *        mostly Source Code Management (<abbr>SCM</abbr>) files
+     */
+    PathMatcher matcher(Collection<String> defaultIncludes, boolean useDefaultExcludes);
 
     /**
      * {@return in which context the source files will be used}.
--- a/impl/maven-impl/src/main/java/org/apache/maven/impl/DefaultSourceRoot.java
+++ b/impl/maven-impl/src/main/java/org/apache/maven/impl/DefaultSourceRoot.java
@@ -19,6 +19,8 @@
 package org.apache.maven.impl;
 
 import java.nio.file.Path;
+import java.nio.file.PathMatcher;
+import java.util.Collection;
 import java.util.List;
 import java.util.Objects;
 import java.util.Optional;
@@ -140,9 +142,9 @@ public final class DefaultSourceRoot implements SourceRoot {
      * @param scope scope of source code (main or test)
      * @param language language of the source code
      * @param directory directory of the source code
-     * @param includes list of patterns for the files to include, or {@code null} if unspecified
-     * @param excludes list of patterns for the files to exclude, or {@code null} if unspecified
-     * */
+     * @param includes patterns for the files to include, or {@code null} or empty if unspecified
+     * @param excludes patterns for the files to exclude, or {@code null} or empty if nothing to exclude
+     */
     public DefaultSourceRoot(
             final ProjectScope scope,
             final Language language,
@@ -183,7 +185,7 @@ public final class DefaultSourceRoot implements SourceRoot {
     }
 
     /**
-     * {@return the list of pattern matchers for the files to include}.
+     * {@return the patterns for the files to include}.
      */
     @Override
     @SuppressWarnings("ReturnOfCollectionOrArrayField") // Safe because unmodifiable
@@ -192,12 +194,28 @@ public final class DefaultSourceRoot implements SourceRoot {
     }
 
     /**
-     * {@return the list of pattern matchers for the files to exclude}.
+     * {@return the patterns for the files to exclude}.
      */
     @Override
     @SuppressWarnings("ReturnOfCollectionOrArrayField") // Safe because unmodifiable
     public List<String> excludes() {
         return excludes;
+    }
+
+    /**
+     * {@return a matcher combining the include and exclude patterns}.
+     *
+     * @param defaultIncludes the default includes if unspecified by the user
+     * @param useDefaultExcludes whether to add the default set of patterns to exclude,
+     *        mostly Source Code Management (<abbr>SCM</abbr>) files
+     */
+    @Override
+    public PathMatcher matcher(Collection<String> defaultIncludes, boolean useDefaultExcludes) {
+        Collection<String> actual = includes();
+        if (actual == null || actual.isEmpty()) {
+            actual = defaultIncludes;
+        }
+        return new PathSelector(directory(), actual, excludes(), useDefaultExcludes).simplify();
     }
 
     /**
--- a/impl/maven-impl/src/main/java/org/apache/maven/impl/PathSelector.java
+++ b/impl/maven-impl/src/main/java/org/apache/maven/impl/PathSelector.java
@@ -453,10 +453,8 @@ public class PathSelector implements PathMatcher {
                     end++; // Ommit the leading slash if there is nothing before it.
                 }
             }
-            if (start > 0) {
-                if (pattern.charAt(--start) != '/') {
-                    continue;
-                }
+            if (start > 0 && pattern.charAt(--start) != '/') {
+                continue;
             }
             String reduced = pattern.substring(0, start) + pattern.substring(end);
             patterns.add(DEFAULT_SYNTAX + reduced);
--- a/impl/maven-impl/src/main/java/org/apache/maven/impl/PathSelector.java
+++ b/impl/maven-impl/src/main/java/org/apache/maven/impl/PathSelector.java
@@ -1,0 +1,622 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.maven.impl;
+
+import java.io.File;
+import java.nio.file.FileSystem;
+import java.nio.file.Path;
+import java.nio.file.PathMatcher;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Collection;
+import java.util.Iterator;
+import java.util.LinkedHashSet;
+import java.util.List;
+import java.util.Set;
+
+/**
+ * Determines whether a path is selected according to include/exclude patterns.
+ * The pathnames used for method parameters will be relative to some base directory
+ * and use {@code '/'} as separator, regardless of the hosting operating system.
+ *
+ * <h2>Syntax</h2>
+ * If a pattern contains the {@code ':'} character and the prefix before is longer than 1 character,
+ * then that pattern is given verbatim to {@link FileSystem#getPathMatcher(String)}, which interprets
+ * the part before {@code ':'} as the syntax (usually {@code "glob"} or {@code "regex"}).
+ * If a pattern does not contain the {@code ':'} character, or if the prefix is one character long
+ * (interpreted as a Windows drive), then the syntax defaults to a reproduction of the Maven 3 behavior.
+ * This is implemented as the {@code "glob"} syntax with the following modifications:
+ *
+ * <ul>
+ *   <li>The platform-specific separator ({@code '\\'} on Windows) is replaced by {@code '/'}.
+ *       Note that it means that the backslash cannot be used for escaping characters.</li>
+ *   <li>Trailing {@code "/"} is completed as {@code "/**"}.</li>
+ *   <li>The {@code "**"} wildcard means "0 or more directories" instead of "1 or more directories".
+ *       This is implemented by adding variants of the pattern without the {@code "**"} wildcard.</li>
+ *   <li>Bracket characters [ ] and { } are escaped.</li>
+ *   <li>On Unix only, the escape character {@code '\\'} is itself escaped.</li>
+ * </ul>
+ *
+ * If above changes are not desired, put an explicit {@code "glob:"} prefix before the pattern.
+ * Note that putting such a prefix is recommended anyway for better performances.
+ *
+ * @author Benjamin Bentmann
+ * @author Martin Desruisseaux
+ *
+ * @see java.nio.file.FileSystem#getPathMatcher(String)
+ */
+public class PathSelector implements PathMatcher {
+    /**
+     * Patterns which should be excluded by default, like <abbr>SCM</abbr> files.
+     *
+     * <p><b>Source:</b> this list is copied from {@code plexus-utils-4.0.2} (released in
+     * September 23, 2024), class {@code org.codehaus.plexus.util.AbstractScanner}.</p>
+     */
+    private static final List<String> DEFAULT_EXCLUDES = List.of(
+            // Miscellaneous typical temporary files
+            "**/*~",
+            "**/#*#",
+            "**/.#*",
+            "**/%*%",
+            "**/._*",
+
+            // CVS
+            "**/CVS",
+            "**/CVS/**",
+            "**/.cvsignore",
+
+            // RCS
+            "**/RCS",
+            "**/RCS/**",
+
+            // SCCS
+            "**/SCCS",
+            "**/SCCS/**",
+
+            // Visual SourceSafe
+            "**/vssver.scc",
+
+            // MKS
+            "**/project.pj",
+
+            // Subversion
+            "**/.svn",
+            "**/.svn/**",
+
+            // Arch
+            "**/.arch-ids",
+            "**/.arch-ids/**",
+
+            // Bazaar
+            "**/.bzr",
+            "**/.bzr/**",
+
+            // SurroundSCM
+            "**/.MySCMServerInfo",
+
+            // Mac
+            "**/.DS_Store",
+
+            // Serena Dimensions Version 10
+            "**/.metadata",
+            "**/.metadata/**",
+
+            // Mercurial
+            "**/.hg",
+            "**/.hg/**",
+
+            // git
+            "**/.git",
+            "**/.git/**",
+            "**/.gitignore",
+
+            // BitKeeper
+            "**/BitKeeper",
+            "**/BitKeeper/**",
+            "**/ChangeSet",
+            "**/ChangeSet/**",
+
+            // darcs
+            "**/_darcs",
+            "**/_darcs/**",
+            "**/.darcsrepo",
+            "**/.darcsrepo/**",
+            "**/-darcs-backup*",
+            "**/.darcs-temp-mail");
+
+    /**
+     * Maximum number of characters of the prefix before {@code ':'} for handling as a Maven syntax.
+     */
+    private static final int MAVEN_SYNTAX_THRESHOLD = 1;
+
+    /**
+     * The default syntax to use if none was specified. Note that when this default syntax is applied,
+     * the user-provided pattern get some changes as documented in class Javadoc.
+     */
+    private static final String DEFAULT_SYNTAX = "glob:";
+
+    /**
+     * Characters having a special meaning in the glob syntax.
+     *
+     * @see FileSystem#getPathMatcher(String)
+     */
+    private static final String SPECIAL_CHARACTERS = "*?[]{}\\";
+
+    /**
+     * A path matcher which accepts all files.
+     *
+     * @see #simplify()
+     */
+    private static final PathMatcher INCLUDES_ALL = (path) -> true;
+
+    /**
+     * String representations of the normalized include filters.
+     * Each pattern shall be prefixed by its syntax, which is {@value #DEFAULT_SYNTAX} by default.
+     *
+     * @see #toString()
+     */
+    private final String[] includePatterns;
+
+    /**
+     * String representations of the normalized exclude filters.
+     * Each pattern shall be prefixed by its syntax, which is {@value #DEFAULT_SYNTAX} by default.
+     * This array may be longer or shorter than the user-supplied excludes, depending on whether
+     * default excludes have been added and whether some unnecessary excludes have been omitted.
+     *
+     * @see #toString()
+     */
+    private final String[] excludePatterns;
+
+    /**
+     * The matcher for includes. The length of this array is equal to {@link #includePatterns} array length.
+     */
+    private final PathMatcher[] includes;
+
+    /**
+     * The matcher for excludes. The length of this array is equal to {@link #excludePatterns} array length.
+     */
+    private final PathMatcher[] excludes;
+
+    /**
+     * The matcher for all directories to include. This array includes the parents of all those directories,
+     * because they need to be accepted before we can walk to the sub-directories.
+     * This is an optimization for skipping whole directories when possible.
+     */
+    private final PathMatcher[] dirIncludes;
+
+    /**
+     * The matcher for directories to exclude. This array does <em>not</em> include the parent directories,
+     * because they may contain other sub-trees that need to be included.
+     * This is an optimization for skipping whole directories when possible.
+     */
+    private final PathMatcher[] dirExcludes;
+
+    /**
+     * The base directory. All files will be relativized to that directory before to be matched.
+     */
+    private final Path baseDirectory;
+
+    /**
+     * Creates a new selector from the given includes and excludes.
+     *
+     * @param directory the base directory of the files to filter
+     * @param includes the patterns of the files to include, or null or empty for including all files
+     * @param excludes the patterns of the files to exclude, or null or empty for no exclusion
+     * @param useDefaultExcludes whether to augment the excludes with a default set of <abbr>SCM</abbr> patterns
+     */
+    public PathSelector(
+            Path directory, Collection<String> includes, Collection<String> excludes, boolean useDefaultExcludes) {
+        includePatterns = normalizePatterns(includes, false);
+        excludePatterns = normalizePatterns(effectiveExcludes(excludes, includePatterns, useDefaultExcludes), true);
+        baseDirectory = directory;
+        FileSystem system = directory.getFileSystem();
+        this.includes = matchers(system, includePatterns);
+        this.excludes = matchers(system, excludePatterns);
+        dirIncludes = matchers(system, directoryPatterns(includePatterns, false));
+        dirExcludes = matchers(system, directoryPatterns(excludePatterns, true));
+    }
+
+    /**
+     * Returns the given array of excludes, optionally expanded with a default set of excludes,
+     * then with unnecessary excludes omitted. An unnecessary exclude is an exclude which will never
+     * match a file because there is no include which would accept a file that could match the exclude.
+     * For example, if the only include is {@code "*.java"}, then the <code>"**&sol;project.pj"</code>,
+     * <code>"**&sol;.DS_Store"</code> and other excludes will never match a file and can be omitted.
+     * Because the list of {@linkplain #DEFAULT_EXCLUDES default excludes} contains many elements,
+     * removing unnecessary excludes can reduce a lot the number of matches tested on each source file.
+     *
+     * <h4>Implementation note</h4>
+     * The removal of unnecessary excludes is done on a best effort basis. The current implementation
+     * compares only the prefixes and suffixes of each pattern, keeping the pattern in case of doubt.
+     * This is not bad, but it does not remove all unnecessary patterns. It would be possible to do
+     * better in the future if benchmarking suggests that it would be worth the effort.
+     *
+     * @param excludes the user-specified excludes, potentially not yet converted to glob syntax
+     * @param includes the include patterns converted to glob syntax
+     * @param useDefaultExcludes whether to expand user exclude with the set of default excludes
+     * @return the potentially expanded or reduced set of excludes to use
+     */
+    private static Collection<String> effectiveExcludes(
+            Collection<String> excludes, final String[] includes, final boolean useDefaultExcludes) {
+        if (excludes == null || excludes.isEmpty()) {
+            if (useDefaultExcludes) {
+                excludes = new ArrayList<>(DEFAULT_EXCLUDES);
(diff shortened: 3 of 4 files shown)


impl/maven-impl/src/main/java/org/apache/maven/impl/DefaultSourceRoot.java
```diff
@@ -39,9 +38,11 @@
 public final class DefaultSourceRoot implements SourceRoot {
     private final Path directory;
 
-    private final List<PathMatcher> includes;
+    private final List<String> includes;
 
-    private final List<PathMatcher> excludes;
+    private final List<String> excludes;
+
+    private transient PathMatcher matcher, matcherWithDefaults;
```
elharo, line 45: one field per declaration please.
also I'm a little surprised these can't be final or local

impl/maven-impl/src/main/java/org/apache/maven/impl/DefaultSourceRoot.java
```diff
@@ -144,13 +143,15 @@ public DefaultSourceRoot(final ProjectScope scope, final Language language, fina
      * @param scope scope of source code (main or test)
      * @param language language of the source code
      * @param directory directory of the source code
+     * @param includes list of patterns for the files to include, or {@code null} if unspecified
```
elharo, line 146: unspecified or nothing to include?

impl/maven-impl/src/main/java/org/apache/maven/impl/DefaultSourceRoot.java
```diff
@@ -144,13 +143,15 @@ public DefaultSourceRoot(final ProjectScope scope, final Language language, fina
...
+     * @param excludes list of patterns for the files to exclude, or {@code null} if unspecified
```
elharo, line 147: unspecified or nothing to exclude?

impl/maven-impl/src/main/java/org/apache/maven/impl/DefaultSourceRoot.java
```diff
@@ -217,23 +186,45 @@ public Path directory() {
...
     @SuppressWarnings("ReturnOfCollectionOrArrayField") // Safe because unmodifiable
-    public List<PathMatcher> excludes() {
+    public List<String> excludes() {
         return excludes;
     }
 
+    /**
+     * {@return a matcher combining the include and exclude patterns}.
+     *
+     * @param useDefaultExcludes whether to add <abbr>SCM</abbr> files to set of exclude patterns
```
elharo, line 209: spell out SCM

impl/maven-impl/src/main/java/org/apache/maven/impl/PathSelector.java
```diff
@@ -0,0 +1,475 @@
...
+        buffer.append(']');
+    }
+
+    /**
+     * {@return a string representation for logging purposes}.
+     * This string representation is reported logged when {@link Cleaner} is executed.
+     */
+    @Override
+    public String toString() {
+        var buffer = new StringBuilder();
```
elharo, line 470: personally I dislike var. Not sure if there's a project setting on this

impl/maven-impl/src/test/java/org/apache/maven/impl/PathSelectorTest.java
```diff
@@ -0,0 +1,92 @@
...
+     */
+    private void filter(final String syntax) throws IOException {
+        var includes = List.of(syntax + "**/*.txt");
+        var excludes = List.of(syntax + "biz/**");
+        var matcher = new PathSelector(directory, includes, excludes, false);
+        filtered = new HashSet<>(Files.walk(directory).filter(matcher::matches).toList());
+    }
+
+    /**
+     * Asserts that the filtered set of paths contains the given item.
```
elharo, line 85: This does more than that. It mutates the collection

elharo (changes requested):

impl/maven-impl/src/main/java/org/apache/maven/impl/DefaultSourceRoot.java
desruisseaux: Separated the declarations and added a comment. The field is non-final because it is a cache computed when first requested (some fields may be never requested).

impl/maven-impl/src/main/java/org/apache/maven/impl/DefaultSourceRoot.java
desruisseaux: Unspecified. The default is plugin-dependent. For example, the compiler plugin will default to the `*.java` includes. Other plugins may have different default.

impl/maven-impl/src/main/java/org/apache/maven/impl/DefaultSourceRoot.java
desruisseaux: Nothing to exclude. Updated the documentation.

impl/maven-impl/src/main/java/org/apache/maven/impl/DefaultSourceRoot.java
desruisseaux: Done (Source Code Management).

impl/maven-impl/src/main/java/org/apache/maven/impl/PathSelector.java
desruisseaux: I do not know if there is a Maven policy. Personally, I use it when it is redundant with what we can read on the same line with no possibility of different result (e.g., the result of `new`), but generally avoid it for the return value of a method call for example, as it is not obvious by reading the line.

impl/maven-impl/src/test/java/org/apache/maven/impl/PathSelectorTest.java
desruisseaux: Updated the documentation.

elharo (commented):
Does this have/need a JIRA issue?

impl/maven-impl/src/main/java/org/apache/maven/impl/PathSelector.java
```diff
@@ -0,0 +1,475 @@
...
+                }
+                buffer.append(patterns[i]);
+            }
+        }
+        buffer.append(']');
+    }
+
+    /**
+     * {@return a string representation for logging purposes}.
+     * This string representation is reported logged when {@link Cleaner} is executed.
```
elharo, line 466: "reported logged" is unclear. maybe just "logged"?

impl/maven-impl/src/test/java/org/apache/maven/impl/PathSelectorTest.java
```diff
@@ -0,0 +1,95 @@
...
+import java.util.Set;
+
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.io.TempDir;
+
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+public class PathSelectorTest {
+    /**
+     * The temporary directory where are the files to test.
```
elharo, line 35: where are the files to test --> containing the files to test

impl/maven-impl/src/test/java/org/apache/maven/impl/PathSelectorTest.java
```diff
@@ -0,0 +1,95 @@
...
+     */
+    private Path directory;
+
+    /**
+     * The filtered set of paths. creates by {@link #filter()}.
```
elharo, line 40: created

impl/maven-impl/src/test/java/org/apache/maven/impl/PathSelectorTest.java
```diff
@@ -0,0 +1,95 @@
...
+        var includes = List.of(syntax + "**/*.txt");
+        var excludes = List.of(syntax + "biz/**");
+        var matcher = new PathSelector(directory, includes, excludes, false);
+        filtered = new HashSet<>(Files.walk(directory).filter(matcher::matches).toList());
+    }
+
+    /**
+     * Asserts that the filtered set of paths contains the given item.
+     * If present, the path is removed from the collection of filtered files.
+     * It allows caller to verify that there is no unexpected elements remaining
```
elharo, line 87: there are

impl/maven-impl/src/test/java/org/apache/maven/impl/PathSelectorTest.java
```diff
@@ -0,0 +1,95 @@
...
+     * after all expected elements have been removed.
+     *
+     * @param path the path to test
+     */
+    private void assertContains(String path) {
```
elharo, line 92: I think this method would be clearer inlined. It's just one line.

elharo (approved):

api/maven-api-core/src/main/java/org/apache/maven/api/SourceRoot.java
```diff
@@ -46,23 +46,39 @@ default Path directory() {
     }
 
     /**
-     * {@return the list of pattern matchers for the files to include}.
+     * {@return the list of patterns for the files to include}.
+     * The path separator is {@code /} on all platforms, including Windows.
+     * The prefix before the {@code :} character, if present and longer than 1 character, is the syntax.
```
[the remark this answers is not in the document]
desruisseaux: It is defined in [Oracle Javadoc](https://docs.oracle.com/en/java/javase/24/docs/api/java.base/java/nio/file/FileSystem.html#getPathMatcher(java.lang.String)). Will add a `@see` tag.

impl/maven-impl/src/main/java/org/apache/maven/impl/DefaultSourceRoot.java
desruisseaux: Actually this "null or empty" (will update the documentation). This is a parameter rather than a return value, accepting both for convenience.

api/maven-api-core/src/main/java/org/apache/maven/api/SourceRoot.java
```diff
@@ -46,23 +46,39 @@ default Path directory() {
     }
 
     /**
-     * {@return the list of pattern matchers for the files to include}.
+     * {@return the list of patterns for the files to include}.
```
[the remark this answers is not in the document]
desruisseaux: Right.

api/maven-api-core/src/main/java/org/apache/maven/api/SourceRoot.java
```diff
@@ -46,23 +46,39 @@ default Path directory() {
...
 
+    /**
+     * {@return a matcher combining the include and exclude patterns}.
+     * The matcher is absent if the includes/excludes lists are empty
+     * and {@code useDefaultExcludes} is {@code false}.
+     *
+     * @param useDefaultExcludes whether to add the default set of patterns to exclude,
+     *        mostly Source Code Management (<abbr>SCM</abbr>) files
+     */
+    Optional<PathMatcher> matcher(boolean useDefaultExcludes);
```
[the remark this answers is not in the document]
desruisseaux: It would rather be a matcher that include everything. I had an hesitation on that one. The current rational is that `java.nio.Files` often returns `Stream<Path>`, in which case the matcher can be applied by a call to `stream.filter((path) -> pm.matches(path))`. If the `Optional` is empty, we can omit the call to `filter` completely. I'm not sure if it makes a big performance difference in practice however.

impl/maven-impl/src/main/java/org/apache/maven/impl/PathSelector.java
desruisseaux: Right, it was a type.

impl/maven-impl/src/main/java/org/apache/maven/impl/PathSelector.java
```diff
@@ -0,0 +1,475 @@
...
+            matchers[i] = fs.getPathMatcher(pattern);
+        }
+        return matchers;
+    }
+
+    /**
+     * {@return a potentially simpler matcher equivalent to this matcher}.
+     * It may be empty if there is no filtering to apply (i.e., all files are included).
+     */
+    @SuppressWarnings("checkstyle:MissingSwitchDefault")
```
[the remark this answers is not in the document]
desruisseaux: (related to a previous comment) because I'm not aware of a way to determine that a `PathMatcher` filter nothing, and therefore a call to `Stream.filter(Predicate)` (for example) is unnecessary. However, I do not know if it would have a significant impact on performance. It is also possible that this case (no filtering) rarely happens in practice, so that the performance concern is irrelevant.

impl/maven-impl/src/test/java/org/apache/maven/impl/PathSelectorTest.java
desruisseaux: That line has some degrees of freedom (the collection, the base directory, the message in case of failure), even if I agree that there is not much (yet) real freedom because there is only one collection and one directory. A purpose of using a method is that, when many calls to `assertContains` are repeated, we know that those things are the same for every lines without having to paid a closer look to the lines.

I will rename as `assertFilteredFilesContains` for more clarity.

api/maven-api-core/src/main/java/org/apache/maven/api/SourceRoot.java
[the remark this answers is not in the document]
desruisseaux: After a second thought, replaced `Optional<PathMatcher> matcher(boolean)` by `PathMatcher matcher(Collection, boolean)`.

impl/maven-impl/src/main/java/org/apache/maven/impl/PathSelector.java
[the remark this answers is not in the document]
desruisseaux: Done as suggested (replaced `Optional` by `PathMatcher` that doesn't filter).

desruisseaux:
JIRA issue created: https://issues.apache.org/jira/browse/MNG-8686

gnodet:
> Following the revert of includes/excludes filters from `PathMatcher` to `String` in #2232, this pull request adds a new `SourceRoot.matcher(boolean)` method which provides the `PathMatcher`. Contrarily to the previous methods, this new method provides a single matcher combining the work of all includes and all excludes. The default implementation is ported from [apache/maven-clean-plugin#243](https://github.com/apache/maven-clean-plugin/pull/243).

We'd need to carefully review the maven-resources-plugins various options and see if they can fit this API. IIRC, there are options such as symlinks following, but there may be others.

desruisseaux:
> We'd need to carefully review the maven-resources-plugins various options and see if they can fit this API. IIRC, there are options such as symlinks following, but there may be others.

Yes. This is another aspect replaced by standard Java now, with `java.nio.file.FileVisitOption.FOLLOW_LINKS`. One issue is that it behaves differently than the current Maven for one kind of link specific to the Windows platform: NTFS "junction links". This is handled by a special case in https://github.com/apache/maven-clean-plugin/pull/243. Thankfully, the `maven-cleaner-plugin` has a quite extensive test suite that covers this case among others.

Symbolic links are not handled by the `PathSelector` class proposed in this pull request. They need to be handled later, in our use of `java.nio.file.FileVisitor`, which is not part of this PR. I was thinking to address `FileVisitor` in a separated PR.

api/maven-api-core/src/main/java/org/apache/maven/api/SourceRoot.java
```diff
@@ -46,23 +47,47 @@ default Path directory() {
...
-    default List<PathMatcher> excludes() {
+    default List<String> excludes() {
         return List.of();
     }
 
+    /**
+     * {@return a matcher combining the include and exclude patterns}.
+     * If the user did not specified any includes, the given {@code defaultIncludes} are used.
+     * These defaults depend on the plugin.
+     * For example, the default include of the Java compiler plugin is <code>"**&sol;*.java"</code>.
```
elharo, line 79: I don't think you need to use &sol; here. A forward slash should be fine if that's what you want.

api/maven-api-core/src/main/java/org/apache/maven/api/SourceRoot.java
```diff
@@ -46,23 +47,47 @@ default Path directory() {
...
+     *
+     * <p>If the user did not specified any excludes, the default can be files generated
```
elharo, line 81: specified --> specify
can be --> is often

impl/maven-impl/src/test/java/org/apache/maven/impl/PathSelectorTest.java
```diff
@@ -0,0 +1,115 @@
...
+import org.junit.jupiter.api.io.TempDir;
+
+import static org.junit.jupiter.api.Assertions.assertFalse;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+public class PathSelectorTest {
+    /**
+     * The temporary directory containing the files to test.
+     */
+    private Path directory;
```
elharo, line 38: This should probably be a local variable since it differs from method to method

impl/maven-impl/src/test/java/org/apache/maven/impl/PathSelectorTest.java
```diff
@@ -0,0 +1,115 @@
...
+
+    /**
+     * The filtered set of paths. Created by {@link #filter()}.
+     */
+    private Set<Path> filtered;
```
elharo, line 43: Make this is a local variable. It's not a fixture since there's no setup, just shared global state between methods. Pass arguments instead.

elharo (changes requested):

api/maven-api-core/src/main/java/org/apache/maven/api/SourceRoot.java
desruisseaux: A plain `/` cannot be used in that particular place, because it is followed by `*` which is interpreted as end of comment. HTML entity `&sol;` is used as a workaround.

impl/maven-impl/src/test/java/org/apache/maven/impl/PathSelectorTest.java
desruisseaux: Done.

impl/maven-impl/src/test/java/org/apache/maven/impl/PathSelectorTest.java
desruisseaux: Done.

api/maven-api-core/src/main/java/org/apache/maven/api/SourceRoot.java
desruisseaux: Done. Will need to be rebased, as there is conflict with other changes added on `master` since that time.

impl/maven-impl/src/test/java/org/apache/maven/impl/PathSelectorTest.java
```diff
@@ -0,0 +1,108 @@
...
+     * @param path      the path to test
+     */
+    private static void assertFilteredFilesContains(Set<Path> filtered, Path directory, String path) {
+        assertTrue(filtered.remove(directory.resolve(path)), path);
+    }
+
+    /**
+     * Tests the omission of unnecessary excludes.
+     *
+     * Note: at the time of writing this test (April 2025), the list of excludes go down from 40 to 17 elements.
```
elharo, line 93: go --> goes

impl/maven-impl/src/test/java/org/apache/maven/impl/PathSelectorTest.java
```diff
@@ -0,0 +1,108 @@
...
+        Files.createFile(directory.resolve("root.txt"));
+        Files.createFile(bar.resolve("leaf.txt"));
+        Files.createFile(biz.resolve("excluded.txt"));
+
+        Set<Path> filtered = filter(directory, "");
+        assertFilteredFilesContains(filtered, directory, "root.txt");
+        assertFilteredFilesContains(filtered, directory, "foo/bar/leaf.txt");
+        assertTrue(filtered.isEmpty(), filtered.toString());
+
+        filtered = filter(directory, "glob:");
```
elharo, line 55: Try not to reuse local variables. It's quite error prone.

impl/maven-impl/src/test/java/org/apache/maven/impl/PathSelectorTest.java
```diff
@@ -0,0 +1,92 @@
...
+        assertFilteredFilesContains(directory, "glob:", "foo/bar/leaf.txt");
+    }
+
+    /**
+     * Asserts that the filtered set of paths contains the given items and nothing more.
+     *
+     * @param directory the temporary directory containing the files to test
+     * @param syntax syntax to test, either an empty string of {@code "glob:"}
+     * @param expected the expected paths
+     * @throws IOException if an error occurred while listing the files.
```
elharo, line 59: nit: no period

impl/maven-impl/src/test/java/org/apache/maven/impl/PathSelectorTest.java
```diff
@@ -0,0 +1,92 @@
...
+     * Creates a temporary directory and checks its list of content based on patterns.
+     *
+     * @param directory temporary directory where to create a tree
+     * @throws IOException if an error occurred while creating a temporary file or directory
+     */
+    @Test
+    public void testTree(final @TempDir Path directory) throws IOException {
+        Path foo = Files.createDirectory(directory.resolve("foo"));
+        Path bar = Files.createDirectory(foo.resolve("bar"));
+        Path biz = Files.createDirectory(directory.resolve("biz"));
```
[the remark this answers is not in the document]
desruisseaux: Done. I wasn't aware of the "biz" slang, thank for the info.

gnodet (approved):

gnodet:
@desruisseaux can you rebase or merge master branch to fix the conflict ?

desruisseaux:
Rebasing done.

impl/maven-impl/src/main/java/org/apache/maven/impl/PathSelector.java
```diff
@@ -0,0 +1,622 @@
...
+                }
+                if (start == 0) {
+                    end++; // Ommit the leading slash if there is nothing before it.
+                }
+            }
+            if (start > 0) {
+                if (pattern.charAt(--start) != '/') {
+                    continue;
+                }
+            }
```
Pankraz76, line 460:
```suggestion
            if (start > 0 && pattern.charAt(--start) != '/') {
                continue;
            }
```
could merge

Pankraz76 (approved):
polish

impl/maven-impl/src/test/java/org/apache/maven/impl/PathSelectorTest.java
```diff
@@ -0,0 +1,92 @@
...
+    @Test
+    public void testExcludeOmission() {
+        Path directory = Path.of("dummy");
+        var includes = List.of("**/*.java");
+        var excludes = List.of("baz/**");
+        var matcher = new PathSelector(directory, includes, excludes, true);
+        String s = matcher.toString();
+        assertTrue(s.contains("glob:**/*.java"));
+        assertFalse(s.contains("project.pj")); // Unnecessary exclusion should have been omitted.
+        assertFalse(s.contains(".DS_Store"));
```
Pankraz76, line 90:
```suggestion
        assertThat(s).contains(".DS_Store");
```
new code, already, considered, outdated - therefore increasing technical dept.

to reach next lvl, we need leadership and team up to standardize `DOD`.

image

consider migration to avoid boilerplate API usage.

- https://github.com/apache/maven/pull/2307
- https://docs.openrewrite.org/recipes/java/testing/assertj/junittoassertj

Pankraz76 (changes requested):
consider Migrate JUnit asserts to AssertJ

impl/maven-impl/src/main/java/org/apache/maven/impl/PathSelector.java
Pankraz76: ideally we could leverage https://kotlinlang.org/docs/basic-syntax.html#variables or https://projectlombok.org/features/val to get rid of `boilerplate`; obtain beloved `immutability`.

https://discuss.kotlinlang.org/t/why-not-have-everything-be-val-by-default/25391

gnodet: > ideally we could leverage `val` https://kotlinlang.org/docs/basic-syntax.html#variables or supplementation https://projectlombok.org/features/val, to get rid of `boilerplate`; obtain beloved `immutability`.
> 
> https://discuss.kotlinlang.org/t/why-not-have-everything-be-val-by-default/25391

We don't use kotlin, nor Lombok, so no.

impl/maven-impl/src/test/java/org/apache/maven/impl/PathSelectorTest.java
gnodet: Let's do the refactoring in the other PR related to AssertJ migration.

impl/maven-impl/src/main/java/org/apache/maven/impl/PathSelector.java
gnodet: @desruisseaux ? you wan to include this suggestion ?

impl/maven-impl/src/test/java/org/apache/maven/impl/PathSelectorTest.java
Pankraz76: yes

Pankraz76: this will serve our need:

https://docs.openrewrite.org/recipes/java/testing/junit5/junit5bestpractices

image

- https://github.com/apache/maven/pull/2320

impl/maven-impl/src/main/java/org/apache/maven/impl/PathSelector.java
desruisseaux: I have no objection. I was waiting for other comments in order to merge them in a single commit.

desruisseaux: I committed, but actually the previous code was done that way for consistency with the previous `if` block. The two blocks were following the same pattern:

```
if (end < length) {
    // Do some check on the suffix.
}
if (start > 0) {
    // Do some check on the prefix.
}
```

The check on the prefix is simpler than the check on the suffix, which is why it could be rewritten. But we lost the symmetry with the suffix case.

Pankraz76: this if looks like if else but thats off-topic, 

thanks for compression.

impl/maven-impl/src/main/java/org/apache/maven/impl/PathSelector.java
```diff
@@ -0,0 +1,620 @@
...
+            suffixes[i] = prefixOrSuffix(include, true);
+        }
+        prefixes = sortByLength(prefixes, false);
+        suffixes = sortByLength(suffixes, true);
+        /*
+         * Keep only the exclude which start with one of the prefixes and end with one of the suffixes.
+         * Note that a prefix or suffix may be the empty string, which match everything.
+         */
+        final Iterator<String> it = excludes.iterator();
+        nextExclude:
```
Pankraz76, line 291: Unused label 'nextExclude' 

image

impl/maven-impl/src/main/java/org/apache/maven/impl/PathSelector.java
```diff
@@ -0,0 +1,620 @@
...
+     * Sorts the given patterns by their length. The main intent is to have the empty string first,
+     * while will cause the loops testing for prefixes and suffixes to stop almost immediately.
+     * Short prefixes or suffixes are also more likely to be matched.
+     *
+     * @param fragments the fragments to sort in-place
+     * @param suffix {@code false} if the specified fragments are prefixes, {@code true} if they are suffixes
+     * @return the given array, or a smaller array if some fragments were discarded because redundant
+     */
+    private static String[] sortByLength(final String[] fragments, final boolean suffix) {
+        Arrays.sort(fragments, (s1, s2) -> s1.length() - s2.length());
```
Pankraz76, line 366: Can be replaced with 'Comparator.comparingInt' 

```suggestion
        Arrays.sort(fragments, Comparator.comparingInt(String::length));
```

image

impl/maven-impl/src/main/java/org/apache/maven/impl/PathSelector.java
```diff
@@ -0,0 +1,620 @@
...
+ * @see java.nio.file.FileSystem#getPathMatcher(String)
+ */
+public class PathSelector implements PathMatcher {
+    /**
+     * Patterns which should be excluded by default, like <abbr>SCM</abbr> files.
+     *
+     * <p><b>Source:</b> this list is copied from {@code plexus-utils-4.0.2} (released in
+     * September 23, 2024), class {@code org.codehaus.plexus.util.AbstractScanner}.</p>
+     */
+    private static final List<String> DEFAULT_EXCLUDES = List.of(
```
Pankraz76, line 71: might use simple alphabetical order sorter like in `.propertie` files - to get sort and grouping out of the box:

The comments mostly DRY.

```suggestion
private static final List<String> DEFAULT_EXCLUDES = List.of(
    "**/*~",
    "**/#*#",
    "**/.#*",
    "**/%*%",
    "**/._*",
    "**/CVS",
    "**/CVS/**",
    "**/.cvsignore",
    "**/RCS",
    "**/RCS/**",
    "**/SCCS",
    "**/SCCS/**",
    "**/vssver.scc",
    "**/project.pj",
    "**/.svn",
    "**/.svn/**",
    "**/.arch-ids",
    "**/.arch-ids/**",
    "**/.bzr",
    "**/.bzr/**",
    "**/.MySCMServerInfo",
    "**/.DS_Store",
    "**/.metadata",
    "**/.metadata/**",
    "**/.hg",
    "**/.hg/**",
    "**/.git",
    "**/.git/**",
    "**/.gitignore",
    "**/BitKeeper",
    "**/BitKeeper/**",
    "**/ChangeSet",
    "**/ChangeSet/**",
    "**/_darcs",
    "**/_darcs/**",
    "**/.darcsrepo",
    "**/.darcsrepo/**",
    "**/-darcs-backup*",
    "**/.darcs-temp-mail"
);
```

Pankraz76 (commented):
warnings

impl/maven-impl/src/main/java/org/apache/maven/impl/PathSelector.java
```diff
@@ -0,0 +1,620 @@
...
+     *
+     * @param directory the directory pathname to test, must not be {@code null}
+     * @return {@code true} if the given directory might contain selected paths, {@code false} if the
+     *         directory will definitively not contain selected paths
+     */
+    public boolean couldHoldSelected(Path directory) {
+        if (baseDirectory.equals(directory)) {
+            return true;
+        }
+        directory = baseDirectory.relativize(directory);
```
Pankraz76, line 584: smell never reassign params as we live in final land.

Pankraz76: image

```suggestion
    public boolean couldHoldSelected(Path directory) {
        if (baseDirectory.equals(directory)) {
            return true;
        }
        return isABoolean(baseDirectory.relativize(directory));
    }

    private boolean isABoolean(Path directory) {
        return dirIncludes.length == 0 || isMatched(directory, dirIncludes)
                && (dirExcludes.length == 0 || !isMatched(directory, dirExcludes));
    }
```

Pankraz76 (changes requested):
checkup

Pankraz76:
leveraging PMD will prevent flaws from being merged.

Pankraz76:
this could not compile with help of https://docs.openrewrite.org/recipes/staticanalysis/finalizemethodarguments

Pankraz76:
rewrite would make the boilerplate way i would prefer silent PMD.

https://adabeat.com/fp/immutability-in-functional-programming/

We make everywhere boilerplate to throw NPE ourselfs. This would really help to make thinks more robust. immutability is/should be the default case. Kotlin is all about being final and avoiding silly NPE. Thats why we want to live in final existence. Trying to reach invinity.

impl/maven-impl/src/main/java/org/apache/maven/impl/PathSelector.java
Pankraz76: https://github.com/checkstyle/checkstyle/pull/16760#discussion_r2084691281

gnodet:
> rewrite would make the boilerplate way i would prefer silent PMD.
> 
> 
> 
> https://adabeat.com/fp/immutability-in-functional-programming/
> 
> 
> 
> We make everywhere boilerplate to throw NPE ourselfs. This would really help to make thinks more robust. immutability is/should be the default case. Kotlin is all about being final and avoiding silly NPE. Thats why we want to live in final existence. Trying to reach invinity.

I don't think we want additional reports. Or that would nice as a GitHub action that runs on PRs.
Additional reports without any automatic way to fix the issue on a big codebase it just useless.

Pankraz76:
> useless

disagree, its called best practise for reason.

agree, with help of rewrite, we can get benefits for free. Enforcing really strong coding guidelines.

Pankraz76:
just need to use what we already got: 

image

I see plugin in pom, but not running. So im this should be out SPOT meaning single point of truth.  

https://pmd.github.io/pmd/pmd_rules_java_bestpractices.html

Then we dont have friction about final params. Im sure @desruisseaux very capable of following patterns to avoid bugs. We all want to best for the code.

gnodet:
> 



> > rewrite would make the boilerplate way i would prefer silent PMD.
> > https://adabeat.com/fp/immutability-in-functional-programming/
> > We make everywhere boilerplate to throw NPE ourselfs. This would really help to make thinks more robust. immutability is/should be the default case. Kotlin is all about being final and avoiding silly NPE. Thats why we want to live in final existence. Trying to reach invinity.
> 
> I don't think we want additional reports. Or that would nice as a GitHub action that runs on PRs. Additional reports without any automatic way to fix the issue on a big codebase it just useless.

Actually, they are available at https://maven.apache.org/ref/4.0.0-rc-3/maven-impl-modules/maven-core/pmd.html but I don't think we'll keep them on the long term.

Pankraz76:
> long term

yes, rewrite has ability to take over and dominate.

impl/maven-impl/src/main/java/org/apache/maven/impl/PathSelector.java
Pankraz76: Many languages use `val` and `var` as type it not first priority. less seems more to me.

- https://github.com/apache/maven/pull/2326
- 

image

desruisseaux:
If the discussion is about avoiding to reassign parameter values, those best practices are guidelines, often improving software but not always. For example, reassigning a parameter with something unrelated is dangerous. In that case, the best practice applies. But reassigning a parameter with a "fixed" (from the perspective of what the method wants to do) value can be good since it guarantees that the potentially broken value is never used. For example, if a method said that it ignores leading and trailing spaces, doing `arg = arg.strip()` at the beginning guarantees the "wrong" (from method's perspective) value with spaces will never be used.

Pankraz76:
> not always

yes, will always allow suppression making exception, for good argument.

impl/maven-impl/src/main/java/org/apache/maven/impl/PathSelector.java
Pankraz76: important `AvoidReassigningParameters` is of course part of `bestpractices `.

Still not encountered in

image

PR #2254, opened 2025-04-11
[MNG-8685] Better support for CI systems

cstamas:
Improve CI detection of Maven, make possible to propagate some options from CI to Maven.

Changes:
* `InvokerRequest` carries new info: `Optional<CIInfo>`: if present, we run on CI, if empty, not.
* `Parser` (creating `InvokerRequest`) already inspects env and assembles invoker request, so make it detect CI as well
* `CIDetector` is a Java Service (this all happens early, no DI yet) that can be extended (like adding CI specific detectors). Core had one implementation that became the "generic" that is what Maven 4 had so far.
* Added "jenkins", "github", "teamcity", "circle" and "travis" support.

---

https://issues.apache.org/jira/browse/MNG-8685


impl/maven-cli/src/main/java/org/apache/maven/cling/invoker/cisupport/GenericCIDetector.java
```diff
@@ -0,0 +1,50 @@
...
+        if (ciEnv != null && !"false".equals(ciEnv)) {
+            return Optional.of(new CISupport() {
+                @Override
+                public String name() {
+                    return NAME;
+                }
+
+                @Override
+                public String message() {
+                    return "Environment variable CI equals \"true\". Disable detection by removing that variable or by setting it to any other value";
```
pzygielo, line 44: That message does not match the check in l. 35.

cstamas: Right, but this message is exactly (along with check) what was in place :smile:

pzygielo: Reviewed it now (https://github.com/apache/maven/pull/1750/files#r2039640411).
Is separate JIRA required for this to be changed? MNG-8283 seems to be RC-released already.

pzygielo: > Is separate JIRA required for this to be changed?

Perhaps, because this PR might or might not be integrated after all.

cstamas: Hm, I _assumed_ this (as is) comes from Maven 3. If not, then yes, let's make it right.

cstamas:
Fun fact: all 3 failing ITs tries to "escape" from CI env by setting `CI=false`, but alas, it does not work anymore, as `GithubCIDetector` kicks in.

impl/maven-cli/src/main/java/org/apache/maven/cling/invoker/cisupport/GenericCIDetector.java
cstamas: Ok, the original comes from https://github.com/apache/maven/commit/ac0bc5541e3d1a19b22c4138c354237c8202e11f

So def not Maven 3.

impl/maven-cli/src/main/java/org/apache/maven/cling/invoker/cisupport/CIDetectorHelper.java
```diff
@@ -0,0 +1,51 @@
...
+ * CI detector helper: it uses service discovery to discover {@link CIDetector}s. If resulting list has more than
+ * one element, it will remove the {@link GenericCIDetector} result, assuming a more specific one is also present.
+ */
+public final class CIDetectorHelper {
+    private CIDetectorHelper() {}
+
+    public static List<CIInfo> detectCI() {
+        List<CIDetector> detectors = ServiceLoader.load(CIDetector.class).stream()
+                .map(ServiceLoader.Provider::get)
+                .toList();
```
pzygielo, line 38: Why not continue with this stream to get `List<CIInfo>` and get rid of the loop?

api/maven-api-cli/src/main/java/org/apache/maven/api/cli/cisupport/CIInfo.java
```diff
@@ -0,0 +1,47 @@
...
+     */
+    @Nonnull
+    default String message() {
+        return "";
+    }
+
+    /**
+     * Some CI systems may allow running jobs in "debug" (or some equivalent) mode.
+     */
+    default boolean isDebug() {
```
slawekjaranowski, line 44: in Maven context we use `verbose`, like `effectiveVerbose` and so on ...

maybe also call it `isVerbose`

impl/maven-cli/src/main/resources/META-INF/services/org.apache.maven.cling.invoker.cisupport.CIDetector
```diff
@@ -0,0 +1,6 @@
+org.apache.maven.cling.invoker.cisupport.CircleCIDetector
+org.apache.maven.cling.invoker.cisupport.GenericCIDetector
+org.apache.maven.cling.invoker.cisupport.GithubCIDetector
+org.apache.maven.cling.invoker.cisupport.JenkinsCIDetector
+org.apache.maven.cling.invoker.cisupport.TeamcityCIDetector
+org.apache.maven.cling.invoker.cisupport.TravisCIDetector
```
slawekjaranowski, line 6: missing new line at the end is by purpose?

impl/maven-cli/src/test/java/org/apache/maven/cling/invoker/cisupport/CIDetectorHelperTest.java
```diff
@@ -0,0 +1,77 @@
...
+import java.util.Collection;
+import java.util.List;
+import java.util.Map;
+
+import org.apache.maven.impl.util.Os;
+import org.junit.jupiter.api.Test;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+public class CIDetectorHelperTest {
```
slawekjaranowski, line 31: Missing tests for Circle, TeamCity, Travis

impl/maven-cli/src/main/java/org/apache/maven/cling/invoker/cisupport/CIDetectorHelper.java
cstamas: Fixed

api/maven-api-cli/src/main/java/org/apache/maven/api/cli/cisupport/CIInfo.java
cstamas: Fixed

impl/maven-cli/src/main/resources/META-INF/services/org.apache.maven.cling.invoker.cisupport.CIDetector
cstamas: Fixed

impl/maven-cli/src/test/java/org/apache/maven/cling/invoker/cisupport/CIDetectorHelperTest.java
cstamas: All missing tests added

impl/maven-cli/src/main/java/org/apache/maven/cling/invoker/cisupport/CIDetectorHelper.java
```diff
@@ -33,15 +33,13 @@ public final class CIDetectorHelper {
...
 
     public static List<CIInfo> detectCI() {
-        List<CIDetector> detectors = ServiceLoader.load(CIDetector.class).stream()
+        List<CIInfo> result = new ArrayList<>(ServiceLoader.load(CIDetector.class).stream()
                 .map(ServiceLoader.Provider::get)
-                .toList();
+                .map(CIDetector::detectCI)
+                .filter(Optional::isPresent)
+                .map(Optional::get)
+                .toList());
```
pzygielo, line 41: https://docs.oracle.com/en/java/javase/21/docs/api/java.base/java/util/stream/Stream.html#toList():
> The returned List is unmodifiable

so not sure about `removeIf` later.

Maybe `Collectors.toCollection(ArrayList::new)`?

slawekjaranowski (approved):

impl/maven-cli/src/main/java/org/apache/maven/cling/invoker/cisupport/CIDetectorHelper.java
cstamas: I think is fine, this happens "on boot", is not some sort of "hot" spot. The reason why I remove "generic" if there are 2 or more is that I assume there is more specific hit (like GH, sets `CI=true` but also sets `GITHUB_RUNNER=true`)... we can refine this later anyway.

cstamas: Also, to keep things simple I don't want to fiddle with order, as other solution would be to test all, and THEN as last, if no hit, "generic"... but that would assume there is no (and never will be) overlap between all CI systems, which IMO we never know and cannot know (future wise).

cstamas:
This change went into the master cli module, so will be out as 4.0.0-rc-4 (next planned release). You can try backport it if you want, and I can help with that, but it will not be a simple thing (like just apply patches). The whole cli was redone in 4...

cstamas:
In short, for mvn3 I'd envision some PR that is more close to your original PR (and adds verbose support for those that support it: GH and Travis IIRC?)

PR #2257, opened 2025-04-13
[MNG-8676] Improve model builder error messages

cstamas:
Only the one that is easy to be mixed up with GAV: DepMgtKey is GAT and not GAV. The rest (XML, PK/GA, REPOID, DIR, GAV) ones are not quite mixable, and string "A:B:C" is usually always assumed is GAV.

The problem with these two are that they look pretty much same (both a "A:B:C", colon segmented string of 3 or 4), but in one case C is dependency type, in other is version.

This PR now lessens the "A:B:C" string overload a bit, and also makes crystal clear that T is not V. The message is precise, but a bit longer. In turn, is not misleading at all now.

---

https://issues.apache.org/jira/browse/MNG-8676

### DIFF
--- a/impl/maven-core/src/test/java/org/apache/maven/project/ProjectBuilderTest.java
+++ b/impl/maven-core/src/test/java/org/apache/maven/project/ProjectBuilderTest.java
@@ -100,8 +100,9 @@ class ProjectBuilderTest extends AbstractCoreMavenComponentTestCase {
                 .build(pomFile, configuration));
         assertThat(
                 e.getResults(),
-                contains(projectBuildingResultWithProblemMessage(
-                        "'dependencies.dependency.version' for org.apache.maven.its:a:jar (GAT) is missing")));
+                contains(
+                        projectBuildingResultWithProblemMessage(
+                                "'dependencies.dependency.version' for g='org.apache.maven.its', a='a', type='jar' is missing")));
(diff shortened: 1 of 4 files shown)


cstamas:
Am unsure about this PR. The reason why it exists is explained in issue, but in short, the "a:b:c" string is _usually implied_ to be GAV, but is also _overloaded_ in Maven as, while many times it _is_ GAV, sometimes is not, but muscle reflex will read it as GAV, and this will cause issues and derailment later on.

Maybe all we need is "just" change the syntax in case of GAT to something else? Like `G:A:V` vs `G:A@T`? To make it visually apparent "is not GAV"?

gnodet:
> Am unsure about this PR. The and and only reason why it exists is explained in issue, but in short, the "a:b:c" string is _usually implied_ to be GAV, but is also _overloaded_ in Maven as, while many times it _is_ GAV, sometimes is not, but muscle reflex will read it as GAV, and this will cause issues and derailment later on.
> 
> Maybe all we need is "just" change the syntax in case of GAT to something else? Like `G:A:V` vs `G:A@T` **in the message only**? To make it visually apparent "is not GAV"?

It would make sense to make sure we use a single syntax everywhere.  Not sure that `G:A@T` would be appropriate, as we sometime use the full syntax `G:A:V:T:E` I think.  So in this case, maybe `G:A::T`

Another problem would be goals...

gnodet (approved):

PR #2272, opened 2025-04-23
[MNG-8694] Fix version interpolation and ternary operator

kamilkrzywanski:
Fix for https://issues.apache.org/jira/browse/MNG-8694
@slawekjaranowski as mentioned today on jug challenge accepted!

### DIFF
--- a/impl/maven-core/src/test/resources/projects/pom-parent-version-from-nested-properties/pom.xml
+++ b/impl/maven-core/src/test/resources/projects/pom-parent-version-from-nested-properties/pom.xml
@@ -1,0 +1,14 @@
+<project xmlns="http://maven.apache.org/POM/4.1.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+  xsi:schemaLocation="http://maven.apache.org/POM/4.1.0 http://maven.apache.org/xsd/maven-4.1.0.xsd">
+  <modelVersion>4.0.0</modelVersion>
+
+  <groupId>gid</groupId>
+  <artifactId>aid</artifactId>
+  <version>${revision}</version>
+  <packaging>pom</packaging>
+
+  <properties>
+    <versionCore>0.1.0</versionCore>
+    <revision>${versionCore}</revision>
+  </properties>
+</project>
--- a/impl/maven-core/src/test/resources/projects/pom-parent-version-from-nested-properties/pom.xml
+++ b/impl/maven-core/src/test/resources/projects/pom-parent-version-from-nested-properties/pom.xml
@@ -11,9 +11,7 @@
 
   <properties>
     <versionCore>0.1.0</versionCore>
-    <releaseProp>-RELEASE</releaseProp>
-    <release>${releaseProp:--DEVELOPER}</release>
-    <revision>${versionCore}${release:--NOT_DEV}</revision>
+    <revision>${versionCore}${release:+:--DEVELOPER}</revision>
   </properties>
 
 </project>
--- a/impl/maven-core/src/test/resources/projects/pom-parent-version-from-nested-properties/pom.xml
+++ b/impl/maven-core/src/test/resources/projects/pom-parent-version-from-nested-properties/pom.xml
@@ -1,6 +1,6 @@
 <project xmlns="http://maven.apache.org/POM/4.1.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
   xsi:schemaLocation="http://maven.apache.org/POM/4.1.0 http://maven.apache.org/xsd/maven-4.1.0.xsd">
-  <modelVersion>4.0.0</modelVersion>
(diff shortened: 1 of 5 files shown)


impl/maven-core/src/test/resources/projects/pom-parent-version-from-nested-properties/pom.xml
```diff
@@ -0,0 +1,14 @@
+<project xmlns="http://maven.apache.org/POM/4.1.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+  xsi:schemaLocation="http://maven.apache.org/POM/4.1.0 http://maven.apache.org/xsd/maven-4.1.0.xsd">
+  <modelVersion>4.0.0</modelVersion>
```
gnodet, line 3: Please change this line to 4.1.0.
This should make the test fail with:
```
org.apache.maven.project.ProjectBuildingException: 1 problem was 
    - [FATAL] 'version' contains an expression but should be a constant. @ line 7, column 3 for project  at .../maven/impl/maven-core/src/test/resources/projects/pom-parent-version-from-nested-properties/pom.xml
```

The reason is that the 4.1.0 model validation enforces the fact that version can be completely resolved.
I think we need to use the `org.apache.maven.impl.model.DefaultInterpolator` service as indicated in the TODO above

gnodet (changes requested):
I don't think the behavior is complete.
It could be solved by a loop, but preferably by using the DefaultInterpolator service.

impl/maven-impl/src/main/java/org/apache/maven/impl/model/DefaultModelBuilder.java
```diff
@@ -597,8 +597,8 @@ String replaceCiFriendlyVersion(Map<String, String> properties, String version)
                         // extract the key inside ${}
                         String key = matcher.group(1);
```
gnodet, line 598: Actually the whole method can be replaced easily with:
```
        String replaceCiFriendlyVersion(Map<String, String> properties, String version) {
            return version != null ? interpolator.interpolate(version, properties::get) : null;
        }
```

impl/maven-core/src/test/resources/projects/pom-parent-version-from-nested-properties/pom.xml
kamilkrzywanski: Right, fails

impl/maven-impl/src/main/java/org/apache/maven/impl/model/DefaultModelBuilder.java
kamilkrzywanski: Right, now works with 4.0.0 and 4.1.0

gnodet: Yes, though I'm not sure it will completely solve the original problem which is to provide the version using a profile.  This may require to use a different mechanism, such as default values with the support for `:-` and `:+` already provided by the interpolator.

gnodet (approved):

kamilkrzywanski:
@gnodet Removed unused regex as well, can you approve again?

slawekjaranowski (approved):

gnodet:
I still think this only partially address the use case from MNG-8694, so IMHO we'd need to extend the test to support a similar use case.  Probably not using a profile, since it won't work, but using a value such as `${coreVersion}${release:--DEVELOPER}` or something like that.

kamilkrzywanski:
@gnodet I've created an example like this:
However, it fails with: org.apache.maven.api.services.InterpolatorException: recursive variable reference: release.
I'll investigate further.
```

<project xmlns="http://maven.apache.org/POM/4.1.0"
         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
         xsi:schemaLocation="http://maven.apache.org/POM/4.1.0 http://maven.apache.org/xsd/maven-4.1.0.xsd"
        >

    <groupId>de.g667.m4-demo</groupId>
    <artifactId>m4-demo-revision</artifactId>
    <version>${revision}</version>
    <packaging>pom</packaging>

    <properties>
        <versionCore>0.1.0</versionCore>
        <release>${release:--DEVELOPER}</release>
        <revision>${versionCore}${release}</revision>
    </properties>

</project>

```

kamilkrzywanski:
@gnodet Test case extebded

gnodet:
> @gnodet Test case extended

Thx, I slightly modified it as one of the problem was the generation of an empty string when `release` is defined to `true` (which is what happens when using `-Drelease` on the command line for example).

This PR now fixes the ternary operator to support expressions such as `${release:+${foo}:-${bar}}` or `${versionCore}${release:+-DEVELOPER}`.

jira-importer:
Resolve #9611

PR #2274, opened 2025-04-24
[MNG-8572] Support DI beans in build extensions

gnodet:
This PR adds support for using the Maven API DI system in build extensions (plugins with extensions=true). It enables build extensions to provide custom artifact type handlers using the new DI mechanism.

## Key changes
- Added an integration test that demonstrates a custom artifact type handler using the Maven API DI system in a build extension
- The test shows how to implement a TypeProvider using `@Named` annotation
- The annotation processor automatically generates the DI index file
- The custom type handler is properly discovered and used during the build

This enhancement allows build extensions to leverage the new Maven API DI system, making it easier to create and maintain extensions that provide custom artifact type handlers.

Fixes: https://issues.apache.org/jira/browse/MNG-8572

### DIFF
--- a/impl/maven-core/src/main/java/org/apache/maven/internal/impl/DefaultTypeRegistry.java
+++ b/impl/maven-core/src/main/java/org/apache/maven/internal/impl/DefaultTypeRegistry.java
@@ -22,16 +22,15 @@ import javax.inject.Inject;
 import javax.inject.Named;
 import javax.inject.Singleton;
 
-import java.util.List;
-import java.util.Map;
+import java.util.Objects;
 import java.util.Optional;
 import java.util.concurrent.ConcurrentHashMap;
-import java.util.stream.Collectors;
 
 import org.apache.maven.api.JavaPathType;
 import org.apache.maven.api.Type;
 import org.apache.maven.api.annotations.Nonnull;
 import org.apache.maven.api.services.LanguageRegistry;
+import org.apache.maven.api.services.Lookup;
 import org.apache.maven.api.services.TypeRegistry;
 import org.apache.maven.api.spi.TypeProvider;
 import org.apache.maven.artifact.handler.ArtifactHandler;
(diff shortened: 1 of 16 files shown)


impl/maven-core/src/main/java/org/apache/maven/internal/impl/DefaultTypeRegistry.java
```diff
@@ -55,11 +53,8 @@ public class DefaultTypeRegistry extends AbstractEventSpy implements TypeRegistr
     private final LegacyArtifactHandlerManager manager;
 
     @Inject
-    public DefaultTypeRegistry(
-            List<TypeProvider> providers, LanguageRegistry languageRegistry, LegacyArtifactHandlerManager manager) {
-        this.types = requireNonNull(providers, "providers cannot be null").stream()
-                .flatMap(p -> p.provides().stream())
-                .collect(Collectors.toMap(Type::id, identity()));
+    public DefaultTypeRegistry(Lookup lookup, LanguageRegistry languageRegistry, LegacyArtifactHandlerManager manager) {
+        this.lookup = lookup;
```
cstamas, line 57: nullable?

cstamas (approved):

impl/maven-core/src/main/java/org/apache/maven/internal/impl/DefaultTypeRegistry.java
gnodet: AFAIK, nullable must be opt-in for injection, so no annotation here means the dependencies are required.

cstamas: I was more about requireNonNull ... As other ctor params

gnodet, line 57:
```suggestion
        this.lookup = requireNonNull(lookup, "lookup cannot be null");
```

jira-importer:
Resolve #9369

PR #2275, opened 2025-04-25
[MNG-8707] Add methods to remove compile and test source roots

gnodet:
JIRA issue: [MNG-8707](https://issues.apache.org/jira/browse/MNG-8707)

# Add methods to remove compile and test source roots

## Overview
This PR adds new methods to the `MavenProject` class to allow removing source roots, complementing the existing methods for adding them. This provides a more complete API for managing source roots in Maven projects.

## Changes
- Added `removeCompileSourceRoot(String)` method to remove a compile source root
- Added `removeTestCompileSourceRoot(String)` method to remove a test compile source root
- Added `removePath(List<String>, String)` utility method to handle path removal
- Extracted common path handling logic into a new `sanitizePath(String)` method
- Added proper Javadoc with `@since 3.9.10` tags to new methods
- Added warning mechanism when source root collections are modified directly
- Improved code readability by replacing `path.length() > 0` with `!path.isEmpty()`

## Implementation Details
- The new methods mirror the behavior of the existing add methods
- Path sanitization is consistent between add and remove operations
- Added a `LoggingList` wrapper that logs warnings when collections are modified directly
- Warnings can be disabled via the system property `maven.project.sourceRoots.warningsDisabled`
- Warning messages indicate that direct modification will not work in Maven 4.0.0

## Motivation
These changes provide a more complete API for managing source roots, which is particularly useful for plugins that need to manipulate the source roots of a project during the build process.

## Compatibility
This change is fully backward compatible. The new methods are marked with `@since 3.9.10` to indicate when they were added.

### DIFF
--- a/maven-core/src/main/java/org/apache/maven/project/MavenProject.java
+++ b/maven-core/src/main/java/org/apache/maven/project/MavenProject.java
@@ -320,7 +320,7 @@ public class MavenProject implements Cloneable {
      * @param path the source root to add
      */
     public void addCompileSourceRoot(String path) {
-        addPath(getCompileSourceRoots(), path);
+        addPath(compileSourceRoots, path);
     }
 
     /**
@@ -330,7 +330,7 @@ public class MavenProject implements Cloneable {
      * @since 3.9.10
      */
     public void removeCompileSourceRoot(String path) {
-        removePath(getCompileSourceRoots(), path);
+        removePath(compileSourceRoots, path);
     }
 
     /**
@@ -339,7 +339,7 @@ public class MavenProject implements Cloneable {
      * @param path the test source root to add
      */
     public void addTestCompileSourceRoot(String path) {
-        addPath(getTestCompileSourceRoots(), path);
+        addPath(testCompileSourceRoots, path);
     }
 
     /**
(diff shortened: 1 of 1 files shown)


maven-core/src/main/java/org/apache/maven/project/MavenProject.java
```diff
@@ -296,14 +296,40 @@ private void addPath(List<String> paths, String path) {
...
+                paths.remove(path);
+            }
+        }
+    }
+
     public void addCompileSourceRoot(String path) {
         addPath(getCompileSourceRoots(), path);
     }
 
+    public void removeCompileSourceRoot(String path) {
```
slawekjaranowski, line 321: javadoc with since will be appreciated

maven-core/src/main/java/org/apache/maven/project/MavenProject.java
```diff
@@ -1120,6 +1188,185 @@ private static String getProjectReferenceId(String groupId, String artifactId, S
...
+        private final String collectionName;
+
+        LoggingList(List<E> delegate, String collectionName) {
+            this.delegate = delegate;
+            this.collectionName = collectionName;
+        }
+
+        private void logWarning(String method) {
+            // Check if warnings are disabled
+            if (Boolean.getBoolean(DISABLE_WARNINGS_PROPERTY)) {
```
slawekjaranowski, line 1211: Can we also check project properties here?
When we only check system properties - users will can only set it on cli ...

gnodet: Done

slawekjaranowski (approved):

PR #2276, opened 2025-04-25
[MNGSITE-393] Remove Maven 1 and 2 references

elharo:


api/maven-api-model/src/main/mdo/maven.mdo
```diff
@@ -3384,10 +3383,6 @@
           <description><![CDATA[
             For a plugin project (packaging is {@code maven-plugin}), the minimum version of
             Maven required to use the resulting plugin.<br>
```
desruisseaux, line 3385: We can remove `<br>` if there is no longer another paragraph after it.

api/maven-api-model/src/main/mdo/maven.mdo
```diff
@@ -3384,10 +3383,6 @@
           <description><![CDATA[
             For a plugin project (packaging is {@code maven-plugin}), the minimum version of
```
desruisseaux, line 3384: Shouldn't `<defaultValue>2.0</defaultValue>` (two lines up) be changed to `<defaultValue>3.0</defaultValue>` then?

gnodet: Not sure, I'd rather avoid any change of default values in model 4.0.0.   But maybe we could change the default to `4.0` for model `4.1.0`.  Though this may not be possible to do that in the parser.  So not sure it's worth it.

api/maven-api-model/src/main/mdo/maven.mdo
elharo: done

api/maven-api-model/src/main/mdo/maven.mdo
elharo: That makes sense, but this is a doc only change. Code changes need a separate issue and PR.

michael-o (approved):

PR #2278, opened 2025-04-27
[MNG-8713] SourceRoot.directory() default value should include the module when present

desruisseaux:
If a `<source>` element in POM does not provide any value for the `<directory>` element, the default value is currently defined as `src/${scope}/${lang}`. This default value provides the familiar `src/main/java` and `src/test/java` default directories. However, if the `<source>` element contains a `<module>` element, the module name must appear somewhere in the default directory name, otherwise the source code of multi-modular projects will collide.

Different conventions are possible:

* OpenJDK uses `src/${module}/` for the main code and a bazar of directories for the tests with no obvious order.
* [Project Jigsaw: Module System Quick-Start Guide](https://openjdk.org/projects/jigsaw/quick-start) uses `src/${module}/` for main code and does not mention tests.
* [junit5-modular-world](https://github.com/junit-team/junit5-samples/tree/main/junit5-modular-world) uses `src/main/${module}/` for the main code and `src/test/${module}/` for the test code. However, it also uses `module-info.java` in test code, which I would like to discourage.
* NetBeans Ant modular project uses `src/${module}/main` for the main code and `src/${module}/test` for the test code.

I propose the NetBeans's convention, which is implemented in this pull request as a default directory of `src/${lang}/${module}/${scope}` when the `<module>` element is present.

## Alternative
If it is considered too premature to adopt a default directory name for modular project, we should at least throw an exception asking user to specify a directory explicitly if a `<module>` element is present.

### DIFF
--- a/api/maven-api-core/src/main/java/org/apache/maven/api/SourceRoot.java
+++ b/api/maven-api-core/src/main/java/org/apache/maven/api/SourceRoot.java
@@ -44,7 +44,7 @@ public interface SourceRoot {
      *     <code>src/{@linkplain #scope() scope}/{@linkplain #language() language}</code>.
      *   </li><li>
      *     If a module name is present, then the default directory is
-     *     <code>src/{@linkplain #language() language}/{@linkplain #module() module}/{@linkplain #scope() scope}</code>.
+     *     <code>src/{@linkplain #module() module}/{@linkplain #scope() scope}/{@linkplain #language() language}</code>.
      *   </li>
      * </ul>
      *
@@ -52,8 +52,11 @@ public interface SourceRoot {
      * Implementation may override with absolute path instead.
      */
     default Path directory() {
-        return module().map((module) -> Path.of("src", language().id(), module, scope().id()))
-                .orElseGet(() -> Path.of("src", scope().id(), language().id()));
+        Path src = Path.of("src");
+        return module().map(src::resolve)
+                .orElse(src)
+                .resolve(scope().id())
+                .resolve(language().id());
     }
 
     /**
--- a/api/maven-api-core/src/main/java/org/apache/maven/api/SourceRoot.java
+++ b/api/maven-api-core/src/main/java/org/apache/maven/api/SourceRoot.java
@@ -37,11 +37,23 @@ public interface SourceRoot {
      * The path is relative to the <abbr>POM</abbr> file.
      *
      * <h4>Default implementation</h4>
-     * The default value is <code>src/{@linkplain #scope() scope}/{@linkplain #language() language}</code>
-     * as a relative path. Implementation classes may override this default with an absolute path instead.
+     * The default value depends on whether a {@linkplain #module() module name} is specified in this source root:
+     * <ul>
+     *   <li>
+     *     If no module name, then the default directory is
+     *     <code>src/{@linkplain #scope() scope}/{@linkplain #language() language}</code>.
+     *   </li><li>
+     *     If a module name is present, then the default directory is
+     *     <code>src/{@linkplain #language() language}/{@linkplain #module() module}/{@linkplain #scope() scope}</code>.
+     *   </li>
+     * </ul>
+     *
+     * These default values are relative directories.
+     * Implementation classes may override this default with absolute paths instead.
      */
     default Path directory() {
-        return Path.of("src", scope().id(), language().id());
+        return module().map((module) -> Path.of("src", language().id(), module, scope().id()))
+                .orElseGet(() -> Path.of("src", scope().id(), language().id()));
     }
 
     /**
--- a/api/maven-api-core/src/main/java/org/apache/maven/api/SourceRoot.java
+++ b/api/maven-api-core/src/main/java/org/apache/maven/api/SourceRoot.java
@@ -48,8 +48,8 @@ public interface SourceRoot {
      *   </li>
      * </ul>
      *
-     * These default values are relative directories.
-     * Implementation classes may override this default with absolute paths instead.
+     * The default value is relative.
+     * Implementation may override with absolute path instead.
      */
     default Path directory() {
         return module().map((module) -> Path.of("src", language().id(), module, scope().id()))
--- a/impl/maven-impl/src/main/java/org/apache/maven/impl/DefaultSourceRoot.java
+++ b/impl/maven-impl/src/main/java/org/apache/maven/impl/DefaultSourceRoot.java
@@ -79,7 +79,12 @@ public final class DefaultSourceRoot implements SourceRoot {
         if (value != null) {
             directory = baseDir.resolve(value);
         } else {
-            directory = baseDir.resolve("src").resolve(scope.id()).resolve(language.id());
+            Path src = baseDir.resolve("src");
+            if (moduleName != null) {
+                directory = src.resolve(language.id()).resolve(moduleName).resolve(scope.id());
+            } else {
+                directory = src.resolve(scope.id()).resolve(language.id());
+            }
         }
 
         value = nonBlank(source.getTargetVersion());
--- a/impl/maven-impl/src/main/java/org/apache/maven/impl/DefaultSourceRoot.java
+++ b/impl/maven-impl/src/main/java/org/apache/maven/impl/DefaultSourceRoot.java
@@ -81,10 +81,9 @@ public final class DefaultSourceRoot implements SourceRoot {
         } else {
             Path src = baseDir.resolve("src");
             if (moduleName != null) {
-                directory = src.resolve(language.id()).resolve(moduleName).resolve(scope.id());
-            } else {
-                directory = src.resolve(scope.id()).resolve(language.id());
+                src = src.resolve(language.id());
             }
+            directory = src.resolve(scope.id()).resolve(language.id());
         }
 
         value = nonBlank(source.getTargetVersion());
--- a/api/maven-api-model/src/main/mdo/maven.mdo
+++ b/api/maven-api-model/src/main/mdo/maven.mdo
@@ -2014,7 +2014,7 @@
         If no source directories are specified, the default values depend on whether module names are specified:
         <ul>
           <li>{@code src/${scope}/${lang}} if no module is specified</li>
-          <li>{@code src/${lang}/${module}/${scope}} if a module is specified</li>
+          <li>{@code src/${module}/${scope}/${lang}} if a module is specified</li>
         </ul>
         where
         {@code ${scope}} is the value of the {@link #scope} element (typically {@code main} or {@code test}),
--- a/api/maven-api-model/src/main/mdo/maven.mdo
+++ b/api/maven-api-model/src/main/mdo/maven.mdo
@@ -2011,9 +2011,15 @@
         usage (main code, tests, <i>etc.</i>) is specified by the {@code scope} element.
 
         <h2>Default source directories</h2>
-        If no source directories are specified, the defaults are {@code src/${scope}/${lang}} where
-        {@code ${scope}} is the value of the {@link #scope} element (typically {@code main} or {@code test}) and
-        {@code ${lang}} is the value of the {@link #lang} element (typically {@code java} or {@code resources}).
+        If no source directories are specified, the default values depend on whether module names are specified:
+        <ul>
+          <li>{@code src/${scope}/${lang}} if no module is specified</li>
+          <li>{@code src/${lang}/${module}/${scope}} if a module is specified</li>
+        </ul>
+        where
+        {@code ${scope}} is the value of the {@link #scope} element (typically {@code main} or {@code test}),
+        {@code ${lang}} is the value of the {@link #lang} element (typically {@code java} or {@code resources}),
+        and {@code ${module}} is the optional {@link #module} element.
         ]]>
       </description>
       <version>4.1.0+</version>


impl/maven-impl/src/main/java/org/apache/maven/impl/DefaultSourceRoot.java
```diff
@@ -79,7 +79,12 @@ public DefaultSourceRoot(final Session session, final Path baseDir, final Source
         if (value != null) {
             directory = baseDir.resolve(value);
         } else {
-            directory = baseDir.resolve("src").resolve(scope.id()).resolve(language.id());
+            var src = baseDir.resolve("src");
```
cstamas, line 82: Do we use `var` in maven codebase? I kinda remember we agreed not to? (but I may be missing some later agreement).

desruisseaux: Right, it was my mistake.

Personally, I sometime use `var` in expressions such as `var x = new Foo()` when there is no way for the type to be something else than what we can read on the line, and usually refrain to use it in other contexts. I do not know if there is a Maven policy.

elharo (changes requested):
A change this significant requires a linked JIRA issue.

desruisseaux:
Done: https://issues.apache.org/jira/browse/MNG-8713

api/maven-api-core/src/main/java/org/apache/maven/api/SourceRoot.java
```diff
@@ -37,11 +37,23 @@ public interface SourceRoot {
      * The path is relative to the <abbr>POM</abbr> file.
      *
      * <h4>Default implementation</h4>
-     * The default value is <code>src/{@linkplain #scope() scope}/{@linkplain #language() language}</code>
-     * as a relative path. Implementation classes may override this default with an absolute path instead.
+     * The default value depends on whether a {@linkplain #module() module name} is specified in this source root:
+     * <ul>
+     *   <li>
+     *     If no module name, then the default directory is
```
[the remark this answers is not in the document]
desruisseaux: Reminder: this pull request does not change the convention of non-modular projects or projects that do not use the [module source hierarchy](https://docs.oracle.com/en/java/javase/24/docs/specs/man/javac.html#module-source-hierarchy): for them, the default stay `src/main/java`. Only for projects that use the module source hierarchy, we would "flip" from _(nothing that exist currently in Maven)_ to `src/java/org.my.module/main`.

The reason for putting `${lang}` before `${module}` is because the module names tend to be language-dependent. For example, Java uses URL domain names (e.g. `org.apache.maven`) while Python uses [shorter module names](https://www.programcreek.com/python/index/module/list), without `org` or `com` prefix. If a Maven subproject mixes Java and Python modules, placing the modules of the two languages in the same directory is confusing.

The reason for putting `${scope}` after `${module}` is for grouping together the files related to each module. This is the same division as Maven subprojects (previously named "modules"): each subproject contains main, test, resource and pom.xml files. Likewise, each Java module would contain its own main, test, resource files, with module-info somewhat playing the role of pom.xml. In other words, we are still splitting the projects in modules, but shifting from a world where modules are managed by Maven to a world were modules are managed by Java.

There is precedents: `${module}/${scope}` is the directory layout of NetBeans's Ant modular project, of [Christian Stein's blog](https://github.com/sormuras/java-module-patching?tab=readme-ov-file#directory-layout-structure) (except for `${lang}`) and of the example given in [Oracle's documentation](https://docs.oracle.com/en/java/javase/24/docs/specs/man/javac.html#module-source-hierarchy) (except for `src` and `${lang}`).

impl/maven-impl/src/main/java/org/apache/maven/impl/DefaultSourceRoot.java
elharo: I strongly prefer to avoid var. IMHO it should never have been added to the language. I want to know the types of things when I'm reading the code. var hides that info. I also don't want method return types to change without explicit consideration and inspection of every call point.

gnodet (approved):

api/maven-api-core/src/main/java/org/apache/maven/api/SourceRoot.java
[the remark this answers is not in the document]
gnodet: That makes sense to me.  @elharo ?

cstamas (approved):

api/maven-api-core/src/main/java/org/apache/maven/api/SourceRoot.java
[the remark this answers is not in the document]
elharo: I am not convinced. That feels very wrong to me, and I expect users will be unnecessarily confused. What we've had in the past is src/main/java/package-structure and the natural extension of that is src/main/java/module/package-structure

If other languages come into play (which I really don;t think will happen or is something we need to worry about) then src/main/python/module or whatever

Changing anything to src/java/main introduces inconsistency we don't have now.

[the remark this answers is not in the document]
gnodet: I'm not sure that's the right angle to look at. 

If a given project has multiple modules, those modules _kinda_ act as subprojects, so it makes sense to group their files together, rather than splitting main / tests.  In that case, each test directory is testing the code of the module, so having them close to each other makes more sense to me.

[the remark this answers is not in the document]
desruisseaux: Indeed, the rational is as Guillaume said. While we try to avoid disruption for users who are not interested in JPMS, those who want to embrace fully JPMS will have to change some habits anyway, some of them out of our control (e.g., the directory layout of `javac` output).

There is a difficulty with resources, which are considered as a language and therefore would be stored apart with a `src/${lang}/${module}/${scope}` layout if the users do not specify the directory themselves. An alternative staying closer to current Maven practice may be `src/${module}/${scope}/${lang}`.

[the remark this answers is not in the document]
gnodet: This layout is closer to the usual maven layout and may be less disruptive.  It just adds the module before the scope.  I like it better actually.

elharo:
Maybe introduce a new modules directory and put all the non-defallt modules there? E.g. modules/{module-name}/src/...

That is, today's src/... would contain only code for the default module, no others.

gnodet:
> Maybe introduce a new modules directory and put all the non-defallt modules there? E.g. modules/{module-name}/src/...
> 
> 
> 
> That is, today's src/... would contain only code for the default module, no others.

I don't think you could mix both in the same maven subproject, so not sure that would help.

api/maven-api-core/src/main/java/org/apache/maven/api/SourceRoot.java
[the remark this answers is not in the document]
desruisseaux: The inconvenient of the alternative proposal is that if a subproject has two or more languages, the modules of different languages will be mixed in the same directory. However, it may not be important if, as suggested in other comments, we choose to not let the multi-languages issue influences the decision.

**Note:** this difficulty come from the fact that in the current non-modular layout, `resources` is implicitly understood as "resources of the *Java* artifact". It is not really an independent "language".

If there is no objection or alternative proposal, I will change the layout to `src/${module}/${scope}/${lang}` in a few days.

elharo:
Since this is a major user facing design decision, this should be written up in a detailed proposal, not in code, and sent to the dev list for broader discussion. Code can come after there's reasonable consensus on the design.

desruisseaux:
An advantage of `src/${module}/${scope}/${lang}` is that it is close to the `javac` output, which is `target/classes/${module}`. The `${module}` part is generated by `javac` itself and cannot be changed.

desruisseaux:
Okay, I will write a document. Do we have a wiki or something like that for this kind of discussion?

desruisseaux:
>> That is, today's src/... would contain only code for the default module, no others.

> don't think you could mix both in the same maven subproject, so not sure that would help.

I confirm. When Java Module Source Hierarchy is used, there is no default module.

gnodet:
> Okay, I will write a document. Do we have a wiki or something like that for this kind of discussion?

https://cwiki.apache.org/confluence/display/MAVEN ?

impl/maven-impl/src/main/java/org/apache/maven/impl/DefaultSourceRoot.java
```diff
@@ -79,7 +79,12 @@ public DefaultSourceRoot(final Session session, final Path baseDir, final Source
         if (value != null) {
             directory = baseDir.resolve(value);
         } else {
-            directory = baseDir.resolve("src").resolve(scope.id()).resolve(language.id());
+            Path src = baseDir.resolve("src");
+            if (moduleName != null) {
+                directory = src.resolve(language.id()).resolve(moduleName).resolve(scope.id());
+            } else {
+                directory = src.resolve(scope.id()).resolve(language.id());
+            }
```
Pankraz76, line 87:
```suggestion
            Path src = baseDir.resolve("src");
            directory = moduleName != null ?
            src.resolve(language.id()).resolve(moduleName).resolve(scope.id()) :
            src.resolve(scope.id()).resolve(language.id());
```

api/maven-api-core/src/main/java/org/apache/maven/api/SourceRoot.java
```diff
@@ -37,11 +37,23 @@ public interface SourceRoot {
      * The path is relative to the <abbr>POM</abbr> file.
      *
      * <h4>Default implementation</h4>
-     * The default value is <code>src/{@linkplain #scope() scope}/{@linkplain #language() language}</code>
-     * as a relative path. Implementation classes may override this default with an absolute path instead.
+     * The default value depends on whether a {@linkplain #module() module name} is specified in this source root:
```
Pankraz76, line 40:
```suggestion
     * The default value depends on whether a {@linkplain #module() module name} is specified in this source root:
```
default value -> `singular`

api/maven-api-core/src/main/java/org/apache/maven/api/SourceRoot.java
```diff
@@ -37,11 +37,23 @@ public interface SourceRoot {
...
+     *   <li>
+     *     If no module name, then the default directory is
+     *     <code>src/{@linkplain #scope() scope}/{@linkplain #language() language}</code>.
+     *   </li><li>
+     *     If a module name is present, then the default directory is
+     *     <code>src/{@linkplain #language() language}/{@linkplain #module() module}/{@linkplain #scope() scope}</code>.
+     *   </li>
+     * </ul>
+     *
+     * These default values are relative directories.
```
Pankraz76, line 51:
```suggestion
     * The default value is relative.
```
default value -> `singular`

api/maven-api-core/src/main/java/org/apache/maven/api/SourceRoot.java
```diff
@@ -37,11 +37,23 @@ public interface SourceRoot {
...
+     * Implementation classes may override this default with absolute paths instead.
```
Pankraz76, line 52:
```suggestion
     * Implementation class may override with absolute path instead.
```

Pankraz76 (changes requested):
spelling: singular becomes plural, why so?

desruisseaux:
> spelling: singular becomes plural, why so?

Probably my broken English, confusing the grammar rules of different languages. Thank for spotting, I will fix.

desruisseaux:
Created a wiki page with a description of the three proposals there:

https://cwiki.apache.org/confluence/display/MAVEN/Default+directory+for+modular+projects

Side note: The "Maven Core Roadmap" page is about 8 years ago. Same for "Maven Internals" (actually empty).

impl/maven-impl/src/main/java/org/apache/maven/impl/DefaultSourceRoot.java
desruisseaux: Will wait to see what is the agreement on the default repository, because it would change that code.

w6et:
Isn't `<source/>` used to configure the subproject directory? 
I think it's two concepts.
1)codebase‘ src
2)maven module/subproject's src
if src/${module}/${scope}/${lang},how to config dependencies between modules?
**maven current:**
${module}/src/${scope}/${lang} can aslo support this like 'src/${module}/src/${scope}/${lang}',the  first 'src' is top directory,not managed by maven.
btw:With the growth of Python\Kotlin,java + python\java +kotlin in a module is a common case

desruisseaux:
> Isn't `<source/>` used to configure the subproject directory?

Yes.

> if `src/${module}/${scope}/${lang}`, how to config dependencies between modules?

The `pom.xml` file (which is a the same level as `src`) specifies all dependencies to place on the class-path or module-path. Then, the `module-info.java` file of each Java *module* (not to be confused with Maven *subproject*) specifies which dependencies are actually used by the module.

In a JPMS world, the class-path and module-path no longer specify the dependencies used by the modules. The module-path specifies which dependencies are *available*  for use by the modules. Whether a module really uses these dependencies or not depends on the `module-info.java`.

> maven current:
> `${module}/src/${scope}/${lang}` can aslo support this like `src/${module}/src/${scope}/${lang}`, the first 'src' is top directory,not managed by maven.

I think that there is a confusion here. Maven current is no longer `${module}/src/${scope}/${lang}`, but `${project}/src/${scope}/${lang}`. "Maven modules" have been renamed "Maven projects" in Maven 4 for avoiding confusion with Java module. The `${module}` component discussed in this pull request is something different, unrelated to Maven subproject (formerly "Maven module").

> btw:With the growth of Python\Kotlin,java + python\java +kotlin in a module is a common case

Yes. If Python uses the same naming convention as Java for module names, a `src/${module}/${scope}/${lang}` pattern works well. In the case of Python, because the naming conventions are different, it may cause an undesired mix. However, we can decide to let Java + Python users sort that out themselves by specifying explicit directory names rather than relying on the default.

impl/maven-impl/src/main/java/org/apache/maven/impl/DefaultSourceRoot.java
desruisseaux: Changed the default to `src/${module}/${scope}/${lang}` which seems to be the preferred layout in the [thread on the developer mailing list](https://lists.apache.org/thread/dzvpyfr667z0bwmwn0tdjn6nlmnx1gcx) (choice 1 in the initial email).

gnodet:
I think the discussion has settled with no major objection to this very issue, so I think we should go ahead.  @desruisseaux ?

desruisseaux:
Yes. I will do the merge. Thanks for the feedback.

PR #2284, opened 2025-05-01
[MNG-8614] Maven Sisu in Maven 4 same as in Maven 3

cstamas:
The Sisu "strict" flag has issues, but also, it is best and simplest to help transitioning by making Sisu in Maven 4 behave in very same way as it behaves in Maven 3.

---

https://issues.apache.org/jira/browse/MNG-8614


kwin:
Can't we fix the underlying issue in sisu rather? I am still convinced that throwing exceptions at least for incompatible bytecode is the right thing to do....

cstamas:
We can fix Sisu and make it strict in 4.1 or alike. I find this change (enabling strict) a bit disrupting for 4.0 that should allow easy transition from 3.x Maven.

slawekjaranowski (approved):

elharo (approved):

PR #2285, opened 2025-05-01
[MNG-8718] Restore Maven 3 compat: ProjectBuilder is unusable for legacy code

cstamas:
Maven 3 codebase is not able to use create new session, and use ProjectBuilder with it, even if it asks for resolver session from Maven itself.

---

https://issues.apache.org/jira/browse/MNG-8718


gnodet (approved):

impl/maven-core/src/main/java/org/apache/maven/DefaultMaven.java
```diff
@@ -408,7 +408,10 @@ private void persistResumptionData(MavenExecutionResult result, MavenSession ses
      */
     @Deprecated
     public RepositorySystemSession newRepositorySession(MavenExecutionRequest request) {
-        return newCloseableSession(request, new MavenChainedWorkspaceReader());
+        RepositorySystemSession result = newCloseableSession(request, new MavenChainedWorkspaceReader());
+        MavenSession session = new MavenSession(result, request, new DefaultMavenExecutionResult());
+        session.setSession(defaultSessionFactory.newSession(session));
```
slawekjaranowski, line 413: Looks like session is not used .... but magic is in `defaultSessionFactory.newSession` which assign MavenSession to RepositorySystemSession

maybe a comment on it?

slawekjaranowski (approved):

impl/maven-core/src/main/java/org/apache/maven/DefaultMaven.java
cstamas: Not just that: perso, _i'd make Maven warn in logs when this method is invoked_, just look at javadoc: "nobody should invoke this method". Takari Plugin testing does, but right now, there is no way to make Maven 3 codebase use ProjectBuilder with controlled session... or I am missing something.... Am on edge with this PR.

@gnodet any opinion?

cstamas: Also, my own code "maven 3 code cannot use project builder" proves am wrong: https://github.com/apache/maven-plugin-tools/blob/master/maven-plugin-report-plugin/src/main/java/org/apache/maven/plugin/plugin/report/PluginReport.java#L312-L329

Still, Takari Plugin testing want to create session "from the scratch" for junit testing the Mojos, and _that_ is impossible right now. When you get a session (already associated with maven 4 session) all is fine.

slawekjaranowski: hym ... versions plugins use a projectBuilder, like in https://github.com/mojohaus/versions/blob/master/versions-maven-plugin/src/main/java/org/codehaus/mojo/versions/DisplayPluginUpdatesMojo.java#L646

and I can execute it by Maven 4 ...
```
mvn versions:display-plugin-updates
```

cstamas: Yes, _received_ session works (like in a plugin I show or like in versions plugin), but _creating a session from the scratch does not_.

jira-importer:
Resolve #9581

PR #2277, opened 2025-04-25
Modernize codebase with Java improvements

gnodet:
This PR includes several improvements to modernize the codebase:

1. Replace IllegalArgumentException with ClassCastException for type cast failures
2. Replace IllegalArgumentException with NullPointerException for null checks
3. Use modern Java collections API (toList() instead of collect(Collectors.toList()))
4. Replace custom null checks with Objects.requireNonNull
5. Add @Nonnull and @Nullable annotations throughout the codebase

These changes make the code more maintainable, safer, and aligned with modern Java practices.

PR #2287, opened 2025-05-02
Modernize codebase with Java improvements - Use modern Java collections API (`toList()` instead of `Collectors.toList()`)

Pankraz76:
Modernize codebase with Java improvements - Use modern Java collections API (`toList()` instead of `Collectors.toList()`)

- https://github.com/apache/maven/pull/2277
- #2290
- https://github.com/checkstyle/checkstyle/issues/17000

### DIFF
--- a/impl/maven-core/src/main/java/org/apache/maven/execution/MavenSession.java
+++ b/impl/maven-core/src/main/java/org/apache/maven/execution/MavenSession.java
@@ -28,7 +28,6 @@ import java.util.Map;
 import java.util.Properties;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ConcurrentMap;
-import java.util.stream.Collectors;
 
 import org.apache.maven.api.Session;
 import org.apache.maven.artifact.repository.ArtifactRepository;
@@ -423,13 +422,13 @@ public class MavenSession implements Cloneable {
                 .localRepository(localRepo != null ? localRepo.getAbsolutePath() : null)
                 .interactiveMode(request.isInteractiveMode())
                 .offline(request.isOffline())
-                .proxies(request.getProxies().stream().map(Proxy::getDelegate).collect(Collectors.toList()))
-                .servers(request.getServers().stream().map(Server::getDelegate).collect(Collectors.toList()))
-                .mirrors(request.getMirrors().stream().map(Mirror::getDelegate).collect(Collectors.toList()))
+                .proxies(request.getProxies().stream().map(Proxy::getDelegate).toList())
+                .servers(request.getServers().stream().map(Server::getDelegate).toList())
+                .mirrors(request.getMirrors().stream().map(Mirror::getDelegate).toList())
                 .profiles(request.getProfiles().stream()
                         .map(Profile::getDelegate)
                         .map(SettingsUtilsV4::convertToSettingsProfile)
-                        .collect(Collectors.toList()))
+                        .toList())
                 .activeProfiles(request.getActiveProfiles())
                 .pluginGroups(request.getPluginGroups())
                 .build());
--- a/impl/maven-impl/src/main/java/org/apache/maven/impl/DefaultChecksumAlgorithmService.java
+++ b/impl/maven-impl/src/main/java/org/apache/maven/impl/DefaultChecksumAlgorithmService.java
@@ -58,7 +58,7 @@ public class DefaultChecksumAlgorithmService implements ChecksumAlgorithmService
     public Collection<String> getChecksumAlgorithmNames() {
         return checksumAlgorithmFactorySelector.getChecksumAlgorithmFactories().stream()
                 .map(ChecksumAlgorithmFactory::getName)
-                .collect(Collectors.toList());
+                .toList();
     }
 
     @Override
--- a/impl/maven-xml/src/main/java/org/apache/maven/internal/xml/DefaultXmlService.java
+++ b/impl/maven-xml/src/main/java/org/apache/maven/internal/xml/DefaultXmlService.java
@@ -312,7 +312,7 @@ public class DefaultXmlService extends XmlService {
                         Iterator<XmlNode> it =
                                 commonChildren.computeIfAbsent(name, n1 -> Stream.of(dominant.children().stream()
                                                 .filter(n2 -> n2.name().equals(n1))
-                                                .collect(Collectors.toList()))
+                                                .toList())
                                         .filter(l -> !l.isEmpty())
                                         .map(List::iterator)
                                         .findFirst()
--- a/impl/maven-cli/src/main/java/org/apache/maven/cling/extensions/BootstrapCoreExtensionManager.java
+++ b/impl/maven-cli/src/main/java/org/apache/maven/cling/extensions/BootstrapCoreExtensionManager.java
@@ -25,7 +25,6 @@ import java.util.List;
 import java.util.NoSuchElementException;
 import java.util.Set;
 import java.util.function.UnaryOperator;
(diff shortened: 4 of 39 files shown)


Pankraz76:
issue: 
image

gnodet (changes requested):
Please only apply to `[root]/api`, `[root]/impl`, but not `[api]/compat`.

Pankraz76:
update: 

image
image

should revert `its`, as well?

impl/maven-impl/src/main/java/org/apache/maven/impl/DefaultChecksumAlgorithmService.java
```diff
@@ -75,9 +74,9 @@ public ChecksumAlgorithm select(String algorithmName) {
     public Collection<ChecksumAlgorithm> select(Collection<String> algorithmNames) {
         nonNull(algorithmNames, "algorithmNames");
         try {
-            return checksumAlgorithmFactorySelector.selectList(new ArrayList<>(algorithmNames)).stream()
+            return new ArrayList<>(checksumAlgorithmFactorySelector.selectList(new ArrayList<>(algorithmNames)).stream()
```
gnodet, line 77: This is wrong.  If a mutable list is required, don't use `toList()`

Pankraz76:
> This is not a drop in replacement. This changes a lot of modifiable lists to immutable lists. In some cases here this is clearly fine — e.g. when all that''s done next is findFirst — but others aren't so obvious. This is really, really hard to debug when it goes wrong. I would limit this PR to cases where the lsist does not escape the method or perhaps the class.

## Immutable Objects and Thread Safety

If someone wants to deviate from the happy path, they should work with a copy of final and secure state - and by happy with that.

### Core Principles:

1. **Final/immutable by Default**:  
   Everything should be `final/immutable` unless there's a compelling reason not to be.

2. **External Manipulation**:  
   If external modification is absolutely necessary:
   - The requester should receive a defensive copy
   - All modifications must go through our controlled interface
   - We maintain the immutable original as the source of truth

### Important Considerations:

- **Strict Data Control**:  
  Every modifiable reference is a potential deviation from our secure, immutable approach.  
  We don't allow external code to directly manipulate our internal state.

- **Side Effect Prevention**:  
  Non-final references can lead to unpredictable side effects and thread safety issues.

### Reference:  
[Immutable Objects - Thread Safety](https://en.wikipedia.org/wiki/Immutable_object#Thread_safety)

### Implementation Strategy:
For production systems:
1. Audit to identify genuine needs for mutability
2. Create exception pathways with proper safeguards:
   - Defensive copying
   - Controlled mutation interfaces
3. Always preserve the immutable source
4. Document all deviation points rigorously

Pankraz76:
in cs `toUnmodifiableList` is `enforced` - so it seems possible and just one abstraction layer away from having final and secure state in theory.

elharo:
You're not working in a green field. You can't change the behavior of APIs and assume existing clients will adapt.

Pankraz76:
> You're not working in a green field. You can't change the behavior of APIs and assume existing clients will adapt.



Yes.

### Which clients are affected?

When will they get in touch? Do they live inside our JVM — at runtime or even compile-time — like within this code repo's scope?  
If so, we should be able to test and adapt accordingly.

If not — for example, clients using the Maven wrapper — they simply execute the `.jar` file and don’t interact with the API in a way where immutability would matter.

The only case I can imagine is another method using the current API and directly manipulating the result.  
In that case, wrapping the result in a simple `new ArrayList<>(...)` is enough to make the list mutable again — that should do the trick.

Whether we return a mutable list or they create one themselves doesn’t really matter.

The scenario where code relies on mutating a shared reference doesn’t apply here, since `toList()` already returns a new list instance.  
Passing around mutable references would be questionable design and could violate data integrity and the getter/setter principle.

The core change here is introducing immutability — which is a best practice and supports defensive programming.  
We also benefit from cleaner code through method argument validation, which is a welcome side effect.

So overall, the change is meaningful in terms of thread safety and data integrity — it's justified.  
We just need to align the code accordingly — `reverting` potential issues or `bridging` them by explicitly creating a new list instance when needed.

See also: [Create mutable list from array - Stack Overflow](https://stackoverflow.com/questions/11659173/create-mutable-list-from-array)

Ideally, failing tests that throw `UnsupportedOperationException` will help identify any breaking points.

https://docs.oracle.com/javase/9/core/creating-immutable-lists-sets-and-maps.htm#JSCOR-GUID-4F3E2B7D-CE90-4862-A78A-414FC08DA6E4

elharo (changes requested):
Sorry, no, I have been burned too badly by this change before. It is not OK to change an existing public method from returning a mutable list to an immutable list. It does break things. I have wasted too many hours of my life tracking down problems caused by an unexpected change from mutable to immutable to be OK with this.

Pankraz76:
of course agree on this.

Then need to make this step by step.

impl/maven-core/src/main/java/org/apache/maven/execution/MavenSession.java
```diff
@@ -423,13 +422,13 @@ private static Settings adaptSettings(MavenExecutionRequest request) {
                 .localRepository(localRepo != null ? localRepo.getAbsolutePath() : null)
                 .interactiveMode(request.isInteractiveMode())
                 .offline(request.isOffline())
-                .proxies(request.getProxies().stream().map(Proxy::getDelegate).collect(Collectors.toList()))
-                .servers(request.getServers().stream().map(Server::getDelegate).collect(Collectors.toList()))
-                .mirrors(request.getMirrors().stream().map(Mirror::getDelegate).collect(Collectors.toList()))
```
Pankraz76, line 428: thats okay as internal.

impl/maven-core/src/main/java/org/apache/maven/execution/MavenSession.java
```diff
@@ -423,13 +422,13 @@ private static Settings adaptSettings(MavenExecutionRequest request) {
...
+                .proxies(request.getProxies().stream().map(Proxy::getDelegate).toList())
+                .servers(request.getServers().stream().map(Server::getDelegate).toList())
+                .mirrors(request.getMirrors().stream().map(Mirror::getDelegate).toList())
                 .profiles(request.getProfiles().stream()
                         .map(Profile::getDelegate)
                         .map(SettingsUtilsV4::convertToSettingsProfile)
-                        .collect(Collectors.toList()))
```
Pankraz76, line 432: this is api

impl/maven-xml/src/main/java/org/apache/maven/internal/xml/DefaultXmlService.java
```diff
@@ -312,7 +312,7 @@ public XmlNode doMerge(XmlNode dominant, XmlNode recessive, Boolean childMergeOv
                         Iterator<XmlNode> it =
                                 commonChildren.computeIfAbsent(name, n1 -> Stream.of(dominant.children().stream()
                                                 .filter(n2 -> n2.name().equals(n1))
-                                                .collect(Collectors.toList()))
```
Pankraz76, line 315: okay, internal.

Pankraz76 (commented):
extract internal.

impl/maven-core/src/main/java/org/apache/maven/execution/MavenSession.java
Pankraz76: can have side effect too when being used outside.

Pankraz76:
Only doing the migration step by step to ensure quick fix, its possible.

As mentioned any change could unpleasant have side effects.

PR #2288, opened 2025-05-02
[MNG-8719] Restore Maven3 compat: model setInherited(boolean)

cstamas:
Added it as deprecated indirection for legacy code.

It was removed 1y ago probably by mistake by this commit:
https://github.com/apache/maven/commit/2a9f39336cec1d8e52d30cc48503d51ed8672536

---

https://issues.apache.org/jira/browse/MNG-8719

### DIFF
--- a/api/maven-api-model/src/main/mdo/maven.mdo
+++ b/api/maven-api-model/src/main/mdo/maven.mdo
@@ -2550,7 +2550,10 @@
     @Deprecated
     public void unsetInheritanceApplied() {
     }
-
+    @Deprecated
+    public void setInherited(boolean value) {
+      setInherited("true");
+    }
             ]]>
           </code>
         </codeSegment>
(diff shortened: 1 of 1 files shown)


gnodet (approved):

elharo (changes requested):
Needs a test. Otherwise it's likely to be deleted again.

cstamas:
> Needs a test. Otherwise it's likely to be deleted again.

Maven never had nor currently have tests for generated model. It was dropped last year by this commit:
https://github.com/apache/maven/commit/2a9f39336cec1d8e52d30cc48503d51ed8672536

cstamas:
I will skip adding test to it, as this is _deprecated_ method anyway, and serves sole purpose of re-adding compatibility for very old codebases (type of inherited with was changed to String long time ago, around 3.0-RC1 or so).

PR #2290, opened 2025-05-02
[chore] Modernize codebase with Java improvements - Replace custom null checks with `Objects.requireNonNull`

Pankraz76:
Modernize codebase with Java improvements - Replace custom null checks with `Objects.requireNonNull`
- #2277
- #2287

### DIFF
--- a/impl/maven-impl/src/main/java/org/apache/maven/impl/AbstractSession.java
+++ b/impl/maven-impl/src/main/java/org/apache/maven/impl/AbstractSession.java
@@ -103,8 +103,8 @@ import org.eclipse.aether.artifact.ArtifactType;
 import org.eclipse.aether.repository.ArtifactRepository;
 import org.eclipse.aether.transfer.TransferResource;
 
+import static java.util.Objects.requireNonNull;
 import static org.apache.maven.impl.ImplUtils.map;
-import static org.apache.maven.impl.ImplUtils.nonNull;
 
 public abstract class AbstractSession implements InternalSession {
 
@@ -134,7 +134,7 @@ public abstract class AbstractSession implements InternalSession {
             List<RemoteRepository> repositories,
             List<org.eclipse.aether.repository.RemoteRepository> resolverRepositories,
             Lookup lookup) {
-        this.session = nonNull(session, "session");
+        this.session = requireNonNull(session, "session");
         this.repositorySystem = repositorySystem;
         this.repositories = getRepositories(repositories, resolverRepositories);
         this.lookup = lookup;
@@ -337,7 +337,7 @@ public abstract class AbstractSession implements InternalSession {
     @Nonnull
     @Override
     public Session withLocalRepository(@Nonnull LocalRepository localRepository) {
-        nonNull(localRepository, "localRepository");
+        requireNonNull(localRepository, "localRepository");
         if (session.getLocalRepository() != null
                 && Objects.equals(session.getLocalRepository().getBasePath(), localRepository.getPath())) {
             return this;
@@ -461,12 +461,12 @@ public abstract class AbstractSession implements InternalSession {
 
     @Override
     public void registerListener(@Nonnull Listener listener) {
-        listeners.add(nonNull(listener));
+        listeners.add(requireNonNull(listener));
     }
 
     @Override
     public void unregisterListener(@Nonnull Listener listener) {
-        listeners.remove(nonNull(listener));
+        listeners.remove(requireNonNull(listener));
     }
 
     @Nonnull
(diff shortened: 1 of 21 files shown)


elharo (approved):

michael-o (approved):

impl/maven-impl/src/main/java/org/apache/maven/impl/ImplUtils.java
```diff
@@ -25,18 +25,21 @@
 import java.util.stream.Collectors;
 
 class ImplUtils {
+
+    /**
+     * @deprecated Use {@link Objects#requireNonNull(Object)} instead.
+     */
+    @Deprecated
```
olamy, line 32: We are still in RC. Maybe we can remove it? (or mark it for removal in 4.1?)

olamy (changes requested):

impl/maven-impl/src/main/java/org/apache/maven/impl/ImplUtils.java
olamy: Something this sure.
But as we are still RC and those methods/classes have been introduced with 4. x, I would even prefer we remove them.

Pankraz76: agree on less is more, so its gone.

impl/maven-impl/src/test/java/org/apache/maven/impl/DefaultModelXmlFactoryTest.java
```diff
@@ -105,7 +105,7 @@ void testNoNamespaceWithModelVersion400() throws Exception {
 
     @Test
     void testNullRequest() {
-        assertThrows(IllegalArgumentException.class, () -> factory.read((XmlReaderRequest) null));
+        assertThrows(NullPointerException.class, () -> factory.read((XmlReaderRequest) null));
```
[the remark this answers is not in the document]
Pankraz76: thx

impl/maven-impl/src/main/java/org/apache/maven/impl/DefaultArtifactCoordinatesFactory.java
```diff
@@ -26,14 +26,14 @@
...
 
-import static org.apache.maven.impl.ImplUtils.nonNull;
+import static java.util.Objects.requireNonNull;
 
 @Named
 @Singleton
 public class DefaultArtifactCoordinatesFactory implements ArtifactCoordinatesFactory {
     @Override
     public ArtifactCoordinates create(@Nonnull ArtifactCoordinatesFactoryRequest request) {
-        nonNull(request, "request");
```
Pankraz76, line 36: this pattern is re-ocurring a lot for defensive programming. Actually the outcome is quite the same, resulting in `NPE`. With the only benefit of giving `NullPointerException("request")` instead of raw `NullPointerException()`.

High invasive on code for little change. Why not giving dedicated abstraction layer for this? Reducing overhead to minimum, getting the message spec for `free`:

image

https://projectlombok.org/features/NonNull

impl/maven-impl/src/main/java/org/apache/maven/impl/AbstractSession.java
```diff
@@ -134,7 +134,7 @@ public AbstractSession(
             List<RemoteRepository> repositories,
             List<org.eclipse.aether.repository.RemoteRepository> resolverRepositories,
             Lookup lookup) {
-        this.session = nonNull(session, "session");
+        this.session = requireNonNull(session, "session");
         this.repositorySystem = repositorySystem;
```
Pankraz76, line 138: Kind of inconsequent. Why not this? Nullable too. Thus `session` not marked `@NonNull` so its kind of magic number?

We produce boilerplate for the only difference to have a little better error message. 

This is high cost for little, to almost no, return.

Bukama:
> Might enable an auto merge after 10 days non rejected.

Not rejected does not mean approved (at least for me) and taken into account how large Maven is, that the Maven team is not a company and the current situation about cyber attacks (including AI generated ones) I don't think auto merges are a good idea.

michael-o:
> > Might enable an auto merge after 10 days non rejected.
> 
> Not rejected does not mean approved (at least for me) and taken into account how large Maven is, that the Maven team is not a company and the current situation about cyber attacks (including AI generated ones) I don't think auto merges are a good idea.

I am totally against any automerges.

Pankraz76:
yes, its a security risk not haven taken into account. Thanks for hint.
If its only possible then with min. approval count increased to 2, better 3, this is altho not pleasant.

Pankraz76:
@olamy any feedback from your side? 

If not might consider merge.

Its just renaming of old method to new one, therefore no delta possible, and removal. No risky s&r brute force.

PR #2292, opened 2025-05-02
Pull #2292: Modernize codebase with Java improvements - SOC `DefaultModelProcessor#read`

Pankraz76:
Pull #2292: `Modernize codebase with Java improvements - functionalize DefaultModelProcessor#read`
- https://github.com/apache/maven/pull/2277

enable for:
- https://github.com/apache/maven/pull/2291

depends on:
- https://github.com/apache/maven/pull/2304

add missing Exception requirement to prevent:
image
image
image
`finally 100`
image

### DIFF
--- a/impl/maven-impl/src/main/java/org/apache/maven/impl/model/DefaultModelProcessor.java
+++ b/impl/maven-impl/src/main/java/org/apache/maven/impl/model/DefaultModelProcessor.java
@@ -29,37 +29,39 @@ import java.util.Objects;
 import java.util.Optional;
 
 import org.apache.maven.api.annotations.Nullable;
-import org.apache.maven.api.di.Inject;
 import org.apache.maven.api.di.Named;
 import org.apache.maven.api.di.Singleton;
 import org.apache.maven.api.model.Model;
+import org.apache.maven.api.services.Source;
 import org.apache.maven.api.services.model.ModelProcessor;
 import org.apache.maven.api.services.xml.ModelXmlFactory;
 import org.apache.maven.api.services.xml.XmlReaderRequest;
 import org.apache.maven.api.spi.ModelParser;
 import org.apache.maven.api.spi.ModelParserException;
 
+import static org.apache.maven.api.spi.ModelParser.STRICT;
+
 /**
  *
  * Note: uses @Typed to limit the types it is available for injection to just ModelProcessor.
- *
+ * <p>
  * This is because the ModelProcessor interface extends ModelLocator and ModelReader. If we
  * made this component available under all its interfaces then it could end up being injected
  * into itself leading to a stack overflow.
- *
+ * <p>
  * A side effect of using @Typed is that it translates to explicit bindings in the container.
  * So instead of binding the component under a 'wildcard' key it is now bound with an explicit
  * key. Since this is a default component; this will be a plain binding of ModelProcessor to
  * this implementation type; that is, no hint/name.
- *
+ * <p>
  * This leads to a second side effect in that any @Inject request for just ModelProcessor in
  * the same injector is immediately matched to this explicit binding, which means extensions
  * cannot override this binding. This is because the lookup is always short-circuited in this
  * specific situation (plain @Inject request, and plain explicit binding for the same type.)
- *
+ * <p>
  * The simplest solution is to use a custom @Named here so it isn't bound under the plain key.
  * This is only necessary for default components using @Typed that want to support overriding.
- *
+ * <p>
  * As a non-default component this now gets a negative priority relative to other implementations
  * of the same interface. Since we want to allow overriding this doesn't matter in this case.
  * (if it did we could add @Priority of 0 to match the priority given to default components.)
@@ -69,75 +71,74 @@ import org.apache.maven.api.spi.ModelParserException;
 public class DefaultModelProcessor implements ModelProcessor {
 
     private final ModelXmlFactory modelXmlFactory;
-    private final List<ModelParser> modelParsers;
+    private final @Nullable List<ModelParser> modelParsers;
 
-    @Inject
     public DefaultModelProcessor(ModelXmlFactory modelXmlFactory, @Nullable List<ModelParser> modelParsers) {
         this.modelXmlFactory = modelXmlFactory;
         this.modelParsers = modelParsers;
     }
 
+    /**
+     * @implNote The ModelProcessor#locatePom never returns null while the ModelParser#locatePom needs to return an existing path!
+     */
     @Override
     public Path locateExistingPom(Path projectDirectory) {
-        // Note that the ModelProcessor#locatePom never returns null
-        // while the ModelParser#locatePom needs to return an existing path!
-        Path pom = modelParsers.stream()
-                .map(m -> m.locate(projectDirectory)
-                        .map(org.apache.maven.api.services.Source::getPath)
-                        .orElse(null))
-                .filter(Objects::nonNull)
+        return modelParsers.stream()
+                .map(parser -> parser.locate(projectDirectory))
+                .filter(Optional::isPresent)
+                .map(Optional::get)
+                .map(Source::getPath)
+                .peek(path -> throwIfWrongProjectDirLocation(path, projectDirectory))
                 .findFirst()
-                .orElseGet(() -> doLocateExistingPom(projectDirectory));
+                .orElseGet(() -> locateExistingPomWithUserDirDefault(projectDirectory));
+    }
+
+    private static void throwIfWrongProjectDirLocation(Path pom, Path projectDirectory) {
         if (pom != null && !pom.equals(projectDirectory) && !pom.getParent().equals(projectDirectory)) {
             throw new IllegalArgumentException("The POM found does not belong to the given directory: " + pom);
         }
-        return pom;
+    }
+
+    private Path locateExistingPomWithUserDirDefault(Path project) {
+        return locateExistingPomInDirOrFile(project != null ? project : Paths.get(System.getProperty("user.dir")));
+    }
+
+    private static Path locateExistingPomInDirOrFile(Path project) {
+        return Files.isDirectory(project) ? isRegularFileOrNull(project.resolve("pom.xml")) : project;
+    }
+
+    private static Path isRegularFileOrNull(Path pom) {
+        return Files.isRegularFile(pom) ? pom : null;
     }
 
     @Override
     public Model read(XmlReaderRequest request) throws IOException {
         Objects.requireNonNull(request, "source cannot be null");
-        Path pomFile = request.getPath();
+        return readPomWithParentInheritance(request, request.getPath());
+    }
+
+    private Model readPomWithParentInheritance(XmlReaderRequest request, Path pomFile) {
+        List<ModelParserException> exceptions = new ArrayList<>();
         if (pomFile != null) {
-            Path projectDirectory = pomFile.getParent();
-            List<ModelParserException> exceptions = new ArrayList<>();
             for (ModelParser parser : modelParsers) {
                 try {
-                    Optional<Model> model =
-                            parser.locateAndParse(projectDirectory, Map.of(ModelParser.STRICT, request.isStrict()));
-                    if (model.isPresent()) {
-                        return model.get().withPomFile(pomFile);
-                    }
-                } catch (ModelParserException e) {
-                    exceptions.add(e);
+                    return parser.locateAndParse(pomFile, Map.of(STRICT, request.isStrict()))
+                            .orElseThrow()
+                            .withPomFile(pomFile);
+                } catch (RuntimeException ex) {
+                    exceptions.add(new ModelParserException(ex));
                 }
             }
-            try {
-                return doRead(request);
-            } catch (IOException e) {
-                exceptions.forEach(e::addSuppressed);
-                throw e;
-            }
-        } else {
-            return doRead(request);
         }
+        return readPomWithSuppressedErrorRethrow(request, exceptions);
     }
 
-    private Path doLocateExistingPom(Path project) {
-        if (project == null) {
-            project = Paths.get(System.getProperty("user.dir"));
+    private Model readPomWithSuppressedErrorRethrow(XmlReaderRequest request, List<ModelParserException> exceptions) {
+        try {
+            return modelXmlFactory.read(request);
+        } catch (RuntimeException ex) {
+            exceptions.forEach(ex::addSuppressed);
+            throw ex;
         }
-        if (Files.isDirectory(project)) {
-            Path pom = project.resolve("pom.xml");
-            return Files.isRegularFile(pom) ? pom : null;
-        } else if (Files.isRegularFile(project)) {
-            return project;
-        } else {
-            return null;
-        }
-    }
-
-    private Model doRead(XmlReaderRequest request) throws IOException {
-        return modelXmlFactory.read(request);
     }
 }
--- a/impl/maven-impl/src/test/java/org/apache/maven/impl/model/DefaultModelProcessorTest.java
+++ b/impl/maven-impl/src/test/java/org/apache/maven/impl/model/DefaultModelProcessorTest.java
@@ -1,0 +1,243 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.maven.impl.model;
+
+import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.util.List;
+import java.util.Optional;
+
+import org.apache.maven.api.model.Model;
+import org.apache.maven.api.services.Source;
+import org.apache.maven.api.services.xml.ModelXmlFactory;
+import org.apache.maven.api.services.xml.XmlReaderRequest;
+import org.apache.maven.api.spi.ModelParser;
+import org.apache.maven.api.spi.ModelParserException;
+import org.junit.jupiter.api.AfterEach;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.io.TempDir;
+
+import static org.assertj.core.api.AssertionsForClassTypes.assertThat;
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertInstanceOf;
+import static org.junit.jupiter.api.Assertions.assertNotNull;
+import static org.junit.jupiter.api.Assertions.assertNull;
+import static org.junit.jupiter.api.Assertions.assertThrows;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+import static org.mockito.Mockito.any;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.when;
+
+class DefaultModelProcessorTest {
+
+    @TempDir
+    Path tempDir;
+
+    Path testProjectDir, testPomFile;
+
+    @AfterEach
+    void cleanup() throws IOException {
+        if (testPomFile != null && Files.exists(testPomFile)) {
+            Files.deleteIfExists(testPomFile);
+        }
+        if (testProjectDir != null && Files.exists(testProjectDir)) {
+            Files.deleteIfExists(testProjectDir);
+        }
+    }
+
+    @Test
+    void readWithValidParserShouldReturnModel() throws Exception {
+        ModelXmlFactory factory = mock(ModelXmlFactory.class);
+        ModelParser parser = mock(ModelParser.class);
+        XmlReaderRequest request = mock(XmlReaderRequest.class);
+        Model model = mock(Model.class);
+        Path path = Path.of("project/pom.xml");
+        when(request.getPath()).thenReturn(path);
+        when(request.isStrict()).thenReturn(true);
+        when(model.withPomFile(path)).thenReturn(model);
+        when(parser.locateAndParse(any(), any())).thenReturn(Optional.of(model));
+        Model result = new DefaultModelProcessor(factory, List.of(parser)).read(request);
+        assertNotNull(result);
+        assertEquals(model, result);
+    }
+
+    @Test
+    void readNullPomPathShouldUseFactoryDirectly() throws Exception {
+        ModelXmlFactory factory = mock(ModelXmlFactory.class);
+        XmlReaderRequest request = mock(XmlReaderRequest.class);
+        Model model = mock(Model.class);
+        when(request.getPath()).thenReturn(null);
+        when(factory.read(request)).thenReturn(model);
+        Model result = new DefaultModelProcessor(factory, List.of()).read(request);
+        assertNotNull(result);
+        assertEquals(model, result);
+    }
+
+    @Test
+    void locateExistingPomWithParsersShouldReturnFirstValid() {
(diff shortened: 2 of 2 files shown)


Pankraz76:
whats the mystery about this? 
```
[ERROR] Errors: 
[ERROR]   EmbeddedMavenExecutorTest.defaultFs3xCaptureOutput(Path) » Timeout defaultFs3xCaptureOutput(java.nio.file.Path) timed out after 15 seconds
[ERROR]   ForkedMavenExecutorTest.dump4(Path, Path) » Timeout dump4(java.nio.file.Path, java.nio.file.Path) timed out after 15 seconds
[ERROR]   ToolboxToolTest.metadataPath3(Mode)[1] » Timeout metadataPath3(org.apache.maven.cling.executor.ExecutorHelper$Mode) timed out after 15 seconds
```

gnodet (changes requested):
This totally breaks the exception handling.

Pankraz76:
> This totally breaks the exception handling.

yes, then we need test, thanks.

impl/maven-impl/src/main/java/org/apache/maven/impl/model/DefaultModelProcessor.java
```diff
@@ -97,30 +95,20 @@ public Path locateExistingPom(Path projectDirectory) {
...
-        } else {
-            return doRead(request);
-        }
+        return readOnSelfOrParent(request.getPath(), request);
+    }
+
+    private Model readOnSelfOrParent(Path pomFile, XmlReaderRequest request) throws IOException {
+        return pomFile == null
+                ? doRead(request)
+                : modelParsers.stream()
```
Pankraz76, line 104: extract concern into readParent() to hide impl. detail and give dedicated concern

elharo (changes requested):
Please don't. The original code is clearer. Lambdas do not improve this.

elharo (changes requested):
Test cases might be missing here, but it looks like this changes functionality in a negative way by not grouping exceptions

Pankraz76:
Yes it’s a test blind spot I want to cover. Sent from my iPhoneOn 5 May 2025, at 12:24, Elliotte Rusty Harold ***@***.***> wrote:﻿
@elharo requested changes on this pull request.

Test cases might be missing here, but it looks like this changes functionality in a negative way by not grouping exceptions

impl/maven-impl/src/test/java/org/apache/maven/impl/model/DefaultModelProcessorTest.java
```diff
@@ -0,0 +1,139 @@
...
+
+import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.util.*;
+
+import static org.junit.jupiter.api.Assertions.*;
+import static org.mockito.Mockito.*;
+
+class DefaultModelProcessorTest {
```
Pankraz76, line 21: tc:
image

image

impl/maven-impl/src/main/java/org/apache/maven/impl/model/DefaultModelProcessor.java
```diff
@@ -137,7 +139,7 @@ private Path doLocateExistingPom(Path project) {
         }
     }
 
-    private Model doRead(XmlReaderRequest request) throws IOException {
```
Pankraz76, line 140: this changes the whole code as this "can not happen"; or if needed then we should change API, if we can, to make code more concise.

image

impl/maven-impl/src/main/java/org/apache/maven/impl/model/DefaultModelProcessor.java
```diff
@@ -97,32 +99,32 @@ public Path locateExistingPom(Path projectDirectory) {
...
                         return model.get().withPomFile(pomFile);
                     }
                 } catch (ModelParserException e) {
                     exceptions.add(e);
                 }
             }
-            try {
-                return doRead(request);
-            } catch (IOException e) {
-                exceptions.forEach(e::addSuppressed);
```
Pankraz76, line 118: Then this is the last branch not covered.

impl/maven-impl/src/test/java/org/apache/maven/impl/model/DefaultModelProcessorTest.java
Pankraz76: image

impl/maven-impl/src/main/java/org/apache/maven/impl/model/DefaultModelProcessor.java
```diff
@@ -97,47 +101,50 @@ public Path locateExistingPom(Path projectDirectory) {
...
+
+    private Path locateExistingPomWithUserDirDefault(Path project) {
+        return locateExistingPomInDirOrFile(project != null ? project : Paths.get(System.getProperty("user.dir")));
+    }
+
+    private static Path locateExistingPomInDirOrFile(Path project) {
         if (Files.isDirectory(project)) {
             Path pom = project.resolve("pom.xml");
             return Files.isRegularFile(pom) ? pom : null;
-        } else if (Files.isRegularFile(project)) {
```
Pankraz76, line 133: It's either a directory (the higher-tier entity), as there is no file without a directory containing it. Which one came first is another topic, but I think it's that simple—either/or. So, we don't need this check, as the null branch is hard to test.

image

impl/maven-impl/src/main/java/org/apache/maven/impl/model/DefaultModelProcessor.java
```diff
@@ -39,27 +39,29 @@
...
+
 /**
  *
  * Note: uses @Typed to limit the types it is available for injection to just ModelProcessor.
- *
+ * <p>
  * This is because the ModelProcessor interface extends ModelLocator and ModelReader. If we
  * made this component available under all its interfaces then it could end up being injected
  * into itself leading to a stack overflow.
- *
```
Pankraz76, line 49: there is an checkstyle rule for this

impl/maven-impl/src/main/java/org/apache/maven/impl/DefaultPluginXmlFactory.java
```diff
@@ -109,7 +110,7 @@ public void write(XmlWriterRequest<PluginDescriptor> request) throws XmlWriterEx
      * @throws XmlReaderException if an error occurs during the parsing
      * @see #toXmlString(Object)
      */
-    public static PluginDescriptor fromXml(@Nonnull String xml) throws XmlReaderException {
```
Pankraz76, line 112: supplement this

impl/maven-impl/src/main/java/org/apache/maven/impl/DefaultPluginXmlFactory.java
```diff
@@ -92,7 +93,7 @@ public void write(XmlWriterRequest<PluginDescriptor> request) throws XmlWriterEx
             } else if (outputStream != null) {
                 new PluginDescriptorStaxWriter().write(outputStream, content);
             } else {
-                try (OutputStream os = Files.newOutputStream(path)) {
```
Pankraz76, line 95: supplement this

Pankraz76 (commented):
SOC

Pankraz76:
is this any good?

impl/maven-impl/src/main/java/org/apache/maven/impl/model/DefaultModelProcessor.java
```diff
@@ -77,17 +79,19 @@ public DefaultModelProcessor(ModelXmlFactory modelXmlFactory, @Nullable List<Mod
...
     public Path locateExistingPom(Path projectDirectory) {
-        // Note that the ModelProcessor#locatePom never returns null
-        // while the ModelParser#locatePom needs to return an existing path!
         Path pom = modelParsers.stream()
                 .map(m -> m.locate(projectDirectory)
                         .map(org.apache.maven.api.services.Source::getPath)
                         .orElse(null))
                 .filter(Objects::nonNull)
                 .findFirst()
-                .orElseGet(() -> doLocateExistingPom(projectDirectory));
```
Pankraz76, line 90: we have already very nice functional clean code style. very good codebase. func over imper

impl/maven-impl/src/main/java/org/apache/maven/impl/DefaultPluginXmlFactory.java
```diff
@@ -92,7 +93,7 @@ public void write(XmlWriterRequest<PluginDescriptor> request) throws XmlWriterEx
...
+                try (OutputStream ignored = Files.newOutputStream(path)) {
```
elharo, line 96: This doesn't look ignored. It looks like it's used?

impl/maven-impl/src/main/java/org/apache/maven/impl/DefaultPluginXmlFactory.java
```diff
@@ -109,7 +110,7 @@ public void write(XmlWriterRequest<PluginDescriptor> request) throws XmlWriterEx
      * @throws XmlReaderException if an error occurs during the parsing
      * @see #toXmlString(Object)
      */
-    public static PluginDescriptor fromXml(@Nonnull String xml) throws XmlReaderException {
+    public static PluginDescriptor fromXml(@Nonnull String xml) throws XmlReaderException, IOException {
         return new DefaultPluginXmlFactory().fromXmlString(xml);
```
elharo, line 114: If the string contains the XML text there shouldn't be IOExceptions here

impl/maven-impl/src/main/java/org/apache/maven/impl/model/DefaultModelProcessor.java
```diff
@@ -97,47 +101,50 @@ public Path locateExistingPom(Path projectDirectory) {
     @Override
     public Model read(XmlReaderRequest request) throws IOException {
         Objects.requireNonNull(request, "source cannot be null");
-        Path pomFile = request.getPath();
+        return readOnSelfOrParent(request, request.getPath());
+    }
+
+    private Model readOnSelfOrParent(XmlReaderRequest request, Path pomFile) throws IOException {
```
elharo, line 107: Not sure what this means. Perhaps readFromSelfOrParent?

impl/maven-impl/src/main/java/org/apache/maven/impl/model/DefaultModelProcessor.java
```diff
@@ -97,47 +101,50 @@ public Path locateExistingPom(Path projectDirectory) {
...
-        } else if (Files.isRegularFile(project)) {
-            return project;
-        } else {
-            return null;
         }
+        return project;
     }
 
-    private Model doRead(XmlReaderRequest request) throws IOException {
+    private Model readOnSelf(XmlReaderRequest request) throws IOException {
```
elharo, line 147: Again, I don't understand "read on self"

impl/maven-impl/src/test/java/org/apache/maven/impl/model/DefaultModelProcessorTest.java
```diff
@@ -0,0 +1,184 @@
...
+        ModelParser parser = mock(ModelParser.class);
+        XmlReaderRequest request = mock(XmlReaderRequest.class);
+        Model fallbackModel = mock(Model.class);
+        Path path = Path.of("project/pom.xml");
+
+        when(request.getPath()).thenReturn(path);
+        when(request.isStrict()).thenReturn(false);
+        when(parser.locateAndParse(any(), any())).thenThrow(new ModelParserException("parse fail"));
+
+        when(factory.read(request)).thenReturn(fallbackModel);
```
elharo, line 78: This mock here seems to make the test meaningless. It looks like all you're testing is what the mock returns when the mocked method is called.

impl/maven-impl/src/test/java/org/apache/maven/impl/model/DefaultModelProcessorTest.java
```diff
@@ -0,0 +1,184 @@
...
+        Path projectDir = Files.createTempDirectory("testproject");
+        Path pomFile = projectDir.resolve("pom.xml");
+        Files.createFile(pomFile);
+
+        DefaultModelProcessor processor = new DefaultModelProcessor(mock(ModelXmlFactory.class), List.of());
+        Path result = processor.locateExistingPom(projectDir);
+
+        assertEquals(pomFile, result);
+
+        Files.deleteIfExists(pomFile);
```
elharo, line 169: cleanup probably belongs in a finally block, or perhaps an @After method

impl/maven-impl/src/test/java/org/apache/maven/impl/model/DefaultModelProcessorTest.java
```diff
@@ -0,0 +1,184 @@
...
+    @Test
+    void locateExistingPomFallbackWithFileAsPathShouldReturnThatFile() throws Exception {
+        Path tempPom = Files.createTempFile("pom", ".xml");
+
+        DefaultModelProcessor processor = new DefaultModelProcessor(mock(ModelXmlFactory.class), List.of());
+        Path result = processor.locateExistingPom(tempPom);
+
+        assertEquals(tempPom, result);
+
+        Files.deleteIfExists(tempPom);
```
elharo, line 182: cleanup probably belongs in a finally block, or perhaps an @After method

impl/maven-impl/src/main/java/org/apache/maven/impl/DefaultPluginXmlFactory.java
Pankraz76: no its not:

image
one time usage:

image

`Value 'outputStream' is always 'null' `

image

this was the idea maybe:

image

impl/maven-impl/src/main/java/org/apache/maven/impl/model/DefaultModelProcessor.java
Pankraz76: It's DRY  anyway, with a lot of abstraction obfuscating the code. I usually overdo the layering. This might actually help bring some cohesion.

impl/maven-impl/src/test/java/org/apache/maven/impl/model/DefaultModelProcessorTest.java
Pankraz76: yes its DRY

impl/maven-impl/src/test/java/org/apache/maven/impl/model/DefaultModelProcessorTest.java
Pankraz76: yes its DRY

impl/maven-impl/src/test/java/org/apache/maven/impl/model/DefaultModelProcessorTest.java
Pankraz76: good eye. yes its boilerplate:

image

Pankraz76: gone

impl/maven-impl/src/main/java/org/apache/maven/impl/model/DefaultModelProcessor.java
```diff
@@ -97,47 +106,44 @@ public Path locateExistingPom(Path projectDirectory) {
...
+                    Optional<Model> parent = readParent(request, pomFile, parser);
+                    if (parent.isPresent()) {
+                        return parent.get().withPomFile(pomFile);
                     }
                 } catch (ModelParserException e) {
                     exceptions.add(e);
                 }
             }
-            try {
-                return doRead(request);
```
Pankraz76, line 116: dry

impl/maven-impl/src/main/java/org/apache/maven/impl/model/DefaultModelProcessor.java
```diff
@@ -97,47 +106,44 @@ public Path locateExistingPom(Path projectDirectory) {
...
-            } catch (IOException e) {
-                exceptions.forEach(e::addSuppressed);
-                throw e;
-            }
-        } else {
-            return doRead(request);
```
Pankraz76, line 122: dry

impl/maven-impl/src/main/java/org/apache/maven/impl/model/DefaultModelProcessor.java
```diff
@@ -97,47 +106,44 @@ public Path locateExistingPom(Path projectDirectory) {
...
-        if (project == null) {
-            project = Paths.get(System.getProperty("user.dir"));
-        }
+    private static Optional<Model> readParent(XmlReaderRequest request, Path pomFile, ModelParser parser) {
+        return parser.locateAndParse(pomFile.getParent(), Map.of(STRICT, request.isStrict()));
+    }
+
+    private Path locateExistingPomWithUserDirDefault(Path project) {
+        return locateExistingPomInDirOrFile(project != null ? project : Paths.get(System.getProperty("user.dir")));
+    }
```
Pankraz76, line 140: SOC, SRP, feature envy, one thing only.

impl/maven-impl/src/main/java/org/apache/maven/impl/model/DefaultModelProcessor.java
```diff
@@ -97,47 +106,44 @@ public Path locateExistingPom(Path projectDirectory) {
     @Override
     public Model read(XmlReaderRequest request) throws IOException {
         Objects.requireNonNull(request, "source cannot be null");
-        Path pomFile = request.getPath();
+        return readPomWithParentInheritance(request, request.getPath());
+    }
+
+    private Model readPomWithParentInheritance(XmlReaderRequest request, Path pomFile) throws IOException {
+        List<ModelParserException> exceptions = new ArrayList<>();
```
Pankraz76, line 113: high coupling -1

impl/maven-impl/src/main/java/org/apache/maven/impl/model/DefaultModelProcessor.java
```diff
@@ -69,25 +71,32 @@
 public class DefaultModelProcessor implements ModelProcessor {
 
     private final ModelXmlFactory modelXmlFactory;
-    private final List<ModelParser> modelParsers;
+    private final @Nullable List<ModelParser> modelParsers;
 
     @Inject
     public DefaultModelProcessor(ModelXmlFactory modelXmlFactory, @Nullable List<ModelParser> modelParsers) {
```
Pankraz76, line 77: everything can be null, why this required?
```suggestion
    public DefaultModelProcessor(ModelXmlFactory modelXmlFactory, List<ModelParser> modelParsers) {
```

impl/maven-impl/src/main/java/org/apache/maven/impl/model/DefaultModelProcessor.java
```diff
@@ -69,25 +72,31 @@
 public class DefaultModelProcessor implements ModelProcessor {
 
     private final ModelXmlFactory modelXmlFactory;
-    private final List<ModelParser> modelParsers;
+    private final @Nullable List<ModelParser> modelParsers;
 
-    @Inject
```
Pankraz76, line 74: its not CDI but dont needed like in spring as its working without (convention over config)

Pankraz76 (commented):
quest

gnodet (changes requested):
Not sure what this PR aims to do, but that's unrelated to its description and goes in lots of directions without any clear purpose.  Adding checked exceptions to the API which relies on unchecked exceptions everywhere is problematic

Pankraz76:
> Adding checked exceptions to the API which relies on unchecked exceptions everywhere is problematic

Thanks for the feedback. But just to clarify — this reflects what the current implementation actually does, right?  
I'm doing this to enforce the current API and approach, so I’d suggest we state things as they are and move forward from there.

Otherwise, I’d consider removing the `IOException` check, since it might be hiding real issues.  
The tests and interface should make the actual behavior and requirements clear.  

I'm not fully aware of all the requirements, but ideally, we can follow `TDD` and `refactor aggressively` to reveal the final version of the code and `surface the core business logic`.

gnodet:
> > Adding checked exceptions to the API which relies on unchecked exceptions everywhere is problematic
> 
> Thanks for the feedback. But just to clarify — this reflects what the current implementation actually does, right? I'm doing this to enforce the current API and approach, so I’d suggest we state things as they are and move forward from there.

Currently, the API is designed to throw XmlReaderException.  Check the implemented interface, there's no IOException thrown here.

> Otherwise, I’d consider removing the `IOException` check — it might actually be causing issues. That’s why I want to enforce and challenge this behavior. The tests and interface should make the actual behavior and requirements clear.
> 
> I'm not fully aware of all the requirements, but ideally, we can follow `TDD` and `refactor aggressively` to reveal the final version of the code and `surface the core business logic`.

Not sure what you mean exactly, but refactoring is not supposed to change the API in any way.

Pankraz76:
> Not sure what you mean exactly, but refactoring is not supposed to change the API in any way.

yes of course will adapt then. sorry

Pankraz76:
> This totally breaks the exception handling.

I'm not sure—when asking AI and trying to mock it, both seem to say the same thing: "Checked exceptions must be declared!"
That's how I remember Java's behavior ever since.

image

image

We shouldn't surprise the consumer by forcing them to handle a checked exception that hasn't been declared. A better experience would be to adapt the API to declare the exception properly, or remove the catch block if the exception can't occur by contract—especially since such cases aren't testable.

> Currently, the API is designed to throw XmlReaderException. Check the implemented interface, there's no IOException thrown here.
> 
> >

If no exception is thrown, then there's no need for a catch block—especially if it's not testable.

```
org.mockito.exceptions.base.MockitoException: 
Checked exception is invalid for this method!
Invalid: java.io.IOException
```
image

Pankraz76:
thats my headache `Exception 'java.io.IOException' is never thrown in the corresponding try block`
image

Pankraz76:
> Currently, the API is designed to throw XmlReaderException. Check the implemented interface, there's no IOException thrown here.
> 
> >

please notice dedicated thread: https://github.com/apache/maven/pull/2304/files#r2079127763

gnodet:
It's not thrown, that's right.  So do not add a throw clause.

impl/maven-impl/src/main/java/org/apache/maven/impl/model/DefaultModelProcessor.java
gnodet: Are you aware that Maven has its own DI framework ? If you remove the required annotations, it will probably break ?
@Nullable can be used to inform the DI that the requirement is optional.
Did you run UTs and ITs ?

gnodet (changes requested):
I' not completely sure you actually understand what you remove.  
So let's keep focused PRs and let us know if you ran the tests successfully first.

impl/maven-impl/src/main/java/org/apache/maven/impl/model/DefaultModelProcessor.java
```diff
@@ -69,75 +71,74 @@
...
+        return Files.isDirectory(project) ? isRegularFileOrNull(project.resolve("pom.xml")) : project;
+    }
+
+    private static Path isRegularFileOrNull(Path pom) {
+        return Files.isRegularFile(pom) ? pom : null;
     }
 
     @Override
     public Model read(XmlReaderRequest request) throws IOException {
         Objects.requireNonNull(request, "source cannot be null");
```
elharo, line 116: source --> request

impl/maven-impl/src/main/java/org/apache/maven/impl/model/DefaultModelProcessor.java
gnodet: You changed the semantics, it's wrong.

impl/maven-impl/src/main/java/org/apache/maven/impl/model/DefaultModelProcessor.java
Pankraz76: tested:

image

impl/maven-impl/src/main/java/org/apache/maven/impl/model/DefaultModelProcessor.java
Pankraz76: > Did you run UTs and ITs ?

always run root `clean install`

> Are you aware that Maven has its own DI framework ? If you remove the required annotations, it will probably break ?

i delete/challenge everything and see if it breaks. All worked out locally thats why i want to lern and clarify, thx.

Pankraz76: image

Pankraz76:
wip reopen

PR #2304, opened 2025-05-08
Pull #2292: `Modernize codebase with Java improvements - test DefaultModelProcessor#read`

Pankraz76:
enable for:

- https://github.com/apache/maven/pull/2292


Pankraz76:
this one important to give liberty on code back to us; making code liquide while logic remains tight and dusty.

Pankraz76:
ideally when merging this, the test will be obsolet in in follow up:
- https://github.com/apache/maven/pull/2292

then we have awareness that coverage is a real deal and not a nice gimmick.

Kindly request your feedback @gnodet @elharo.

impl/maven-impl/src/main/java/org/apache/maven/impl/model/DefaultModelProcessor.java
```diff
@@ -114,9 +114,9 @@ public Model read(XmlReaderRequest request) throws IOException {
             }
             try {
                 return doRead(request);
-            } catch (IOException e) {
-                exceptions.forEach(e::addSuppressed);
-                throw e;
```
Pankraz76, line 119: would suggest forbidding one char vars only allowing in for loops like in cs itself.

impl/maven-impl/src/main/java/org/apache/maven/impl/model/DefaultModelProcessor.java
```diff
@@ -126,18 +126,14 @@ public Model read(XmlReaderRequest request) throws IOException {
...
             project = Paths.get(System.getProperty("user.dir"));
-        }
-        if (Files.isDirectory(project)) {
+        } else if (Files.isDirectory(project)) {
             Path pom = project.resolve("pom.xml");
             return Files.isRegularFile(pom) ? pom : null;
-        } else if (Files.isRegularFile(project)) {
-            return project;
-        } else {
-            return null;
```
Pankraz76, line 136: IMHO, it's either a directory (the higher-tier entity), as there is no file without a directory containing it. Which one came first is another topic, but I think it's that simple—either/or. So, we don't need this check, as the null branch is hard to test.

image

https://github.com/apache/maven/pull/2292/files#r2077436936

impl/maven-impl/src/main/java/org/apache/maven/impl/model/DefaultModelProcessor.java
```diff
@@ -126,18 +126,14 @@ public Model read(XmlReaderRequest request) throws IOException {
...
         }
+        return project;
     }
 
-    private Model doRead(XmlReaderRequest request) throws IOException {
```
Pankraz76, line 140: This is `wrong/counterfeit` because:
> Currently, the API is designed to throw XmlReaderException. Check the implemented interface, there's no IOException thrown here.
> 
> >

https://github.com/apache/maven/pull/2292#issuecomment-2860229309

impl/maven-impl/src/test/java/org/apache/maven/impl/model/DefaultModelProcessorTest.java
```diff
@@ -0,0 +1,349 @@
...
+class DefaultModelProcessorTest {
+
+    @TempDir
+    Path tempDir;
+
+    Path testProjectDir, testPomFile;
+
+    @AfterEach
+    void cleanup() throws IOException {
+        if (testPomFile != null && Files.exists(testPomFile)) {
```
elharo, line 59: probably don't need to double check if the file exists

impl/maven-impl/src/main/java/org/apache/maven/impl/model/DefaultModelProcessor.java
```diff
@@ -114,9 +114,9 @@ public Model read(XmlReaderRequest request) throws IOException {
             }
             try {
                 return doRead(request);
-            } catch (IOException e) {
-                exceptions.forEach(e::addSuppressed);
-                throw e;
+            } catch (RuntimeException ex) {
```
elharo, line 117: Per Effective Java don't catch a raw runtime exception, only the specific subclass you expect to be thrown

elharo (changes requested):

impl/maven-impl/src/test/java/org/apache/maven/impl/model/DefaultModelProcessorTest.java
Pankraz76: yes had same idea, then nested needed, as not all use this.

impl/maven-impl/src/main/java/org/apache/maven/impl/model/DefaultModelProcessor.java
```diff
@@ -94,50 +97,45 @@ public Path locateExistingPom(Path projectDirectory) {
...
-                    Optional<Model> model =
-                            parser.locateAndParse(projectDirectory, Map.of(ModelParser.STRICT, request.isStrict()));
-                    if (model.isPresent()) {
-                        return model.get().withPomFile(pomFile);
+                    Optional<Model> parent = parent(request, parser, pomFile.getParent());
+                    if (parent.isPresent()) {
+                        return parent.get().withPomFile(pomFile);
                     }
                 } catch (ModelParserException e) {
                     exceptions.add(e);
```
Pankraz76, line 122: is it mandatory `to continue if an error occurs` to give full error report? Or can we have here early exit if the branch kicks in like above in its counterpart?

impl/maven-impl/src/main/java/org/apache/maven/impl/model/DefaultModelProcessor.java
```diff
@@ -94,50 +97,45 @@ public Path locateExistingPom(Path projectDirectory) {
         return pom;
     }
 
+    private Path doLocateExistingPom(Path project) {
```
elharo, line 100: no need to use "do" on methods like this. locateExistingPom is still a verb and more natural.

impl/maven-impl/src/main/java/org/apache/maven/impl/model/DefaultModelProcessor.java
```diff
@@ -94,50 +97,45 @@ public Path locateExistingPom(Path projectDirectory) {
...
+            return Files.isRegularFile(pom) ? pom : null;
+        }
+        return project;
+    }
+
     @Override
     public Model read(XmlReaderRequest request) throws IOException {
-        Objects.requireNonNull(request, "source cannot be null");
-        Path pomFile = request.getPath();
+        Path pomFile = requireNonNull(request, "source cannot be null").getPath();
```
elharo, line 112: source --> request

impl/maven-impl/src/main/java/org/apache/maven/impl/model/DefaultModelProcessor.java
Pankraz76: yes sorry i expected this form main we on the pretty same page thats outdates somehow sorry.


Can i pitch that we switch to main BTW? 

https://github.com/github/renaming
https://www.theserverside.com/feature/Why-GitHub-renamed-its-master-branch-to-main

could be quite simple just coordinating from longer name `master` to simpler and not de facto standard `main`.

impl/maven-impl/src/test/java/org/apache/maven/impl/model/DefaultModelProcessorTest.java
```diff
@@ -0,0 +1,343 @@
...
+        Model originalModel = mock(Model.class);
+        Model modelWithPom = mock(Model.class);
+        when(originalModel.withPomFile(pomPath)).thenReturn(modelWithPom);
+        when(parser.locateAndParse(any(), any())).thenReturn(Optional.of(originalModel));
+        assertSame(modelWithPom, new DefaultModelProcessor(factory, List.of(parser)).read(request));
+        verify(originalModel).withPomFile(pomPath);
+    }
+
+    @Nested
+    class WithTmpFiles {
```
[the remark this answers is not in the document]
Pankraz76: SOC is kind of crazy at first, but it can be handy—and necessary for special test patterns. It simplifies complex scenarios without forcing you to massively copy or abstract test code, as logic can now be shared and layered in one place.  

It’s similar to a dedicated class but with inheritance, allowing special sub-conditions. That said, it’s flexible and therefore has significant trade-offs (both good and bad side effects).  

### JUnit Nested Classes: Quick Pros & Cons  

**Pros:**  
✅ Better organization  
✅ Clearer test grouping  
✅ Flexible setup per group  
✅ Reduces duplication  
✅ Good for stateful testing  

**Cons:**  
❌ Adds complexity  
❌ IDE navigation may suffer  
❌ Risk of over-nesting  
❌ Must be non-static  
❌ Learning curve for new devs  

**Best for:** Complex tests with shared setups or multiple states.

impl/maven-impl/src/main/java/org/apache/maven/impl/model/DefaultModelProcessor.java
gnodet: The fact it's hard to test does not mean the case does not exist.

impl/maven-impl/src/main/java/org/apache/maven/impl/model/DefaultModelProcessor.java
```diff
@@ -94,50 +97,45 @@ public Path locateExistingPom(Path projectDirectory) {
...
-        }
-        if (Files.isDirectory(project)) {
-            Path pom = project.resolve("pom.xml");
-            return Files.isRegularFile(pom) ? pom : null;
-        } else if (Files.isRegularFile(project)) {
-            return project;
-        } else {
-            return null;
-        }
+    private static Optional<Model> parent(XmlReaderRequest request, ModelParser parser, Path projectDirectory) {
```
gnodet, line 130: A parent usually refers to the parent model, so I'd rather use parentDirectory to make it clear what we're looking for.

impl/maven-impl/src/main/java/org/apache/maven/impl/model/DefaultModelProcessor.java
```diff
@@ -94,50 +97,45 @@ public Path locateExistingPom(Path projectDirectory) {
...
+        return parser.locateAndParse(projectDirectory, Map.of(STRICT, request.isStrict()));
     }
 
-    private Model doRead(XmlReaderRequest request) throws IOException {
-        return modelXmlFactory.read(request);
+    private static void throwErrorsSuppressed(List<ModelParserException> exceptions) throws IOException {
```
gnodet, line 134: We're not using the IOException in the API, so not sure why we'd want to build one here.  Model parsing exception aren't necessarily IoExceptions.

gnodet (changes requested):

impl/maven-impl/src/main/java/org/apache/maven/impl/model/DefaultModelProcessor.java
Pankraz76: yes thanks, will restore and cover.

impl/maven-impl/src/main/java/org/apache/maven/impl/model/DefaultModelProcessor.java
```diff
@@ -77,67 +80,69 @@ public DefaultModelProcessor(ModelXmlFactory modelXmlFactory, @Nullable List<Mod
...
     public Path locateExistingPom(Path projectDirectory) {
-        // Note that the ModelProcessor#locatePom never returns null
-        // while the ModelParser#locatePom needs to return an existing path!
-        Path pom = modelParsers.stream()
-                .map(m -> m.locate(projectDirectory)
-                        .map(org.apache.maven.api.services.Source::getPath)
-                        .orElse(null))
-                .filter(Objects::nonNull)
-                .findFirst()
-                .orElseGet(() -> doLocateExistingPom(projectDirectory));
```
Pankraz76, line 90: unpleasant do prefix is fix for naming collision @elharo

impl/maven-impl/src/main/java/org/apache/maven/impl/model/DefaultModelProcessor.java
```diff
@@ -77,67 +80,69 @@ public DefaultModelProcessor(ModelXmlFactory modelXmlFactory, @Nullable List<Mod
...
+                modelParsers.stream()
+                        .map(m -> m.locate(projectDirectory)
+                                .map(org.apache.maven.api.services.Source::getPath)
+                                .orElse(null))
+                        .filter(Objects::nonNull)
+                        .findFirst()
+                        .orElseGet(() -> locateExistingPomWithFallback(projectDirectory)));
+    }
+
+    private static Path throwIfWrongProjectDirLocation(Path projectDirectory, Path pom) {
```
Pankraz76, line 99: SOC

impl/maven-impl/src/main/java/org/apache/maven/impl/model/DefaultModelProcessor.java
```diff
@@ -77,67 +80,69 @@ public DefaultModelProcessor(ModelXmlFactory modelXmlFactory, @Nullable List<Mod
...
         if (pom != null && !pom.equals(projectDirectory) && !pom.getParent().equals(projectDirectory)) {
             throw new IllegalArgumentException("The POM found does not belong to the given directory: " + pom);
         }
         return pom;
     }
 
+    private Path locateExistingPomWithFallback(Path project) {
```
Pankraz76, line 106: SRP

impl/maven-impl/src/main/java/org/apache/maven/impl/model/DefaultModelProcessor.java
Pankraz76: covered.

impl/maven-impl/src/test/java/org/apache/maven/impl/model/DefaultModelProcessorTest.java
```diff
@@ -0,0 +1,400 @@
...
+import static org.mockito.Mockito.when;
+
+class DefaultModelProcessorTest {
+
+    @TempDir
+    Path tempDir;
+
+    @Test
+    void locateExistingPomShouldHandleRegularFileInput() throws IOException {
+        // Create a temporary file
```
Pankraz76, line 61: meldown

Pankraz76 (commented):
finally covered, small polish.

Pankraz76:
wip reopen

PR #2303, opened 2025-05-07
Pull #3003: `fix DefaultPluginXmlFactory#write`

Pankraz76:
Pull #3003: `fix DefaultPluginXmlFactory#write`

- https://github.com/apache/maven/pull/2292
- https://github.com/apache/maven/pull/2292/files#r2077643882

PR #2306, opened 2025-05-09
Fix `redundant` code in `DefaultPluginXmlFactory:write`

Pankraz76:
Fix `redundant` code in `DefaultPluginXmlFactory:write` - avoiding `DRY` increasing `cohesion` lowering `coupling`

- https://github.com/apache/maven/pull/2303

fix `unreachable` code:

https://github.com/apache/maven/blob/6be7a126d5778e781102445658da0b54b97df0e2/impl/maven-impl/src/main/java/org/apache/maven/impl/DefaultModelXmlFactory.java#L103

imho, this branch can never happen:

image

## because of early exit check before

This is too defensive over here. The last `if-else` is the actual final possible condition and therefore should just return:

Its `double-trouble` anyway and therefore `DRY violation` -  its a smell in eyes. 🎃

image

image

image

### DIFF
--- a/impl/maven-impl/src/main/java/org/apache/maven/impl/DefaultPluginXmlFactory.java
+++ b/impl/maven-impl/src/main/java/org/apache/maven/impl/DefaultPluginXmlFactory.java
@@ -38,26 +38,20 @@ import org.apache.maven.api.services.xml.XmlWriterRequest;
 import org.apache.maven.plugin.descriptor.io.PluginDescriptorStaxReader;
 import org.apache.maven.plugin.descriptor.io.PluginDescriptorStaxWriter;
 
-import static java.util.Objects.requireNonNull;
+import static org.apache.maven.impl.ImplUtils.nonNull;
 import static org.apache.maven.impl.StaxLocation.getLocation;
 import static org.apache.maven.impl.StaxLocation.getMessage;
 
 @Named
 @Singleton
 public class DefaultPluginXmlFactory implements PluginXmlFactory {
-
     @Override
     public PluginDescriptor read(@Nonnull XmlReaderRequest request) throws XmlReaderException {
-        return read(
-                request,
-                requireNonNull(request).getPath(),
-                request.getURL(),
-                request.getReader(),
-                request.getInputStream());
-    }
-
-    private static PluginDescriptor read(
-            XmlReaderRequest request, Path path, URL url, Reader reader, InputStream inputStream) {
+        nonNull(request, "request");
+        Path path = request.getPath();
+        URL url = request.getURL();
+        Reader reader = request.getReader();
+        InputStream inputStream = request.getInputStream();
         if (path == null && url == null && reader == null && inputStream == null) {
             throw new IllegalArgumentException("path, url, reader or inputStream must be non null");
         }
@@ -68,9 +62,14 @@ public class DefaultPluginXmlFactory implements PluginXmlFactory {
                 return xml.read(inputStream, request.isStrict());
             } else if (reader != null) {
                 return xml.read(reader, request.isStrict());
-            }
-            try (InputStream is = Files.newInputStream(requireNonNull(path))) {
-                return xml.read(is, request.isStrict());
+            } else if (path != null) {
+                try (InputStream is = Files.newInputStream(path)) {
+                    return xml.read(is, request.isStrict());
+                }
+            } else {
+                try (InputStream is = url.openStream()) {
+                    return xml.read(is, request.isStrict());
+                }
             }
         } catch (Exception e) {
             throw new XmlReaderException("Unable to read plugin: " + getMessage(e), getLocation(e), e);
@@ -79,14 +78,11 @@ public class DefaultPluginXmlFactory implements PluginXmlFactory {
 
     @Override
     public void write(XmlWriterRequest<PluginDescriptor> request) throws XmlWriterException {
-        write(
-                request.getWriter(),
-                request.getOutputStream(),
-                request.getPath(),
-                requireNonNull(requireNonNull(request, "request").getContent(), "content"));
-    }
-
-    private static void write(Writer writer, OutputStream outputStream, Path path, PluginDescriptor content) {
+        nonNull(request, "request");
+        PluginDescriptor content = nonNull(request.getContent(), "content");
+        Path path = request.getPath();
+        OutputStream outputStream = request.getOutputStream();
+        Writer writer = request.getWriter();
         if (writer == null && outputStream == null && path == null) {
             throw new IllegalArgumentException("writer, outputStream or path must be non null");
         }
@@ -106,7 +102,7 @@ public class DefaultPluginXmlFactory implements PluginXmlFactory {
     }
 
     /**
-     * Parse the given XML string.
+     * Simply parse the given xml string.
      *
      * @param xml the input XML string
      * @return the parsed object
@@ -118,7 +114,7 @@ public class DefaultPluginXmlFactory implements PluginXmlFactory {
     }
 
     /**
-     * Convert the given content to an XML string.
+     * Simply converts the given content to an XML string.
      *
      * @param content the object to convert
      * @return the XML string representation
--- a/impl/maven-impl/src/main/java/org/apache/maven/impl/DefaultPluginXmlFactory.java
+++ b/impl/maven-impl/src/main/java/org/apache/maven/impl/DefaultPluginXmlFactory.java
@@ -66,10 +66,9 @@ public class DefaultPluginXmlFactory implements PluginXmlFactory {
                 try (InputStream is = Files.newInputStream(path)) {
                     return xml.read(is, request.isStrict());
                 }
-            } else {
-                try (InputStream is = url.openStream()) {
-                    return xml.read(is, request.isStrict());
-                }
+            }
+            try (InputStream is = url.openStream()) {
+                return xml.read(is, request.isStrict());
             }
         } catch (Exception e) {
             throw new XmlReaderException("Unable to read plugin: " + getMessage(e), getLocation(e), e);
--- a/impl/maven-impl/src/main/java/org/apache/maven/impl/DefaultPluginXmlFactory.java
+++ b/impl/maven-impl/src/main/java/org/apache/maven/impl/DefaultPluginXmlFactory.java
@@ -25,6 +25,7 @@ import java.io.Writer;
 import java.net.URL;
 import java.nio.file.Files;
 import java.nio.file.Path;
+import java.util.Objects;
 
 import org.apache.maven.api.annotations.Nonnull;
 import org.apache.maven.api.di.Named;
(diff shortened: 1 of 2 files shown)


Pankraz76:
So no, this is not a dead branch - it serves an important purpose in the API's flexibility.

Conclusion:
✅ The else block is reachable and necessary.
It serves as a fallback when only a URL is provided in the XmlReaderRequest.

Would you like help adding a unit test to validate this branch?

Pankraz76:
But it's redundant then, as the behavior is now the same, and the end result is just without being overly complex and duplicating it—making the default exception explicit.

When the default behavior is the same in any case, it’s kind of simpler and should turn out equal in the end, right? So, classic redundancy—time for TDD (aka refactoring).

impl/maven-impl/src/main/java/org/apache/maven/impl/DefaultPluginXmlFactory.java
```diff
@@ -62,14 +63,9 @@ public PluginDescriptor read(@Nonnull XmlReaderRequest request) throws XmlReader
...
                 return xml.read(reader, request.isStrict());
-            } else if (path != null) {
-                try (InputStream is = Files.newInputStream(path)) {
-                    return xml.read(is, request.isStrict());
-                }
-            } else {
-                try (InputStream is = url.openStream()) {
-                    return xml.read(is, request.isStrict());
-                }
+            }
```
elharo, line 66: This throws an NPE if the path is null. I don't think that's what's wanted here

impl/maven-impl/src/main/java/org/apache/maven/impl/DefaultPluginXmlFactory.java
```diff
@@ -62,14 +63,9 @@ public PluginDescriptor read(@Nonnull XmlReaderRequest request) throws XmlReader
...
+            try (InputStream is = Files.newInputStream(Objects.requireNonNull(path))) {
+                return xml.read(is, request.isStrict());
             }
         } catch (Exception e) {
```
elharo, line 70: we shouldn't be catching raw exceptions here; probably fix that first

impl/maven-impl/src/main/java/org/apache/maven/impl/DefaultPluginXmlFactory.java
```diff
@@ -93,7 +89,7 @@ public void write(XmlWriterRequest<PluginDescriptor> request) throws XmlWriterEx
                 new PluginDescriptorStaxWriter().write(outputStream, content);
             } else {
                 try (OutputStream os = Files.newOutputStream(path)) {
-                    new PluginDescriptorStaxWriter().write(outputStream, content);
+                    new PluginDescriptorStaxWriter().write(os, content);
```
elharo, line 92: That line does look like a bug

impl/maven-impl/src/test/java/org/apache/maven/impl/DefaultPluginXmlFactoryTest.java
```diff
@@ -0,0 +1,260 @@
...
+            <?xml version="1.0" encoding="UTF-8"?>
+            <plugin>
+              <name>Sample Plugin</name>
+              <groupId>org.example</groupId>
+              <artifactId>sample-plugin</artifactId>
+              <version>1.0.0</version>
+            </plugin>
+            """;
+
+    private final DefaultPluginXmlFactory sut = new DefaultPluginXmlFactory();
```
elharo, line 61: sut --> factory

impl/maven-impl/src/test/java/org/apache/maven/impl/DefaultPluginXmlFactoryTest.java
```diff
@@ -0,0 +1,260 @@
...
+                .inputStream(new ByteArrayInputStream(SAMPLE_PLUGIN_XML.getBytes()))
+                .build());
+        assertEquals("Sample Plugin", descriptor.getName());
+        assertEquals("org.example", descriptor.getGroupId());
+        assertEquals("sample-plugin", descriptor.getArtifactId());
+        assertEquals("1.0.0", descriptor.getVersion());
+    }
+
+    @Test
+    void readFromReaderParsesPluginDescriptorCorrectly() {
```
elharo, line 78: This is a little long. Drop "Correctly" or maybe just testParsePlugin

impl/maven-impl/src/test/java/org/apache/maven/impl/DefaultPluginXmlFactoryTest.java
```diff
@@ -0,0 +1,260 @@
...
+        assertEquals(
+                "Sample Plugin",
+                sut.read(XmlReaderRequest.builder()
+                                .reader(new StringReader(SAMPLE_PLUGIN_XML))
+                                .build())
+                        .getName());
+    }
+
+    @Test
+    void readFromPathParsesPluginDescriptorCorrectly(@TempDir Path tempDir) throws Exception {
```
elharo, line 88: similar

impl/maven-impl/src/test/java/org/apache/maven/impl/DefaultPluginXmlFactoryTest.java
```diff
@@ -0,0 +1,260 @@
...
+                                .build())
+                        .build()));
+        assertTrue(exception.getMessage().contains("Unable to write plugin"));
+        assertInstanceOf(RuntimeException.class, exception.getCause());
+    }
+
+    @Test
+    void writeWithNoTargetThrowsIllegalArgumentException() {
+        assertEquals(
+                "writer, outputStream or path must be non null",
```
elharo, line 214: comma after outputStream and outputStream --> output stream

impl/maven-impl/src/test/java/org/apache/maven/impl/DefaultPluginXmlFactoryTest.java
```diff
@@ -0,0 +1,260 @@
...
+
+    @Test
+    void readMalformedXmlThrowsXmlReaderException() {
+        XmlReaderException exception = assertThrows(
+                XmlReaderException.class,
+                () -> sut.read(XmlReaderRequest.builder()
+                        .inputStream(new ByteArrayInputStream("<plugin><name>Broken Plugin".getBytes()))
+                        .build()));
+        assertTrue(exception.getMessage().contains("Unable to read plugin"));
+        assertInstanceOf(Exception.class, exception.getCause());
```
elharo, line 233: This assert is almost meaningless

impl/maven-impl/src/test/java/org/apache/maven/impl/DefaultPluginXmlFactoryTest.java
```diff
@@ -0,0 +1,260 @@
...
+
+    @Test
+    void readFromUrlParsesPluginDescriptorCorrectly(@TempDir Path tempDir) throws Exception {
+        Path xmlFile = tempDir.resolve("plugin.xml");
+        Files.write(xmlFile, SAMPLE_PLUGIN_XML.getBytes());
+        URL url = xmlFile.toUri().toURL();
+
+        // Create request with URL using reflection since builder doesn't have url() method
+        XmlReaderRequest request =
+                XmlReaderRequest.builder().inputStream(url.openStream()).build();
```
elharo, line 251: How is this different than just parsing any other input stream?

elharo (changes requested):

impl/maven-impl/src/main/java/org/apache/maven/impl/DefaultPluginXmlFactory.java
Pankraz76: yes already done:
- https://github.com/apache/maven/pull/2303

This PR is follow up. Thx

impl/maven-impl/src/main/java/org/apache/maven/impl/DefaultPluginXmlFactory.java
Pankraz76: see L:56:

`path == null`

This case is not possible. Only in our mind or if JVM is corrupt.

Pankraz76: early exit then tho IDE makes warning here too, requesting `requireNonNull`.

If path is null like you tell, whats `impossible` imho, then it will exit here again with runtime error.

image

impl/maven-impl/src/test/java/org/apache/maven/impl/DefaultPluginXmlFactoryTest.java
Pankraz76: will fix in prerequisite: 
- https://github.com/apache/maven/pull/2303 
thanks

impl/maven-impl/src/test/java/org/apache/maven/impl/DefaultPluginXmlFactoryTest.java
Pankraz76: `testParsePlugin`

impl/maven-impl/src/test/java/org/apache/maven/impl/DefaultPluginXmlFactoryTest.java
Pankraz76: `assertInstanceOf(WstxEOFException.class, exception.getCause());`

impl/maven-impl/src/test/java/org/apache/maven/impl/DefaultPluginXmlFactoryTest.java
elharo: "writer, output stream, or path must be non-null",

It's called an Oxford comma

Pankraz76: thx

Pankraz76: done thx.

impl/maven-impl/src/test/java/org/apache/maven/impl/DefaultPluginXmlFactoryTest.java
Pankraz76: gone.

Pankraz76:
https://github.com/apache/maven/pull/2312

impl/maven-impl/src/main/java/org/apache/maven/impl/DefaultPluginXmlFactory.java
Pankraz76: > If this case really isn't possible

4 params 4 cases thats mandatory.

Now the code has dedicated design making it clear.

impl/maven-impl/src/main/java/org/apache/maven/impl/DefaultPluginXmlFactory.java
Pankraz76: `IOException | XMLStreamException e`

need this: 
https://checkstyle.sourceforge.io/checks/coding/illegalcatch.html

image

Pankraz76: Of course, it's better API design to be precise and not broad—especially in this case.

If possible we should ensure this standard as well. In `checkstyle` its the same (of course as all the checkers are on), so i know this quite well. Its a good thing to strengthen the design.

https://github.com/apache/maven-shared-resources/pull/73

PR #2311, opened 2025-05-09
Fix for Java 24 w/ upcoming new Eclipse Sisu

cstamas:
The upcoming Eclipse Sisu will not provide anymore "shaded" JARs (with out without ASM).
Also, fixed the `maven-executor` test that failed on "unexpected stderr output" that now happens with Java 24 due Guava.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -829,6 +829,17 @@ under the License.</licenseText>
             </dependency>
           </dependencies>
         </plugin>
+        <plugin>
+          <groupId>org.apache.maven.plugins</groupId>
+          <artifactId>maven-pmd-plugin</artifactId>
+          <dependencies>
+            <dependency>
+              <groupId>net.sourceforge.pmd</groupId>
+              <artifactId>pmd-core</artifactId>
+              <version>7.12.0</version>
+            </dependency>
+          </dependencies>
+        </plugin>
       </plugins>
     </pluginManagement>
     <plugins>


cstamas:
The "full-build" CI job builds Maven Site, that contains PMD report that craps out on Java 24:
```
	at net.sourceforge.pmd.lang.java.types.TypeOps.isSubClassOfNoInterface(TypeOps.java:1957)
	at net.sourceforge.pmd.lang.java.types.TypeOps.isSubClassOfNoInterface(TypeOps.java:1957)
	at net.sourceforge.pmd.lang.java.types.TypeOps.isSubClassOfNoInterface(TypeOps.java:1957)
	at net.sourceforge.pmd.lang.java.types.TypeOps.isSubClassOfNoInterface(TypeOps.java:1957)
Error: [ERROR] Unknown error occurred while executing a PmdRunnable: **java.lang.StackOverflowError**
```

cstamas:
The IT failures:
```
[ERROR] Errors: 
[ERROR]   MavenITmng7045DropUselessAndOutdatedCdiApiTest.testShouldNotLeakCdiApi:41 » Verification Exit code was non-zero: 1; command line and log = 
[ERROR]   MavenITmng7160ExtensionClassloader.testVerify:44 » Verification Text not found in log: xpp3 -> mvn
[ERROR]   MavenITmng7587Jsr330.test:57 » Verification Exit code was non-zero: 1; command line and log = 
```

7045 craps out due old Groovy on Java 24, the latter two due lack of sisu index, but the build does not calls index plugin explicitly...

slawekjaranowski:
maybe it is time for:
- https://github.com/apache/maven-parent/issues/254
or even though
- https://github.com/apache/maven-parent/issues/238

impl/maven-executor/src/main/java/org/apache/maven/cling/executor/internal/ToolboxTool.java
```diff
@@ -122,27 +122,29 @@ private ExecutorRequest.Builder mojo(ExecutorRequest.Builder builder, String moj
...
         }
+        return request;
     }
 
     /**
      * Performs "sanity check" for output, making sure no insane values like empty strings are returned.
      */
-    private String validateOutput(boolean shave, ByteArrayOutputStream stdout, ByteArrayOutputStream stderr) {
+    private String validateOutput(
+            ExecutorRequest request, boolean shave, ByteArrayOutputStream stdout, ByteArrayOutputStream stderr) {
```
slawekjaranowski, line 140: Where a `request` parameter is used in this method?

cstamas: Nowhere, you are right. The original problem was that on "unexpected stderr output" this method was throwing, but on Java 24 that "unexpected output" is present. Change can be simplified by dropping this method.

(originally I wanted to send content to caller, just realize that this class is setting the streams, not caller, caller is really interested in success (exit code = 0) and stdout).

cstamas: Removed all the changes in this class except the needed one.

slawekjaranowski (approved):

slachiewicz (approved):

impl/maven-executor/src/main/java/org/apache/maven/cling/executor/internal/ToolboxTool.java
```diff
@@ -142,7 +142,7 @@ private String validateOutput(boolean shave, ByteArrayOutputStream stdout, ByteA
         }
         // sanity checks: stderr has any OR result is empty string (no method should emit empty string)
         if (stderr.size() > 0 || result.trim().isEmpty()) {
-            throw new ExecutorException(
+            System.err.println(
```
Pankraz76, line 145: do we have logger for this? using system directly would cause spotbugs violations. We having this code the pattern a lot so its a bigger thing kind of.

https://stackoverflow.com/questions/8601831/do-not-use-system-out-println-in-server-side-code#:~:text=71-,System.,until%20the%20println%20has%20finished.

Why cannot we throw anymore, please? Of course we dont want to lose the information. As we having error we can not tread, how and why continue? 
Thanks.

cstamas: We cannot throw, as this code (whole `maven-executor`) purpose is to execute Maven (embedded in JVM, or forked as in forked child process) _real Maven mvn command_ in IT-like scenarios (this very code is used in Maven ITs). This codebase supports Maven 3 and Maven 4 transparently. The goal of PR is to "prepare it for Java 24".

All existing Maven versions released so far (including Maven 4 rc releases) **emit warning on Java 24**, and there is nothing to be done here (the MNG-8399 and/or alike). Hence, to make this code work on Java 24 AND be able to use current releases (ie. 3.9.9 or 4-rc-3), it cannot throw anymore, as err logging is part of "normal operation" on those Maven versions on Java 24. Or in other words, if we'd throw here (as before), we would not be able to run Maven 3.9.9 and 4-rc-3 and older Maven versions using Java 24, only "fixed" Maven 3.9.10 and upcoming 4-rc-4, and that would narrow the capabilities of this module too much.

The point of this PR is to prepare codebase be able to run on Java 24 as is, and this throw was one of the obstacles preventing it.

pom.xml
```diff
@@ -829,6 +829,17 @@ under the License.</licenseText>
             </dependency>
           </dependencies>
         </plugin>
+        <plugin>
+          <groupId>org.apache.maven.plugins</groupId>
+          <artifactId>maven-pmd-plugin</artifactId>
```
Pankraz76, line 834: why do we have cost of carry? this plugin is not in use:

impl/maven-executor/src/main/java/org/apache/maven/cling/executor/internal/ToolboxTool.java
Pankraz76: thx.

PR #2315, opened 2025-05-10
Include module `core-its`

Pankraz76:
enabler for: 
- https://github.com/apache/maven/pull/2319

I wonder why it’s not included. I was thinking about this, assuming it might be related to integration tests—and it turned out to be true.

Why not provide this information upfront?

If it’s resource-intensive, it should only run in a dedicated stage.

Ignore whats actually there seems strange:

image

`use` it or `loose` it.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -103,6 +103,7 @@ under the License.
     <module>impl</module>
     <module>compat</module>
     <module>apache-maven</module>
+    <module>its</module>
   </modules>
 
   <scm>


Pankraz76:
wip

gnodet (commented):
The `its` are activated in a profile.
This could be a good idea, but I don't think the default config would use the distribution built earlier in the build, bur rather the distribution used to actually build maven.  If you find a way to fix that, that would be nice.
As to wether ITs should be activated by default, it would be doable, but the CI scripts would have to be adjusted, and we'd need a way to not include then (i.e. turn the profile to activate the ITs unless a property is set, or something like that).

gnodet (changes requested):
The its are activated in a profile.
This could be a good idea, but I don't think the default config would use the distribution built earlier in the build, bur rather the distribution used to actually build maven. If you find a way to fix that, that would be nice.
As to wether ITs should be activated by default, it would be doable, but the CI scripts would have to be adjusted, and we'd need a way to not include then (i.e. turn the profile to activate the ITs unless a property is set, or something like that).

gnodet:
Please refrain from raising PRs when you have questions.  Just go on slack and ask there, there will be plenty of people to answer questions.

Pankraz76:
Assuming the tight build time has only increased by a few seconds, it seems that nothing has changed except for including the project, making it directly visible in the Project Explorer.

Since these are integration tests, they should not be excluded by default due to resource limitations.

I just mean—try the CI build. It shouldn't break anything and will only slightly increase the local build time by a few seconds.

`clean, install` is all we need to care on local machine, so to me looks good.

image


image

https://maven.apache.org/surefire/maven-failsafe-plugin/

Pankraz76:
might consider folder alignment as follow up:

image

gnodet:
I actually don't like folder alignment proposal.
On the contrary I think the module maven-api-core should be in the directory maven/api/core and not maven/maven-api/maven-api-core

gnodet (changes requested):
I don't think it's a good idea to integrate the its subprojects in the build (even without running ITs) unless we can actually run ITs, as the assumption would be that we can.

Pankraz76:
> unless we can actually run ITs, as the assumption would be that we can.

Why Should Developers Run This High-Intensity Task? Its excluded in dedicated profile, like expected for good reason. 

**Intent**:  
- Make the project visible/accessible  
- Prevent compilation issues  

**Key Points**:  
- Only adds ~30 seconds to runtime for significant benefits  
- ITs will execute `as before` in cloud environment
- ITs remain disabled by default locally (aligns with convention-over-configuration principle)  

**Our Approach**:  
- Minimal implementation  
- Focuses only on real issues (e.g., non-compliant code)  
- Avoids unnecessary overhead  
  - nothing changes nothing breaks **until real issue given**

Pankraz76:
> I actually don't like folder alignment proposal.

Whats the matter?

**Current Problems**:
- Out of sync implementation and design module id and actual module name aka. dir structure causing double rendering in IDE.
- Causes unnecessary overhead

**Assessment**:
- This particular module is stable and safe to adjust
- While there's always risk when modifying implementations/APIs (especially regarding backward compatibility)

**Key Argument**:
- Misalignment in code is not acceptable
- Should be targeted for rework
- Consistency should be prioritized

gnodet:
I agree with consistency.  That does not necessarily mean aligning directory structure with artifact id.

Pankraz76:
its possible, having only small benefit yes, but idk why we should hide the code.

its there, lets admin and embrace.

gnodet:
> > I actually don't like folder alignment proposal.
> 
> 
> 
> Whats the matter?
> 
> 
> 
> **Current Problems**:
> 
> - Out of sync implementation and design module id and actual module name aka. dir structure causing double rendering in IDE.

That' an IDE problem.

It hinders ease of navigation in the console for example.  Going to a subdirectory or building a subproject needs more typing because the completion with tab in your shell cannot complete directly.  You first need to autocomplete the unambiguous part for example.

> 
> - Causes unnecessary overhead
> 
> 
> 
> **Assessment**:
> 
> - This particular module is stable and safe to adjust
> 
> - While there's always risk when modifying implementations/APIs (especially regarding backward compatibility)
> 
> 
> 
> **Key Argument**:
> 
> - Misalignment in code is not acceptable
> 
> - Should be targeted for rework
> 
> - Consistency should be prioritized

gnodet:
> > unless we can actually run ITs, as the assumption would be that we can.
> 
> 
> 
> Why Should Developers Run This High-Intensity Task? Its excluded in dedicated profile, like expected for good reason. 
> 
> 
> 
> **Intent**:  
> 
> - Make the project visible/accessible  
> 
> - Prevent compilation issues  
> 
> 
> 
> **Key Points**:  
> 
> - Only adds ~30 seconds to runtime for significant benefits  
> 
> - ITs will execute `as before` in cloud environment
> 
> - ITs remain disabled by default locally (aligns with convention-over-configuration principle)  
> 
>   - only ensuring everything tight; production ready 
> 
> 
> 
> **Our Approach**:  
> 
> - Minimal implementation  
> 
> - Focuses only on real issues (e.g., non-compliant code)  
> 
> - Avoids unnecessary overhead  
> 
>   - nothing changes nothing breaks **until real issue given**

Again I'm fine _if_ we can make the ITs also run.  It's have been merged into the core project only a few months ago, they were in a separate git repo for ages.
Having the ability to run ITs does not mean we should by default.
And breaking compilation of ITs almost never happen.  The reason is that they are not even compiled against the jars that are being built.  So there's almost no benefit in including them by default (even without actually running the ITs).

Pankraz76:
> make the ITs also run

yes, this the next step imho to have clear history.

But still it will be the same like now clean install basically just compiling. No one will ever run these on daily bases due to heavy load. If they run now on some cloud they should run everywhere. 

its dedicated concern. I will run locally then we can go on.

Pankraz76:
item:

Caused by: org.eclipse.aether.resolution.ArtifactResolutionException: The following artifacts could not be resolved: org.apache.maven.its.mng5639:a:pom:0.1 (absent): Could not find artifact org.apache.maven.its.mng5639:a:pom:0.1 in central (https://repo.maven.apache.org/maven2)



it takes 9 mins as expected even if build success makes no difference because this it designed to run online on large scale only for dev on subset locally. But this is different topic.


considering MakeItWorkMakeItRightMakeItFast

merging it. then trying to go the next step. How is it working ATM are there 2 versions? A working setup should be assumed for such high critical system.

https://wiki.c2.com/?MakeItWorkMakeItRightMakeItFast

Pankraz76:
wip reopen when ci green

gnodet (changes requested):
Adding ITs to the build is wrong, unless they can actually run with the being built assembly.  The argument about avoid possible compilation breakage is wrong, as the ITs don't use the jars that are being built, but much older versions usually.
Please fix the first point or abandon the idea.  The second point should not be attempted at this point, as the policy is not to upgrade ITs to latest dependencies, but only test with the latest maven.

Pankraz76:
thanks for info.

Pankraz76:
[ERROR] Could not find goal 'addPluginArtifactMetadata' in plugin org.apache.maven.plugins:maven-plugin-plugin:0.1-stub-SNAPSHOT among available goals descriptor -> [Help 1]

this seems fixable.

Any suggestion how to handle this?

PR #2281, opened 2025-05-01
[POC] Migrate to `Java 21` - code

Pankraz76:
[POC] Migrate to `Java 21` - code

depends on:
- https://github.com/apache/maven/pull/2319

apply:
- https://docs.openrewrite.org/recipes/java/migrate/upgradetojava21

PR #2319, opened 2025-05-10
[POC] Migrate to `Java 21` - infrastructure

Pankraz76:
[POC] Migrate to `Java 21` - infrastructure

depends on:
- https://github.com/apache/maven/pull/2315

enabler for: 
- https://github.com/apache/maven/pull/2281
- https://github.com/apache/maven/pull/2281#pullrequestreview-2813199057

> We need a separate PR for bumping the required JDK to 21. Other PRs can follow and be based on that one.

WOMM

image

### DIFF
--- a/.github/workflows/maven.yml
+++ b/.github/workflows/maven.yml
@@ -33,7 +33,7 @@ jobs:
       - name: Set up JDK
         uses: actions/setup-java@v4
         with:
-          java-version: 17
+          java-version: 21
           distribution: 'temurin'
 
       - name: Checkout maven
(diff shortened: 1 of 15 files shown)


Pankraz76:
this is a nice example why we should integrate all modules.

- https://github.com/apache/maven/pull/2315

Now it `its` remains outdated and out of sync. 

need to pay special attention. When integrated, its done in one way.

Pankraz76:
kindly request feedback.

gnodet:
I don't think it's a good idea to integrate the its subprojects in the build (even without running ITs) unless we can actually run ITs, as the assumption would be that we can.

gnodet:
The decision on switching to JDK 21 has not been settled yet.
And there's already another PR for it.

PR #2320, opened 2025-05-11
[POC] JUnit Jupiter best practices

Pankraz76:
[POC] JUnit Jupiter best practices

- https://docs.openrewrite.org/recipes/java/testing/junit5/junit5bestpractices

### DIFF
--- a/api/maven-api-core/src/test/java/org/apache/maven/api/MonotonicClockTest.java
+++ b/api/maven-api-core/src/test/java/org/apache/maven/api/MonotonicClockTest.java
@@ -37,7 +37,7 @@ class MonotonicClockTest {
 
     @Test
     @DisplayName("MonotonicClock singleton instance should always return the same instance")
-    void testSingletonInstance() {
+    void singletonInstance() {
         MonotonicClock clock1 = MonotonicClock.get();
         MonotonicClock clock2 = MonotonicClock.get();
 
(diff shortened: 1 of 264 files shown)


elharo (changes requested):
I strongly prefer to prefix method names with "test". method names are verbs that describe what the method does. That's the best practice to follow. A method that tests foo should be named testFoo, not foo.

Pankraz76:
yes everyone has some opinion.
Thats the reason for best practice, not having to discuss every detail over and over again.

This idea(l) is known as https://en.wikipedia.org/wiki/Convention_over_configuration

As the whole world is probably following, its hart to outsmart them.

Pankraz76:
> I strongly prefer to prefix method names with "test". method names are verbs that describe what the method does. That's the best practice to follow. A method that tests foo should be named testFoo, not foo.

being outdated is no problem, it will catch us all. Its only a problem when not evolving therefore leaving being risking being obsolet.

image

- https://junit.org/junit5/docs/current/user-guide/#writing-tests-parallel-execution-synchronization
- https://thejavaguy.org/posts/011-dont-prefix-junit-tests-with-the-word-test/

Pankraz76:
principle here might be [DRY](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself)

as first thing we read is trash just like the null checks 

very first attention already going into void. Thats bad design. We tend to apply functional style, reducing everything. Then we are left over with only real business logic and not fluff.

Pankraz76:
FYI

PR #2328, opened 2025-05-12
[MNG-8727] Prepare for Java 24

cstamas:
Simple changes that makes build/ITs work with Java 24.

PR does not contains:
* action addition of Java 24 to CI
* updated Eclipse Sisu version (needed for that above)

---

https://issues.apache.org/jira/browse/MNG-8727

### DIFF
--- a/impl/maven-executor/src/main/java/org/apache/maven/cling/executor/internal/ToolboxTool.java
+++ b/impl/maven-executor/src/main/java/org/apache/maven/cling/executor/internal/ToolboxTool.java
@@ -142,7 +142,7 @@ public class ToolboxTool implements ExecutorTool {
         }
         // sanity checks: stderr has any OR result is empty string (no method should emit empty string)
         if (stderr.size() > 0 || result.trim().isEmpty()) {
-            throw new ExecutorException(
+            System.err.println(
                     "Unexpected stdout[" + stdout.size() + "]=" + stdout + "; stderr[" + stderr.size() + "]=" + stderr);
         }
         return result;
--- a/its/core-it-suite/src/test/resources/mng-7045/pom.xml
+++ b/its/core-it-suite/src/test/resources/mng-7045/pom.xml
@@ -43,13 +43,13 @@ javax.enterprise.inject.Instance.class.getDeclaredMethod("stream")]]></script>
           <dependency>
             <groupId>org.apache.groovy</groupId>
             <artifactId>groovy-ant</artifactId>
-            <version>4.0.15</version>
+            <version>4.0.26</version>
             <scope>runtime</scope>
           </dependency>
           <dependency>
             <groupId>org.apache.groovy</groupId>
             <artifactId>groovy</artifactId>
-            <version>4.0.15</version>
+            <version>4.0.26</version>
             <scope>runtime</scope>
           </dependency>
         </dependencies>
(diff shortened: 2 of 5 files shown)


pom.xml
```diff
@@ -796,6 +796,17 @@ under the License.</licenseText>
             </excludes>
           </configuration>
         </plugin>
+        <plugin>
+          <groupId>org.apache.maven.plugins</groupId>
+          <artifactId>maven-pmd-plugin</artifactId>
```
gnodet, line 801: Is that related ?

cstamas: Yes, older PMD ctaps out on Java 24

impl/maven-executor/src/main/java/org/apache/maven/cling/executor/internal/ToolboxTool.java
```diff
@@ -142,7 +142,7 @@ private String validateOutput(boolean shave, ByteArrayOutputStream stdout, ByteA
         }
         // sanity checks: stderr has any OR result is empty string (no method should emit empty string)
         if (stderr.size() > 0 || result.trim().isEmpty()) {
-            throw new ExecutorException(
+            System.err.println(
```
michael-o, line 145: Why?

cstamas: To not have it lost. Without this, the output is lost, this way you have it at least in surefire report.
And with throw it would just fail (as on Java 24 mvn4 always emits warning on stderr).

michael-o: I see, makes sense.

its/core-it-suite/src/test/resources/mng-7045/pom.xml
```diff
@@ -43,13 +43,13 @@ javax.enterprise.inject.Instance.class.getDeclaredMethod("stream")]]></script>
...
             <artifactId>groovy-ant</artifactId>
-            <version>4.0.15</version>
+            <version>4.0.26</version>
             <scope>runtime</scope>
           </dependency>
           <dependency>
             <groupId>org.apache.groovy</groupId>
             <artifactId>groovy</artifactId>
-            <version>4.0.15</version>
+            <version>4.0.26</version>
```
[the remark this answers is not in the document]
cstamas: this is an IT and this version moves first time since IT was added...  I may be a property, but ... meh

michael-o (approved):

its/core-it-suite/src/test/resources/mng-7045/pom.xml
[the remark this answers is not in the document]
cstamas: Groovy 4.0.26 contains ASM 9.7.1, and no newer release so far. I guess there will be once Java 26 is out? So this version will change once we move CI to Java 26+

pom.xml
Pankraz76: nice to see things going in same direction

impl/maven-executor/src/main/java/org/apache/maven/cling/executor/internal/ToolboxTool.java
Pankraz76: leftover...

image

thats why we to rely on tools, not on human try&error:
https://docs.openrewrite.org/recipes/staticanalysis/unnecessarythrows

PR #2310, opened 2025-05-09
RemoveUnusedPrivateMethods

Pankraz76:
- https://docs.openrewrite.org/recipes/staticanalysis/removeunusedprivatemethods
- https://github.com/checkstyle/checkstyle/issues/16960
- https://github.com/checkstyle/checkstyle/pull/16894

PR #2334, opened 2025-05-13
chore: drop obsolete `DefaultXmlService#findNodeById`

Pankraz76:
- https://github.com/apache/maven/pull/2310
- https://github.com/apache/maven/pull/2364

### DIFF
--- a/impl/maven-xml/src/main/java/org/apache/maven/internal/xml/DefaultXmlService.java
+++ b/impl/maven-xml/src/main/java/org/apache/maven/internal/xml/DefaultXmlService.java
@@ -392,34 +392,6 @@ public class DefaultXmlService extends XmlService {
         return !isEmpty(value) ? value : DEFAULT_CHILDREN_COMBINATION_MODE;
     }
 
-    @Nullable
-    private static XmlNode findNodeById(@Nonnull List<XmlNode> nodes, @Nonnull String id) {
-        return nodes.stream()
-                .filter(n -> id.equals(n.attribute(ID_COMBINATION_MODE_ATTRIBUTE)))
-                .findFirst()
-                .orElse(null);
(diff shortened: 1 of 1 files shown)


impl/maven-core/src/main/java/org/apache/maven/internal/transformation/impl/DefaultConsumerPomArtifactTransformer.java
```diff
@@ -117,17 +116,6 @@ private void deferDeleteFile(Path generatedFile) {
         toDelete.add(generatedFile.toAbsolutePath());
     }
 
-    @PreDestroy
-    private void doDeleteFiles() {
```
Pankraz76, line 121: need to test. 

should break break.

impl/maven-xml/src/main/java/org/apache/maven/internal/xml/DefaultXmlService.java
```diff
@@ -392,34 +392,6 @@ private static String getChildCombinationMode(Map<String, String> attributes) {
         return !isEmpty(value) ? value : DEFAULT_CHILDREN_COMBINATION_MODE;
     }
 
-    @Nullable
-    private static XmlNode findNodeById(@Nonnull List<XmlNode> nodes, @Nonnull String id) {
```
Pankraz76, line 396: one time usage so might be lost:

image

Pankraz76: its private and never used.

use or loose: its all alone, waiting to be avoided.

https://github.com/search?q=repo%3Aapache%2Fmaven%20findNodeById&type=code

@elharo

PR #2336, opened 2025-05-14
[MNG-8727] Prepare for Java 24 - Unnecessary throws

Pankraz76:
- https://github.com/apache/maven/pull/2328
-


Pankraz76:
leftover...

image

thats why CPU `out-smart` human `try&error`, not the other way around, right?:
https://docs.openrewrite.org/recipes/staticanalysis/unnecessarythrows

michael-o:
Is this an unchecked exception?

Pankraz76:
sorry. No its Runtime. Why this makes sense then? and if important then make in interface to enforce.

image

impl/maven-executor/src/main/java/org/apache/maven/cling/executor/internal/ToolboxTool.java
```diff
@@ -100,8 +100,7 @@ public String artifactPath(ExecutorRequest.Builder executorRequest, String gav,
     }
 
     @Override
-    public String metadataPath(ExecutorRequest.Builder executorRequest, String gav, String repositoryId)
-            throws ExecutorException {
+    public String metadataPath(ExecutorRequest.Builder executorRequest, String gav, String repositoryId) {
```
gnodet, line 103: The interface actually throws the exception.  Given this is a public method, I'd rather keep it, as it may forbid derived classes to actually throw the exception.

gnodet: Ah, it's an unchecked exception, so the argument is moot.

michael-o:
Unchecked ones should be in Javadoc only, not on the signature.

impl/maven-executor/src/main/java/org/apache/maven/api/cli/ExecutorException.java
```diff
@@ -48,4 +48,8 @@ public ExecutorException(@Nullable String message) {
     public ExecutorException(@Nullable String message, @Nullable Throwable cause) {
         super(message, cause);
     }
+
+    public RuntimeException toRuntime() {
```
Pankraz76, line 52:
```suggestion
    public RuntimeException toRethrow() {
```

Pankraz76:
> Unchecked ones should be in Javadoc only, not on the signature.

kindly request critics.

impl/maven-executor/src/main/java/org/apache/maven/api/cli/ExecutorException.java
```diff
@@ -29,7 +29,7 @@
  * @since 4.0.0
  */
 @Experimental
-public class ExecutorException extends RuntimeException {
```
Pankraz76, line 32: sounds `important` enough, being handled, receiving dedication and segregation.

impl/maven-executor/src/main/java/org/apache/maven/cling/executor/embedded/EmbeddedMavenExecutor.java
```diff
@@ -176,7 +176,7 @@ protected void disposeRuntimeCreatedRealms(Context context) {
                 }
             }
         } catch (Exception e) {
-            throw new ExecutorException("Failed to dispose runtime created realms", e);
+            throw new RuntimeException(new ExecutorException("Failed to dispose runtime created realms", e));
```
hgschmie, line 179: this makes zero sense. First make the actual exception a checked exception, then wrap it into a (untyped) runtime exception.

Now there is no longer a ExecutorException thrown but a RuntimeException.

AI Slop.

hgschmie (changes requested):

jira-importer:
Resolve #9569

PR #2341, opened 2025-05-16
[MNG-8731] Use https for xsi:schemaLocation in generated descriptors

slawekjaranowski:
Attribute xml.schemaLocation is used by modello to generate documentations

We have mixed http/https, so it should be fixed to use everywhere https


gnodet (approved):

jira-importer:
Resolve #9568

PR #2342, opened 2025-05-16
[MNG-8730] Fix versions of new items in settings.xml

slawekjaranowski:
https://issues.apache.org/jira/browse/MNG-8730


slawekjaranowski:
ok, looks more complicated we use new settings items in old code ....

gnodet:
I'm not really sure we need to upgrade to 2.0.0.  Or is that to leave room to upgrade Maven 3.x with only a subset of those ?

jira-importer:
Resolve #9347

PR #2343, opened 2025-05-16
[MNG-8731] Use https for xsi:schemaLocation in generated descriptors

slawekjaranowski:
Attribute xml.schemaLocation is used by modello to generate documentations

Also update modello descriptor to 2.0.0


jira-importer:
Resolve #9568

PR #2347, opened 2025-05-18
[MNG-8696] Hide the cache from DefaultDependencyResolverResult constructor

desruisseaux:
This is an alternative to pull request #2219 resolving the same issue, but with more emphasis that `PathModularizationCache` should not be created in the `DefaultDependencyResolverResult` constructor, and that doing so is a temporary hack that may be removed in a future version. More specifically:

* Make package-private the `DefaultDependencyResolverResult` constructor having a `PathModularizationCache` argument, because the latter is package-private.
* Add a public constructor without the cache argument for codes in other packages that need to instantiate `DefaultDependencyResolverResult`  directly.
  * Put a warning in the javadoc saying that this constructor may be removed in a future version.
* Add `PathModularizationCache` private field in `DefaultDependencyResolver`. This is partially for performance reasons (see below), but also for expressing the intend that the cache has a longer lifetime than `DependencyResolverResult`.

## Rational
Initializing `PathModularizationCache` inside the `DependencyResolverResult` constructor is close to useless, because a cache is useful only when the cached values are reused. The way that the `DependencyResolverResult` is coded, the same result instance is unlikely to ask the same cached value twice. The cache become useful only when many `DependencyResolverResult` are instantiated while reusing the same cache.

The "useless" approach is nevertheless used by `DependencyResolverResult`, but this is only because I do not yet know how to manage a session-wide cache. The current way to create the cache was intended to be temporary.

I'm not sure if it happens often that the same `DependencyResolverResultBuilder` is reused for creating more than one `DependencyResolverResult`. If not, moving the cache inside `DependencyResolverResultBuilder`, as done in this commit, may not bring real performance improvement. However, it makes clearer that `PathModularizationCache` is expected to have a longer lifetime than `DependencyResolverResult`. The discussion in #2219 gives the impression that such clarification is worth to do.

### DIFF
--- a/impl/maven-impl/src/main/java/org/apache/maven/impl/DefaultDependencyResolver.java
+++ b/impl/maven-impl/src/main/java/org/apache/maven/impl/DefaultDependencyResolver.java
@@ -67,14 +67,15 @@ import static org.apache.maven.impl.ImplUtils.nonNull;
 @Named
 @Singleton
 public class DefaultDependencyResolver implements DependencyResolver {
+
     /**
      * Cache of information about the modules contained in a path element.
      * This cache is created when first needed. It may be never created.
      *
      * <p><b>TODO:</b> This field should not be in this class, because the cache should be global to the session.
-     * This field exists here only temporarily, until we clarified where to store session-wide caches.</p>
+     * This field exists here only temporarily, until clarified where to store session-wide caches.</p>
      *
-     * @see moduleCache()
+     * @see #moduleCache()
      */
     private PathModularizationCache moduleCache;
 
@@ -245,7 +246,7 @@ public class DefaultDependencyResolver implements DependencyResolver {
      * {@return the cache of information about the modules contained in a path element}.
      *
      * <p><b>TODO:</b> This method should not be in this class, because the cache should be global to the session.
-     * This method exists here only temporarily, until we clarified where to store session-wide caches.</p>
+     * This method exists here only temporarily, until clarified where to store session-wide caches.</p>
      */
     private PathModularizationCache moduleCache() {
         if (moduleCache == null) {
--- a/impl/maven-impl/src/main/java/org/apache/maven/impl/DefaultDependencyResolver.java
+++ b/impl/maven-impl/src/main/java/org/apache/maven/impl/DefaultDependencyResolver.java
@@ -67,6 +67,21 @@ import static org.apache.maven.impl.ImplUtils.nonNull;
 @Named
 @Singleton
 public class DefaultDependencyResolver implements DependencyResolver {
+    /**
+     * Cache of information about the modules contained in a path element.
+     * This cache is created when first needed. It may be never created.
+     *
+     * <p><b>TODO:</b> This field should not be in this class, because the cache should be global to the session.
+     * This field exists here only temporarily, until we clarified where to store session-wide caches.</p>
+     *
+     * @see moduleCache()
+     */
+    private PathModularizationCache moduleCache;
+
+    /**
+     * Creates an initially empty resolver.
+     */
+    public DefaultDependencyResolver() {}
 
     @Nonnull
     @Override
@@ -126,7 +141,11 @@ public class DefaultDependencyResolver implements DependencyResolver {
                 final CollectResult result =
                         session.getRepositorySystem().collectDependencies(systemSession, collectRequest);
                 return new DefaultDependencyResolverResult(
-                        null, null, result.getExceptions(), session.getNode(result.getRoot(), request.getVerbose()), 0);
+                        null,
+                        moduleCache(),
+                        result.getExceptions(),
+                        session.getNode(result.getRoot(), request.getVerbose()),
+                        0);
             } catch (DependencyCollectionException e) {
                 throw new DependencyResolverException("Unable to collect dependencies", e);
             }
@@ -191,18 +210,14 @@ public class DefaultDependencyResolver implements DependencyResolver {
                         .map(Artifact::toCoordinates)
                         .collect(Collectors.toList());
                 Predicate<PathType> filter = request.getPathTypeFilter();
+                DefaultDependencyResolverResult resolverResult = new DefaultDependencyResolverResult(
+                        null, moduleCache(), collectorResult.getExceptions(), collectorResult.getRoot(), nodes.size());
                 if (request.getRequestType() == DependencyResolverRequest.RequestType.FLATTEN) {
-                    DefaultDependencyResolverResult flattenResult = new DefaultDependencyResolverResult(
-                            null, null, collectorResult.getExceptions(), collectorResult.getRoot(), nodes.size());
                     for (Node node : nodes) {
-                        flattenResult.addNode(node);
+                        resolverResult.addNode(node);
                     }
-                    result = flattenResult;
+                    result = resolverResult;
                 } else {
-                    PathModularizationCache cache =
-                            new PathModularizationCache(); // TODO: should be project-wide cache.
-                    DefaultDependencyResolverResult resolverResult = new DefaultDependencyResolverResult(
-                            null, cache, collectorResult.getExceptions(), collectorResult.getRoot(), nodes.size());
                     ArtifactResolverResult artifactResolverResult =
                             session.getService(ArtifactResolver.class).resolve(session, coordinates, repositories);
                     for (Node node : nodes) {
@@ -224,6 +239,19 @@ public class DefaultDependencyResolver implements DependencyResolver {
         } finally {
             RequestTraceHelper.exit(trace);
         }
+    }
+
+    /**
+     * {@return the cache of information about the modules contained in a path element}.
+     *
+     * <p><b>TODO:</b> This method should not be in this class, because the cache should be global to the session.
+     * This method exists here only temporarily, until we clarified where to store session-wide caches.</p>
+     */
+    private PathModularizationCache moduleCache() {
+        if (moduleCache == null) {
+            moduleCache = new PathModularizationCache();
+        }
+        return moduleCache;
     }
 
     private static DependencyResolverException cannotReadModuleInfo(final Path path, final IOException cause) {
--- a/impl/maven-impl/src/main/java/org/apache/maven/impl/DefaultDependencyResolver.java
+++ b/impl/maven-impl/src/main/java/org/apache/maven/impl/DefaultDependencyResolver.java
@@ -70,19 +70,19 @@ public class DefaultDependencyResolver implements DependencyResolver {
 
     /**
      * Cache of information about the modules contained in a path element.
-     * This cache is created when first needed. It may be never created.
      *
      * <p><b>TODO:</b> This field should not be in this class, because the cache should be global to the session.
      * This field exists here only temporarily, until clarified where to store session-wide caches.</p>
-     *
-     * @see #moduleCache()
      */
-    private PathModularizationCache moduleCache;
+    private final PathModularizationCache moduleCache;
 
     /**
      * Creates an initially empty resolver.
      */
-    public DefaultDependencyResolver() {}
+    public DefaultDependencyResolver() {
+        // TODO: the cache should not be instantiated here, but should rather be session-wide.
+        moduleCache = new PathModularizationCache();
+    }
 
     @Nonnull
     @Override
@@ -143,7 +143,7 @@ public class DefaultDependencyResolver implements DependencyResolver {
                         session.getRepositorySystem().collectDependencies(systemSession, collectRequest);
                 return new DefaultDependencyResolverResult(
                         null,
-                        moduleCache(),
+                        moduleCache,
                         result.getExceptions(),
                         session.getNode(result.getRoot(), request.getVerbose()),
                         0);
@@ -191,8 +191,8 @@ public class DefaultDependencyResolver implements DependencyResolver {
         InternalSession session =
                 InternalSession.from(nonNull(request, "request").getSession());
         RequestTraceHelper.ResolverTrace trace = RequestTraceHelper.enter(session, request);
+        DependencyResolverResult result;
         try {
-            DependencyResolverResult result;
             DependencyResolverResult collectorResult = collect(request);
             List<RemoteRepository> repositories = request.getRepositories() != null
                     ? request.getRepositories()
@@ -212,12 +212,11 @@ public class DefaultDependencyResolver implements DependencyResolver {
                         .collect(Collectors.toList());
                 Predicate<PathType> filter = request.getPathTypeFilter();
                 DefaultDependencyResolverResult resolverResult = new DefaultDependencyResolverResult(
-                        null, moduleCache(), collectorResult.getExceptions(), collectorResult.getRoot(), nodes.size());
+                        null, moduleCache, collectorResult.getExceptions(), collectorResult.getRoot(), nodes.size());
                 if (request.getRequestType() == DependencyResolverRequest.RequestType.FLATTEN) {
                     for (Node node : nodes) {
                         resolverResult.addNode(node);
                     }
-                    result = resolverResult;
                 } else {
                     ArtifactResolverResult artifactResolverResult =
                             session.getService(ArtifactResolver.class).resolve(session, coordinates, repositories);
@@ -233,26 +232,13 @@ public class DefaultDependencyResolver implements DependencyResolver {
                             throw cannotReadModuleInfo(path, e);
                         }
                     }
-                    result = resolverResult;
                 }
+                result = resolverResult;
             }
-            return result;
         } finally {
             RequestTraceHelper.exit(trace);
         }
-    }
-
-    /**
-     * {@return the cache of information about the modules contained in a path element}.
-     *
-     * <p><b>TODO:</b> This method should not be in this class, because the cache should be global to the session.
-     * This method exists here only temporarily, until clarified where to store session-wide caches.</p>
-     */
-    private PathModularizationCache moduleCache() {
-        if (moduleCache == null) {
-            moduleCache = new PathModularizationCache();
-        }
-        return moduleCache;
+        return result;
     }
 
     private static DependencyResolverException cannotReadModuleInfo(final Path path, final IOException cause) {
--- a/impl/maven-impl/src/main/java/org/apache/maven/impl/DefaultDependencyResolverResult.java
+++ b/impl/maven-impl/src/main/java/org/apache/maven/impl/DefaultDependencyResolverResult.java
@@ -370,7 +370,7 @@ public class DefaultDependencyResolverResult implements DependencyResolverResult
 
     @Override
     public List<Exception> getExceptions() {
-        return exceptions;
+        return Collections.unmodifiableList(exceptions);
     }
 
     @Override
@@ -380,22 +380,22 @@ public class DefaultDependencyResolverResult implements DependencyResolverResult
 
     @Override
     public List<Node> getNodes() {
-        return nodes;
+        return Collections.unmodifiableList(nodes);
     }
 
     @Override
     public List<Path> getPaths() {
-        return paths;
+        return Collections.unmodifiableList(paths);
     }
 
     @Override
     public Map<PathType, List<Path>> getDispatchedPaths() {
-        return dispatchedPaths;
+        return Collections.unmodifiableMap(dispatchedPaths);
     }
 
     @Override
     public Map<Dependency, Path> getDependencies() {
-        return dependencies;
+        return Collections.unmodifiableMap(dependencies);
     }
 
     @Override
--- a/impl/maven-impl/src/main/java/org/apache/maven/impl/DefaultDependencyResolverResult.java
+++ b/impl/maven-impl/src/main/java/org/apache/maven/impl/DefaultDependencyResolverResult.java
@@ -27,6 +27,7 @@ import java.util.Collections;
 import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.Objects;
 import java.util.Optional;
 import java.util.Set;
 import java.util.function.Predicate;
@@ -56,6 +57,7 @@ public class DefaultDependencyResolverResult implements DependencyResolverResult
      * The corresponding request.
      */
     private final DependencyResolverRequest request;
+
     /**
      * The exceptions that occurred while building the dependency graph.
      */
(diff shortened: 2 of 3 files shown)


impl/maven-impl/src/main/java/org/apache/maven/impl/DefaultDependencyResolver.java
```diff
@@ -67,6 +67,21 @@
 @Named
 @Singleton
 public class DefaultDependencyResolver implements DependencyResolver {
+    /**
+     * Cache of information about the modules contained in a path element.
+     * This cache is created when first needed. It may be never created.
+     *
+     * <p><b>TODO:</b> This field should not be in this class, because the cache should be global to the session.
+     * This field exists here only temporarily, until we clarified where to store session-wide caches.</p>
```
Pankraz76, line 75:
```suggestion
     * This field exists here only temporarily, until clarified where to store session-wide caches.</p>
```

impl/maven-impl/src/main/java/org/apache/maven/impl/DefaultDependencyResolver.java
```diff
@@ -67,6 +67,21 @@
...
+     *
+     * @see moduleCache()
+     */
+    private PathModularizationCache moduleCache;
+
+    /**
+     * Creates an initially empty resolver.
+     */
+    public DefaultDependencyResolver() {}
```
Pankraz76, line 84: who need this? will create very fragile object.

impl/maven-impl/src/main/java/org/apache/maven/impl/DefaultDependencyResolverResult.java
```diff
@@ -97,6 +99,24 @@ public class DefaultDependencyResolverResult implements DependencyResolverResult
      */
     private final PathModularizationCache cache;
 
+    /**
+     * Creates an initially empty result with a temporary cache.
+     * Callers should add path elements by calls to {@link #addDependency(Node, Dependency, Predicate, Path)}.
+     *
+     * <p><b>WARNING: this constructor may be removed in a future Maven release.</b>
+     * The reason is because {@code DefaultDependencyResolverResult} needs a cache, which should
+     * preferably by session-wide. But we have not yet clarified how such caches should be managed.</p>
```
Pankraz76, line 108:
```suggestion
     * preferably by session-wide. How to manage such caches, has not yet been clarified.</p>
```

impl/maven-impl/src/main/java/org/apache/maven/impl/DefaultDependencyResolverResult.java
```diff
@@ -97,6 +99,24 @@ public class DefaultDependencyResolverResult implements DependencyResolverResult
...
+     *
+     * @param request the corresponding request
+     * @param exceptions the exceptions that occurred while building the dependency graph
+     * @param root the root node of the dependency graph
+     * @param count estimated number of dependencies
+     */
+    public DefaultDependencyResolverResult(
+            DependencyResolverRequest request, List<Exception> exceptions, Node root, int count) {
+        this(request, new PathModularizationCache(), exceptions, root, count);
```
Pankraz76, line 117: yes this gives dedication and option.

impl/maven-impl/src/main/java/org/apache/maven/impl/DefaultDependencyResolver.java
```diff
@@ -126,7 +141,11 @@ public DependencyResolverResult collect(@Nonnull DependencyResolverRequest reque
                 final CollectResult result =
                         session.getRepositorySystem().collectDependencies(systemSession, collectRequest);
                 return new DefaultDependencyResolverResult(
-                        null, null, result.getExceptions(), session.getNode(result.getRoot(), request.getVerbose()), 0);
+                        null,
+                        moduleCache(),
```
Pankraz76, line 145: as its dry, and static, inline. Ide shoud give warning on method or constructor that value is always same.

impl/maven-impl/src/main/java/org/apache/maven/impl/DefaultDependencyResolverResult.java
Pankraz76:
```suggestion
        this(request, moduleCache(), exceptions, root, count);
```

impl/maven-impl/src/main/java/org/apache/maven/impl/DefaultDependencyResolverResult.java
```diff
@@ -107,14 +127,14 @@ public class DefaultDependencyResolverResult implements DependencyResolverResult
...
-    public DefaultDependencyResolverResult(
+    DefaultDependencyResolverResult(
             DependencyResolverRequest request,
             PathModularizationCache cache,
             List<Exception> exceptions,
             Node root,
             int count) {
         this.request = request;
-        this.cache = cache;
+        this.cache = Objects.requireNonNull(cache);
```
Pankraz76, line 137: or here 
```suggestion
        this.cache = Objects.requireNonNull(moduleCache());
```
one of the suggestions is right. as `moduleCache()` is SSOT.

Pankraz76 (changes requested):
spelling, potential static inline.

impl/maven-impl/src/main/java/org/apache/maven/impl/DefaultDependencyResolver.java
desruisseaux: This constructor was already there before, only implicit (when a Java file does not declare any constructor at all). This addition only makes it explicit.

impl/maven-impl/src/main/java/org/apache/maven/impl/DefaultDependencyResolver.java
desruisseaux: Thanks, fixed.

impl/maven-impl/src/main/java/org/apache/maven/impl/DefaultDependencyResolverResult.java
desruisseaux: Thanks, fixed.

impl/maven-impl/src/main/java/org/apache/maven/impl/DefaultDependencyResolverResult.java
desruisseaux: The `moduleCache()` method is defined in another class. We cannot access that method from here.

impl/maven-impl/src/main/java/org/apache/maven/impl/DefaultDependencyResolver.java
desruisseaux: I'm not sure to understand. Inline the call to `moduleCache()`? This is uneasy, as that method does lazy instantiation.

impl/maven-impl/src/main/java/org/apache/maven/impl/DefaultDependencyResolverResult.java
desruisseaux: The `moduleCache()` method is defined in another class. We cannot access that method from here.

However, this is indeed the kind of code that we would have after we clarified how to manage session-wide caches. The `moduleCache()` method would be in the class managing those caches. We are just not there yet.

impl/maven-impl/src/main/java/org/apache/maven/impl/DefaultDependencyResolver.java
Pankraz76: yes right, thanks. But is in use? Had you compile issues, therefore created and sure about usage?
Might not needed before. Want to make sure.

Pankraz76: image

impl/maven-impl/src/main/java/org/apache/maven/impl/DefaultDependencyResolver.java
```diff
@@ -67,6 +67,21 @@
...
 @Singleton
 public class DefaultDependencyResolver implements DependencyResolver {
+    /**
+     * Cache of information about the modules contained in a path element.
+     * This cache is created when first needed. It may be never created.
+     *
+     * <p><b>TODO:</b> This field should not be in this class, because the cache should be global to the session.
+     * This field exists here only temporarily, until clarified where to store session-wide caches.</p>
+     *
+     * @see moduleCache()
```
Pankraz76, line 77:
```suggestion
     * @see #moduleCache()
```
image

cs should fail this build.

Pankraz76 (commented):
doc link.

impl/maven-impl/src/main/java/org/apache/maven/impl/DefaultDependencyResolver.java
```diff
@@ -67,6 +67,21 @@
 @Named
 @Singleton
 public class DefaultDependencyResolver implements DependencyResolver {
+    /**
```
Pankraz76, line 70:
```suggestion

    /**
```
cs can/must fail this.

impl/maven-impl/src/main/java/org/apache/maven/impl/DefaultDependencyResolver.java
```diff
@@ -191,18 +210,14 @@ public DependencyResolverResult resolve(DependencyResolverRequest request)
                         .map(Artifact::toCoordinates)
                         .collect(Collectors.toList());
                 Predicate<PathType> filter = request.getPathTypeFilter();
+                DefaultDependencyResolverResult resolverResult = new DefaultDependencyResolverResult(
+                        null, moduleCache(), collectorResult.getExceptions(), collectorResult.getRoot(), nodes.size());
                 if (request.getRequestType() == DependencyResolverRequest.RequestType.FLATTEN) {
-                    DefaultDependencyResolverResult flattenResult = new DefaultDependencyResolverResult(
```
Pankraz76, line 195: this scope was right. this field concern lives only here. Even could dedicate whole block.

impl/maven-impl/src/main/java/org/apache/maven/impl/DefaultDependencyResolver.java
desruisseaux: I think that it is instantiated via reflection, as the class is annotated with `@Named` and `@Singleton` from the `org.apache.maven.api.di` package.

impl/maven-impl/src/main/java/org/apache/maven/impl/DefaultDependencyResolver.java
desruisseaux: Side note: no need for CheckStyle, as `javac` can do this verification itself with the `-Xdoclint` option. It is not enabled in Maven maybe for historical reasons. Or maybe because we want to be able to use a `@see`  tag to another module without causing build failure. The latter is the kind of problem that "multi-modules project" resolves. But this is for another day, maybe far in the future.

impl/maven-impl/src/main/java/org/apache/maven/impl/DefaultDependencyResolver.java
desruisseaux: The change was for removing code duplication. The exact same code was inside the two branches of the `if … else` statement, with only different variable names.

impl/maven-impl/src/main/java/org/apache/maven/impl/DefaultDependencyResolver.java
desruisseaux: Javadoc link fixed.

impl/maven-impl/src/main/java/org/apache/maven/impl/DefaultDependencyResolver.java
desruisseaux: Done.

impl/maven-impl/src/main/java/org/apache/maven/impl/DefaultDependencyResolver.java
Pankraz76: then will see on IT test latest. or does clean install locally cover this?

impl/maven-impl/src/main/java/org/apache/maven/impl/DefaultDependencyResolver.java
Pankraz76: yes see. 

Common part can be extracted from 'if' 

its same obtain the warning:

image


```java
    @Override
    public DependencyResolverResult resolve(DependencyResolverRequest request)
            throws DependencyResolverException, ArtifactResolverException {
        InternalSession session =
                InternalSession.from(nonNull(request, "request").getSession());
        try {
            DependencyResolverResult collectorResult = collect(request);
            DependencyResolverResult result = collectorResult;
            List<RemoteRepository> repositories = request.getRepositories() != null
                    ? request.getRepositories()
                    : request.getProject().isPresent()
                    ? session.getService(ProjectManager.class)
                    .getRemoteProjectRepositories(
                            request.getProject().get())
                    : session.getRemoteRepositories();
            if (request.getRequestType()!= DependencyResolverRequest.RequestType.COLLECT) {
                List<Node> nodes = flatten(session, collectorResult.getRoot(), request.getPathScope());
                DefaultDependencyResolverResult resolverResult = new DefaultDependencyResolverResult(
                        null, moduleCache(), collectorResult.getExceptions(), collectorResult.getRoot(), nodes.size());
                if (request.getRequestType() == DependencyResolverRequest.RequestType.FLATTEN) {
                    for (Node node : nodes) {
                        resolverResult.addNode(node);
                    }
                } else {
                    for (Node node : nodes) {
                        Path path = (node.getArtifact() != null)
                                ? session.getService(ArtifactResolver.class).resolve(session, nodes.stream()
                                        .map(Node::getDependency)
                                        .filter(Objects::nonNull)
                                        .map(Artifact::toCoordinates)
                                        .collect(Collectors.toList()), repositories)
                                .getResult(node.getArtifact().toCoordinates())
                                .getPath()
                                : null;
                        try {
                            resolverResult.addDependency(node, node.getDependency(), request.getPathTypeFilter(), path);
                        } catch (IOException e) {
                            throw cannotReadModuleInfo(path, e);
                        }
                    }
                }
                result = resolverResult;
            }
            return result;
        } finally {
            RequestTraceHelper.exit(RequestTraceHelper.enter(session, request));
        }
    }
```
is the finally an issue inlining exit ?

impl/maven-impl/src/main/java/org/apache/maven/impl/DefaultDependencyResolver.java
```diff
@@ -68,6 +68,22 @@
...
+    /**
+     * Cache of information about the modules contained in a path element.
+     * This cache is created when first needed. It may be never created.
+     *
+     * <p><b>TODO:</b> This field should not be in this class, because the cache should be global to the session.
+     * This field exists here only temporarily, until clarified where to store session-wide caches.</p>
+     *
+     * @see #moduleCache()
+     */
+    private PathModularizationCache moduleCache;
```
Pankraz76, line 80: except of static this is the same outcome.
```suggestion
    private final static PathModularizationCache MODULE_CACHE = new PathModularizationCache();
```

impl/maven-impl/src/main/java/org/apache/maven/impl/DefaultDependencyResolver.java
Pankraz76: sorry wrong assumption. This is called every time as this the only return therefore default.

impl/maven-impl/src/main/java/org/apache/maven/impl/DefaultDependencyResolver.java
```diff
@@ -226,6 +242,19 @@ public DependencyResolverResult resolve(DependencyResolverRequest request)
...
+     * <p><b>TODO:</b> This method should not be in this class, because the cache should be global to the session.
+     * This method exists here only temporarily, until clarified where to store session-wide caches.</p>
+     */
+    private PathModularizationCache moduleCache() {
+        if (moduleCache == null) {
+            moduleCache = new PathModularizationCache();
+        }
+        return moduleCache;
+    }
+
```
Pankraz76, line 257:
```suggestion
```
making field final is the same result. static would change to all classes change same cache. If no collision risk then might be possible.

impl/maven-impl/src/main/java/org/apache/maven/impl/DefaultDependencyResolver.java
```diff
@@ -68,6 +68,22 @@
 @Singleton
 public class DefaultDependencyResolver implements DependencyResolver {
 
+    /**
+     * Cache of information about the modules contained in a path element.
+     * This cache is created when first needed. It may be never created.
```
Pankraz76, line 73:
```suggestion
```
imho no, return case will always call.

image

Pankraz76 (commented):
quest.

impl/maven-impl/src/main/java/org/apache/maven/impl/DefaultDependencyResolver.java
desruisseaux: The IT already passed with this change (explicit public constructor). To be sure, we could test with a breakpoint in the debugger. But I don't think that it is necessary. There is no other constructor and this class is instantiated somewhere for sure since `maven-compiler-plugin` uses it.

Pankraz76: im confused. this is new making explicit whats implicit. 
This is the first constructor introduced for no **compiler** reason given.

image

Pankraz76: > test

delete, rebuild, pass. 

It was done there before and it is without we having to make up whats already there.

desruisseaux: I'm not sure which change is requested here. Is it to remove the constructor? It was added only for making explicit what was implicit before. So this is mostly for documentation purpose.

# Demonstration
Save the following file somewhere:

```java
/**
 * A dummy comment.
 */
public class Test {
    /**
     * Another dummy comment.
     */
    public Test() {
    }
}
```

Compile as below with a recent Java version (tested with Java 24). The `-Xdoclint` option is for having warnings about the documentation. The `-g:none` option is for omitting debugging information, because the line numbers will change in this demo.

```bash
javac -g:none -Xdoclint:all Test.java
```

Compilation succeed with no warning. Save the output file:

```bash
mv Test.class Test.bak
```

Remove the constructor (including its comment) and recompile with the same command. Note that we now get the following warning:

```
Test.java:4: warning: use of default constructor, which does not provide a comment
public class Test {
       ^
1 warning
```

Compare the output files:

```
diff Test.bak Test.class
```

They are identical. Making the implicit constructor explicit changes nothing to the compiled code, but is nevertheless recommended according above warning.

Pankraz76: Yes provided sorry. Then lets merge and local build should fail when removed, right?
Then we simply know riving removal PR. it must fail build.

gnodet:
@desruisseaux for consistency and to honor the fact that the {{DepfaultDependencyResolverResult}} implements {{DependencyResolverResult}} which is annotated with {{@Immutable}}, I'd rather have {{DepfaultDependencyResolverResult}}  being immutable.  Is there a way to use a temporary intermediate class to store the computation data, and only keep the final computed state in this class ?

desruisseaux:
>  Is there a way to use a temporary intermediate class to store the computation data (…snip…)?

Yes, ideally the cache should be in some session-wide place. The only reason why the cache is declared in `DepfaultDependencyResolverResult` is because I do not know if there is a central place in Maven 4 where we can put stuff that we would like to cache for the duration of the session.

impl/maven-impl/src/main/java/org/apache/maven/impl/DefaultDependencyResolver.java
desruisseaux: > Then lets merge and local build should fail when removed, right?

We could make the build fail when the explicit constructor is absent by adding the `-Werror` compiler options. But Maven is not ready for that yet. For now, this change is just resolving one documentation warning among the many the Maven would have if the  `-Xdoclint:all` option was provided.

impl/maven-impl/src/main/java/org/apache/maven/impl/DefaultDependencyResolver.java
desruisseaux: > is the finally an issue inlining exit ?

Judging by the method names, it would probably not work. `RequestTraceHelper.enter` probably needs to be invoked before to start the execution of the code inside the `try` block. Inlining would cause `enter` to be invoked too late.

impl/maven-impl/src/main/java/org/apache/maven/impl/DefaultDependencyResolver.java
desruisseaux: A collision is possible if a JAR file changes between two executions. It may happen in particular if a JAR is the output of another project. It can also be a memory leak if a Maven process runs for a long time (the daemon?) and JAR paths change often (e.g. snapshot with timestamps). This is why the desired lifetime is a session (e.g. a call to `mvn compile`), with a fresh cache rebuilt every time that `mvn compile` is executed from the root of a multi-projects.

impl/maven-impl/src/main/java/org/apache/maven/impl/DefaultDependencyResolver.java
desruisseaux: > imho no, as always have return call, on default.

Right. Checking again, a see no code path where the cache would not be created. I will remove the lazy instantiation aspect.

impl/maven-impl/src/main/java/org/apache/maven/impl/DefaultDependencyResolver.java
Pankraz76: thx.

impl/maven-impl/src/main/java/org/apache/maven/impl/DefaultDependencyResolverResult.java
gnodet: Fwiw, session wide informations can be stored in `Session.getData()`.

jira-importer:
Resolve #9351

Pankraz76 (approved):

desruisseaux:
Since we had no more comment for more than a month, I presume that this pull request can be merged. I will process.

PR #2349, opened 2025-05-18
UnnecessaryParentheses

Pankraz76:
- https://pmd.github.io/pmd/pmd_rules_java_codestyle.html#uselessparentheses

- UselessParentheses (37)
- image

### DIFF
--- a/api/maven-api-core/src/main/java/org/apache/maven/api/JavaPathType.java
+++ b/api/maven-api-core/src/main/java/org/apache/maven/api/JavaPathType.java
@@ -263,7 +263,7 @@ public enum JavaPathType implements PathType {
         if (option == null) {
             throw new IllegalStateException("No option is associated to this path type.");
         }
-        String prefix = (moduleName == null) ? "" : (moduleName + '=');
+        String prefix = moduleName == null ? "" : moduleName + '=';
         StringJoiner joiner = new StringJoiner(File.pathSeparator, prefix, "");
         joiner.setEmptyValue("");
         for (Path p : paths) {
--- a/compat/maven-artifact/src/main/java/org/apache/maven/artifact/resolver/MultipleArtifactsNotFoundException.java
+++ b/compat/maven-artifact/src/main/java/org/apache/maven/artifact/resolver/MultipleArtifactsNotFoundException.java
@@ -93,7 +93,7 @@ public class MultipleArtifactsNotFoundException extends ArtifactResolutionExcept
         int counter = 0;
 
         for (Artifact artifact : artifacts) {
-            String message = (++counter) + ") " + artifact.getId();
+            String message = ++counter + ") " + artifact.getId();
 
             buffer.append(constructMissingArtifactMessage(
                     message,
--- a/compat/maven-artifact/src/main/java/org/apache/maven/artifact/versioning/ComparableVersion.java
+++ b/compat/maven-artifact/src/main/java/org/apache/maven/artifact/versioning/ComparableVersion.java
@@ -219,7 +219,7 @@ public class ComparableVersion implements Comparable<ComparableVersion> {
 
         @Override
         public int hashCode() {
-            return (int) (value ^ (value >>> 32));
+            return (int) (value ^ value >>> 32);
         }
 
         @Override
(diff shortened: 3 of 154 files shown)


hgschmie (changes requested):
Literally all of these changes make the code less readable.

compat/maven-artifact/src/main/java/org/apache/maven/artifact/resolver/MultipleArtifactsNotFoundException.java
```diff
@@ -93,7 +93,7 @@ private static String constructMessage(List<Artifact> artifacts) {
         int counter = 0;
 
         for (Artifact artifact : artifacts) {
-            String message = (++counter) + ") " + artifact.getId();
+            String message = ++counter + ") " + artifact.getId();
```
Pankraz76, line 96: **What's the benefit of having parentheses here?**  

For some mathematical terms, parentheses are important and have logical implications. Thus, it can be misleading if parentheses are included without being necessary for an actual logical operation. This makes understanding the code much harder, as one must decipher the deeper reasoning behind them—if there even is any.  

For example, in math:  
- `1 + 1` is equivalent to `(1) + (1)`  

While technically the same, the unnecessary parentheses add visual noise, requiring extra effort to interpret and distracting from the core logic. The key issue is that if parentheses are used inconsistently or without clear intent, it becomes difficult to distinguish between meaningful logic and arbitrary formatting.

compat/maven-artifact/src/main/java/org/apache/maven/artifact/versioning/ComparableVersion.java
```diff
@@ -118,7 +118,7 @@ public boolean isNull() {
         @Override
         public int compareTo(Item item) {
             if (item == null) {
-                return (value == 0) ? 0 : 1; // 1.0 == 1, 1.1 > 1
+                return value == 0 ? 0 : 1; // 1.0 == 1, 1.1 > 1
```
Pankraz76, line 121: **What's the benefit of having parentheses here?**  

For some mathematical terms, parentheses are important and have logical implications. Thus, it can be misleading if parentheses are included without being necessary for an actual logical operation. This makes understanding the code much harder, as one must decipher the deeper reasoning behind them—if there even is any.  

For example, in math:  
- `1 + 1` is equivalent to `(1) + (1)`  

While technically the same, the unnecessary parentheses add visual noise, requiring extra effort to interpret and distracting from the core logic. The key issue is that if parentheses are used inconsistently or without clear intent, it becomes difficult to distinguish between meaningful logic and arbitrary formatting.

api/maven-api-core/src/main/java/org/apache/maven/api/JavaPathType.java
```diff
@@ -263,7 +263,7 @@ final String[] format(String moduleName, Iterable<? extends Path> paths) {
         if (option == null) {
             throw new IllegalStateException("No option is associated to this path type.");
         }
-        String prefix = (moduleName == null) ? "" : (moduleName + '=');
```
Pankraz76, line 266: ### **Parentheses: Explicit Clarity vs. Unnecessary Overhead**  
Parentheses are useful for turning *implicit* assumptions into *explicit* guarantees. They enforce clarity, especially in complex logic, by making intent unambiguous. When the outcome is rigorously defined, reasoning about the code becomes easier.  

However, if the language’s behavior is already well-specified (e.g., operator precedence rules), redundant parentheses add *visual noise* and *cognitive overhead*. `Inconsistent` use—where some groupings are meaningful while others are arbitrary—creates a `misleading` pattern. Those extra characters (`(` and `)`) become syntactic bloat, making code harder to skim and implying intentional grouping where none exists.  

**Key Trade-off:**  
- **Explicit:** Essential for disambiguating non-obvious logic or overriding default precedence.  
- **Overhead:** Distracting when rules are universally known (e.g., `(a) + (b)` instead of `a + b`).  

The core challenge is balancing *intent* (highlighting critical logic) against *brevity* (avoiding clutter in trivial cases).

Pankraz76 (commented):
Inconsistent use—where some groupings are meaningful while others are arbitrary—creates a misleading pattern.

Pankraz76:
> Literally all of these changes make the code less readable.

deeply appreciate your feedback. Yes sometimes its better to be explicit. First priority on code is accessibility. Might be achievable with normalization.

Its possible to debate over this particular rule and break with convention, giving override for this rule in general, or only let existing code **horse around**.

slachiewicz:
Thank You. We are not going to continue with the proposal

Pankraz76:
- https://sonarsource.github.io/rspec/#/rspec/S1110
- https://sonarsource.github.io/rspec/#/rspec/S1611

yes of course, its just demo.

PR #2350, opened 2025-05-18
UnusedAssignment

Pankraz76:
- https://pmd.github.io/pmd/pmd_rules_java_bestpractices.html#unusedassignment

- UnusedAssignment
image

image

### DIFF
--- a/compat/maven-compat/src/main/java/org/apache/maven/repository/metadata/MetadataGraphEdge.java
+++ b/compat/maven-compat/src/main/java/org/apache/maven/repository/metadata/MetadataGraphEdge.java
@@ -30,9 +30,9 @@ import org.apache.maven.artifact.ArtifactScopeEnum;
 public class MetadataGraphEdge {
     String version;
     ArtifactScopeEnum scope;
-    int depth = -1;
-    int pomOrder = -1;
-    boolean resolved = true;
+    int depth;
+    int pomOrder;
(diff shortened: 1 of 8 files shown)


Pankraz76:
maybe flaky:
Caused by: org.apache.maven.api.services.ArtifactResolverException: The following artifacts could not be resolved: eu.maveniverse.maven.parent:parent:pom:28 (absent): Could not transfer artifact eu.maveniverse.maven.parent:parent:pom:28 from/to central (https://repo.maven.apache.org/maven2): HTTP connect timed out

compat/maven-compat/src/main/java/org/apache/maven/repository/metadata/MetadataGraphEdge.java
```diff
@@ -30,9 +30,9 @@
 public class MetadataGraphEdge {
     String version;
     ArtifactScopeEnum scope;
-    int depth = -1;
-    int pomOrder = -1;
-    boolean resolved = true;
+    int depth;
```
hgschmie, line 33: removing those is wrong. Currently, calling "getPomOrder()" in a subclass constructor (which is called from super()) will return -1. After this change, it will return 0. 

This is not just a simple cleanup, but changing the actual logic of the class.

hgschmie (changes requested):

PR #2312, opened 2025-05-09
[MNG-8729] Use correct `outputStream` destination; `request` instead of `path` in `DefaultPluginXmlFactory#write`

Pankraz76:
JIRA issue: [MNG-8729](https://issues.apache.org/jira/browse/MNG-8729)

follow up of:
- https://github.com/apache/maven/pull/2303

finally 100

image

PR #2357, opened 2025-05-20
Revert "[MNG-8729] Use correct `outputStream` destination; `request` instead of `path` in `DefaultPluginXmlFactory#write`"

Pankraz76:
Reverts apache/maven#2312

- No one notice this issue, being an indication, that this code is probably outdated as its not working. Assuming this situation can, or will, never happen this coding seems obsolete.
- Not important to prevent any issues, therefore embrace it.


jira-importer:
Resolve #9348

PR #2363, opened 2025-05-21
chore: Simplify immutable collections implementation by using standard Java APIs

gnodet:
- Replace custom List implementation with standard List.of() and List.copyOf()
- Remove unnecessary custom list implementations (List1, List2, ListN, etc.)
- Remove ROProperties implementation as it's no longer needed
- Add documentation note referencing JDK issue JDK-8323729
- Reduce code complexity and maintenance burden by leveraging JDK standard APIs

This change aligns with the goal of using standard Java APIs where possible
instead of maintaining custom implementations.


src/mdo/java/ImmutableCollections.java
```diff
@@ -79,32 +65,7 @@ public int size() {
...
-    }
-
-    @SuppressWarnings("unchecked")
-    static <E> List<E> emptyList() {
-        return (List<E>) EMPTY_LIST;
-    }
-
-    static <E> List<E> singletonList(E element) {
-        return new List1<>(element);
+        return collection == null ? List.of() : List.copyOf(collection);
```
Bukama, line 68: Why not `Collections.emptyList()`? Does the empty list needs to be mutable?

gnodet: List.of() also returns an immutable list (although not the same objectas Collections.emptyList()).  
It's a more concise form, and I think the best practice will be to use List.of() and Map.of() instead of the similar methods from Collections.

Bukama (approved):

src/mdo/java/ImmutableCollections.java
```diff
@@ -79,32 +65,7 @@ public int size() {
     };
 
     static <E> List<E> copy(Collection<E> collection) {
```
cstamas, line 67:
```suggestion
    static <E1, E2 extends E1> List<E1> copy(Collection<E2> collection) {
```

cstamas (approved):

src/mdo/java/ImmutableCollections.java
mbien: another advantage of using `Foo.of()` is that any defensive `Foo.copyOf()` downstream becomes a no-op.

https://github.com/openjdk/jdk/blob/a300c356555019a42c19bf0c16184f6dee4ad96e/src/java.base/share/classes/java/util/ImmutableCollections.java#L186-L187

pzygielo:
> one landmine to watch out for is that JDK's immutable collections can't contain null.

https://bugs.openjdk.org/browse/JDK-8265905

slachiewicz (approved):

PR #2370, opened 2025-05-21
fix: debounce `while(true)` in `CheckThreadSafetyMojo` PMD-Rule: `EmptyControlStatement`

Pankraz76:
- https://github.com/apache/maven/pull/2381

### DIFF
--- a/its/core-it-support/core-it-plugins/maven-it-plugin-active-collection/src/main/java/org/apache/maven/plugin/coreit/CheckThreadSafetyMojo.java
+++ b/its/core-it-support/core-it-plugins/maven-it-plugin-active-collection/src/main/java/org/apache/maven/plugin/coreit/CheckThreadSafetyMojo.java
@@ -95,7 +95,10 @@ public class CheckThreadSafetyMojo extends AbstractMojo {
                     getLog().info("[MAVEN-CORE-IT-LOG] Thread " + this + " uses " + tccl);
                     Thread.currentThread().setContextClassLoader(tccl);
                     while (go.isEmpty()) {
-                        // wait for start
+                        try {
+                            Thread.sleep(100); // wait for the start
+                        } catch (InterruptedException ignored) {
+                        }
                     }
                     for (int j = 0; j < 10 * 1000; j++) {
                         try {


its/core-it-support/core-it-plugins/maven-it-plugin-active-collection/src/main/java/org/apache/maven/plugin/coreit/CheckThreadSafetyMojo.java
```diff
@@ -95,7 +95,11 @@ public void run() {
                     getLog().info("[MAVEN-CORE-IT-LOG] Thread " + this + " uses " + tccl);
                     Thread.currentThread().setContextClassLoader(tccl);
                     while (go.isEmpty()) {
-                        // wait for start
```
Pankraz76, line 98: no need for debounce?

potentially `while(true)` cloud max out CPU.

desruisseaux: I have not checked the context. But maybe `java.util.concurrent.CountDownLatch` could be used instead of that loop.

its/core-it-support/core-it-plugins/maven-it-plugin-active-collection/src/main/java/org/apache/maven/plugin/coreit/CheckThreadSafetyMojo.java
```diff
@@ -95,7 +95,11 @@ public void run() {
...
+                        try {
+                            Thread.sleep(100); // wait for the start
+                        } catch (InterruptedException e) {
```
elharo, line 100: No, don't throw a raw RuntimeException. probably just continue here

elharo (changes requested):

its/core-it-support/core-it-plugins/maven-it-plugin-active-collection/src/main/java/org/apache/maven/plugin/coreit/CheckThreadSafetyMojo.java
Pankraz76: ide finding: `'continue' is unnecessary as the last statement in a loop `
image

its/core-it-support/core-it-plugins/maven-it-plugin-active-collection/src/main/java/org/apache/maven/plugin/coreit/CheckThreadSafetyMojo.java
Pankraz76: > CountDownLatch

thanks for ideation.

Pankraz76: https://github.com/apache/maven/pull/2381

its/core-it-support/core-it-plugins/maven-it-plugin-active-collection/src/main/java/org/apache/maven/plugin/coreit/CheckThreadSafetyMojo.java
```diff
@@ -95,7 +95,10 @@ public void run() {
                     getLog().info("[MAVEN-CORE-IT-LOG] Thread " + this + " uses " + tccl);
                     Thread.currentThread().setContextClassLoader(tccl);
                     while (go.isEmpty()) {
-                        // wait for start
+                        try {
+                            Thread.sleep(100); // wait for the start
+                        } catch (InterruptedException ignored) {
```
gnodet, line 100: Empty blocks are illegal wrt checkstyle.

Pankraz76: yes this rules must be absent or bug. As we have this several times i assume its misconfigured.

- https://checkstyle.sourceforge.io/checks/blocks/emptyblock.html
- https://github.com/apache/maven-shared-resources/blob/bad8df9df24313a66dd5968194b631305ac499dd/src/main/resources/config/maven_checks_nocodestyle.xml#L106
- so its a bug. Again one reason to chain the tools.

Pankraz76: > Empty blocks are illegal wrt checkstyle.

if this is true then this whole PR would not be possible as the root cause is https://pmd.github.io/pmd/pmd_rules_java_codestyle.html#emptycontrolstatement.




[INFO] BUILD SUCCESS

its/core-it-support/core-it-plugins/maven-it-plugin-active-collection/src/main/java/org/apache/maven/plugin/coreit/CheckThreadSafetyMojo.java
```diff
@@ -95,7 +95,10 @@ public void run() {
                     getLog().info("[MAVEN-CORE-IT-LOG] Thread " + this + " uses " + tccl);
                     Thread.currentThread().setContextClassLoader(tccl);
                     while (go.isEmpty()) {
-                        // wait for start
+                        try {
+                            Thread.sleep(100); // wait for the start
```
Pankraz76, line 99:
```suggestion
                            MILLISECONDS.sleep(100); // wait for the start
```
https://docs.oracle.com/javase/9/docs/api/java/util/concurrent/TimeUnit.html#sleep-long-

Pankraz76:
- https://github.com/apache/maven/pull/2381

its/core-it-support/core-it-plugins/maven-it-plugin-active-collection/src/main/java/org/apache/maven/plugin/coreit/CheckThreadSafetyMojo.java
adangel: Note: [EmptyControlStatement](https://pmd.github.io/pmd/pmd_rules_java_codestyle.html#emptycontrolstatement) and [EmptyCatchBlock](https://pmd.github.io/pmd/pmd_rules_java_errorprone.html#emptycatchblock) are two different rules. Also note, that by default, if the exception variable name is "ignored" or "expected", then the empty catch block won't be reported.

PR #2376, opened 2025-05-22
chore: `DefaultUpdateCheckManager` Fix variable 'lock' is never used (PMD: UnusedPrivateField)

Pankraz76:


compat/maven-compat/src/main/java/org/apache/maven/repository/legacy/DefaultUpdateCheckManager.java
```diff
@@ -50,8 +50,6 @@ public class DefaultUpdateCheckManager extends AbstractLogEnabled implements Upd
 
     private static final String ERROR_KEY_SUFFIX = ".error";
 
-    public DefaultUpdateCheckManager() {}
```
elharo, line 53: don't change the API

compat/maven-compat/src/main/java/org/apache/maven/repository/legacy/DefaultUpdateCheckManager.java
```diff
@@ -303,7 +301,7 @@ private Properties read(File touchfile) {
                 Properties props = new Properties();
 
                 try (FileInputStream in = new FileInputStream(touchfile)) {
-                    try (FileLock lock = in.getChannel().lock(0, Long.MAX_VALUE, true)) {
+                    try (FileLock ignored = in.getChannel().lock(0, Long.MAX_VALUE, true)) {
```
elharo, line 304: It's not ignored. Is PMD simply wrong here?

impl/maven-core/src/main/java/org/apache/maven/lifecycle/internal/MojoExecutor.java
```diff
@@ -306,7 +306,7 @@ private void doExecute(MavenSession session, MojoExecution mojoExecution, Depend
 
         ensureDependenciesAreResolved(mojoDescriptor, session, dependencyContext);
 
-        try (NoExceptionCloseable lock = getProjectLock(session, mojoDescriptor)) {
+        try (NoExceptionCloseable ignore = getProjectLock(session, mojoDescriptor)) {
```
elharo, line 309: Yes, looks like a bug in PMD

elharo (changes requested):

compat/maven-compat/src/main/java/org/apache/maven/repository/legacy/DefaultUpdateCheckManager.java
desruisseaux: Another reason for keeping this constructor is that this class is annotated with `@Named` and `@Singleton`. So there is chances that this constructor is invoked by reflection, even if the IDE reports no usage of this constructor.

Pankraz76: yes, we apply use (test) or loose.

compat/maven-compat/src/main/java/org/apache/maven/repository/legacy/DefaultUpdateCheckManager.java
```diff
@@ -303,7 +301,7 @@ private Properties read(File touchfile) {
                 Properties props = new Properties();
 
                 try (FileInputStream in = new FileInputStream(touchfile)) {
-                    try (FileLock lock = in.getChannel().lock(0, Long.MAX_VALUE, true)) {
```
Pankraz76, line 306: the block is super small its **defacto** unused or im not Vincent Potucek.

image

compat/maven-compat/src/main/java/org/apache/maven/repository/legacy/DefaultUpdateCheckManager.java
Pankraz76: really? check out the usage. Its **unused for sure**, or the last 15 years i have done everything wrong, not even understood basics in pointer and reference.

https://www.geeksforgeeks.org/reference-variable-in-java/

Pankraz76 (commented):
the block is super small, its **defacto** unused or im not Vincent Potucek.

compat/maven-compat/src/main/java/org/apache/maven/repository/legacy/DefaultUpdateCheckManager.java
Pankraz76: extract thin into dedicated pr just wanted to give attention.

impl/maven-core/src/main/java/org/apache/maven/lifecycle/internal/MojoExecutor.java
Pankraz76: really? Show you IDE screen not point out the same:

Looks like a bug in maven to me.

image

compat/maven-compat/src/main/java/org/apache/maven/repository/legacy/DefaultUpdateCheckManager.java
Pankraz76: its is an error indication, therefore should be named according to convention.

desruisseaux:
It is used implicitly in a implicit `finally` block, even if we don't see it in the code.

impl/maven-core/src/main/java/org/apache/maven/lifecycle/internal/MojoExecutor.java
desruisseaux: It is used implicitly by the "try with resource" constructor. The `lock` name is fine, as it is a meaningful name if a future version decides to use it explicitly.

Pankraz76: This code makes no sense then. Why declare a field not used?

Then simply remove the overhead not needed:

image

Pankraz76: set dedicate this:
- https://github.com/apache/maven/pull/2384

Pankraz76:
idk this is not how java works. The scope of the field `lock` is limited to the block `{}`. not reachable in `final` or `catch`

image

its 2 different kind: just see highligh:

image

Pankraz76:
> It is used implicitly in a implicit `finally` block, even if we don't see it in the code.

yes, then it should be named accordingly.

Pankraz76:
dont we have test for this lock? afaik the lock only gets closed but not like open or checked if its ready to go. So calling this should do the trick as well:
`getProjectLock(session, mojoDescriptor).lock();`

- https://github.com/apache/maven/pull/2384/files#r2104747698
- https://github.com/apache/maven/pull/2384

compat/maven-compat/src/main/java/org/apache/maven/repository/legacy/DefaultUpdateCheckManager.java
Pankraz76: @desruisseaux point out correctly that its used `indirectly`, having the implizit `close()` call.

As IDEA and PMD both point this out is seems ok to rename to give proper context. Its ignored as handled by default.

Other way should be an inline comment for suppression.

elharo (changes requested):

compat/maven-compat/src/main/java/org/apache/maven/repository/legacy/DefaultUpdateCheckManager.java
Pankraz76: indeed. This is the only use, which is an exception of the normal pattern (using is explizit).

To make this clear the convention seems naming it `ignroed`, at least considering IDEA and PMD, both recognize the variable as marked explizit, as only used implicit.

Now using the resource which the statement is all about, seems like an error, thats why its pointes out by these tools.
Im altho wondering why its not used, allowing a comment to explain too.
Then the comment end up in the code again, to be self explaining.

PR #2404, opened 2025-05-28
[MNG-8746] Preserve property insertion order in WrapperProperties

gnodet:
## Description

This PR fixes MNG-8746 by implementing property insertion order preservation in Maven's WrapperProperties class. The issue was that properties accessed through `model.getProperties()` were not maintaining their insertion order, leading to inconsistent behavior.

## Problem

The WrapperProperties class in the Maven compat layer was not preserving the insertion order of properties. When properties were added through the Properties API and then accessed via methods like `keySet()` or `entrySet()`, they would be returned in an arbitrary order (often alphabetical) rather than the order they were inserted.

## Solution

Implemented a comprehensive order preservation mechanism:

### Key Changes:

1. **Enhanced WrapperProperties Template** (`src/mdo/java/WrapperProperties.java`):
   - Added internal `OrderedProperties` class that maintains insertion order
   - Implemented caching mechanism to preserve order across read/write operations
   - Modified all read operations to use the ordered cache
   - Simplified write operations to work directly with the ordered properties

2. **OrderedProperties Implementation**:
   - Uses `ArrayList<Object> keyOrder` to track key insertion order
   - Overrides `put()`, `setProperty()`, and `remove()` methods to maintain order
   - Custom `KeySet` and `EntrySet` implementations that iterate in insertion order
   - Removed unnecessary synchronized wrappers that interfered with ordering

3. **Architecture Simplification**:
   - WrapperProperties maintains its own ordered state independently
   - Only syncs content (not order) with the v4 API model when needed
   - Eliminated complex cache invalidation logic that was causing order loss

## Testing

Added comprehensive tests:
- `PropertiesOrderTest`: Tests v4 API model order preservation
- `WrapperPropertiesOrderTest`: Tests compat layer order preservation
- Covers various scenarios including property modification and initial properties

## Verification

The implementation ensures that:
```java
Properties props = model.getProperties();
props.setProperty("key3", "value3");
props.setProperty("key1", "value1");
props.setProperty("key2", "value2");

// keySet() now returns ["key3", "key1", "key2"] in insertion order
// instead of ["key1", "key2", "key3"] in alphabetical order
```

## Impact

- ✅ Preserves property insertion order for consistent behavior
- ✅ Maintains backward compatibility
- ✅ No breaking changes to existing APIs
- ✅ Improves reproducible builds

Fixes: https://issues.apache.org/jira/browse/MNG-8746

---

### DIFF
--- a/src/mdo/java/WrapperProperties.java
+++ b/src/mdo/java/WrapperProperties.java
@@ -51,7 +51,7 @@ class WrapperProperties extends Properties {
     final Supplier<Map<String, String>> getter;
     final Consumer<Properties> setter;
     private final OrderedProperties orderedProps = new OrderedProperties();
-    private boolean initialized = false;
+    private boolean initialized;
 
     WrapperProperties(Supplier<Map<String, String>> getter, Consumer<Properties> setter) {
         this.getter = getter;
--- a/src/mdo/java/WrapperProperties.java
+++ b/src/mdo/java/WrapperProperties.java
@@ -60,8 +60,7 @@ class WrapperProperties extends Properties {
 
     private synchronized void ensureInitialized() {
         if (!initialized) {
-            Map<String, String> initial = getter.get();
-            orderedProps.putAll(initial);
+            orderedProps.putAll(getter.get());
             initialized = true;
         }
     }
--- a/src/mdo/java/WrapperProperties.java
+++ b/src/mdo/java/WrapperProperties.java
@@ -27,14 +27,19 @@ import java.io.PrintWriter;
 import java.io.Reader;
 import java.io.Writer;
 import java.util.AbstractSet;
+import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.Enumeration;
 import java.util.InvalidPropertiesFormatException;
 import java.util.Iterator;
+import java.util.LinkedHashSet;
+import java.util.List;
 import java.util.Map;
+import java.util.Objects;
 import java.util.Properties;
 import java.util.Set;
+import java.util.concurrent.CopyOnWriteArrayList;
 import java.util.function.BiConsumer;
 import java.util.function.BiFunction;
 import java.util.function.Consumer;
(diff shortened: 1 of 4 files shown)


cstamas (approved):

src/mdo/java/WrapperProperties.java
```diff
@@ -377,4 +369,131 @@ private Object writeReplace() throws java.io.ObjectStreamException {
...
+                };
+            }
+
+            @Override
+            public int size() {
+                return keyOrder.size();
+            }
+        }
+
+        private class KeySet extends AbstractSet<Object> {
```
Pankraz76, line 457: this can stay internal.
```suggestion
        private static class KeySet extends AbstractSet<Object> {
```
will work too, right?

src/mdo/java/WrapperProperties.java
```diff
@@ -377,4 +369,131 @@ private Object writeReplace() throws java.io.ObjectStreamException {
...
+            keyOrder.remove(key);
+            return super.remove(key);
+        }
+
+        @Override
+        public synchronized void forEach(BiConsumer<? super Object, ? super Object> action) {
+            entrySet().forEach(e -> action.accept(e.getKey(), e.getValue()));
+        }
+
+        private class EntrySet extends AbstractSet<Map.Entry<Object, Object>> {
```
Pankraz76, line 418: this can stay internal.

src/mdo/java/WrapperProperties.java
```diff
@@ -377,4 +369,131 @@ private Object writeReplace() throws java.io.ObjectStreamException {
         props.putAll(getter.get());
         return props;
     }
+
+    private class OrderedProperties extends Properties {
```
Pankraz76, line 373: might plz consider give dedicated file, as this is quite packed already.

src/mdo/java/WrapperProperties.java
```diff
@@ -45,30 +50,44 @@ class WrapperProperties extends Properties {
...
+            Map<String, String> initial = getter.get();
+            orderedProps.putAll(initial);
+            initialized = true;
+        }
+    }
+
     @Override
     public String getProperty(String key) {
-        return getter.get().get(key);
+        ensureInitialized();
```
Pankraz76, line 71: why not call once in constructor, post construct, as we mimic CDI?
this seems redundant.

- https://stackoverflow.com/questions/8519187/spring-postconstruct-vs-init-method-attribute
- https://docs.oracle.com/javase/8/docs/api/javax/annotation/PostConstruct.html

Pankraz76 (changes requested):
can we use PostConstruct ?

src/mdo/java/WrapperProperties.java
```diff
@@ -45,30 +50,44 @@ class WrapperProperties extends Properties {
...
 
     WrapperProperties(Supplier<Map<String, String>> getter, Consumer<Properties> setter) {
         this.getter = getter;
         this.setter = setter;
     }
 
+    private synchronized void ensureInitialized() {
+        if (!initialized) {
+            Map<String, String> initial = getter.get();
+            orderedProps.putAll(initial);
```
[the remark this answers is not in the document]
Pankraz76: - https://github.com/pmd/pmd/issues/5771
- https://github.com/pmd/pmd/issues/5770

src/mdo/java/WrapperProperties.java
Pankraz76: ok sorry, i always aim to make static final.

PR #2411, opened 2025-05-30
[MNG-8759] Restore toString method in DefaultJavaToolchain

slawekjaranowski:
https://issues.apache.org/jira/browse/MNG-8759


jira-importer:
Resolve #9531

PR #2414, opened 2025-05-31
Enable Github Issues (3.9.x)

Bukama:
PR automation already there

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -9,7 +9,7 @@ contribution quickly and easily:
   This may not always be possible but is a best-practice.
 - [ ] Run `mvn verify` to make sure basic checks pass.
   A more thorough check will be performed on your pull request automatically.
-- [ ] You have run the integration tests successfully (`mvn -Prun-its verify`).
+- [ ] You have run the [Core IT][core-its] successfully.
 
 If your pull request is about ~20 lines of code you don't need to sign an
 [Individual Contributor License Agreement](https://www.apache.org/licenses/icla.pdf) if you are unsure
@@ -21,3 +21,5 @@ you have to acknowledge this by using the following check-box.
 
 - [ ] I hereby declare this contribution to be licenced under the [Apache License Version 2.0, January 2004](http://www.apache.org/licenses/LICENSE-2.0)
 - [ ] In any other case, please file an [Apache Individual Contributor License Agreement](https://www.apache.org/licenses/icla.pdf).
+
+[core-its]: https://maven.apache.org/core-its/core-it-suite/


.github/pull_request_template.md
```diff
@@ -1,18 +1,15 @@
...
- - [ ] Run `mvn clean verify` to make sure basic checks pass. A more thorough check will
-       be performed on your pull request automatically.
- - [ ] You have run the [Core IT][core-its] successfully.
+- [ ] Your pull request should address just one issue, without pulling in other changes.
+- [ ] Write a pull request description that is detailed enough to understand what the pull request does, how, and why.
+- [ ] Each commit in the pull request should have a meaningful subject line and body.
+  Note that commits might be squashed by a maintainer on merge.
+- [ ] Write unit tests that match behavioral changes, where the tests fail if the changes to the runtime are not applied.
+  This may not always be possible but is a best-practice.
+- [ ] Run `mvn verify` to make sure basic checks pass.
```
Pankraz76, line 10:
```suggestion
- [ ] Run `mvn verify` to pass basic checks.
```

.github/pull_request_template.md
```diff
@@ -1,18 +1,15 @@
...
+  A more thorough check will be performed on your pull request automatically.
+- [ ] You have run the integration tests successfully (`mvn -Prun-its verify`).
```
Pankraz76, line 12:
```suggestion
- [ ] You have run the integration tests successfully (`mvn verify -Prun-its`).
```
goal (more) important, profile is context.

Pankraz76: BTW: they dont run on my system.
just like main build fails mostly with 3 test reaching 15 sec. timeout.

Pankraz76 (changes requested):

.github/pull_request_template.md
Bukama: Two thoughts here: To my English knowledge (I'm not native speaker) the meaning of your suggestion is another as the original. Do you agree?

As were are thinking about changing the PR template and the way it is applied in https://github.com/apache/maven-gh-actions-shared/issues/165 and just align the templates now while activating GHissues I think I'll leave it as it is for now.

.github/pull_request_template.md
Bukama: As stated in the main PR the profile does not exists, so it fails

slawekjaranowski (approved):
ok, we can start I'm not sure if PR template will be respected per branch

olamy:
> ok, we can start I'm not sure if PR template will be respected per branch

I think only template from default branch is used. (when you create an issue is not against a branch ;) )

PR #2395, opened 2025-05-27
chore: singularize `AbstractMojo#createFile(File)`

Pankraz76:
- https://github.com/apache/maven/pull/2396
- https://pmd.github.io/pmd/pmd_rules_java_bestpractices.html#usetrywithresources

22 times the same code:

image

PR #2417, opened 2025-06-01
centralize domain logic `ArtifactRepositoryPolicy#getEffectivePolicy`

Pankraz76:
- https://github.com/apache/maven/pull/2395/files#r2117679994

what do you think about this? @elharo

### DIFF
--- a/compat/maven-artifact/src/main/java/org/apache/maven/artifact/repository/ArtifactRepositoryPolicy.java
+++ b/compat/maven-artifact/src/main/java/org/apache/maven/artifact/repository/ArtifactRepositoryPolicy.java
@@ -19,6 +19,7 @@
 package org.apache.maven.artifact.repository;
 
 import java.util.Calendar;
+import java.util.Collection;
 import java.util.Date;
 
 /**
@@ -143,7 +144,7 @@ public class ArtifactRepositoryPolicy {
         return buffer.toString();
     }
 
-    public void merge(ArtifactRepositoryPolicy policy) {
+    public ArtifactRepositoryPolicy merge(ArtifactRepositoryPolicy policy) {
         if (policy != null && policy.isEnabled()) {
             setEnabled(true);
 
@@ -155,6 +156,14 @@ public class ArtifactRepositoryPolicy {
                 setUpdatePolicy(policy.getUpdatePolicy());
             }
         }
+        return policy;
+    }
+
+    public static ArtifactRepositoryPolicy getEffectivePolicy(Collection<ArtifactRepositoryPolicy> policies) {
+        return policies.stream()
+                .map(ArtifactRepositoryPolicy::new)
+                .reduce(ArtifactRepositoryPolicy::merge)
+                .orElse(null);
     }
 
     private int ordinalOfCksumPolicy(String policy) {
--- a/compat/maven-compat/src/main/java/org/apache/maven/repository/legacy/LegacyRepositorySystem.java
+++ b/compat/maven-compat/src/main/java/org/apache/maven/repository/legacy/LegacyRepositorySystem.java
@@ -25,7 +25,6 @@ import javax.inject.Singleton;
 import java.io.File;
 import java.io.IOException;
 import java.util.ArrayList;
-import java.util.Collection;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.LinkedHashMap;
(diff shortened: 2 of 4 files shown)


compat/maven-compat/src/main/java/org/apache/maven/repository/legacy/LegacyRepositorySystem.java
```diff
@@ -387,20 +388,6 @@ public List<ArtifactRepository> getEffectiveRepositories(List<ArtifactRepository
         return effectiveRepositories;
     }
 
-    private ArtifactRepositoryPolicy getEffectivePolicy(Collection<ArtifactRepositoryPolicy> policies) {
-        ArtifactRepositoryPolicy effectivePolicy = null;
-
-        for (ArtifactRepositoryPolicy policy : policies) {
```
Pankraz76, line 393: dry.

impl/maven-core/src/main/java/org/apache/maven/bridge/MavenRepositorySystem.java
```diff
@@ -656,20 +657,6 @@ public List<ArtifactRepository> getEffectiveRepositories(List<ArtifactRepository
         return effectiveRepositories;
     }
 
-    private ArtifactRepositoryPolicy getEffectivePolicy(Collection<ArtifactRepositoryPolicy> policies) {
-        ArtifactRepositoryPolicy effectivePolicy = null;
-
-        for (ArtifactRepositoryPolicy policy : policies) {
```
Pankraz76, line 662: dry.

compat/maven-artifact/src/main/java/org/apache/maven/artifact/repository/ArtifactRepositoryPolicy.java
```diff
@@ -155,6 +157,25 @@ public void merge(ArtifactRepositoryPolicy policy) {
                 setUpdatePolicy(policy.getUpdatePolicy());
             }
         }
+        return policy;
+    }
+
+    public static ArtifactRepositoryPolicy getEffectivePolicy(Collection<ArtifactRepositoryPolicy> policies) {
```
Pankraz76, line 162: cohesion.

compat/maven-artifact/src/main/java/org/apache/maven/artifact/repository/ArtifactRepositoryPolicy.java
```diff
@@ -155,6 +157,25 @@ public void merge(ArtifactRepositoryPolicy policy) {
                 setUpdatePolicy(policy.getUpdatePolicy());
             }
         }
+        return policy;
```
Pankraz76, line 159: https://de.wikipedia.org/wiki/Fluent_Interface

compat/maven-compat/src/main/java/org/apache/maven/repository/legacy/LegacyRepositorySystem.java
```diff
@@ -358,20 +359,20 @@ public List<ArtifactRepository> getEffectiveRepositories(List<ArtifactRepository
                 mirroredRepos.addAll(aliasedRepo.getMirroredRepositories());
             }
 
-            ArtifactRepositoryPolicy releasePolicy = getEffectivePolicy(releasePolicies);
```
Pankraz76, line 361: coupling.
- https://checkstyle.sourceforge.io/checks/coding/variabledeclarationusagedistance.html
- https://github.com/pmd/pmd/issues/5771
- https://en.wikipedia.org/wiki/Coupling_(computer_programming)

Pankraz76:
thx. So no changes at all in this module? Then it could be ignored from git, right?

gnodet:
> thx. So no changes at all in this module? Then it could be ignored from git, right?

I don't think we can just ignore them, as they are part of the build and are still needed for Maven 3 compatibility, and even still used, but the plan is to not use them anymore in the coming months and get rid of them in Maven 5.

elharo:
They absolutely cannot and should not be ignored in git. .gitignore is for files that are not committed to the repo, exactly the opposite of these files.

Pankraz76:
yes, it would be nifty workaround, aka hack. 

If only a freeze is the goal then `.gignore`, even if treaded unintentionally, serves dedicated purpose without any further.

Same counts for potential PMD analysis, which needed to be excluded from this code, as frozen. 

when needed again, then simple just remove `temp-ignore`.

PR #2426, opened 2025-06-03
use `try-with-resources` statement in `LookupWagonMojo`

Pankraz76:
- https://github.com/apache/maven/pull/2424

ref 
- https://github.com/apache/maven/pull/2396#issuecomment-2936965168

### DIFF
--- a/its/core-it-support/core-it-plugins/maven-it-plugin-uses-wagon/src/main/java/org/apache/maven/plugin/coreit/LookupWagonMojo.java
+++ b/its/core-it-support/core-it-plugins/maven-it-plugin-uses-wagon/src/main/java/org/apache/maven/plugin/coreit/LookupWagonMojo.java
@@ -92,21 +92,11 @@ public class LookupWagonMojo extends AbstractMojo {
 
         getLog().info("[MAVEN-CORE-IT-LOG] Creating output file " + outputFile);
 
-        OutputStream out = null;
-        try {
-            outputFile.getParentFile().mkdirs();
-            out = new FileOutputStream(outputFile);
+        outputFile.getParentFile().mkdirs();
+        try (OutputStream out = new FileOutputStream(outputFile)) {
             loaderProperties.store(out, "MAVEN-CORE-IT-LOG");
         } catch (IOException e) {
-            throw new MojoExecutionException("Output file could not be created: " + outputFile, e);
-        } finally {
-            if (out != null) {
-                try {
-                    out.close();
-                } catch (IOException e) {
-                    // just ignore
-                }
-            }
+            throw new MojoExecutionException(e);
         }
 
         getLog().info("[MAVEN-CORE-IT-LOG] Created output file " + outputFile);


its/core-it-support/core-it-plugins/maven-it-plugin-uses-wagon/src/main/java/org/apache/maven/plugin/coreit/LookupWagonMojo.java
```diff
@@ -92,21 +92,11 @@ public void execute() throws MojoExecutionException, MojoFailureException {
 
         getLog().info("[MAVEN-CORE-IT-LOG] Creating output file " + outputFile);
 
-        OutputStream out = null;
-        try {
-            outputFile.getParentFile().mkdirs();
```
Pankraz76, line 97: without this its not possible.  This confuses the code as has nothing to to with the try catch.

When looking into this argument relativized as used in fluid API melding down boilerplate even more

- https://github.com/apache/maven/pull/2395

10 LOC bare boilerplate less.

lees seems more. 

stay lean to get clean.

@elharo @gnodet

its/core-it-support/core-it-plugins/maven-it-plugin-uses-wagon/src/main/java/org/apache/maven/plugin/coreit/LookupWagonMojo.java
```diff
@@ -92,21 +92,11 @@ public void execute() throws MojoExecutionException, MojoFailureException {
...
-            out = new FileOutputStream(outputFile);
+        outputFile.getParentFile().mkdirs();
+        try (OutputStream out = new FileOutputStream(outputFile)) {
```
Pankraz76, line 96:
```suggestion
        try (OutputStream out = new FileOutputStream(mkDir(outputFile))) {
```
fluid api & SOC

its/core-it-support/core-it-plugins/maven-it-plugin-uses-wagon/src/main/java/org/apache/maven/plugin/coreit/LookupWagonMojo.java
```diff
@@ -92,21 +92,11 @@ public void execute() throws MojoExecutionException, MojoFailureException {
...
             loaderProperties.store(out, "MAVEN-CORE-IT-LOG");
         } catch (IOException e) {
             throw new MojoExecutionException("Output file could not be created: " + outputFile, e);
```
elharo: not necessarily true

Pankraz76:
```suggestion
            throw new MojoExecutionException(e);
```
same same. 
consider this less beneficial, as the real cause matters.

- https://github.com/apache/maven/pull/2427#discussion_r2126362897

elharo (approved):

PR #2425, opened 2025-06-03
[MNG-8764] Sort injected lists by @Priority annotation

gnodet:
When injecting List<T> dependencies, the DI container now sorts the
bindings by their @Priority annotation value in descending order
(highest priority first) to ensure deterministic ordering.

This change ensures that components with higher priority values
appear first in injected lists, providing predictable behavior
for dependency injection scenarios where order matters.

- Modified InjectorImpl.doGetCompiledBinding() to sort bindings
  by priority before creating supplier lists
- Added comprehensive tests for priority-based list ordering
- Includes tests for mixed priorities and default priority handling

JIRA issue: [MNG-8764](https://issues.apache.org/jira/browse/MNG-8764)

PR #2428, opened 2025-06-04
evolve Binding to Comparable following solid design patterns avoiding redundancies leveraging OOP Fix sorted bindings

Pankraz76:
- https://github.com/apache/maven/pull/2425
- https://github.com/apache/maven/pull/2425#issuecomment-2938746732


impl/maven-di/src/main/java/org/apache/maven/di/impl/InjectorImpl.java
```diff
@@ -426,29 +425,14 @@ private void removeFromResolutionStack(Key<?> key) {
...
-                                synchronized (this) {
-                                    if (instance == null) {
-                                        instance = unscoped.get();
-                                    }
-                                }
-                            }
-                            return instance;
-                        }
-                    });
+            return (java.util.function.Supplier<T>) cache.computeIfAbsent(key, k -> unscoped);
```
rmannibucau: why making a singleton prototype?

impl/maven-core/src/main/java/org/apache/maven/internal/impl/SisuDiBridgeModule.java
```diff
@@ -176,17 +175,14 @@ public <Q> Supplier<Q> getCompiledBinding(Dependency<Q> dep) {
...
-            // Add Plexus bindings
             for (var bean : locator.get().locate(toGuiceKey(key))) {
                 if (isPlexusBean(bean)) {
                     list.add(new BindingToBeanEntry<>(key).toBeanEntry(bean).prioritize(bean.getRank()));
                 }
             }
             if (!list.isEmpty()) {
-                list.sort(getBindingComparator());
-                //noinspection unchecked
+                list.sort(null); // Uses Binding's compareTo
```
Pankraz76: reverse

impl/maven-core/src/main/java/org/apache/maven/internal/impl/SisuDiBridgeModule.java
```diff
@@ -24,7 +24,6 @@
 import java.lang.annotation.Annotation;
```
rmannibucau, line 24: is there any benefit in this change except adding mor responsability to binding than it should have? (btw OOP is outdated these days we tend to move to something closer to functional since it evolves smoother in practise, exactly as your PR shows)

a good example is that if we get interceptors for ex, sorting will be reversed so your comparable impl will not work and will need to be reverted to this code or will make the code no more even

impl/maven-di/src/main/java/org/apache/maven/di/impl/Binding.java
```diff
@@ -128,14 +128,19 @@ public Annotation getScope() {
...
         return dependencies.stream().map(Dependency::getDisplayString).collect(joining(", ", "[", "]"));
     }
 
     public Key<?> getOriginalKey() {
         return originalKey;
     }
 
+    @Override
+    public int compareTo(Binding<?> objectBinding) {
+        return Integer.compare(this.priority, objectBinding.priority);
```
gnodet: Reverse the logic, bindings are always sorted in descending order to get the ones with highest priority first.

impl/maven-core/src/main/java/org/apache/maven/internal/impl/SisuDiBridgeModule.java
gnodet: Yes, that's questionable.  A middle ground would be have a public static method in `Binding` returning the reversed priority comparator.

impl/maven-core/src/main/java/org/apache/maven/internal/impl/SisuDiBridgeModule.java
```diff
@@ -176,17 +175,14 @@ public <Q> Supplier<Q> getCompiledBinding(Dependency<Q> dep) {
 
         private <Q> Supplier<Q> getBeanSupplier(Dependency<Q> dep, Key<Q> key) {
             List<Binding<?>> list = new ArrayList<>();
-            // Add DI bindings
```
gnodet: I don't think comments in code are completely useless.

impl/maven-core/src/main/java/org/apache/maven/internal/impl/SisuDiBridgeModule.java
```diff
@@ -253,11 +244,6 @@ private <Q> Q getInstance(Binding<Q> binding) {
             return compile(binding).get();
         }
 
-        private static Comparator<Binding<?>> getBindingComparator() {
```
gnodet, line 256: As suggested above, a good alternative could be to move this method to `Binding` to make it reusable.

impl/maven-di/src/main/java/org/apache/maven/di/impl/InjectorImpl.java
gnodet: Wow, don't change the logic here !  You're just completely breaking the _singleton scope_ which would not memoize the values anymore.   Your proposal just returns the original `unscoped` supplier, so `unscoped.get()` will be called every time the returned supplier is invoked, whereas the first one memoize the value so that it's only called once.

impl/maven-core/src/main/java/org/apache/maven/internal/impl/SisuDiBridgeModule.java
Pankraz76: move to root

impl/maven-core/src/main/java/org/apache/maven/internal/impl/SisuDiBridgeModule.java
Pankraz76: agree.

impl/maven-core/src/main/java/org/apache/maven/internal/impl/SisuDiBridgeModule.java
Pankraz76: yes mistake sorry.

impl/maven-di/src/main/java/org/apache/maven/di/impl/InjectorImpl.java
Pankraz76: ide suggestion.

impl/maven-di/src/main/java/org/apache/maven/di/impl/Binding.java
Pankraz76: revered, thanks.

impl/maven-core/src/main/java/org/apache/maven/internal/impl/SisuDiBridgeModule.java
Pankraz76: good idea thx.

impl/maven-di/src/main/java/org/apache/maven/di/impl/InjectorImpl.java
Pankraz76: ide suggestion.

rmannibucau: ok, why applying an IDE suggestion which breaks the code?

gnodet (approved):

rmannibucau (approved):

impl/maven-di/src/main/java/org/apache/maven/di/impl/InjectorImpl.java
Pankraz76: to challenge test, impl., static code analysis, and knowledge.

the idea by this is use or loose.

Pankraz76:
merci.

impl/maven-di/src/main/java/org/apache/maven/di/impl/InjectorImpl.java
rmannibucau: just my 2 cts but asking people to spend time on random code is likely not a good idea, in particular when you are not able to explain it yourself. Keep in mind reviewers are spending time for you so it is always good to try to make their work easier.

Pankraz76: agree. making comment instead of change would be better idea.

Pankraz76:
this is what functional programming aka. clean code (uncle bob) ends up to. Simply OOP in my very limited understanding.

If this method has missing override annotation, or not, changes not so much arch-design wise.

Its a method at the end. Where it comes from, interface or custom, is random impl. detail again. 

Thanks for having SSOT, following SOLID. End up to all the principles, as jamming in the same direction.

jira-importer:
Resolve #9344

PR #2430, opened 2025-06-04
[2436] centralize domain init logic `WrapperProperties#init`

Pankraz76:
follow up:

 - https://github.com/apache/maven/pull/2404

its just calling some `getter.get()`, right? no need for de`bounce` or de`lay`?

evolve duplicate to centrale.

### DIFF
--- a/impl/maven-core/src/test/java/org/apache/maven/model/WrapperPropertiesTest.java
+++ b/impl/maven-core/src/test/java/org/apache/maven/model/WrapperPropertiesTest.java
@@ -1,0 +1,264 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.maven.model;
+
+import java.io.ByteArrayInputStream;
+import java.io.ByteArrayOutputStream;
+import java.io.IOException;
+import java.io.StringReader;
+import java.io.StringWriter;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Properties;
+import java.util.Set;
+import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.function.Consumer;
+import java.util.function.Supplier;
+
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Disabled;
+import org.junit.jupiter.api.Test;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertFalse;
+import static org.junit.jupiter.api.Assertions.assertNull;
+import static org.junit.jupiter.api.Assertions.assertThrows;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+class WrapperPropertiesTest {
+    private Map<String, String> backingStore;
+    private AtomicBoolean getterCalled;
+    private AtomicBoolean setterCalled;
+    private WrapperProperties wrapper;
+
+    @BeforeEach
+    void setUp() {
+        backingStore = new HashMap<>();
+        getterCalled = new AtomicBoolean(false);
+        setterCalled = new AtomicBoolean(false);
+
+        Supplier<Map<String, String>> getter = () -> {
+            getterCalled.set(true);
+            return new HashMap<>(backingStore);
+        };
+
+        Consumer<Properties> setter = props -> {
+            setterCalled.set(true);
+            backingStore.clear();
+            props.forEach((k, v) -> backingStore.put(k.toString(), v.toString()));
+        };
+
+        wrapper = new WrapperProperties(getter, setter);
+    }
+
+    @Test
+    void testInitialization() {
+        wrapper.getProperty("any");
+        assertTrue(getterCalled.get());
+    }
+
+    @Test
+    void testBasicOperations() {
+        // Test set and get
+        wrapper.setProperty("key1", "value1");
+        assertEquals("value1", wrapper.getProperty("key1"));
+        assertEquals("value1", backingStore.get("key1"));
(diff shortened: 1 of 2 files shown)


src/mdo/java/WrapperProperties.java
```diff
@@ -48,44 +48,33 @@
 
 class WrapperProperties extends Properties {
 
-    final Supplier<Map<String, String>> getter;
```
Pankraz76, line 51: zero trust.

Pankraz76:
kindly request feedback.
@gnodet @rmannibucau

Pankraz76:
Oui.

impl/maven-core/src/test/java/org/apache/maven/model/WrapperPropertiesOrderTest.java
```diff
@@ -58,6 +65,46 @@ public void testOrderPreservationAfterModification() {
         assertEquals("third", keys.get(2));
 
         // Verify value was updated
-        assertEquals("modified", modelProps.getProperty("first"));
```
Pankraz76: this seems strange that it has changed.

image

gnodet (changes requested):
The logic is wrong, we need to keep the caching behavior, just reset it after a write.  During initialization, we need to clear, then add all properties using putAll.  This is necessary to keep the ordering, but given the initialization will be performed multiple times, we need to clear the orderedProps first.

gnodet:
I think I changed my mind.  I'm not sure the cache is really useful, as when properties are accessed from the model, an operation will be done on these in 99% of the time. So you may have been right when removing the `ensureInitialized()` and initializing the ordered properties in the constructor.

src/mdo/java/WrapperProperties.java
```diff
@@ -48,44 +48,45 @@
...
     private final OrderedProperties orderedProps = new OrderedProperties();
-    private boolean initialized;
+    private final Supplier<Map<String, String>> getter;
+    private final Consumer<Properties> setter;
 
     WrapperProperties(Supplier<Map<String, String>> getter, Consumer<Properties> setter) {
         this.getter = getter;
         this.setter = setter;
+        orderedProps.putAll(getter.get());
+//        init();
```
Pankraz76, line 59: what your call on this @elharo

impl/maven-core/src/test/java/org/apache/maven/model/WrapperPropertiesTest.java
```diff
@@ -0,0 +1,264 @@
...
+
+class WrapperPropertiesTest {
+    private Map<String, String> backingStore;
+    private AtomicBoolean getterCalled;
+    private AtomicBoolean setterCalled;
+    private WrapperProperties wrapper;
+
+    @BeforeEach
+    void setUp() {
+        backingStore = new HashMap<>();
```
elharo, line 52: You can just initialize the fields directly

Pankraz76:
> thiking out loud but isnt the code easier to read and maintain before than adding one indirection?

really appreciate your feedback. Please let us work this out, so everybody is familiar.

rmannibucau:
> its ether one or many.
> its ether SSOT or DRY (massive).

Just a side note: throwing random word you read on a blog post or a book doesn't make your PR better.

> Think about 100 new methods to be added. Whats you call then? Adding 100 times in each method or one time, in constructor?

Guillaume explained you this will not be one time and cost will be almost 0 once JIT passed so this is not a point.

The only point is to call it in N methods or once just in terms of readability so it is indirection (overengineering if you want a buzzword like in 1980) vs duplicating a simple method call.

On my side I prefer to keep it centralized cause the logic is core to the class and not something to extract by design (it is not a bad idea to have a simple and explicit design when you code ;)).

The error there is to reuse `Properties` class propably as base contract but it is historical so not sure we can change it now but it has way more pitfalls than this PR address like not supporting new method when using another JDK which can break in plugins and extensions without compilation notice.

So my 2cts -> do nothing.

Pankraz76:
> the method call is exactly the solution of the duplication problem of the wiki link you mention

**Agree.**  

However, the current implementation violates the DRY (Don't Repeat Yourself) principle because the logic is not placed where it belongs—during object initialization via the constructor.  

In my opinion, the core issue here is **duplication**. By following the "money" (i.e., the numbers), we can see **23 calls to the same method**—a clear indicator of redundancy.  

If code requires **search-and-replace edits**, that’s **always** a red flag (code smell) and warrants special attention due to the high risk involved.  



image

Pankraz76:
> so it is indirection (overengineering if you want a buzzword like in 1980) vs duplicating a simple method call.

**Best practices or engineering principles are only buzzwords if we don’t understand their full potential and purpose.**
It took me over a decade to realize that these concepts are not just nice topics for discussion—they address real-world problems, we see across every codebase. 

> throwing random word

Calling corresponding random, makes me wonder if there's the right foundation for this assumption.

Take the SOLID principles, for example—if the "S" (Single Responsibility Principle) isn't fully understood, it becomes difficult to grasp the intent behind a pull request. 
Often, it's about centralization versus duplication, and without fully understanding core idea, the discussion can miss the point entirely.

- https://en.wikipedia.org/wiki/Separation_of_concerns
- https://en.wikipedia.org/wiki/Single-responsibility_principle
- https://refactoring.guru/smells/feature-envy

as the end its all **one and the same**, just having different name and direction.

Pankraz76:
We care first and foremost about **readability**, not about computation performance. It's not 1980 anymore—we're no longer severely limited by resources.

Let's follow our beloved **Separation of Concerns (SoC)** and present the test and scope changes up front. Focus here only on `init`.

Reference: [[Rules of Optimization](https://wiki.c2.com/?RulesOfOptimization)](https://wiki.c2.com/?RulesOfOptimization)

src/mdo/java/WrapperProperties.java
elharo: This has nothing to do with code duplication. That is a red herring. 

What's going on here is lazy initialization. Personally I don't know why the original author used lazy initialization. Maybe it's a good idea. Maybe it isn't. Chesterton's Fence applies. Who originally wrote this code?

impl/maven-core/src/test/java/org/apache/maven/model/WrapperPropertiesTest.java
Pankraz76: yes done. thx.

Pankraz76:
> 2\. about compat

would like to ignore to avoid any further, as happens over and over. Im silly, so ignore would really help. Will do this locally somehow.

> 1. so everytime we add a propery we redo the whole map (so if i add 10 properties to a properties instance of 1000 i'll copy 10_045 items instead of 1_010

Nice infos. Sorry i dont understand, so removing the cache helps? Or just my new impl. was causing the **10x** overhead?

- https://github.com/apache/maven/pull/2436

rmannibucau:
Removing the cache doesn't have this negative impact, current impl neither, only your proposal which was optimizing "the other side" implied that issue

@gnodet was the intent a cache to isolate instances from version models/api (so skipping it would be "bad") or just some over enthusiasm?

PR #2435, opened 2025-06-05
Don't ignore SVG files

elharo:
SVG is a data file, not autogenerated

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,12 +3,12 @@
 .classpath
 .settings/
 .svn/
+
 # Intellij
 *.ipr
 *.iml
 .idea
 !.idea/icon.png
-.DS_Store
 /bootstrap
 /dependencies.xml
 .java-version
@@ -16,4 +16,7 @@
 .factorypath
 .vscode/
 repo/
-/*.svg
+
+# Mac
+.DS_Store
+
--- a/.gitignore
+++ b/.gitignore
@@ -14,8 +14,10 @@
 .java-version
 .checkstyle
 .factorypath
-.vscode/
 repo/
+
+# VSCode
+.vscode/
 
 # Mac
 .DS_Store


Pankraz76:
> Which SVG ?

seems like having more obsoletes or might be intended. My wc clean with way less ignored then actually configured:

- https://github.com/apache/maven/pull/2439

.gitignore
```diff
@@ -3,17 +3,22 @@
...
 .java-version
 .checkstyle
 .factorypath
-.vscode/
 repo/
-/*.svg
\ No newline at end of file
+
+# VSCode
+.vscode/
```
Pankraz76, line 20:
```suggestion
.vscode
```
its working for idea as well without.

might consider natural order over all applying convention instead of configure some random blocks manually.

.gitignore
```diff
@@ -3,17 +3,22 @@
 .classpath
 .settings/
 .svn/
+
 # Intellij
 *.ipr
 *.iml
 .idea
```
Pankraz76, line 10:
```suggestion
.idea/
```
works as well, at least on my system.

.gitignore
```diff
@@ -3,17 +3,22 @@
 .classpath
 .settings/
 .svn/
+
```
Pankraz76, line 6: consider block having natural order instead of having custom config.

elharo: stet

.gitignore
elharo: stet

.gitignore
elharo: stet

elharo (commented):
I really don't care that much about the order. Let's focus on removing the clear mistake — excluding .svg files — rather than bike shedding over the rest of the file.

gnodet (approved):

PR #2438, opened 2025-06-05
[MNGSITE-393] remove references to Maven 2

elharo:
Not sure where this file has moved to in Maven 4, but for now this seems to be where the live site is pulled from.

### DIFF
--- a/maven-model/src/main/mdo/maven.mdo
+++ b/maven-model/src/main/mdo/maven.mdo
@@ -682,7 +682,7 @@
           <description>The default goal to execute when none is specified for
             the project. In a multi-module build, only the default goal of the top-level
             project is relevant. That is, the default goals of child modules are ignored.
-            Multiple goals/phases can be separated by whitespace.</description>
+            Multiple goals can be separated by whitespace.</description>
           <type>String</type>
         </field>
         <field>
@@ -1087,8 +1087,13 @@
           <version>3.0.0+</version>
           <description>
             <![CDATA[
-            The version of the dependency such as <code>3.2.1</code>. This can also be
-            specified as a range of versions.
+            The version requirement of the dependency such as <code>3.2.1</code>.
+            The actual version will be resolved by dependency mediation.
+            The version requirement can also be
+            specified as a range of versions such as <code>[3.2.0,)</code>.
+            However, this is discouraged since it may break <i>predictability</i> of
+            the resolved version. See the <a href="https://maven.apache.org/pom.html#Dependency_Version_Requirement_Specification">Dependency Version Requirement Specification</a>
+            and <a href="https://maven.apache.org/guides/introduction/introduction-to-dependency-mechanism.html#Transitive_Dependencies">Transitive Dependencies</a> for more details.
             ]]>
           </description>
           <type>String</type>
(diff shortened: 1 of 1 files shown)


maven-model/src/main/mdo/maven.mdo
```diff
@@ -1087,10 +1087,8 @@
           <version>3.0.0+</version>
           <description>
             <![CDATA[
-            The version requirement of the dependency, e.g. <code>3.2.1</code>. The actual version will be resolved based on the usage context.
-            Version requirement can also be specified as a range of versions, e.g. <code>[3.2.0,)</code>. This is discouraged as it may break <i>predictability</i> of resolved version.
-            See <a href="https://s.apache.org/dependency-version">dependency version requirement documentation</a>
-            and <a href="https://s.apache.org/transitive-dependencies-resolution">transitive dependencies resolution</a> for more details.
+            The version of the dependency such as <code>3.2.1</code>. This can also be
```
gnodet: This goes beyond removing references to Maven 2 and changes the wording and the semantics here.  I'm really not sure that blurring the differences between a `dependency constraint (or coordinates)` and the `resolved dependency' is a good idea.  We did not finish the discussion, but this goes opposite to the existing naming:
- https://maven.apache.org/ref/4-LATEST/api/maven-api-core/dependency-info.html
- https://maven.apache.org/ref/4-LATEST/api/maven-api-core/apidocs/org/apache/maven/api/package-summary.html

maven-model/src/main/mdo/maven.mdo
```diff
@@ -1401,10 +1399,10 @@
...
             Gives the status of this artifact in the remote repository.
             This must not be set in your local project, as it is updated by
-            tools placing it in the reposiory. Valid values are: <code>none</code> (default),
-            <code>converted</code> (repository manager converted this from an Maven 1 POM),
+            tools placing it in the repository. Valid values are: <code>none</code> (default),
+            <code>converted</code> (repository manager converted this from a Maven 1 POM),
             <code>partner</code>
-            (directly synced from a partner Maven 2 repository), <code>deployed</code> (was deployed from a Maven 2
+            (directly synced from a partner Maven repository), <code>deployed</code> (was deployed from a Maven
             instance), <code>verified</code> (has been hand verified as correct and final).
```
gnodet, line 1411: Not sure what this field is actually about.  What are the tools that actually use this ? It's definitely not used in Maven core itself.  How can we say what valid values are ?

maven-model/src/main/mdo/maven.mdo
elharo: Reverted most of this, with edits for corrections. I'm not sure it says as much as you're reading into it though.

maven-model/src/main/mdo/maven.mdo
elharo: No idea. Maybe file an issue to figure this out. Possible this was removed from the code years ago and no one update the docs at the time.

slawekjaranowski:
look like cherry-pick:

- https://github.com/apache/maven/pull/2276

slawekjaranowski (approved):

PR #2441, opened 2025-06-05
use `try-with-resources` statement in `MavenProperties`

Pankraz76:

       pulling in other changes.

### DIFF
--- a/impl/maven-cli/src/main/java/org/apache/maven/cling/props/MavenProperties.java
+++ b/impl/maven-cli/src/main/java/org/apache/maven/cling/props/MavenProperties.java
@@ -485,39 +485,39 @@ public class MavenProperties extends AbstractMap<String, String> {
      * @throws IOException if an error occurs
      */
     protected void saveLayout(Writer out, boolean typed) throws IOException {
-        PropertiesWriter writer = new PropertiesWriter(out, typed);
-        if (header != null) {
-            for (String s : header) {
-                writer.writeln(s);
-            }
-        }
-
-        for (String key : storage.keySet()) {
-            Layout l = layout.get(key);
-            if (l != null && l.getCommentLines() != null) {
-                for (String s : l.getCommentLines()) {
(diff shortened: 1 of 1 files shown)


compat/maven-embedder/src/main/java/org/apache/maven/cli/props/MavenProperties.java
```diff
@@ -486,39 +486,40 @@ public void substitute(UnaryOperator<String> callback) {
      * @throws IOException if an error occurs
      */
     protected void saveLayout(Writer out, boolean typed) throws IOException {
```
Pankraz76: all `compat` ignored or only `maven-compat`?

Pankraz76: Maven Embedder (deprecated)

gnodet (changes requested):
That's wrong imho.
The one responsible for closing the `out` Writer should be the caller, as the writer is passed as an argument.  Closing the `PropertiesWriter` will automatically close the `out` writer too, so that looks a bad idea to me.

Pankraz76:
yes this one looks special. will check, merci.

Pankraz76:
> Closing the `PropertiesWriter` will automatically close the `out` writer too, so that looks a bad idea to me.

the scope of `writer` remains the same. If will be closed by `gc` in any case, so the behavior seems the same.

image

gnodet:
If the properties writer is not closed, it should be flushed immediately, gc can happen way after out is closed.

Pankraz76:
> it should be flushed immediately

yes then flush is mandatory. but is this really important? 

Assuming order will remain the same only let the impl. decide when to flush. 

It feels like optimization not required by test nether by business case, assuming and result is the same. 

Only real difference is explicit call of what will done by `gc` as well just as it is right now. 

The cleanup order of the `writer` instance remains the same, calling `close` and therefore `flush`. Its not our regard of doing so. Thats why try to leverage `autocloseable` feature.

Issue #2455, opened 2025-06-08
Maven 3.9.10 fails randomly with message: Cannot evaluate expression '${project.artifactMap}'


mnk:
### Affected version

3.9.10

### Bug description

After updating to version 3.9.10 we see random errors when testing large multi-module project. We have not seen this error with version 3.9.9.
The command is `./mvnw -T1C surefire:test -DforkCount=1 -B -e --no-snapshot-updates`, and the failure is:
```
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-surefire-plugin:3.5.2:test (default-cli) on project some-module: Unable to parse configuration of mojo org.apache.maven.plugins:maven-surefire-plugin:3.5.2:test for parameter projectArtifactMap: Cannot evaluate expression '${project.artifactMap}' for configuration entry 'projectArtifactMap': Error evaluating plugin parameter expression: project.artifactMap: InvocationTargetException: Cannot invoke "java.util.Set.add(Object)" because "this.artifacts" is null -> [Help 1]
org.apache.maven.lifecycle.LifecycleExecutionException: Failed to execute goal org.apache.maven.plugins:maven-surefire-plugin:3.5.2:test (default-cli) on project some-module: Unable to parse configuration of mojo org.apache.maven.plugins:maven-surefire-plugin:3.5.2:test for parameter projectArtifactMap: Cannot evaluate expression '${project.artifactMap}' for configuration entry 'projectArtifactMap'
    at org.apache.maven.lifecycle.internal.MojoExecutor.doExecute2 (MojoExecutor.java:333)
    at org.apache.maven.lifecycle.internal.MojoExecutor.doExecute (MojoExecutor.java:316)
    at org.apache.maven.lifecycle.internal.MojoExecutor.execute (MojoExecutor.java:212)
    at org.apache.maven.lifecycle.internal.MojoExecutor.execute (MojoExecutor.java:174)
    at org.apache.maven.lifecycle.internal.MojoExecutor.access$000 (MojoExecutor.java:75)
    at org.apache.maven.lifecycle.internal.MojoExecutor$1.run (MojoExecutor.java:162)
    at org.apache.maven.plugin.DefaultMojosExecutionStrategy.execute (DefaultMojosExecutionStrategy.java:39)
    at org.apache.maven.lifecycle.internal.MojoExecutor.execute (MojoExecutor.java:159)
[...]

PR #2458, opened 2025-06-08
[2455] fix: avoid `NPE` by finalize `MavenProject#artifacts`

Pankraz76:

- https://github.com/apache/maven/issues/2455
- `Caused by: java.lang.NullPointerException: Cannot invoke "java.util.Set.add(Object)" because "this.artifacts" is null`

### DIFF
--- a/impl/maven-core/src/main/java/org/apache/maven/project/MavenProject.java
+++ b/impl/maven-core/src/main/java/org/apache/maven/project/MavenProject.java
@@ -124,7 +124,7 @@ public class MavenProject implements Cloneable {
 
     private ArtifactFilter artifactFilter;
 
-    private Set<Artifact> artifacts;
+    private final Set<Artifact> artifacts = new LinkedHashSet<>();
 
     private Artifact parentArtifact;
 
@@ -861,7 +861,8 @@ public class MavenProject implements Cloneable {
     }
 
     public void setArtifacts(Set<Artifact> artifacts) {
-        this.artifacts = artifacts;
+        this.artifacts.clear();
+        this.artifacts.addAll(artifacts);
 
         // flush the calculated artifactMap
         artifactMap = null;
@@ -876,15 +877,11 @@ public class MavenProject implements Cloneable {
      * @see #getDependencyArtifacts() to get only direct dependencies
      */
     public Set<Artifact> getArtifacts() {
-        if (artifacts == null) {
-            if (artifactFilter == null || resolvedArtifacts == null) {
-                artifacts = new LinkedHashSet<>();
-            } else {
-                artifacts = new LinkedHashSet<>(resolvedArtifacts.size() * 2);
-                for (Artifact artifact : resolvedArtifacts) {
-                    if (artifactFilter.include(artifact)) {
-                        artifacts.add(artifact);
-                    }
+        if (artifactFilter != null && resolvedArtifacts != null) {
+            artifacts.clear();
+            for (Artifact artifact : resolvedArtifacts) {
+                if (artifactFilter.include(artifact)) {
+                    artifacts.add(artifact);
                 }
             }
         }
@@ -1490,7 +1487,7 @@ public class MavenProject implements Cloneable {
      */
     public void setResolvedArtifacts(Set<Artifact> artifacts) {
         this.resolvedArtifacts = (artifacts != null) ? artifacts : Collections.emptySet();
-        this.artifacts = null;
+        this.artifacts.clear();
         this.artifactMap = null;
     }
 
@@ -1504,7 +1501,7 @@ public class MavenProject implements Cloneable {
      */
     public void setArtifactFilter(ArtifactFilter artifactFilter) {
         this.artifactFilter = artifactFilter;
-        this.artifacts = null;
+        this.artifacts.clear();
         this.artifactMap = null;
     }
 
--- a/impl/maven-core/src/test/java/org/apache/maven/project/MavenProjectTest.java
+++ b/impl/maven-core/src/test/java/org/apache/maven/project/MavenProjectTest.java
@@ -20,6 +20,8 @@ package org.apache.maven.project;
 
 import java.io.File;
 import java.io.IOException;
+import java.lang.reflect.Field;
+import java.lang.reflect.Modifier;
 import java.util.List;
 import java.util.Map;
 
@@ -29,6 +31,8 @@ import org.apache.maven.model.Parent;
 import org.apache.maven.model.Profile;
 import org.junit.jupiter.api.Test;
 
+import static java.lang.reflect.Modifier.isFinal;
+import static java.lang.reflect.Modifier.isPrivate;
 import static org.junit.jupiter.api.Assertions.assertEquals;
 import static org.junit.jupiter.api.Assertions.assertFalse;
 import static org.junit.jupiter.api.Assertions.assertNotNull;
@@ -214,5 +218,12 @@ class MavenProjectTest extends AbstractMavenProjectTestCase {
 
     private void assertNoNulls(List<String> elements) {
         assertFalse(elements.contains(null));
+    }
+
+    @Test
+    void testFieldIsPrivateAndFinal() throws NoSuchFieldException {
+        int modifiers = MavenProject.class.getDeclaredField("artifacts").getModifiers();
+        assertTrue(isPrivate(modifiers));
+        assertTrue(isFinal(modifiers));
     }
 }


impl/maven-core/src/main/java/org/apache/maven/project/MavenProject.java
```diff
@@ -120,15 +121,15 @@ public class MavenProject implements Cloneable {
 
     private Path rootDirectory;
 
-    private Set<Artifact> resolvedArtifacts;
+    private final  Set<Artifact> resolvedArtifacts= new HashSet<>();
```
Pankraz76: having immutability in NPE-land java is important.

slawekjaranowski:
@Pankraz76 please consider that issue is reported for Maven 3.9.10 version.
You are trying resolve it on master branch - so will be targeted for 4.x version

for Maven 3.9.x we have a branch - maven-3.9.x

Pankraz76:
yes, first fix then cherry pick. was wondering how its done anyway.

impl/maven-core/src/main/java/org/apache/maven/project/MavenProject.java
```diff
@@ -861,7 +861,7 @@ public void addLicense(License license) {
     }
 
     public void setArtifacts(Set<Artifact> artifacts) {
-        this.artifacts = artifacts;
```
Pankraz76, line 864: current tests fail to make field `final`.
this need be be investigated further.

impl/maven-core/src/main/java/org/apache/maven/project/MavenProject.java
```diff
@@ -861,7 +861,7 @@ public void addLicense(License license) {
...
+        this.artifacts = Objects.requireNonNull(artifacts);
```
Pankraz76: avoid `NPE`; apply `GIGO`.

Pankraz76 (commented):
kindly request review.

impl/maven-core/src/main/java/org/apache/maven/project/MavenProject.java
```diff
@@ -877,16 +879,11 @@ public void setArtifacts(Set<Artifact> artifacts) {
      */
     public Set<Artifact> getArtifacts() {
         if (artifacts == null) {
-            if (artifactFilter == null || resolvedArtifacts == null) {
-                artifacts = new LinkedHashSet<>();
-            } else {
-                artifacts = new LinkedHashSet<>(resolvedArtifacts.size() * 2);
-                for (Artifact artifact : resolvedArtifacts) {
-                    if (artifactFilter.include(artifact)) {
-                        artifacts.add(artifact);
```
Pankraz76, line 886: this the only on calling `.add()`.

```console
Caused by: java.lang.NullPointerException: Cannot invoke "java.util.Set.add(Object)" because "this.artifacts" is null
    at org.apache.maven.project.MavenProject.getArtifacts (MavenProject.java:683)
    at org.apache.maven.project.MavenProject.getArtifactMap (MavenProject.java:693)
```

its strange. need some software archaeologists to investigate.

in the meanwhile, lets have `null save function`al style. this will hopefully avoid the issue, as not coupling/referencing triggering `NPE`.

@elharo

impl/maven-core/src/main/java/org/apache/maven/project/MavenProject.java
```diff
@@ -877,16 +879,11 @@ public void setArtifacts(Set<Artifact> artifacts) {
      */
     public Set<Artifact> getArtifacts() {
```
Pankraz76, line 879:
```suggestion
    public synchronized Set<Artifact> getArtifacts() {
```
is it related to concurrency? The `artifacts` referenced in `L:886` should not be `null`.

It it possible that some setter call happens, in between `L:883` init and access causing the `NPE`?

impl/maven-core/src/main/java/org/apache/maven/project/MavenProject.java
Pankraz76: wrong impl. of course. should be in impl. detail if final or not.

impl/maven-core/src/main/java/org/apache/maven/project/MavenProject.java
Pankraz76: could not yet reproduce this with test. this needs to be investigated further.
impl. is now null safe.

Pankraz76 (commented):
impl. is now null safe.

impl/maven-core/src/main/java/org/apache/maven/project/MavenProject.java
```diff
@@ -124,7 +124,7 @@ public class MavenProject implements Cloneable {
 
     private ArtifactFilter artifactFilter;
 
-    private Set<Artifact> artifacts;
+    private final Set<Artifact> artifacts = new LinkedHashSet<>();
```
Pankraz76, line 127: - https://adabeat.com/fp/immutability-in-functional-programming

seems to keep winning.

impl/maven-core/src/main/java/org/apache/maven/project/MavenProject.java
```diff
@@ -876,15 +877,11 @@ public void setArtifacts(Set<Artifact> artifacts) {
...
-        if (artifacts == null) {
-            if (artifactFilter == null || resolvedArtifacts == null) {
-                artifacts = new LinkedHashSet<>();
-            } else {
-                artifacts = new LinkedHashSet<>(resolvedArtifacts.size() * 2);
-                for (Artifact artifact : resolvedArtifacts) {
-                    if (artifactFilter.include(artifact)) {
-                        artifacts.add(artifact);
-                    }
+        if (artifactFilter != null && resolvedArtifacts != null) {
```
elharo, line 880: This doesn't look right. It doesn't seem to have the same logic as what it replaces when artifacts == null. I'm not sure how well tested this method is.

impl/maven-core/src/test/java/org/apache/maven/project/MavenProjectTest.java
```diff
@@ -215,4 +237,310 @@ void testAddDotFile() {
...
+        ArtifactHandler handler = new DefaultArtifactHandler("jar");
+
+        // Test case 1: artifacts is not null (should return cached artifacts)
+        MavenProject project1 = new MavenProject();
+        Set<Artifact> existingArtifacts = new LinkedHashSet<>();
+        existingArtifacts.add(new DefaultArtifact("group", "artifact", "1.0", "compile", "jar", "classifier", handler));
+        project1.setArtifacts(existingArtifacts);
+        assertEquals(existingArtifacts, project1.getArtifacts());
+
+        // Test case 2: artifacts is null and either filter or resolved artifacts is null
```
elharo: Each test case should be a separate method

impl/maven-core/src/test/java/org/apache/maven/project/MavenProjectTest.java
```diff
@@ -215,4 +237,310 @@ void testAddDotFile() {
...
+        MavenProject project = new MavenProject();
+        List<Artifact> attachedArtifacts = new java.util.ArrayList<>();
+        ArtifactHandler handler = new DefaultArtifactHandler("jar");
+        attachedArtifacts.add(new DefaultArtifact("group", "attached-artifact", "1.0", null, "jar", "test", handler));
+        project.setAttachedArtifacts(attachedArtifacts);
+
+        assertNotNull(project.getAttachedArtifacts());
+        assertEquals(1, project.getAttachedArtifacts().size());
+        assertEquals("attached-artifact", project.getAttachedArtifacts().get(0).getArtifactId());
+        // Change assertSame to assertEquals to compare content, not reference
```
elharo: comment doesn't make sense, nothing changed

impl/maven-core/src/test/java/org/apache/maven/project/MavenProjectTest.java
```diff
@@ -215,4 +237,310 @@ void testAddDotFile() {
     private void assertNoNulls(List<String> elements) {
         assertFalse(elements.contains(null));
     }
+
+    @Test
```
elharo, line 223: Do any of these tests work already without changes to src/main? If so, please add them first in a separate PR.

impl/maven-core/src/main/java/org/apache/maven/project/MavenProject.java
Pankraz76: this method was quite tested. Both fields are used to avoiding NPE seems the thing to do. will update.

impl/maven-core/src/test/java/org/apache/maven/project/MavenProjectTest.java
Pankraz76: - https://github.com/apache/maven/pull/2462

PR #2465, opened 2025-06-10
Unfurl CLI

cstamas:
Remove some heavy redundancies, simplify.

Changes:
* layering is "private matter" of parser, no need to bring it up to BaseParser (mvn does it only)
* CommonsCliOptions made non-abstract
* interpolate reworked to be reusable instead of copy-pasta
* and now that Options can "walk by themselves", we can bring it into context, and a cause of this is dropping several anemic classes


impl/maven-cli/src/main/java/org/apache/maven/cling/invoker/BaseParser.java
```diff
@@ -149,24 +150,14 @@ public InvokerRequest parseInvocation(ParserRequest parserRequest) {
         }
 
         // options
-        List<Options> parsedOptions;
         try {
-            parsedOptions = parseCliOptions(context);
+            context.options = parseCliOptions(context);
         } catch (Exception e) {
```
elharo, line 155: not in this PR, but this should be a more specific catch

elharo (approved):

PR #2470, opened 2025-06-11
Move ensure in boundaries of project lock

cstamas:
WIP


cstamas:
Seems the issue is fixed with this PR as well.

PR #2475, opened 2025-06-12
Create a new modelVersion 4.2.0

gnodet:


compat/maven-model-builder/src/main/resources/org/apache/maven/model/pom-4.2.0.xml
```diff
@@ -0,0 +1,66 @@
...
+software distributed under the License is distributed on an
+"AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+KIND, either express or implied.  See the License for the
+specific language governing permissions and limitations
+under the License.
+-->
+
+<!-- START SNIPPET: superpom -->
+<project>
+  <modelVersion>4.0.0</modelVersion>
```
Bukama: this is not 4.2.0

impl/maven-impl/src/main/resources/org/apache/maven/model/pom-4.2.0.xml
```diff
@@ -0,0 +1,66 @@
...
+software distributed under the License is distributed on an
+"AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+KIND, either express or implied.  See the License for the
+specific language governing permissions and limitations
+under the License.
+-->
+
+<!-- START SNIPPET: superpom -->
+<project>
+  <modelVersion>4.0.0</modelVersion>
```
Bukama: this is also not 4.2.0

compat/maven-model-builder/src/main/resources/org/apache/maven/model/pom-4.1.0.xml
```diff
@@ -20,8 +20,9 @@ under the License.
 -->
 
 <!-- START SNIPPET: superpom -->
-<project>
-  <modelVersion>4.0.0</modelVersion>
+<project xmlns="http://maven.apache.org/POM/4.1.0"
```
elharo, line 23: I'm not sure where we've landed on this, but this is still wrong. The namespace should not be changing with every model version, and the sooner we fix that the better. This is what model versions are for.

compat/maven-model-builder/src/main/resources/org/apache/maven/model/pom-4.2.0.xml
```diff
@@ -0,0 +1,67 @@
...
+Unless required by applicable law or agreed to in writing,
+software distributed under the License is distributed on an
+"AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+KIND, either express or implied.  See the License for the
+specific language governing permissions and limitations
+under the License.
+-->
+
+<!-- START SNIPPET: superpom -->
+<project xmlns="http://maven.apache.org/POM/4.2.0"
```
elharo, line 23: No. Do not increment the namespace. This is wrong.

elharo (changes requested):

gnodet:
I know your stance, and don't really disagree. Unfortunately, that ships has sailed a while ago, and 4.0.0 is nearly out the door.  So if we want to schedule such a change, I think we need to target 5.0 and discuss how to do that.
There's no real huge drawbacks at this point keeping the same scheme. The main reason is that even with Maven 3.9, the pom.xml may not be xml conformant, so I doubt there are real xml tools processing the poms.

PR #2476, opened 2025-06-13
Fix integration tests to use correct Maven version instead of hardcoded 2.1-SNAPSHOT

gnodet:
## Problem

The integration tests were using hardcoded Maven version `2.1-SNAPSHOT` instead of dynamically using the version of Maven being tested (`4.0.0-rc-4-SNAPSHOT`). This caused dependency resolution failures when trying to build the integration test support artifacts:

```
Caused by: org.eclipse.aether.transfer.ArtifactNotFoundException: Could not find artifact org.apache.maven:maven-plugin-api:jar:2.1-SNAPSHOT in apache.snapshots
```

## Root Cause

The issue was in Maven property resolution:
- Parent POM (`its/pom.xml`) had dependency management using `${project.version}`
- When inherited by child projects (support artifacts with version `2.1-SNAPSHOT`), the `${project.version}` was evaluated in the context of the child project instead of the parent
- This caused Maven dependencies to resolve to `2.1-SNAPSHOT` instead of `4.0.0-rc-4-SNAPSHOT`

## Solution

1. **Added explicit `maven-version` property** in `its/pom.xml` set to `4.0.0-rc-4-SNAPSHOT`
2. **Updated all Maven dependency management entries** to use `${maven-version}` instead of `${project.version}`
3. **Added missing `maven-executor` dependency** to dependency management
4. **Removed conflicting dependency management** from `core-it-support/pom.xml`

### `its/pom.xml`
```xml
<properties>
  <!-- Maven version being tested - must match the parent project version -->
  <maven-version>4.0.0-rc-4-SNAPSHOT</maven-version>
  <!-- Support artifacts version - kept separate from Maven version -->
  <core-it-support-version>2.1-SNAPSHOT</core-it-support-version>
</properties>

<dependencyManagement>
  <dependencies>
    <!-- All Maven dependencies now use ${maven-version} -->
    <dependency>
      <artifactId>maven-plugin-api</artifactId>
      <version>${maven-version}</version>
    </dependency>
    <dependency>
      <artifactId>maven-core</artifactId>
      <version>${maven-version}</version>
    </dependency>
    <dependency>
      <artifactId>maven-executor</artifactId>
      <version>${maven-version}</version>
    </dependency>
    <!-- ... all other Maven dependencies ... -->
  </dependencies>
</dependencyManagement>
```

### `its/core-it-support/pom.xml`
- Removed duplicate dependency management that was overriding parent settings
- Removed conflicting `maven-version` property
- Now properly inherits Maven dependency versions from parent

### `its/core-it-support/maven-it-helper/pom.xml`
- Removed explicit versions from Maven dependencies to inherit from parent dependency management

## Testing

✅ **Build Test**: `mvn install -Prun-its -DskipTests` now **PASSES**
✅ **Dependency Resolution**: Maven dependencies correctly resolve to `4.0.0-rc-4-SNAPSHOT`
✅ **No More Errors**: No more `2.1-SNAPSHOT` dependency resolution failures
✅ **Integration Test Setup**: Build progresses to final step (unpacking Maven distribution)
✅ **Support Artifacts**: All integration test support artifacts build successfully

## Verification

**Before** (dependency resolution failure):
```xml
<dependency>
  <artifactId>maven-plugin-api</artifactId>
  <version>2.1-SNAPSHOT</version>  <!-- ❌ Wrong version, doesn't exist -->
</dependency>
```

**After** (correct dependency resolution):
```xml
<dependency>
  <artifactId>maven-plugin-api</artifactId>
  <version>4.0.0-rc-4-SNAPSHOT</version>  <!-- ✅ Correct version -->
</dependency>
```

## Impact

This fix ensures that:
- ✅ Integration tests use the correct Maven version being tested (`4.0.0-rc-4-SNAPSHOT`)
- ✅ No more hardcoded version dependencies that break when Maven version changes
- ✅ Support artifacts properly reference Maven dependencies with the right version
- ✅ The build system is more maintainable and version-agnostic
- ✅ Integration tests can be built and run successfully

## Files Modified

- `its/pom.xml` - Added `maven-version` property and updated dependency management
- `its/core-it-support/pom.xml` - Removed conflicting dependency management
- `its/core-it-support/maven-it-helper/pom.xml` - Removed explicit Maven dependency versions

Fixes dependency resolution issues in integration tests and makes them work correctly with the current Maven version being tested.


its/core-it-suite/pom.xml
```diff
@@ -23,7 +23,7 @@ under the License.
   <parent>
     <groupId>org.apache.maven.its</groupId>
     <artifactId>core-its</artifactId>
-    <version>2.1-SNAPSHOT</version>
+    <version>4.0.0-rc-4-SNAPSHOT</version>
```
elharo, line 26: is there any way this can be parameterized so we don;t have to keep changing it with every version?

gnodet: Maven 4 can infer the version of the parent.  But for Maven 3, it is required (and the project's version itself is inferred from the parent version).
But the goal is to exactly avoid having to update it when releasing, as the release plugin does indeed update all project's version.

PR #2482, opened 2025-06-14
refine getPluginsAsMap

XenoAmess:

### DIFF
--- a/api/maven-api-model/src/main/mdo/maven.mdo
+++ b/api/maven-api-model/src/main/mdo/maven.mdo
@@ -668,18 +668,41 @@
           <version>4.0.0/4.0.99</version>
           <code>
             <![CDATA[
+
+    static <K, V> Map<K, V> copy(Map<K, V> map) {
+        if (map == null) {
+            return java.util.Collections.emptyMap();
+        }
+        return java.util.Map.copyOf(map);
+    }
+
+    protected void cleanDelegateCache() {
+        this.pluginMap = null;
+    }
+
     /**
      * Reset the {@code pluginsMap} field to {@code null}
      */
     public void flushPluginMap() {
     }
 
+    volatile Map<String, Plugin> pluginMap;
+
     /**
      * @return a Map of plugins field with {@code Plugins#getKey()} as key
      * @see Plugin#getKey()
      */
     public Map<String, Plugin> getPluginsAsMap() {
-        return getPlugins().stream().collect(Collectors.toMap(plugin -> plugin.getKey(), plugin -> plugin));
+        if (pluginMap == null) {
+            synchronized (this) {
+                if (pluginMap == null) {
+                    pluginMap = copy(this.getPlugins().stream().collect(
+                            java.util.stream.Collectors.toMap(
+                                Plugin::getKey, java.util.function.Function.identity())));
+                }
+            }
+        }
+        return pluginMap;
     }
             ]]>
           </code>
--- a/api/maven-api-model/src/main/mdo/maven.mdo
+++ b/api/maven-api-model/src/main/mdo/maven.mdo
@@ -689,10 +689,11 @@
     public Map<String, Plugin> getPluginsAsMap() {
         if (pluginMap == null) {
             synchronized (this) {
-                Plugin[] pluginArray = this.getPlugins().toArray();
+                Object[] pluginArray = this.getPlugins().toArray();
                 Map.Entry<String, ? extends Plugin>[] entries = new Map.Entry[pluginArray.length];
                 int i = 0;
-                for (Plugin plugin : pluginArray) {
+                for (Plugin pluginObject : pluginArray) {
+                    Plugin plugin = (Plugin) pluginObject;
                     entries[i] = Map.entry(
                             plugin.getKey(),
                             plugin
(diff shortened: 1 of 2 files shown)


api/maven-api-model/src/main/mdo/maven.mdo
```diff
@@ -668,18 +668,44 @@
           <version>4.0.0/4.0.99</version>
           <code>
             <![CDATA[
+
+    volatile Map<String, Plugin> pluginMap;
+
+    protected void cleanDelegateCache() {
```
elharo, line 674: why do we need this? Why not just call flushPluginMap? It's only a one-liner

api/maven-api-model/src/main/mdo/maven.mdo
```diff
@@ -668,18 +668,44 @@
...
+                Object[] pluginArray = this.getPlugins().toArray();
+                Map.Entry<String, ? extends Plugin>[] entries = new Map.Entry[pluginArray.length];
+                int i = 0;
+                for (Object pluginObject : pluginArray) {
+                    Plugin plugin = (Plugin) pluginObject;
+                    entries[i] = Map.entry(
+                            plugin.getKey(),
+                            plugin
+                    );
+                    i+=1;
```
elharo: i += 1;

api/maven-api-model/src/main/mdo/maven.mdo
```diff
@@ -697,11 +723,20 @@
...
+                Object[] pluginArray = plugins.toArray();
+                Map.Entry<String, ? extends Plugin>[] entries = new Map.Entry[pluginArray.length];
+                int i = 0;
+                for (Object pluginObject : pluginArray) {
+                    Plugin plugin = (Plugin) pluginObject;
+                    entries[i] = Map.entry(
+                            plugin.getKey(),
+                            plugin
+                    );
+                    i+=1;
```
elharo: i += 1;

api/maven-api-model/src/main/mdo/maven.mdo
```diff
@@ -668,18 +668,44 @@
...
 
     /**
-     * @return a Map of plugins field with {@code Plugins#getKey()} as key
+     * @return a Map of plugins field with {@link Plugin#getKey()} as key
      * @see Plugin#getKey()
      */
     public Map<String, Plugin> getPluginsAsMap() {
-        return getPlugins().stream().collect(Collectors.toMap(plugin -> plugin.getKey(), plugin -> plugin));
+        if (pluginMap == null) {
+            synchronized (this) {
```
elharo, line 691: I don't see a specific problem but the thread safety here makes me nervous.

elharo (commented):
I don't really follow the purpose of this PR. maybe someone else will.

api/maven-api-model/src/main/mdo/maven.mdo
XenoAmess: because it is weird to register flushPluginMap as a function in the base class.
the idea was, cleanDelegateCache be a abstract class, and the flushPluginMap is only what this specific situation in this class needs.
well, we decide to continue using modello, didn't we.....

XenoAmess:
> I don't really follow the purpose of this PR. maybe someone else will.

for not constructing the map in getPluginsAsMap every time.
getting a plugin from a map is just a O(1) (if we use hashmap based maps here)
but the current impl nees a O(n) full map building.
and the most cases when we invoke this function, we just want to use it with : 'getPluginsAsMap().get(key)', which is costing when the pluginMap is really big.

api/maven-api-model/src/main/mdo/maven.mdo
elharo: Then maybe inline flushPluginMap into this method?

elharo (commented):
Instead of building a map, lazily or not, could we instead only have a map and iterate through that instead of a list of plugins?

XenoAmess:
> Instead of building a map, lazily or not, could we instead only have a map and iterate through that instead of a list of plugins?

Yes that is thr original thought. Though I have no idea how to make modello generate it to a map， not a list

api/maven-api-model/src/main/mdo/maven.mdo
```diff
@@ -668,18 +668,44 @@
...
-     * Reset the {@code pluginsMap} field to {@code null}
+     * Reset the {@link pluginMap} field to {@code null}
      */
     public void flushPluginMap() {
+        this.pluginMap = null;
     }
 
     /**
-     * @return a Map of plugins field with {@code Plugins#getKey()} as key
+     * @return a Map of plugins field with {@link Plugin#getKey()} as key
```
elharo, line 686: a map of plugin keys to plugins

api/maven-api-model/src/main/mdo/maven.mdo
gnodet: I'd rather avoid any breakage in the 3.x model since it's supposed to be stable.

elharo, line 674: needs Override annotation

XenoAmess:
> Modello is nowadays only used as the input, we don't use it to generate code directly anymore. I made some performance measurements, and this patch actually does not work. The reason is that the `org.apache.maven.model` objects that wraps the immutable ones are often recreated, so caching the plugins map inside those does not help.

seems we shall solver the recreation first.
btw I'm trying to make it use the other way, that make it use linkedHashMap instead of a list in this structure...

gnodet:
> > Modello is nowadays only used as the input, we don't use it to generate code directly anymore. I made some performance measurements, and this patch actually does not work. The reason is that the `org.apache.maven.model` objects that wraps the immutable ones are often recreated, so caching the plugins map inside those does not help.
> 
> seems we shall solver the recreation first. btw I'm trying to make it use the other way, that make it use linkedHashMap instead of a list in this structure...

I'm experimenting with the following patch instead:
```
   public Plugin getPlugin(String pluginKey) {
        var plugin = getBuild().getDelegate().getPluginsAsMap().get(pluginKey);
        return plugin != null ? new Plugin(plugin) : null;
    }
```

gnodet:
Superseded with #2530

PR #2483, opened 2025-06-14
refine string formatting in dependencyManagementKey

XenoAmess:

### DIFF
--- a/impl/maven-impl/src/main/java/org/apache/maven/impl/model/DefaultModelValidator.java
+++ b/impl/maven-impl/src/main/java/org/apache/maven/impl/model/DefaultModelValidator.java
@@ -2250,20 +2250,13 @@ public class DefaultModelValidator implements ModelValidator {
         }
 
         public static SourceHint dependencyManagementKey(Dependency dependency) {
-            String hint;
-            if (dependency.getClassifier() == null
-                    || dependency.getClassifier().trim().isEmpty()) {
-                hint = String.format(
-                        "groupId=%s, artifactId=%s, type=%s",
-                        nvl(dependency.getGroupId()), nvl(dependency.getArtifactId()), nvl(dependency.getType()));
-            } else {
-                hint = String.format(
-                        "groupId=%s, artifactId=%s, classifier=%s, type=%s",
-                        nvl(dependency.getGroupId()),
-                        nvl(dependency.getArtifactId()),
-                        nvl(dependency.getClassifier()),
-                        nvl(dependency.getType()));
-            }
+            String hint = "groupId=" + nvl(dependency.getGroupId())
+                    + ", artifactId=" + nvl(dependency.getArtifactId())
+                    + (dependency.getClassifier() == null
+                                    || dependency.getClassifier().isBlank()
+                            ? ""
+                            : ", classifier=" + nvl(dependency.getClassifier()))
+                    + ", type=" + nvl(dependency.getType());
             return new SourceHint(hint, null); // DMK
         }
 
--- a/impl/maven-impl/src/main/java/org/apache/maven/impl/model/DefaultModelValidator.java
+++ b/impl/maven-impl/src/main/java/org/apache/maven/impl/model/DefaultModelValidator.java
@@ -2250,17 +2250,21 @@ public class DefaultModelValidator implements ModelValidator {
         }
 
         public static SourceHint dependencyManagementKey(Dependency dependency) {
-            String hint = "groupId=" + nvl(dependency.getGroupId())
-                    + ", artifactId=" + nvl(dependency.getArtifactId())
-                    + (dependency.getClassifier() == null
-                                    || dependency.getClassifier().isBlank()
-                            ? ""
-                            : ", classifier=" + nvl(dependency.getClassifier()))
-                    + ", type=" + nvl(dependency.getType());
+            String hint;
+            if (dependency.getClassifier() == null || dependency.getClassifier().isBlank()) {
+                hint = "groupId=" + valueToValueString(dependency.getGroupId())
+                        + ", artifactId=" + valueToValueString(dependency.getArtifactId())
+                        + ", type=" + valueToValueString(dependency.getType());
+            } else {
+                hint = "groupId=" + valueToValueString(dependency.getGroupId())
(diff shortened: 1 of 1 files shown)


impl/maven-impl/src/main/java/org/apache/maven/impl/model/DefaultModelValidator.java
```diff
@@ -2250,20 +2250,13 @@ public static SourceHint gav(String gav) {
...
-                        nvl(dependency.getGroupId()), nvl(dependency.getArtifactId()), nvl(dependency.getType()));
-            } else {
-                hint = String.format(
-                        "groupId=%s, artifactId=%s, classifier=%s, type=%s",
-                        nvl(dependency.getGroupId()),
-                        nvl(dependency.getArtifactId()),
-                        nvl(dependency.getClassifier()),
-                        nvl(dependency.getType()));
-            }
+            String hint = "groupId=" + nvl(dependency.getGroupId())
```
elharo: This is even harder to read than what it replaced. I can't look at this code and figure out what it does. So three suggestions:

1. Break it up with multiple local variables
2. rename the nvl method
3. Add detailed  doc comment that specifies the format

XenoAmess: rework as you suggested. please tell me if this make it read easier.

impl/maven-impl/src/main/java/org/apache/maven/impl/model/DefaultModelValidator.java
```diff
@@ -2251,23 +2251,20 @@ public static SourceHint gav(String gav) {
 
         public static SourceHint dependencyManagementKey(Dependency dependency) {
             String hint;
-            if (dependency.getClassifier() == null
-                    || dependency.getClassifier().trim().isEmpty()) {
-                hint = String.format(
-                        "groupId=%s, artifactId=%s, type=%s",
-                        nvl(dependency.getGroupId()), nvl(dependency.getArtifactId()), nvl(dependency.getType()));
+            if (dependency.getClassifier() == null || dependency.getClassifier().isBlank()) {
+                hint = "groupId=" + valueToValueString(dependency.getGroupId())
```
elharo, line 2255: so this can really have groupId= rather than groupId="" ?
That seems strange

impl/maven-impl/src/main/java/org/apache/maven/impl/model/DefaultModelValidator.java
```diff
@@ -2251,23 +2251,20 @@ public static SourceHint gav(String gav) {
...
+                        + ", artifactId=" + valueToValueString(dependency.getArtifactId())
+                        + ", classifier=" + valueToValueString(dependency.getClassifier())
+                        + ", type=" + valueToValueString(dependency.getType());
             }
             return new SourceHint(hint, null); // DMK
         }
 
-        private static String nvl(String value) {
+        private static String valueToValueString(String value) {
```
elharo, line 2267: Maybe just quote as a method name
Is it OK to have groupId=""? instead?

impl/maven-impl/src/main/java/org/apache/maven/impl/model/DefaultModelValidator.java
```diff
@@ -2251,23 +2251,20 @@ public static SourceHint gav(String gav) {
 
         public static SourceHint dependencyManagementKey(Dependency dependency) {
```
elharo, line 2252: This needs javadoc

impl/maven-impl/src/main/java/org/apache/maven/impl/model/DefaultModelValidator.java
```diff
@@ -2251,23 +2251,20 @@ public static SourceHint gav(String gav) {
...
             String hint;
-            if (dependency.getClassifier() == null
-                    || dependency.getClassifier().trim().isEmpty()) {
-                hint = String.format(
-                        "groupId=%s, artifactId=%s, type=%s",
-                        nvl(dependency.getGroupId()), nvl(dependency.getArtifactId()), nvl(dependency.getType()));
+            if (dependency.getClassifier() == null || dependency.getClassifier().isBlank()) {
```
elharo, line 2254: This might work better with a StringBuilder that adds a piece at a time. Then there'd be less duplicate code.

XenoAmess: > This might work better with a StringBuilder that adds a piece at a time. Then there'd be less duplicate code.

yes, but it would make the codes harder to understand as you pointed out before...

impl/maven-impl/src/main/java/org/apache/maven/impl/model/DefaultModelValidator.java
XenoAmess: > so this can really have groupId= rather than groupId="" ? 

according to the original codes, yes.

> That seems strange

if you ask me for my ideas, yes, it is strange :)

gnodet (approved):
No objection, though the gain will be really small.  The code is a bit nicer though.
If we were to change the output format, I'll defer to 4.1.0.

gnodet:
> No objection, though the gain will be really small. The code is a bit nicer though. If we were to change the output format, I'll defer to 4.1.0.

@XenoAmess 

Actually, I retract ;-)
There are at dozens of calls to `SourceHint.dependencyManagementKey(d)` for the same dependency, and the results are actually rarely used.  This is very time consuming, and we pass a `Supplier<SourceHint>` instead, so that we lazily compute it only when needed. 
This could be done in a separate PR, but I think this would be the right place.

gnodet:
> > There are at dozens of calls to `SourceHint.dependencyManagementKey(d)` for the same dependency
> 
> yes, too many. this be the reason why I think this should be refined.
> 
> > we pass a Supplier instead
> 
> good idea! I'm afraid it would break bc though. anyway we havn't release 4.
> 
> ps: maybe it be better for SourceHint having a cache for the string it built to avoid duplicate calculation?(if one of them would be called for multiple times)(though I have no idea if it worth the memory storage.)

See https://github.com/apache/maven/pull/2518

gnodet:
Superseded by #2518

Issue #2487, opened 2025-06-16
Maven 3.9.10 duplicates projects in parallel build


alzimmermsft:
### Affected version

3.9.10

### Bug description

We have a build attempting to parallelize work, using `-T 1C`, while only building part of a larger set of projects, using `--pl`, which is now hitting an odd bug where the same project is built multiple times.

The following command is being ran (seen both on Linux in CI and on Window on my local machine):

```
mvn clean verify -f pom.xml --fail-at-end "-Dmaven.javadoc.skip=true" "-Dcodesnippet.skip=true" "-Dspotbugs.skip=true" "-Dcheckstyle.skip=true" "-Drevapi.skip=true" -DtrimStackTrace=false "-Dspotless.apply.skip=true" "-Dspotless.check.skip=true" -pl "com.azure:azure-core-http-vertx,com.azure:azure-core-tracing-opentelemetry-samples,com.azure:azure-core-perf,com.azure:perf-test-core,com.azure:azure-core" -DAZURE_TEST_DEBUG=false -Dparallel-test-playback=true -DcreateSourcesJar=false -T 1C
```

On the following POM file:

https://github.com/Azure/azure-sdk-for-java/blob/main/pom.xml

At some point in the build I see the following output:

![Image](https://github.com/user-attachments/assets/458c19a8-a5fd-4a50-9b86-80d719d754c5)

Where `azure-core-perf` is being built three times, and seeing 6/5 projects being built, which eventually results in the build failing.

Running the build without parallelization works, as well as reverting back to Mavan 3.9.6 (the other version I have on my local machine).

Additionally, what is odd, adding more projects to the project list doesn't run into this problem either:

```
[...]

PR #2489, opened 2025-06-17
Deduplicate filtered dependency graph

alzimmermsft:
Fixes https://github.com/apache/maven/issues/2487

Adds logic to de-duplicate the list of `MavenProject`s being returned by `FilteredProjectDependencyGraph.applyFilter` to prevent building the same project multiple times in a parallel, multi-module build.

A lengthy comment is added to the main change explaining a situation where the same project could be added multiple times, which I'll repeat here.

Given the following, simplified, aggregate POM where project E has a direct dependency on all of A, B, C, and D.

```xml
<modules>
  <module>A</module>
  <module>B</module>
  <module>C</module>
  <module>D</module>
  <module>E</module>
<modules>
```

If a parallel build was using `--pl A,C,E` if `FilteredProjectDependencyGraph.applyFilter` was called with the aggregate POM's `MavenProject`s (A, B, C, D, and E) and non-transitive and downstream dependencies were being filtered for, the following would happen.

1. A, C, and E would be added to the `filtered` result as they're in the whitelist.
2. B and D would be scanned for their downstream projects as they aren't in the whitelist and transitive is false.
3. The result of B and D's scanning would find project E as it depends on both projects.

This would result in project E being added to `filtered` three separate times. Once for being in the whitelist itself and twice more for being a downstream project of two projects not in the whitelist.

I also added a small bit of logic to short circuit being distincting `filtered` when it would effectively be a wasted no-op, when there is zero or one project in the list.

### DIFF
--- a/maven-core/src/main/java/org/apache/maven/graph/FilteredProjectDependencyGraph.java
+++ b/maven-core/src/main/java/org/apache/maven/graph/FilteredProjectDependencyGraph.java
@@ -149,7 +149,24 @@ class FilteredProjectDependencyGraph implements ProjectDependencyGraph {
                 filtered.addAll(upstream ? getUpstreamProjects(project, false) : getDownstreamProjects(project, false));
             }
         }
-        return filtered;
+        if (filtered.isEmpty() || filtered.size() == 1) {
+            // Optimization to skip streaming, distincting, and collecting to a new list when there is zero or one
+            // project, aka there can't be duplicates.
+            return filtered;
+        }
+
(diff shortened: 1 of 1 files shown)


slawekjaranowski (approved):

slawekjaranowski:
look ok for me @cstamas @gnodet please also look

We need do the same in master.

cstamas:
Thank you for contribution! :beers:

alzimmermsft:
Glad to help @cstamas!

If this is also an issue in master, I'm happy to file a PR against it as well.

olamy (approved):

PR #2493, opened 2025-06-18
Deduplicate filtered dependency graph

alzimmermsft:
Port of https://github.com/apache/maven/pull/2489 against master.

### DIFF
--- a/impl/maven-core/src/main/java/org/apache/maven/graph/FilteredProjectDependencyGraph.java
+++ b/impl/maven-core/src/main/java/org/apache/maven/graph/FilteredProjectDependencyGraph.java
@@ -25,6 +25,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.Objects;
 import java.util.concurrent.ConcurrentHashMap;
+import java.util.stream.Collectors;
 
 import org.apache.maven.execution.ProjectDependencyGraph;
 import org.apache.maven.project.MavenProject;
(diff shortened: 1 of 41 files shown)


impl/maven-core/src/main/java/org/apache/maven/graph/FilteredProjectDependencyGraph.java
```diff
@@ -123,7 +124,24 @@ private List<MavenProject> applyFilter(
...
+        // Given a scenario where there is an aggregate POM with modules A, B, C, D, and E and project E depends on
+        // A, B, C, and D. If the aggregate POM is being filtered for non-transitive and downstream dependencies where
+        // only A, C, and E are whitelisted duplicates will occur. When scanning projects A, C, and E, those will be
+        // added to 'filtered' as they are whitelisted. When scanning B and D, they are not whitelisted, and since
+        // transitive is false, their downstream dependencies will be added to 'filtered'. E is a downstream dependency
+        // of A, B, C, and D, so when scanning B and D, E will be added again 'filtered'.
+        //
+        // Without de-duplication, the final list would contain E three times, once for E being in the projects and
+        // whitelisted, and twice more for E being a downstream dependency of B and D.
+        return filtered.stream().distinct().collect(Collectors.toList());
```
alzimmermsft: Added and used import of `Collectors.toList()` here instead of `Stream.toList()` as they have mutability differences.

`Stream.toList()` returns an immutable list, akin to `Collections.unmodifiableList()`. 

`Stream.collect(Collectors.toList())` returns a mutable list, akin to `ArrayList.addAll()` or `Arrays.asList()`.

I'm uncertain whether mutability or immutability is preferred here. So, I went with a straight copy, but if immutability is preferred I'll change this to `Stream.toList()`.

alzimmermsft: FYI @cstamas

slawekjaranowski: if `Stream.toList()` can not be used, I would like to add such info also to code.

@alzimmermsft Did you try with immutable list here by executing ITs

alzimmermsft: I didn't, I just copied what I had in the other PR.

Will change to `Stream.toList()` and test locally. If it passes, I'll use it to match the other stream collections in the file.

alzimmermsft: @slawekjaranowski switched to `Stream.toList()` as the ITs executed without issue.

cstamas (approved):

PR #2504, opened 2025-06-21
Execute GitHub action - Java CI on maven-4.0.x branch

slawekjaranowski:
now we have disable action on branch ....

### DIFF
--- a/.github/workflows/maven.yml
+++ b/.github/workflows/maven.yml
@@ -19,9 +19,9 @@ name: Java CI
 
 on:
   push:
-    branches: [ master ]
+    branches: [ maven-4.0.x ]
   pull_request:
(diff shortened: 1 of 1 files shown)


Bukama (approved):
That needs to be documented in the guidelines.

Bukama:
> > That needs to be documented in the guidelines.
> 
> We don't have a guideline for multi branches project ....
> 
> We should know to: when create new branch after or before last release and so .... what should be update in new brach, what should be updated in master ...

That‘s why we should create on or add to the GitHub convention page

gnodet (approved):

slawekjaranowski:
@gnodet which milestone should I use ... 4.0.0-rc-5, 4.0.0?

PR #2506, opened 2025-06-24
Feature/improve cache

gnodet:


impl/maven-impl/src/main/java/org/apache/maven/impl/model/DefaultModelObjectPool.java
```diff
@@ -0,0 +1,331 @@
...
+    /**
+     * Gets the default reference type from system properties.
+     */
+    private static Cache.ReferenceType getDefaultReferenceType() {
+        try {
+            String referenceTypeProperty =
+                    System.getProperty(Constants.MAVEN_MODEL_PROCESSOR_REFERENCE_TYPE, Cache.ReferenceType.HARD.name());
+            return Cache.ReferenceType.valueOf(referenceTypeProperty.toUpperCase());
+        } catch (IllegalArgumentException e) {
+            System.err.println("Unknown default reference type, using HARD");
```
elharo: don't use println; log.debug operhaps

impl/maven-core/src/test/java/org/apache/maven/project/DefaultMavenProjectBuilderTest.java
```diff
@@ -345,79 +364,152 @@ void testActivatedProfileBySource() throws Exception {
...
+
+            // This demonstrates the cross-platform path behavior:
+            // - On Unix systems, paths use forward slashes (/)
+            // - On Windows systems, paths use backslashes (\)
+            // - The actual file system being used determines the separator
+            String actualLocation = location.getSource().getLocation();
+            String expectedPath = "pom-with-profiles" + separator + "pom.xml";
+
+            // Log the actual vs expected for debugging
+            System.out.println("=== Cross-Platform Path Test [" + fsName + "] ===");
```
elharo: make this part of the failure message instead

impl/maven-core/src/test/java/org/apache/maven/project/DefaultMavenProjectBuilderTest.java
```diff
@@ -345,79 +364,152 @@ void testActivatedProfileBySource() throws Exception {
...
+            assertThat(location.getColumnNumber(), greaterThan(0));
+            assertNotNull(location.getSource());
+            assertThat(location.getSource().getLocation(), containsString("pom-with-profiles/pom.xml"));
+
+            // This demonstrates the cross-platform path behavior for the POM file
+            String actualLocation = location.getSource().getLocation();
+            String expectedPath = "pom-with-profiles" + separator + "pom.xml";
+
+            // Log the actual vs expected for debugging
+            System.out.println("=== Cross-Platform Path Test [" + fsName + "] - External Profile ===");
```
elharo: move to failure message

impl/maven-impl/src/main/java/org/apache/maven/impl/model/DefaultModelObjectPool.java
```diff
@@ -0,0 +1,331 @@
...
+
+        // Check for per-type configuration first
+        String perTypeProperty = Constants.MAVEN_MODEL_PROCESSOR_REFERENCE_TYPE_PREFIX + className;
+        String perTypeValue = System.getProperty(perTypeProperty);
+
+        if (perTypeValue != null) {
+            try {
+                return Cache.ReferenceType.valueOf(perTypeValue.toUpperCase());
+            } catch (IllegalArgumentException e) {
+                System.err.println("Unknown reference type for " + className + ": " + perTypeValue + ", using default");
```
elharo: log.debug or ignore

impl/maven-impl/src/test/java/org/apache/maven/impl/cache/ReferenceTypeStatisticsIntegrationTest.java
```diff
@@ -0,0 +1,117 @@
...
+
+        // Capture the formatted output (not used in this test, but could be useful for future enhancements)
+
+        // Use reflection to call the private formatCacheStatistics method
+        try {
+            var method = DefaultRequestCache.class.getDeclaredMethod("formatCacheStatistics", CacheStatistics.class);
+            method.setAccessible(true);
+            String output = (String) method.invoke(null, statistics);
+
+            System.out.println("=== Enhanced Cache Statistics Output ===");
```
elharo: remove

impl/maven-impl/src/test/java/org/apache/maven/impl/cache/ReferenceTypeStatisticsIntegrationTest.java
```diff
@@ -0,0 +1,117 @@
...
+            assertTrue(output.contains("accesses"), "Should show access count");
+            assertTrue(output.contains("hit ratio"), "Should show hit ratio");
+
+            // Verify that different hit ratios are shown correctly
+            assertTrue(
+                    output.contains("66.7%") || output.contains("66.6%"), "Should show HARD/HARD hit ratio (~66.7%)");
+            assertTrue(output.contains("33.3%"), "Should show SOFT/WEAK hit ratio (33.3%)");
+            assertTrue(output.contains("0.0%"), "Should show WEAK/SOFT hit ratio (0.0%)");
+
+        } catch (Exception e) {
```
elharo: catch block not needed, just declare the exception

impl/maven-impl/src/test/java/org/apache/maven/impl/cache/ReferenceTypeStatisticsIntegrationTest.java
```diff
@@ -0,0 +1,117 @@
...
+            String output = (String) method.invoke(null, statistics);
+
+            System.out.println("=== Memory Pressure Analysis ===");
+            System.out.println(output);
+
+            // Should show high usage of hard references
+            assertTrue(output.contains("HARD/HARD:"), "Should show hard reference usage");
+            assertTrue(output.contains("1000 accesses"), "Should show high access count for hard references");
+
+        } catch (Exception e) {
```
elharo: catch block not needed

src/mdo/java/InputLocation.java
```diff
@@ -33,6 +33,28 @@ public class InputLocation implements Serializable, InputLocationTracker {
...
     private final Map<Object, InputLocation> locations;
 
+    private static final InputLocation EMPTY = new InputLocation(-1, -1);
+
+    public static InputLocation of() {
+        return EMPTY;
+    }
+
+    public static InputLocation of(int lineNumber, int columnNumber) {
+        return new InputLocation(lineNumber, columnNumber);
```
elharo: If these aren't being pooled, then a constructor is more custmary and obvious.

src/site/markdown/cache-configuration.md
```diff
@@ -0,0 +1,188 @@
...
+specific language governing permissions and limitations
+under the License.
+-->
+# Maven Cache Configuration Enhancement
+
+This document describes the enhanced cache configuration functionality in Maven's DefaultRequestCache.
+
+## Overview
+
+The DefaultRequestCache has been enhanced to support configurable reference types and cache scopes through user-defined selectors. This allows fine-grained control over caching behavior for different request types.
```
elharo: has been enhanced to support --> supports

src/site/markdown/cache-configuration.md
```diff
@@ -0,0 +1,188 @@
...
+
+## Key Features
+
+### 1. Early Return for ProtoSession
+- The `doCache` method now returns early when the request session is not a `Session` instance (e.g., `ProtoSession`)
+- This prevents caching attempts for non-session contexts
```
elharo: period

src/site/markdown/cache-configuration.md
```diff
@@ -0,0 +1,188 @@
...
+This document describes the enhanced cache configuration functionality in Maven's DefaultRequestCache.
+
+## Overview
+
+The DefaultRequestCache has been enhanced to support configurable reference types and cache scopes through user-defined selectors. This allows fine-grained control over caching behavior for different request types.
+
+## Key Features
+
+### 1. Early Return for ProtoSession
+- The `doCache` method now returns early when the request session is not a `Session` instance (e.g., `ProtoSession`)
```
elharo: period

src/site/markdown/cache-configuration.md
```diff
@@ -0,0 +1,188 @@
...
+- This prevents caching attempts for non-session contexts
+
+### 2. Configurable Cache Behavior
+- Cache scope and reference type can be configured via session user properties
```
elharo: period here and below

src/site/markdown/cache-configuration.md
```diff
@@ -0,0 +1,188 @@
...
+- Configuration uses CSS-like selectors to match request types
+- Supports parent-child request relationships
+
+### 3. Backward Compatibility
+- Existing `CacheMetadata` interface still supported
```
elharo: in general, write in present tense. The reader will want to know how this operates. That is, "this is what the cache does" rather than "this is how the cache has changed".

src/mdo/java/InputLocation.java
gnodet: The static factory methods are now a common pattern when building immutable objects: `Optional`, `Path`, `List`, `Set`, `Map`, `Stream`, `LocalDate`, `Duration`, `Period`, etc...

elharo: Some of those have reasons for doing this that simply don't apply here. This does nothing the existing constructor doesn't do, and just adds code and API complexity. When the basics work, stick to the basics.

gnodet: Of course, there are different reasons for using a common pattern.  Some of the above are interfaces, so they are clearly different.  However, the other ones are trivial, have a look at `Optional` for example.
```
    private Optional(T value) {
        this.value = value;
    }

    public static <T> Optional<T> of(T value) {
        return new Optional<>(Objects.requireNonNull(value));
    }
```
This is clearly the same pattern I want to reuse here for `InputSource` and `InputLocation`.
The fact that they currently have a constructor means we cannot plug in the `ModelObjectPool`, so we need an indirection, and the usual _static factory method_ pattern clearly provides it here.

_Effective Java, Third Edition, Chapter 2, Item 1_: Consider using factory methods instead of constructors
This is the _very first_ item of this book, and explains the various benefits of this pattern.

I know you don't like introducing things that are seen as _not necessary_.  But if we had those in the first place, I would not have had to introduce breaking changes in this PR: given we have public constructors, we'll have to deprecate them in 4.0 so that they can be made private/protected in 4.1.

In addition, I think having an homogeneous API is key for ease of use.  So I'd rather change all InputLocation/InputSource to use factory methods, even if some of them simply delegate to the constructors.

Fwiw, I intend to split this PR in two, because it's too big: one for the `ModelObjectProcessor`, and another one for the `Cache`.

elharo: Better yet, just don't deprecate the constructors and don't introduce factory methods. Effective Java lists *4* reasons to consider factory methods. Exactly *zero* of those apply here. In fact, rereading that chapter I notice that one of the reasons is so that constructors with different arguments have meaningful, dissimilar names. And yet here all four methods are named the undescriptive "of".

Plugging in ModelObjectPool could be a reason to use factory methods, but this PR isn't doing that.

elharo (changes requested):

src/mdo/java/InputLocation.java
gnodet: The main reason is coherence.  We need it for the model InputLocation/InputSource.  I agree we don't really need that abstraction for the other models which don't use ModelObjectPool.
But the templates `merger.vm`, `model.vm` and `stax-reader.vm` generates classes for all modello models. If we don't bring coherence in those two classes, we'll need to add another boolean to differentiate the models and use either constructors/factory methods.  I already extracted those 2 classes (InputSource/InputLocation) to avoid having big if/then/else in templates for the java classes, as those java classes are generated using VM templates, but the only change is the package name.
I don't think the pain is worth it here.

gnodet: Fwiw, I won't split unless really required, as `ModelObjectProcessor` does require `Cache`...

gnodet:
> Is this to be revived?

Could be.  I need to rebase on latest master and run the tests at https://github.com/gnodet/maven-bench to get proper results.

api/maven-api-core/src/main/java/org/apache/maven/api/Constants.java
```diff
@@ -662,5 +662,66 @@ public final class Constants {
...
+     * when the build completes.
+     *
+     * @since 4.1.0
+     */
+    @Config(type = "java.lang.Boolean", defaultValue = "false")
+    public static final String MAVEN_CACHE_STATS = "maven.cache.stats";
+
+    /**
+     * User property to configure separate reference types for cache keys and values.
+     * Format: "key:value" where key and value can be NONE, SOFT, WEAK, or HARD.
```
elharo: would it make sense to have separate propertied for keys and values? That's effectively what's happening here.

api/maven-api-model/src/main/java/org/apache/maven/api/model/ModelObjectProcessor.java
```diff
@@ -0,0 +1,118 @@
...
+                return object;
+            }
+        }
+
+        try {
+            ServiceLoader<ModelObjectProcessor> loader = ServiceLoader.load(ModelObjectProcessor.class);
+            for (ModelObjectProcessor processor : loader) {
+                return processor;
+            }
+        } catch (Exception e) {
```
elharo: can this be a more specific exception type?

impl/maven-impl/src/main/java/org/apache/maven/impl/cache/CacheSelectorParser.java
```diff
@@ -0,0 +1,202 @@
...
+        }
+
+        Matcher ruleMatcher = RULE_PATTERN.matcher(configString);
+        while (ruleMatcher.find()) {
+            try {
+                CacheSelector selector = parseRule(ruleMatcher);
+                if (selector != null) {
+                    selectors.add(selector);
+                }
+            } catch (Exception e) {
```
elharo: can this be a more specific type

impl/maven-impl/src/main/java/org/apache/maven/impl/cache/CacheSelectorParser.java
```diff
@@ -0,0 +1,202 @@
...
+    private static PartialCacheConfig parseProperties(String properties) {
+        CacheRetention scope = null;
+        Cache.ReferenceType referenceType = null;
+
+        Matcher propMatcher = PROPERTY_PATTERN.matcher(properties);
+        while (propMatcher.find()) {
+            String key = propMatcher.group(1);
+            String value = propMatcher.group(2);
+
+            switch (key.toLowerCase()) {
```
elharo: add Locale

impl/maven-impl/src/main/java/org/apache/maven/impl/cache/CacheSelectorParser.java
```diff
@@ -0,0 +1,202 @@
...
+
+        // Return partial configuration (null values are allowed)
+        return new PartialCacheConfig(scope, referenceType);
+    }
+
+    /**
+     * Parses a scope string into CacheRetention.
+     */
+    private static CacheRetention parseScope(String value) {
+        return switch (value.toLowerCase()) {
```
elharo: add locale

impl/maven-impl/src/main/java/org/apache/maven/impl/cache/CacheSelectorParser.java
```diff
@@ -0,0 +1,202 @@
...
+                yield CacheRetention.REQUEST_SCOPED;
+            }
+        };
+    }
+
+    /**
+     * Parses a reference type string into Cache.ReferenceType.
+     */
+    private static Cache.ReferenceType parseReferenceType(String value) {
+        return switch (value.toLowerCase()) {
```
elharo: add locale

impl/maven-impl/src/test/java/org/apache/maven/impl/cache/CacheStatisticsTest.java
```diff
@@ -0,0 +1,211 @@
...
+/**
+ * Test for cache statistics functionality with the improved cache architecture.
+ */
+class CacheStatisticsTest {
+
+    private CacheStatistics statistics;
+
+    @BeforeEach
+    void setUp() {
+        statistics = new CacheStatistics();
```
elharo: could just init the field

impl/maven-impl/src/test/java/org/apache/maven/impl/cache/CacheStatisticsTest.java
```diff
@@ -0,0 +1,211 @@
...
+
+        // Test property parsing behavior
+        System.setProperty(Constants.MAVEN_CACHE_STATS, "true");
+        String statsProperty = System.getProperty(Constants.MAVEN_CACHE_STATS);
+        boolean showStats = Boolean.parseBoolean(statsProperty);
+        assertTrue(showStats, "Cache stats should be enabled when property is true");
+
+        System.setProperty(Constants.MAVEN_CACHE_STATS, "false");
+        statsProperty = System.getProperty(Constants.MAVEN_CACHE_STATS);
+        showStats = Boolean.parseBoolean(statsProperty);
```
elharo: try not to reset local variables in tests; it's error prone. Use a new variable or a separate test method instead

impl/maven-impl/src/test/java/org/apache/maven/impl/cache/RefConcurrentMapTest.java
```diff
@@ -182,19 +226,80 @@ void shouldHandleSoftReferences() throws InterruptedException {
...
+
+        // Force garbage collection multiple times
+        for (int i = 0; i < 5; i++) {
+            System.gc();
+            Thread.sleep(50);
+            // Trigger cleanup by calling a method that calls expungeStaleEntries()
+            softMap.size();
+        }
+
+        // The map should eventually clean up the garbage collected entries
```
elharo, line 313: Can we move this to a separate test suite or IT or something? I really don't want to introduce known flaky tests

impl/maven-impl/src/test/java/org/apache/maven/impl/cache/ReferenceTypeStatisticsTest.java
```diff
@@ -0,0 +1,167 @@
...
+import static org.junit.jupiter.api.Assertions.assertNotNull;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+class ReferenceTypeStatisticsTest {
+
+    private CacheStatistics statistics;
+
+    @BeforeEach
+    void setUp() {
+        statistics = new CacheStatistics();
```
elharo: init the field

impl/maven-impl/src/test/java/org/apache/maven/impl/cache/ReferenceTypeStatisticsTest.java
```diff
@@ -0,0 +1,167 @@
...
+        statistics.recordCacheCreation("SOFT", "WEAK", CacheRetention.SESSION_SCOPED);
+        statistics.recordCacheCreation("HARD", "SOFT", CacheRetention.REQUEST_SCOPED);
+        statistics.recordCacheCreation("SOFT", "WEAK", CacheRetention.SESSION_SCOPED);
+
+        // Record cache accesses
+        statistics.recordCacheAccess("SOFT", "WEAK", true); // hit
+        statistics.recordCacheAccess("SOFT", "WEAK", false); // miss
+        statistics.recordCacheAccess("HARD", "SOFT", true); // hit
+
+        var refTypeStats = statistics.getReferenceTypeStatistics();
```
elharo: var has some surprising failure modes. An explicit type is safer

impl/maven-impl/src/test/java/org/apache/maven/impl/cache/ReferenceTypeStatisticsTest.java
```diff
@@ -0,0 +1,167 @@
...
+        }
+
+        // Simulate some soft reference usage
+        for (int i = 0; i < 100; i++) {
+            statistics.recordCacheAccess("SOFT", "SOFT", i % 2 == 0);
+        }
+
+        String output = DefaultRequestCache.formatCacheStatistics(statistics);
+
+        System.out.println("=== Memory Pressure Analysis ===");
```
elharo: don't print in non-failing tests

impl/maven-impl/src/test/java/org/apache/maven/impl/model/DefaultModelObjectPoolTest.java
```diff
@@ -0,0 +1,267 @@
...
+            assertSame(result1, result2);
+
+            // Non-dependency objects should not be pooled (pass through)
+            String str1 = "test";
+            String str2 = processor.process(str1);
+            assertSame(str1, str2); // Same instance because it's not pooled
+
+        } finally {
+            if (originalPooledTypes != null) {
+                System.setProperty(Constants.MAVEN_MODEL_PROCESSOR_POOLED_TYPES, originalPooledTypes);
```
elharo: Depending on system property setting can cause problems in parallel tests

impl/maven-impl/src/test/java/org/apache/maven/impl/cache/CacheStatisticsTest.java
gnodet: The `setUp` has a `@BeforeEach`, not a `@BeforeAll`, so that the `CacheStatistics` are cleaned between tests.

elharo: Yes, and you don't need that. JUnit very deliberately creates a new object for every test method. setUp methods are only needed when there's more to do than call a constructor to init an instance field.

cstamas (approved):

PR #2514, opened 2025-06-25
Fix StackOverflowError in InputLocation.hashCode() due to circular references

gnodet:
## Summary

Fixes a critical StackOverflowError in `InputLocation.hashCode()` that occurs when InputLocation objects contain circular references in their locations map.

## Problem

The `InputLocation` class can contain self-references in its `locations` map, which causes infinite recursion during `hashCode()` and `equals()` calculations, leading to StackOverflowError.

## Solution

- **Added safe `equals()` method** with `safeLocationsEquals()` helper that detects and handles self-references
- **Added safe `hashCode()` method** with `safeHash()` helper that skips hash calculation for self-references
- **Added cached hashCode** with volatile field for thread-safe performance optimization
- **Added Objects import** for proper hash code calculation

## Key Features

- **Circular Reference Detection**: Both `equals()` and `hashCode()` detect self-references and handle them safely
- **Performance Optimization**: Hash codes are cached to avoid repeated calculations
- **Thread Safety**: Uses volatile field for safe concurrent access
- **Backward Compatibility**: Normal operation without circular references continues to work correctly

## Testing

The fix resolves the failing test:
- `DefaultDependencyManagementImporterTest#testUpdateWithImportedFromMultiLevelImportedFromSetChanged`

## Impact

- **Risk**: Very Low - Only adds missing `equals()` and `hashCode()` methods with safe circular reference handling
- **Scope**: Isolated to `InputLocation.java` only
- **Compatibility**: Fully backward compatible

This is a critical bug fix that should be prioritized for inclusion.

---

### DIFF
--- a/api/maven-api-model/src/main/java/org/apache/maven/api/model/InputLocation.java
+++ b/api/maven-api-model/src/main/java/org/apache/maven/api/model/InputLocation.java
@@ -23,6 +23,7 @@ import java.util.Collection;
 import java.util.Collections;
 import java.util.LinkedHashMap;
 import java.util.Map;
+import java.util.Objects;
 
 /**
  * Represents the location of an element within a model source file.
@@ -39,6 +40,8 @@ public class InputLocation implements Serializable, InputLocationTracker {
     private final InputSource source;
     private final Map<Object, InputLocation> locations;
     private final InputLocation importedFrom;
+
+    private volatile int hashCode = 0; // Cached hashCode for performance
 
     public InputLocation(InputSource source) {
         this.lineNumber = -1;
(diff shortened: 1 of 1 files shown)


api/maven-api-model/src/main/java/org/apache/maven/api/model/InputLocation.java
```diff
@@ -40,6 +41,8 @@ public class InputLocation implements Serializable, InputLocationTracker {
     private final Map<Object, InputLocation> locations;
     private final InputLocation importedFrom;
 
+    private volatile int hashCode = 0; // Cached hashCode for performance
```
elharo: Is there a benchmark showing this actually matters? Otherwise, I'd prefer to not bother, mpost especially if this object is mutable

cstamas:
@elharo on GH PR == issue and we moved off JIRA, so unsure what you ask for.

elharo:
One more time PR != issue. They are **NOT** the same thing and do not have the same purpose. There is not even a 1:1 relationship between them. An issue can often require multiple PRs to fix. 

Issues *describe* a problem or concern.
PRs provide **code**

Prose and code are two different things, and we need both.

gnodet:
> > One more time PR != issue. They are **NOT** the same thing and do not have the same purpose. There is not even a 1:1 relationship between them. An issue can often require multiple PRs to fix.
> > Issues _describe_ a problem or concern. PRs provide **code**
> > Prose and code are two different things, and we need both.
> 
> We don't. While issues cannot contain code, PRs can contain prose. So, it's very true that if someone raises an issue, we'll need a PR to actually fix/implement it. But the opposite is NOT true. Everything that is in an issue can be in a PR. They are assigned to milestones, are part of release notes, etc... So we _do_ need an issue _when_ we don't start from a PR, else I don't see the point at all.

That's proven by the very fact we're discussing here on the PR. ;-)

elharo:
Issue tracking != release notes != PRs != commit history.

In a  well-maintained project, developers think about and communicate decisions in the relevant locations. They don't just write code. They write prose. And they do that in an organized, searchable system like the issue tracker that records and documents decisions. But many developers are uncomfortable with communication, so they try to avoid communicating. For instance, instead of thinking about a bug and laying out what the bug is, what its symptoms are, and how it can be verified, they just ship some code. This approach is fast and reckless, and leads to poor quality software.

Projects that practice good issue tracking as a separate step have better results. Over the years I've noticed a very high correlation between solid issue tracking and software quality. I suspect the relationship is near linear and causal. I've seen poor quality software that did use issue tracking, but I've never seen high quality software that didn't; and badly maintained issue trackers are often a signal of a project that's in trouble.

PR #2515, opened 2025-06-26
Bump resolver to 2.0.10

slawekjaranowski:


cstamas (approved):

its/core-it-suite/src/test/resources/.mvn/readme.txt
```diff
@@ -0,0 +1 @@
+ITs should not use project .mvn directory, so create an empty one here
```
gnodet, line 1: Can we move that to a different PR ? that looks unrelated, right ?

gnodet: Or repurpose this PR with this only change and use the dependabot for upgrading resolver version.

slawekjaranowski:
suppressed by:
 - #2544

PR #2528, opened 2025-06-29
Split user and system properties

gnodet:
## Summary

This PR configures Maven 4 to use IPC (Inter-Process Communication) locks instead of the default file-based locks for the resolver's named locking mechanism. This change provides significantly better performance, especially in multi-threaded build scenarios and when multiple Maven processes run simultaneously.

## Changes

- Modified `apache-maven/src/assembly/maven/conf/maven.properties` to set:
  - `aether.syncContext.named.factory = ipc` (use IPC locks)
  - Keep default `file-gav` name mapper for optimal lock granularity

## Background

The Maven resolver uses named locks for concurrent access control during artifact resolution. The current default uses file-based locks (`file-lock` factory) which can become a bottleneck in high-concurrency scenarios.

IPC locks provide superior performance characteristics:
- **Better concurrency**: Reduced lock contention between threads and processes
- **Faster resolution**: Lower overhead compared to file system operations
- **Multi-process efficiency**: Optimized for scenarios where multiple Maven instances run in parallel
- **Scalability**: Better performance scaling with increased parallelism
- **Efficient batching**: IPC can handle multiple named locks in a single call

## Why Not file-static Name Mapper?

Initially considered using `file-static` name mapper, but this would actually **reduce** throughput in multi-threaded builds because:
- `file-static` serializes all lock requests to a single static name
- This eliminates the benefits of parallel artifact resolution
- IPC locks are already optimized for handling multiple named locks efficiently

By keeping the default `file-gav` name mapper, we maintain proper lock granularity while benefiting from IPC's superior performance characteristics.

## Performance Benefits

According to the [Maven resolver documentation](https://maven.apache.org/resolver/configuration.html), IPC locks are designed for high-performance scenarios and provide significant improvements over file-based locking, especially when:
- Running multi-threaded builds (`-T` flag)
- Multiple Maven processes execute simultaneously
- High artifact resolution activity occurs
- CI/CD environments with parallel builds

## Compatibility

This change is backward compatible and does not affect build correctness, only performance. The IPC lock factory is available in Maven Resolver 2.0+ and provides the same locking semantics as file-based locks.

## Testing

This is a configuration change that affects the internal locking mechanism of the resolver. The change maintains the same locking semantics while improving performance characteristics.

## References

- [Maven Resolver Configuration Documentation](https://maven.apache.org/resolver/configuration.html)
- [Maven Artifact Resolver Named Locks using IPC](https://maven.apache.org/resolver/maven-resolver-named-locks-ipc/)
- [Named Locks Documentation](https://maven.apache.org/resolver/maven-resolver-named-locks/)

### DIFF
--- a/apache-maven/src/assembly/maven/conf/maven-user.properties
+++ b/apache-maven/src/assembly/maven/conf/maven-user.properties
@@ -27,8 +27,5 @@
 # Comma-separated list of files to include.
 # Each item may be enclosed in quotes to gracefully include spaces. Items are trimmed before being loaded.
 # If the first character of an item is a question mark, the load will silently fail if the file does not exist.
-# Note: the maven.properties will be removed in Maven 4.1.0 and only maven-user.properties will be loaded!
 ${includes} = ?"${maven.user.conf}/maven-user.properties", \
-              ?"${maven.user.conf}/maven.properties", \
-              ?"${maven.project.conf}/maven-user.properties", \
-              ?"${maven.project.conf}/maven.properties",
+              ?"${maven.project.conf}/maven-user.properties"


gnodet:
I haven't measured the exact performance difference with this patch, but I end up with lots of sleeps, even though I have a single JVM running Maven.  @cstamas thoughts?
```
java.lang.Thread.sleep(Native Method)
org.eclipse.aether.named.support.Retry.retry(Retry.java:76)
org.eclipse.aether.named.support.FileLockNamedLock.doLockExclusively(FileLockNamedLock.java:95)
org.eclipse.aether.named.support.NamedLockSupport.lockExclusively(NamedLockSupport.java:86)
org.eclipse.aether.internal.impl.synccontext.named.NamedLockFactoryAdapter$AdaptedLockSyncContext.acquire(NamedLockFactoryAdapter.java:232)
org.eclipse.aether.internal.impl.DefaultInstaller.install(DefaultInstaller.java:142)
org.eclipse.aether.internal.impl.DefaultInstaller.install(DefaultInstaller.java:107)
org.eclipse.aether.internal.impl.DefaultRepositorySystem.install(DefaultRepositorySystem.java:371)
org.apache.maven.plugins.install.InstallMojo.installProject(InstallMojo.java:177)
org.apache.maven.plugins.install.InstallMojo.execute(InstallMojo.java:132)
org.apache.maven.plugin.DefaultBuildPluginManager.executeMojo(DefaultBuildPluginManager.java:146)
org.apache.maven.lifecycle.internal.MojoExecutor.doExecute2(MojoExecutor.java:339)
org.apache.maven.lifecycle.internal.MojoExecutor.doExecute(MojoExecutor.java:310)
org.apache.maven.lifecycle.internal.MojoExecutor.execute(MojoExecutor.java:214)
org.apache.maven.lifecycle.internal.MojoExecutor.execute(MojoExecutor.java:179)
org.apache.maven.lifecycle.internal.MojoExecutor$1.run(MojoExecutor.java:168)
org.apache.maven.plugin.DefaultMojosExecutionStrategy.execute(DefaultMojosExecutionStrategy.java:39)
org.apache.maven.lifecycle.internal.MojoExecutor.execute(MojoExecutor.java:165)
org.apache.maven.lifecycle.internal.LifecycleModuleBuilder.buildProject(LifecycleModuleBuilder.java:110)
org.apache.maven.lifecycle.internal.builder.multithreaded.MultiThreadedBuilder.lambda$createBuildCallable$1(MultiThreadedBuilder.java:191)
java.util.concurrent.FutureTask.run(FutureTask.java)
java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:539)
java.util.concurrent.FutureTask.run(FutureTask.java:264)
java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1136)
java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:635)
java.lang.Thread.run(Thread.java:840)
```

gnodet:
> I haven't measured the exact performance difference with this patch, but I end up with lots of sleeps, even though I have a single JVM running Maven. @cstamas thoughts?

I see we can configure the sleep delay using `aether.named.file-lock.sleepMillis`.   However, [the one on the `FileLockNamedLock` cannot](https://github.com/apache/maven-resolver/blob/089796ba66d6f5f1a9d4fa7f797f613bae4f5373/maven-resolver-named-locks/src/main/java/org/eclipse/aether/named/support/FileLockNamedLock.java#L52)

gnodet:
> > I haven't measured the exact performance difference with this patch, but I end up with lots of sleeps, even though I have a single JVM running Maven. @cstamas thoughts?
> 
> I see we can configure the sleep delay using `aether.named.file-lock.sleepMillis`. However, [the one on the `FileLockNamedLock` cannot](https://github.com/apache/maven-resolver/blob/089796ba66d6f5f1a9d4fa7f797f613bae4f5373/maven-resolver-named-locks/src/main/java/org/eclipse/aether/named/support/FileLockNamedLock.java#L52)

The IPC lock factory is actually much faster... so we could try using that one.

cstamas:
IPC locking (same as Redisson or Hazelcast) is NOT pulled in transitively, needs to be added to `apache-maven` module to put it into `/lib`.

impl/maven-impl/src/main/java/org/apache/maven/impl/model/DefaultModelBuilder.java
```diff
@@ -1402,7 +1402,7 @@ Model doReadFileModel() throws ModelBuilderException {
                 } else {
                     properties.putAll(model.getProperties());
                 }
-                properties.putAll(session.getUserProperties());
+                properties.putAll(session.getEffectiveProperties());
```
cstamas, line 1405: @gnodet this one here is wonky, pls doublecheck

gnodet: That sounds wrong.  If we want system properties, they should be added before model properties I think.

PR #2530, opened 2025-06-30
Fix MavenProject#getPlugin(String) performances

gnodet:


cstamas:
RAO is down for HW migration https://status.apache.org/

impl/maven-core/src/main/java/org/apache/maven/project/MavenProject.java
```diff
@@ -1240,7 +1240,8 @@ public String getDefaultGoal() {
     }
 
     public Plugin getPlugin(String pluginKey) {
-        return getBuild().getPluginsAsMap().get(pluginKey);
+        var plugin = getBuild().getDelegate().getPluginsAsMap().get(pluginKey);
```
cstamas: nit: we agreed to avoid `var`, and it is in this case `Plugin`?

impl/maven-core/src/test/java/org/apache/maven/project/PluginConnectionSimpleTest.java
```diff
@@ -0,0 +1,119 @@
...
+import org.apache.maven.model.Model;
+import org.apache.maven.model.Plugin;
+import org.junit.jupiter.api.Test;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertNotNull;
+
+/**
+ * Simple test to verify that Plugin objects returned by MavenProject.getPlugin() are connected to the project model.
+ * This test specifically verifies the fix for the issue where getPlugin() was returning disconnected Plugin objects.
```
elharo: Include a URL for the issue.

impl/maven-core/src/main/java/org/apache/maven/project/MavenProject.java
```diff
@@ -1240,7 +1240,9 @@ public String getDefaultGoal() {
     }
 
     public Plugin getPlugin(String pluginKey) {
-        return getBuild().getPluginsAsMap().get(pluginKey);
+        org.apache.maven.api.model.Plugin plugin =
```
elharo, line 1243: why do we have both  org.apache.maven.model.Plugin and  org.apache.maven.api.model.Plugin? I'm not sure, but this might be a symptom of too many packages, over-abstraction, or both. Do we need multiple plugin types?

elharo (approved):

impl/maven-core/src/main/java/org/apache/maven/project/MavenProject.java
gnodet: We're translating from the Maven 3 api to the Maven 4 api here.  `org.apache.maven.model.Plugin` is the Maven 3 api, with mutable objects.  `org.apache.maven.api.model.Plugin` is the (now underlying) Maven 4 api with immutable objects.
This is somewhat temporary, the plan is to get away with the Maven 3 api in Maven 5.  But we need it to support Maven 3 plugins in Maven 4 and not break everything.

gnodet:

| Status | Branch | Result |
|:------:|:------:|:------|
|✅|maven-4.0.x|[](https://github.com/apache/maven/pull/2573)|

### Questions ?

PR #2548, opened 2025-07-02
Split system and user properties from maven.properties

gnodet:


cstamas (approved):

impl/maven-cli/src/main/java/org/apache/maven/cling/invoker/BaseParser.java
```diff
@@ -393,6 +395,31 @@ protected Map<String, String> populateSystemProperties(LocalContext context) {
         String mavenBuildVersion = CLIReportingUtils.createMavenVersionString(buildProperties);
         systemProperties.setProperty(Constants.MAVEN_BUILD_VERSION, mavenBuildVersion);
 
+        Path mavenConf;
+        if (systemProperties.getProperty(Constants.MAVEN_INSTALLATION_CONF) != null) {
+            mavenConf = context.installationDirectory.resolve(
+                    systemProperties.getProperty(Constants.MAVEN_INSTALLATION_CONF));
+        } else if (systemProperties.getProperty("maven.conf") != null) {
```
gnodet, line 402: Do we need all those hardcoded defaults ? Cling is new code, I'm not sure it's worth trying to mimic completely the old behavior... @cstamas ?

PR #2574, opened 2025-07-04
Augment version range resolution used repositories

cstamas:
Ability to control which repositories to be queried to resolve ranges. Introduced new property:

`maven.versionRangeResolutionNature` that is used in VersionRangeResolver, and means:
* not set; behave as before (default)
* "auto" string; will check lower and upper bounds, and if any is snapshot, will querty snapshot reposes otherwise not
* any valid value of resolver Metadata.Nature enum; then will use that

Fixes: #2558 

Note: spotless insisted on reformatting the unrelated class, unsure why? and now CI fails due formatting???

### DIFF
--- a/maven-resolver-provider/src/main/java/org/apache/maven/repository/internal/DefaultVersionRangeResolver.java
+++ b/maven-resolver-provider/src/main/java/org/apache/maven/repository/internal/DefaultVersionRangeResolver.java
@@ -71,14 +71,15 @@ import org.eclipse.aether.version.VersionScheme;
 @Singleton
 public class DefaultVersionRangeResolver implements VersionRangeResolver, Service {
     /**
-     * User property for version range handling. It may contain values of {@link Metadata.Nature} enum, or
-     * value "auto" to decide based on range: if any of the boundary version is snapshot, snapshots will be
-     * allowed, otherwise not. Default (unset) is "behave as before", query all repositories available in scope,
-     * so even snapshot repositories.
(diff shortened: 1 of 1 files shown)


maven-resolver-provider/src/main/java/org/apache/maven/repository/internal/DefaultVersionRangeResolver.java
```diff
@@ -68,6 +70,15 @@
...
 public class DefaultVersionRangeResolver implements VersionRangeResolver, Service {
+    /**
+     * User property for version range handling. It may contain values of {@link Metadata.Nature} enum, or
+     * value "auto" to decide based on range: if any of the boundary version is snapshot, snapshots will be
+     * allowed, otherwise not. Default (unset) is "behave as before", query all repositories available in scope,
+     * so even snapshot repositories.
+     *
+     * @since 3.9.11
+     */
+    public static final String MAVEN_VERSION_RANGE_HANDLING = "maven.versionRangeResolutionNature";
```
gnodet: Usually, the constant name matches the value...

gnodet (changes requested):

maven-resolver-provider/src/main/java/org/apache/maven/repository/internal/DefaultVersionRangeResolver.java
cstamas: Fixed, also improved javadoc

gnodet (approved):

slawekjaranowski (approved):

PR #10897, opened 2025-07-08
Pin GitHub action versions by hash

slawekjaranowski:
From security reason we should use a hash for GitHub action versions


elharo (commented):
did you open this PR twice?

cstamas (approved):

Bukama (approved):

PR #10899, opened 2025-07-09
add mavenVersion to error message when rejecting modelVersions

### LINKED ISSUE #10891: improve the wording of the error message for modelVersion mismatch

The message generated when the modelVersion is too high for the currently used maven version is a bit confusing at the moment, because the maven version and the modelVersion are quite similar.

Only after some experimentation and reading the source of `DefaultModelValidator.java` the meaning is clear.
The fact that the values array (`[4.0.0]`) is a singleton array for all `3.x.y` maven versions does not help.

I propose the wording 

```java
"addViolation(
                    problems,
                    Severity.FATAL,
                    Version.V20,
                    "modelVersion",
                    null,
                    "of '" + string + "' is newer than the versions supported by this version of Maven. The supported versions are: " + values
                            + ". Building this project requires a newer version of Maven.",
                    tracker);
``` 

It would be helpful to add the actual Maven version as well, but that needs an additional parameter or a known method to obtain the version of this Maven.

homberghp:
fixes #10891.

In this fix, the mavenVersion is clearly distinguished from the modelVersion, to avoid confusion and to better identify the mavenVersion in action.

In some IDEs, in Apache NetBeans, in particular, the bundled Maven version is used to validate the pom file. At the moment, (2025-07-09, NetBeans 26), this is maven-3.9.10. This impairs experiments with the combination of Maven, for instance maven 4.0.0, and NetBeans, and makes it impossible to work on feautures and improvements in that combination.

### DIFF
--- a/maven-model-builder/src/main/java/org/apache/maven/model/validation/DefaultModelValidator.java
+++ b/maven-model-builder/src/main/java/org/apache/maven/model/validation/DefaultModelValidator.java
@@ -23,6 +23,7 @@ import javax.inject.Named;
 import javax.inject.Singleton;
 
 import java.io.File;
+import java.io.IOException;
 import java.io.InputStream;
 import java.util.Arrays;
 import java.util.Deque;
@@ -1757,7 +1758,7 @@ public class DefaultModelValidator implements ModelValidator {
             if (resourceAsStream != null) {
                 properties.load(resourceAsStream);
             }
-        } catch (Exception e) {
+        } catch (IOException e) {
             System.err.println("Unable determine version from JAR file: " + e.getMessage());
         }
 


maven-model-builder/src/main/java/org/apache/maven/model/validation/DefaultModelValidator.java
```diff
@@ -1489,8 +1491,10 @@ private boolean validateModelVersion(
                     Version.V20,
                     "modelVersion",
                     null,
-                    "of '" + string + "' is newer than the versions supported by this version of Maven: " + values
-                            + ". Building this project requires a newer version of Maven.",
+                    "of '" + string + "' is newer than the versions supported by this Maven version("
```
desruisseaux: Is a space missing before the parenthesis?

cstamas:
```suggestion
                    "of '" + string + "' is newer than the versions supported by this Maven version ("
```

maven-model-builder/src/main/java/org/apache/maven/model/validation/DefaultModelValidator.java
```diff
@@ -1489,8 +1491,10 @@ private boolean validateModelVersion(
...
+                            + getMavenVersion()
+                            + ").\n Supported modelVersions are: " + values
+                            + ".\n Building this project requires a newer version of Maven.",
```
desruisseaux: I do not know the details of how the message is used, but I presume that `\n` should be `System.lineSeparator()`. Maybe there is no space at the beginning of the new line?

maven-model-builder/src/main/java/org/apache/maven/model/validation/DefaultModelValidator.java
```diff
@@ -1733,4 +1737,20 @@ private static Severity getSeverity(int validationLevel, int errorThreshold) {
             return Severity.ERROR;
         }
     }
+
+    private String getMavenVersion() {
```
cstamas, line 1752: ~~Nope, this is a managed component. Why not inject the RT instead?~~

cstamas: See https://github.com/apache/maven/blob/a8a44dc0ad3b950ce7e265c9337b2eb9aee55ff2/maven-core/src/main/java/org/apache/maven/rtinfo/RuntimeInformation.java

maven-model-builder/src/main/java/org/apache/maven/model/validation/DefaultModelValidator.java
```diff
@@ -1733,4 +1737,20 @@ private static Severity getSeverity(int validationLevel, int errorThreshold) {
...
+        Properties properties = new Properties();
```
desruisseaux, line 1753: I do not know this part of Maven API well, but it seems that the Maven version is provided by `org.apache.maven.rtinfo.RuntimeInformation`. But I do not know how to get an instance of `RuntimeInformation` from `ModelValidator` of Maven 3.

cstamas: Simple: DefaultValidator is a JSR300 component and uses ctor injecton, just add to ctor RuntimeInformation as well and it will be injected (as RuntimeInformation is also a managed component).

maven-model-builder/src/main/java/org/apache/maven/model/validation/DefaultModelValidator.java
homberghp: I tried that, but that leads to a circular dependencies in the poms.

maven-model-builder/src/main/java/org/apache/maven/model/validation/DefaultModelValidator.java
homberghp: @cstamas 
Let me try that.
Is that not the same circular dependency problem, since 

```shell
[ERROR] [ERROR] The projects in the reactor contain a cyclic reference: Edge between 'Vertex{label='org.apache.maven:maven-core:3.9.11-SNAPSHOT'}' and 'Vertex{label='org.apache.maven:maven-model-builder:3.9.11-SNAPSHOT'}' introduces to cycle in the graph org.apache.maven:maven-model-builder:3.9.11-SNAPSHOT --> org.apache.maven:maven-core:3.9.11-SNAPSHOT --> org.apache.maven:maven-model-builder:3.9.11-SNAPSHOT @ 
[ERROR] The projects in the reactor contain a cyclic reference: Edge between 'Vertex{label='org.apache.maven:maven-core:3.9.11-SNAPSHOT'}' and 'Vertex{label='org.apache.maven:maven-model-builder:3.9.11-SNAPSHOT'}' introduces to cycle in the graph org.apache.maven:maven-model-builder:3.9.11-SNAPSHOT --> org.apache.maven:maven-core:3.9.11-SNAPSHOT --> org.apache.maven:maven-model-builder:3.9.11-SNAPSHOT -> [Help 1]
[ERROR] 
[ERROR] To see the full stack trace of the errors, re-run Maven with the -e switch.
[ERROR] Re-run Maven using the -X switch to enable full debug logging.
[ERROR] 
[ERROR] [Help 1] http://cwiki.apache.org/confluence/display/MAVEN/ProjectCycleException
✘
```

Let me try with **compat**.

homberghp: @cstamas 
compat does not help either. Both depend on model builder and having model builder depend on them creates a cycle.

maven-model-builder/src/main/java/org/apache/maven/model/validation/DefaultModelValidator.java
cstamas: oh true, i forgot that RT is in core and model validator is in model builder.... 

Then ignore my "nope" for now.

desruisseaux: May be worth to add a comment saying that this method exists because of circular dependencies. That way, if we organize Maven modules differently in the future, we can revisit if we still need that method.

homberghp:
The 3.9.x line, and most likely anything before that, lacks a proper API part, AFAIK. This is repaired in maven 4.
That may also be the reason that the integration of the Maven 3 line in NetBeans is way more complex than it should have been.

In later versions of this improvement, one would of course use the RuntimeInformation interface, but in the 3.9 line that is not properly reachable.

maven-model-builder/src/main/java/org/apache/maven/model/validation/DefaultModelValidator.java
```diff
@@ -1733,4 +1746,21 @@ private static Severity getSeverity(int validationLevel, int errorThreshold) {
...
+                getClass().getResourceAsStream("/org/apache/maven/messages/build.properties")) {
+
+            if (resourceAsStream != null) {
+                properties.load(resourceAsStream);
+            }
+        } catch (Exception e) {
+            System.err.println("Unable determine version from JAR file: " + e.getMessage());
+        }
+
+        return properties.getProperty("version");
```
slawekjaranowski, line 1765: ~As `DefaultModelValidator` is singleton we can cache this value in filed.~

maven-model-builder/src/main/java/org/apache/maven/model/validation/DefaultModelValidator.java
```diff
@@ -1733,4 +1746,21 @@ private static Severity getSeverity(int validationLevel, int errorThreshold) {
...
+    private String getMavenVersion() {
+        Properties properties = new Properties();
+
+        try (InputStream resourceAsStream =
+                getClass().getResourceAsStream("/org/apache/maven/messages/build.properties")) {
+
+            if (resourceAsStream != null) {
+                properties.load(resourceAsStream);
+            }
+        } catch (Exception e) {
```
slawekjaranowski: `Exception` -> `IOException`

maven-model-builder/src/main/java/org/apache/maven/model/validation/DefaultModelValidator.java
slawekjaranowski: ok, method is used only during error message generating, so not necessary to cache

maven-model-builder/src/main/java/org/apache/maven/model/validation/DefaultModelValidator.java
gnodet: I'm fine with the hack in 3.9.x branch.  We need to use `Session#getMavenVersion()` when porting to 4.0.x branch.

maven-model-builder/src/main/java/org/apache/maven/model/validation/DefaultModelValidator.java
gnodet: > May be worth to add a comment saying that this method exists because of circular dependencies. That way, if we organize Maven modules differently in the future, we can revisit if we still need that method.

It should not be needed with 4.0.

homberghp:
This PR is less urgent now, at least for the NetBeans team. They deal with the issue on their side for the moment, as suggested by Neil C Smith of the NetBeans team.

PR #10903, opened 2025-07-10
Implement Injector.dispose() and add cleanup test

arturobernalg:
Dispose() clears all internal state including bindings, scopes, singleton caches, and loaded URLs Add unit test verifying that injector no longer provides disposed bindings

- [X ] Write unit tests that match behavioral changes, where the tests fail if the changes to the runtime are not applied.
- [ X] You have run the [Core IT][core-its] successfully.

### DIFF
--- a/impl/maven-di/src/main/java/org/apache/maven/di/Injector.java
+++ b/impl/maven-di/src/main/java/org/apache/maven/di/Injector.java
@@ -161,4 +161,16 @@ public interface Injector {
      */
     @Nonnull
     <T> T getInstance(@Nonnull Key<T> key);
+
+    /**
+     * Disposes this Injector, clearing all internal state (bindings, caches, scopes, etc.).
+     * After calling this, the Injector should not be used again.
+     * @since 4.1
+     */
+    default void dispose() {
+        // delegate to the implementation
+        if (this instanceof InjectorImpl) {
+            ((InjectorImpl) this).dispose();
+        }
+    }
 }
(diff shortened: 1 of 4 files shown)


impl/maven-di/src/main/java/org/apache/maven/di/Injector.java
```diff
@@ -161,4 +161,16 @@ static Injector create() {
...
     <T> T getInstance(@Nonnull Key<T> key);
+
+    /**
+     * Disposes this Injector, clearing all internal state (bindings, caches, scopes, etc.).
+     * After calling this, the Injector should not be used again.
+     * @since 4.1
+     */
+    default void dispose() {
+        // delegate to the implementation
+        if (this instanceof InjectorImpl) {
```
gnodet, line 172: I think we can use `if (this instanceof InjectorImpl impl)` and avoid the cast below.

gnodet (approved):
LGTM, thx !
This looks more like a bug fix though, and I think we may want to backport that to maven-4.0.x branch.  In that case, the `@since` tags will have to be adjusted.

impl/maven-di/src/main/java/org/apache/maven/di/Injector.java
arturobernalg: I used the explicit cast intentionally: it keeps the delegation to `InjectorImpl.dispose()` obvious at first glance, and it still compiles on older JDKs without pattern-matching. Happy to switch to the newer syntax if we want to modernise, but the explicit cast was just for clarity.

PR #10909, opened 2025-07-12
Port the bug fixes identified when using that class in Maven clean and compiler plugin.

desruisseaux:
This is a port of the following pull requests:

* https://github.com/apache/maven-compiler-plugin/pull/945
* https://github.com/apache/maven-clean-plugin/pull/266

After this fix is released, the `PathSelector` class of the Maven clean and compiler plugins can be removed.

# Question
Are `maven-clean-plugin` and `maven-compiler-plugin` allowed to use directly a class from `maven-impl` instead of `maven-api`? If not, we may need to move `PathSelector` in another package or module. But where?

### DIFF
--- a/impl/maven-impl/src/main/java/org/apache/maven/impl/PathSelector.java
+++ b/impl/maven-impl/src/main/java/org/apache/maven/impl/PathSelector.java
@@ -28,6 +28,7 @@ import java.util.Collection;
 import java.util.Iterator;
 import java.util.LinkedHashSet;
 import java.util.List;
+import java.util.Objects;
 import java.util.Set;
 
 /**
@@ -168,6 +169,7 @@ public class PathSelector implements PathMatcher {
     /**
      * String representations of the normalized include filters.
      * Each pattern shall be prefixed by its syntax, which is {@value #DEFAULT_SYNTAX} by default.
+     * An empty array means to include all files.
      *
      * @see #toString()
      */
@@ -175,7 +177,8 @@ public class PathSelector implements PathMatcher {
 
     /**
      * String representations of the normalized exclude filters.
-     * Each pattern shall be prefixed by its syntax, which is {@value #DEFAULT_SYNTAX} by default.
+     * Each pattern shall be prefixed by its syntax. If no syntax is specified,
+     * the default is a Maven 3 syntax similar, but not identical, to {@value #DEFAULT_SYNTAX}.
      * This array may be longer or shorter than the user-supplied excludes, depending on whether
      * default excludes have been added and whether some unnecessary excludes have been omitted.
      *
@@ -185,6 +188,7 @@ public class PathSelector implements PathMatcher {
 
     /**
      * The matcher for includes. The length of this array is equal to {@link #includePatterns} array length.
+     * An empty array means to include all files.
      */
     private final PathMatcher[] includes;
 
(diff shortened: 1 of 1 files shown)


impl/maven-impl/src/main/java/org/apache/maven/impl/PathSelector.java
```diff
@@ -225,11 +237,12 @@ public PathSelector(
         includePatterns = normalizePatterns(includes, false);
         excludePatterns = normalizePatterns(effectiveExcludes(excludes, includePatterns, useDefaultExcludes), true);
         baseDirectory = directory;
-        FileSystem system = directory.getFileSystem();
-        this.includes = matchers(system, includePatterns);
-        this.excludes = matchers(system, excludePatterns);
-        dirIncludes = matchers(system, directoryPatterns(includePatterns, false));
-        dirExcludes = matchers(system, directoryPatterns(excludePatterns, true));
+        FileSystem fs = directory.getFileSystem();
```
elharo, line 240: fs --> fileSystem

impl/maven-impl/src/main/java/org/apache/maven/impl/PathSelector.java
```diff
@@ -516,6 +533,21 @@ private static String[] directoryPatterns(final String[] patterns, final boolean
         return simplify(directories, excludes);
     }
 
+    /**
+     * Returns {@code true} if at least one pattern requires path to be relativized before to be matched.
```
elharo: to be --> being

impl/maven-impl/src/main/java/org/apache/maven/impl/PathSelector.java
```diff
@@ -528,12 +560,20 @@ private static PathMatcher[] matchers(final FileSystem fs, final String[] patter
         return matchers;
     }
 
+    /**
+     * {@return whether there is no include or exclude filters}.
```
elharo: is --> are

impl/maven-impl/src/main/java/org/apache/maven/impl/PathSelector.java
```diff
@@ -528,12 +560,20 @@ private static PathMatcher[] matchers(final FileSystem fs, final String[] patter
...
+     * In such case, this {@code PathSelector} instance should be ignored.
+     */
+    public boolean isEmpty() {
+        return (includes.length | excludes.length) == 0;
```
elharo: I'm not sure this is right, and even if it is, it's hard to think about bitwise operations. Prefer a simple

if includes.length == 0 && excludes.length == 0

gnodet:
**Update**: I've modified PR #10923 to avoid conflicts! 🚀

After reviewing your PR more carefully, I realized you already include the PathSelector fix needed for exclude-only patterns (`if (includes.length == 0) { return excludes; }`). 

**Changes made to PR #10923:**
- ✅ **Removed PathSelector changes** to avoid conflicts
- ✅ **Kept the API method** that plugins need
- ✅ **Updated tests** to document dependency on your PR

**Result:**
- **Your PR #10909**: Provides the PathSelector fixes
- **My PR #10923**: Provides the missing `createExcludeOnlyMatcher` API method
- **Together**: Complete solution for exclude-only functionality
- **Separately**: No conflicts, can be merged independently

This way we avoid duplicating your excellent work while still providing the API method that plugins will need. Both PRs complement each other perfectly! 👍

Thanks for the great work on identifying and fixing these PathSelector issues!

gnodet:
I raised #10935 as I could not push to this branch to resolve the conflict.

PR #10925, opened 2025-07-16
Add skipMavenRc to ExecutorRequest and use it in ITs

slawekjaranowski:
ITs should be isolated from the executing system, so:

- skip mavenrc in forked mode
- exclude MAVEN_ARGS and MAVEN_OPTS from test environments


impl/maven-executor/src/main/java/org/apache/maven/api/cli/ExecutorRequest.java
```diff
@@ -144,6 +144,13 @@ public interface ExecutorRequest {
      */
     Optional<OutputStream> stdErr();
 
+    /**
+     * Indicate if {@code ~/.mavenrc} should be skipped during execution.
+     * <br/>
```
gnodet:
```suggestion
     * <p>
 ```

File: impl/maven-executor/src/main/java/org/apache/maven/api/cli/ExecutorRequest.java [hunk 2]
```diff
@@ -333,6 +345,11 @@ public Builder stdErr(OutputStream stdErr) {
             return this;
         }
 
+        public Builder skipMavenRc(boolean skipMavenRc) {
```
gnodet, line 349:
```suggestion
        @Nonnull
        public Builder skipMavenRc(boolean skipMavenRc) {
```

gnodet (approved):
LGTM, two minor suggestions

PR #10935, opened 2025-07-18
Port the bug fixes identified when using that class in Maven clean and compiler plugin

gnodet:
- **Port the bug fixes identified when using that class in Maven clean and compiler plugin.**
- **Apply corrections for comments posted on the pull request.**
- **Rename fs -> fileSystem**


gnodet:
Resolves conflict in #10909

PR #10938, opened 2025-07-18
Consolidate ArtifactDescriptorReaderDelegate into DefaultArtifactDescriptorReader

gnodet:
## Summary

This PR consolidates the `ArtifactDescriptorReaderDelegate` functionality directly into `DefaultArtifactDescriptorReader` in the main implementation (`impl/maven-impl`), while preserving the original delegate pattern in the deprecated compat module for backward compatibility.

## Background

The `ArtifactDescriptorReaderDelegate` class was marked as "work in progress" and could be "changed or removed without notice" since version 3.2.4. The current implementation uses a delegate pattern with an unused extension mechanism through session config properties that appears to have no practical usage.

### Main Implementation (`impl/maven-impl`)

- **Consolidated functionality**: Moved all methods from `ArtifactDescriptorReaderDelegate` as private methods in `DefaultArtifactDescriptorReader`
- **Removed delegate pattern**: Eliminated the delegate field and its initialization
- **Simplified extension mechanism**: Removed the unused session config properties lookup for custom delegates
- **Maintained functionality**: All existing behavior is preserved for normal usage
- **Deleted**: `ArtifactDescriptorReaderDelegate.java`

### Compat Module (unchanged)

- **Preserved original pattern**: The deprecated compat module retains the original delegate pattern
- **Backward compatibility**: Existing code using the compat module continues to work
- **Deprecation path**: Since the compat module is already deprecated for removal in 4.0.0, this provides a clean migration path

## Benefits

1. **Cleaner codebase**: Removes unnecessary abstraction that was marked as work-in-progress
2. **Better encapsulation**: Conversion methods are now private implementation details
3. **Simplified maintenance**: Fewer classes to maintain in the main implementation
4. **Preserved compatibility**: No breaking changes for normal usage
5. **Clear migration path**: Users can migrate from compat to main implementation when ready

## Backward Compatibility

This change maintains backward compatibility for normal usage:
- The public API of `DefaultArtifactDescriptorReader` remains unchanged
- All functionality is identical, just moved from a separate class to private methods
- The only potential breaking change is for code using the undocumented extension mechanism through session config properties, which appears to be unused

## Future Extension Points

If a proper extension point is needed in the future, it can be designed more explicitly with:
- Proper interfaces and contracts
- Clear documentation
- Well-defined extension mechanisms
- Proper testing and validation

This consolidation removes the work-in-progress delegate pattern and provides a cleaner foundation for any future extension needs.

---


impl/maven-impl/src/main/java/org/apache/maven/impl/resolver/DefaultArtifactDescriptorReader.java
```diff
@@ -331,4 +339,99 @@ private int getPolicy(RepositorySystemSession session, Artifact a, ArtifactDescr
         }
         return policy.getPolicy(session, new ArtifactDescriptorPolicyRequest(a, request.getRequestContext()));
     }
+
+    private void populateResult(InternalSession session, ArtifactDescriptorResult result, Model model) {
+        ArtifactTypeRegistry stereotypes = session.getSession().getArtifactTypeRegistry();
+
+        for (Repository r : model.getRepositories()) {
```
elharo: r --> repository

impl/maven-impl/src/main/java/org/apache/maven/impl/resolver/DefaultArtifactDescriptorReader.java
```diff
@@ -331,4 +339,99 @@ private int getPolicy(RepositorySystemSession session, Artifact a, ArtifactDescr
...
+            result.addRepository(session.toRepository(
+                    session.getService(RepositoryFactory.class).createRemote(r)));
+        }
+
+        for (org.apache.maven.api.model.Dependency dependency : model.getDependencies()) {
+            result.addDependency(convert(dependency, stereotypes));
+        }
+
+        DependencyManagement mgmt = model.getDependencyManagement();
```
elharo: management or dependencyManagement

impl/maven-impl/src/main/java/org/apache/maven/impl/resolver/DefaultArtifactDescriptorReader.java
```diff
@@ -331,4 +339,99 @@ private int getPolicy(RepositorySystemSession session, Artifact a, ArtifactDescr
...
+
+        result.setProperties(properties);
+
+        setArtifactProperties(result, model);
+    }
+
+    private Dependency convert(org.apache.maven.api.model.Dependency dependency, ArtifactTypeRegistry stereotypes) {
+        ArtifactType stereotype = stereotypes.get(dependency.getType());
+        if (stereotype == null) {
+            // TODO: this here is fishy
```
elharo: can this be fixed in this PR?

impl/maven-impl/src/main/java/org/apache/maven/impl/resolver/DefaultArtifactDescriptorReader.java
```diff
@@ -331,4 +339,99 @@ private int getPolicy(RepositorySystemSession session, Artifact a, ArtifactDescr
...
+            stereotype = new DefaultType(dependency.getType(), Language.NONE, "", null, false);
+        }
+
+        boolean system = dependency.getSystemPath() != null
+                && !dependency.getSystemPath().isEmpty();
+
+        Map<String, String> props = null;
```
elharo: properties

impl/maven-impl/src/main/java/org/apache/maven/impl/resolver/DefaultArtifactDescriptorReader.java
```diff
@@ -331,4 +339,99 @@ private int getPolicy(RepositorySystemSession session, Artifact a, ArtifactDescr
...
+            props = Collections.singletonMap(MavenArtifactProperties.LOCAL_PATH, dependency.getSystemPath());
+        }
+
+        Artifact artifact = new DefaultArtifact(
+                dependency.getGroupId(),
+                dependency.getArtifactId(),
+                dependency.getClassifier(),
+                null,
+                dependency.getVersion(),
+                props,
```
elharo: null is OK, not an empty map?

impl/maven-impl/src/main/java/org/apache/maven/impl/resolver/DefaultArtifactDescriptorReader.java
```diff
@@ -331,4 +339,99 @@ private int getPolicy(RepositorySystemSession session, Artifact a, ArtifactDescr
...
+                exclusions);
+    }
+
+    private Exclusion convert(org.apache.maven.api.model.Exclusion exclusion) {
+        return new Exclusion(exclusion.getGroupId(), exclusion.getArtifactId(), "*", "*");
+    }
+
+    private void setArtifactProperties(ArtifactDescriptorResult result, Model model) {
+        String downloadUrl = null;
+        DistributionManagement distMgmt = model.getDistributionManagement();
```
elharo: distributionManagement

impl/maven-impl/src/main/java/org/apache/maven/impl/resolver/DefaultArtifactDescriptorReader.java
```diff
@@ -331,4 +339,99 @@ private int getPolicy(RepositorySystemSession session, Artifact a, ArtifactDescr
...
+        if (distMgmt != null) {
```
elharo: combine the ifs

impl/maven-impl/src/main/java/org/apache/maven/impl/resolver/DefaultArtifactDescriptorReader.java
gnodet: `null` is supported

impl/maven-impl/src/main/java/org/apache/maven/impl/resolver/DefaultArtifactDescriptorReader.java
cstamas: Short story: this cannot happen (null type).

Longer story: this class is resolver component implementation, where in resolver apis the resolver artifact type registry may return null. But in case of Maven, artifact type registry is delegated to old Maven artifact handler registry, that never returns null. Still, i may be wrong here, as we may split the two, i kinda remember something along these lines .. @gnodet ?

gnodet: I can see two implementations:
* [`MavenArtifactTypeRegistry`](https://github.com/apache/maven/blob/b61b3bd0b0e473ac2a62f41ec635c7df11396b57/impl/maven-core/src/main/java/org/apache/maven/RepositoryUtils.java#L314) which is not used
* [`TypeRegistryAdapter`](https://github.com/apache/maven/blob/8134aa24b1f2afd9374e73f860542384c89354a9/impl/maven-core/src/main/java/org/apache/maven/internal/aether/TypeRegistryAdapter.java#L30) which is the one used in [`DefaultRepositorySystemSessionFactory`](https://github.com/apache/maven/blob/6dc2178651634c69ea0a313b836d5305d783053b/impl/maven-core/src/main/java/org/apache/maven/internal/aether/DefaultRepositorySystemSessionFactory.java#L166)

So ... not sure

gnodet: Ah, the code in `TypeRegistryAdapter` is misleading.  It calls `typeRegistry.require`, which always return a non null value.  So yes, we should be safe.

gnodet: I've raised #10941

gnodet:
@elharo another look ?

Issue #10939, opened 2025-07-18
Location tracking can not be disabled through the DefaultModelXmlFactory


Giovds:
### Affected version

4.0.0-rc-4

### Bug description

While working on moving the `maven-help-plugin` to Maven 4 I noticed we no longer use the `MavenStaxWriter` directly. This is now done through the `DefaultModelXmlFactory`. 

The `DefaultModelXmlFactory` however does not provide a way to define `addLocationInformation` on the `MavenStaxWriter`. The `XmlWriterRequest` defines a way to set the `InputLocationFormatter` which is related to the `addLocationInformation`. The formatter is (currently) only used when locationtracking is requested.

My suggestion would be:

- Set `addLocationInformation` to `false` by default in `DefaultModelXmlFactory`.
- If the `XmlWriterRequest` specifies an `InputLocationFormatter`, set `addLocationInformation` to `true` and apply the formatter.

PR #10940, opened 2025-07-18
[#10939] Enable addLocationInformation in DefaultModelXmlFactory base on InputLocationFormatter presence

Giovds:
Show example for [#10939] 

Set addLocationInformation to false by default in DefaultModelXmlFactory.
If the XmlWriterRequest specifies an InputLocationFormatter, set addLocationInformation to true and apply the formatter.

---

### DIFF
--- a/compat/maven-model/src/main/java/org/apache/maven/model/io/xpp3/MavenXpp3Writer.java
+++ b/compat/maven-model/src/main/java/org/apache/maven/model/io/xpp3/MavenXpp3Writer.java
@@ -44,8 +44,7 @@ public class MavenXpp3Writer {
     // - Methods -/
     // -----------/
 
-    public MavenXpp3Writer() {
-    }
+    public MavenXpp3Writer() {}
 
     protected MavenXpp3Writer(Function<org.apache.maven.api.model.InputLocation, String> formatter) {
         delegate.setStringFormatter(formatter);
--- a/compat/maven-model/src/main/java/org/apache/maven/model/io/xpp3/MavenXpp3Writer.java
+++ b/compat/maven-model/src/main/java/org/apache/maven/model/io/xpp3/MavenXpp3Writer.java
@@ -45,7 +45,6 @@ public class MavenXpp3Writer {
     // -----------/
 
     public MavenXpp3Writer() {
-        this((inputLocation) -> inputLocation.toString());
     }
 
     protected MavenXpp3Writer(Function<org.apache.maven.api.model.InputLocation, String> formatter) {
--- a/compat/maven-model/src/main/java/org/apache/maven/model/io/xpp3/MavenXpp3Writer.java
+++ b/compat/maven-model/src/main/java/org/apache/maven/model/io/xpp3/MavenXpp3Writer.java
@@ -23,13 +23,14 @@ import javax.xml.stream.XMLStreamException;
 import java.io.IOException;
 import java.io.OutputStream;
 import java.io.Writer;
+import java.util.function.Function;
 
 import org.apache.maven.model.InputLocation;
 import org.apache.maven.model.Model;
 import org.apache.maven.model.v4.MavenStaxWriter;
 
 /**
- * @deprecated Use MavenStaxWriter instead
+ * @deprecated Use {@link MavenStaxWriter} instead
  */
 @Deprecated
 public class MavenXpp3Writer {
@@ -44,11 +45,11 @@ public class MavenXpp3Writer {
     // -----------/
 
     public MavenXpp3Writer() {
-        this(false);
+        this((inputLocation) -> inputLocation.toString());
     }
 
-    protected MavenXpp3Writer(boolean addLocationInformation) {
-        delegate.setAddLocationInformation(addLocationInformation);
+    protected MavenXpp3Writer(Function<org.apache.maven.api.model.InputLocation, String> formatter) {
+        delegate.setStringFormatter(formatter);
     }
 
     /**
--- a/compat/maven-model/src/main/java/org/apache/maven/model/io/xpp3/MavenXpp3WriterEx.java
+++ b/compat/maven-model/src/main/java/org/apache/maven/model/io/xpp3/MavenXpp3WriterEx.java
@@ -24,9 +24,10 @@ import java.io.Writer;
 
 import org.apache.maven.model.InputLocation;
 import org.apache.maven.model.Model;
+import org.apache.maven.model.v4.MavenStaxWriter;
 
 /**
- * @deprecated Use MavenStaxWriter instead
+ * @deprecated Use {@link MavenStaxWriter} instead
  */
 @Deprecated
 public class MavenXpp3WriterEx extends MavenXpp3Writer {
@@ -36,7 +37,7 @@ public class MavenXpp3WriterEx extends MavenXpp3Writer {
     // -----------/
 
     public MavenXpp3WriterEx() {
-        super(true);
+        super((inputLocation) -> inputLocation.toString());
     }
 
     /**
--- a/compat/maven-model/src/test/java/org/apache/maven/model/v4/ModelXmlTest.java
+++ b/compat/maven-model/src/test/java/org/apache/maven/model/v4/ModelXmlTest.java
@@ -89,7 +89,6 @@ class ModelXmlTest {
     String toXml(Model model) throws IOException, XMLStreamException {
         StringWriter sw = new StringWriter();
         MavenStaxWriter writer = new MavenStaxWriter();
-        writer.setAddLocationInformation(false);
         writer.write(sw, model);
         return sw.toString();
     }
--- a/compat/maven-settings/src/main/java/org/apache/maven/settings/io/xpp3/SettingsXpp3Writer.java
+++ b/compat/maven-settings/src/main/java/org/apache/maven/settings/io/xpp3/SettingsXpp3Writer.java
@@ -38,7 +38,6 @@ public class SettingsXpp3Writer {
 
     public SettingsXpp3Writer() {
         delegate = new SettingsStaxWriter();
-        delegate.setAddLocationInformation(false);
     }
     /**
      * Method setFileComment.
--- a/compat/maven-toolchain-model/src/main/java/org/apache/maven/toolchain/model/io/xpp3/MavenToolchainsXpp3Writer.java
+++ b/compat/maven-toolchain-model/src/main/java/org/apache/maven/toolchain/model/io/xpp3/MavenToolchainsXpp3Writer.java
@@ -41,7 +41,6 @@ public class MavenToolchainsXpp3Writer {
 
     public MavenToolchainsXpp3Writer() {
         delegate = new MavenToolchainsStaxWriter();
-        delegate.setAddLocationInformation(false);
     }
 
     /**
--- a/impl/maven-impl/src/main/java/org/apache/maven/impl/DefaultModelXmlFactory.java
+++ b/impl/maven-impl/src/main/java/org/apache/maven/impl/DefaultModelXmlFactory.java
@@ -127,7 +127,9 @@ public class DefaultModelXmlFactory implements ModelXmlFactory {
         }
         try {
             MavenStaxWriter w = new MavenStaxWriter();
+            w.setAddLocationInformation(false);
             if (inputLocationFormatter != null) {
+                w.setAddLocationInformation(true);
                 w.setStringFormatter((Function) inputLocationFormatter);
             }
             if (writer != null) {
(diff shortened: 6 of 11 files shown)


Giovds:
Current behaviour without a formatter will result in:

```xml
    <modelVersion>4.0.0</modelVersion><!-- 2 -->
    <groupId>org.example</groupId><!-- 3 -->
    <artifactId>example-artifact</artifactId><!-- 4 -->
    <version>1.0.0</version><!-- 5 -->
    <name>Example Project</name><!-- 6 -->
    <properties>
      <example.property>value</example.property><!-- 8 -->
    </properties>
  </project>
  ```

File: impl/maven-impl/src/main/java/org/apache/maven/impl/DefaultModelXmlFactory.java [hunk 1]
```diff
@@ -127,7 +127,9 @@ public void write(XmlWriterRequest<Model> request) throws XmlWriterException {
         }
         try {
             MavenStaxWriter w = new MavenStaxWriter();
+            w.setAddLocationInformation(false);
```
elharo: maybe  use an else block for this; what's the default?

impl/maven-impl/src/test/java/org/apache/maven/impl/DefaultModelXmlFactoryTest.java
```diff
@@ -122,4 +126,91 @@ void testMalformedModelVersion() throws Exception {
         Model model = factory.read(request);
         assertEquals("invalid.version", model.getModelVersion());
     }
+
+    @Test
+    void testWriteModelWithoutInputLocationTracking() {
+        String model =
+                """
+                <project xmlns="http://maven.apache.org/POM/4.0.0">
```
elharo, line 133: I don't think this works in Java 11. Are we requiring Java 17 now?

impl/maven-impl/src/test/java/org/apache/maven/impl/DefaultModelXmlFactoryTest.java
```diff
@@ -19,13 +19,17 @@
...
 
+import org.apache.maven.api.model.InputLocation;
 import org.apache.maven.api.model.Model;
 import org.apache.maven.api.services.xml.XmlReaderException;
 import org.apache.maven.api.services.xml.XmlReaderRequest;
+import org.apache.maven.api.services.xml.XmlWriterRequest;
 import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
 
+import static org.assertj.core.api.Assertions.assertThat;
```
elharo: please use standard JUnit asserts only. I'd like to keep third party dependencies to a minimum

impl/maven-impl/src/test/java/org/apache/maven/impl/DefaultModelXmlFactoryTest.java
```diff
@@ -122,4 +126,91 @@ void testMalformedModelVersion() throws Exception {
...
+                    <properties>
+                        <example.property>value</example.property>
+                    </properties>
+                </project>""";
+
+        String expected =
+                """
+                <?xml version="1.0" encoding="UTF-8"?>
+                <project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 https://maven.apache.org/xsd/maven-4.0.0.xsd">
+                  <modelVersion>4.0.0</modelVersion><!--some formatter: n/a @ 2:3-->
```
elharo, line 191: I do not understand why someone would want this in their output. It feels very special case. This is not what I thought location information was, and I'm tempted to take this out.

elharo (changes requested):

impl/maven-impl/src/main/java/org/apache/maven/impl/DefaultModelXmlFactory.java
Giovds: The `MavenStaxWriter` is initialized with `true` therefore I explicitly set it to `false`. I personally prefer to keep nesting brackets to a minimum, as this makes it less verbose and easier to understand what's happening. The result would be the same?

impl/maven-impl/src/test/java/org/apache/maven/impl/DefaultModelXmlFactoryTest.java
Giovds: Aren't we compiling against 17 nowadays? The text blocks are used in other locations (within the same test class) as well.

impl/maven-impl/src/test/java/org/apache/maven/impl/DefaultModelXmlFactoryTest.java
Giovds: Shouldn't we remove this dependency entirely in that case?

impl/maven-impl/src/test/java/org/apache/maven/impl/DefaultModelXmlFactoryTest.java
Giovds: I've mainly written this test to:
1. Add a test scenario in the first place as there were non
2. Document the behaviour that this class exposes

impl/maven-impl/src/test/java/org/apache/maven/impl/DefaultModelXmlFactoryTest.java
gnodet: Yes, Maven 4 has switched to 17 a while ago, see https://github.com/apache/maven/issues/10806

impl/maven-impl/src/test/java/org/apache/maven/impl/DefaultModelXmlFactoryTest.java
gnodet: > please use standard JUnit asserts only. I'd like to keep third party dependencies to a minimum

I agree with the goal, but AssertJ is much more expressive than JUnit.  In addition, this is a test-time dependency, so much less risk involved.

impl/maven-impl/src/test/java/org/apache/maven/impl/DefaultModelXmlFactoryTest.java
gnodet: > I do not understand why someone would want this in their output. It feels very special case. This is not what I thought location information was, and I'm tempted to take this out.

Take out location information ? are you kidding?

Of course those examples are very basic, location information is only interesting if it refers to non trivial things, especially effective poms/settings, which undergo several merges, property substitution, etc...

This is needed for `help:effective-pom` and `help:effective-settings` which are really important tools for debugging.

impl/maven-impl/src/main/java/org/apache/maven/impl/DefaultModelXmlFactory.java
elharo: In that case, no need to set it to true below. Use an else block to set it to false

impl/maven-impl/src/test/java/org/apache/maven/impl/DefaultModelXmlFactoryTest.java
elharo: Maybe for some things, but not in this case. assertEquals is just fine

impl/maven-impl/src/test/java/org/apache/maven/impl/DefaultModelXmlFactoryTest.java
elharo: My concern is this is getting added to comments in the serialized output. That's not how I expected this to be presented.

impl/maven-impl/src/main/java/org/apache/maven/impl/DefaultModelXmlFactory.java
Giovds: That would make it depend on the implementation of the `MavenStaxWriter`. If it changes the default this will affect our class.

I was thinking the same as @gnodet:
> I wonder if this should be moved further down in the MavenStaxWriter and remove the addLocationInformation completely in that case.

If we use the proposed solution of moving this behaviour to the `MavenStaxWriter` this might solve both the need to set the value, and other XML factory implementations.

This is a generic generated class (through`writer-stax.vm`), so all `ZZZStaxWriter` classes (which are about 8 classes) will use this behaviour from then on. I'm not entirely sure about that change, but probably will be the better option as it will solve the issues for other `DefaultZZZXmlFactory` classes as well that use one of the `ZZZStaxWriter`'s.

impl/maven-impl/src/test/java/org/apache/maven/impl/DefaultModelXmlFactoryTest.java
Giovds: For the sake of this PR I will make it use `assertEquals` from JUnit. 

I do however think that there should be a statement from the (core) team to either:

- Remove the dependency entirely so it can not be used in the first place.
- Agree to keep the dependency and accept the use of it.

This distracts from the feedback on actual change, and will be a recurring discussion on other PR's.

I'm happy to open a thread to discuss this on the mailing list. WDYT?

impl/maven-impl/src/test/java/org/apache/maven/impl/DefaultModelXmlFactoryTest.java
Giovds: It seems like something that is valid be discussed, but I suggest to raise a new issue/discussion on this as it won't be fixed in this PR. 

This is the current implementation on location tracking when a formatter is present. This PR changes _when_ it should trigger the formatting, not _how_ it is formatted. Whether it should be parsed as a new element, comment, attribute, or something else is not part of this change.

compat/maven-model/src/main/java/org/apache/maven/model/io/xpp3/MavenXpp3Writer.java
```diff
@@ -43,12 +44,10 @@ public class MavenXpp3Writer {
     // - Methods -/
     // -----------/
 
-    public MavenXpp3Writer() {
-        this(false);
-    }
+    public MavenXpp3Writer() {}
 
-    protected MavenXpp3Writer(boolean addLocationInformation) {
```
gnodet, line 50: No API breakage in Maven 3 compatibility layer.

compat/maven-model/src/main/java/org/apache/maven/model/io/xpp3/MavenXpp3WriterEx.java
```diff
@@ -36,7 +37,7 @@ public class MavenXpp3WriterEx extends MavenXpp3Writer {
     // -----------/
 
     public MavenXpp3WriterEx() {
-        super(true);
+        super((inputLocation) -> inputLocation.toString());
```
gnodet, line 40: Undo

compat/maven-model/src/test/java/org/apache/maven/model/v4/ModelXmlTest.java
```diff
@@ -89,7 +89,6 @@ void testNamespaceInXmlNode() throws XMLStreamException {
     String toXml(Model model) throws IOException, XMLStreamException {
         StringWriter sw = new StringWriter();
         MavenStaxWriter writer = new MavenStaxWriter();
-        writer.setAddLocationInformation(false);
```
gnodet, line 92: Undo

compat/maven-settings/src/main/java/org/apache/maven/settings/io/xpp3/SettingsXpp3Writer.java
```diff
@@ -38,7 +38,6 @@ public class SettingsXpp3Writer {
 
     public SettingsXpp3Writer() {
         delegate = new SettingsStaxWriter();
-        delegate.setAddLocationInformation(false);
```
gnodet, line 41: Undo

compat/maven-toolchain-model/src/main/java/org/apache/maven/toolchain/model/io/xpp3/MavenToolchainsXpp3Writer.java
```diff
@@ -41,7 +41,6 @@ public class MavenToolchainsXpp3Writer {
 
     public MavenToolchainsXpp3Writer() {
         delegate = new MavenToolchainsStaxWriter();
-        delegate.setAddLocationInformation(false);
```
gnodet, line 44: Undo

gnodet (changes requested):
We cannot break compatibility with Maven 3 API here.
Undo the changes in compat/* modules.

compat/maven-model/src/main/java/org/apache/maven/model/io/xpp3/MavenXpp3Writer.java
Giovds: Check. Won’t be able to work on this due to holiday for the next couple weeks. Feel free to adjust it, otherwise I’ll look into this when I’m back.

PR #10952, opened 2025-07-19
[MNG-8537] Keep Maven Namespace the same

### LINKED ISSUE #10692: [MNG-8537] Maven 4 namespace should not change
**[Elliotte Rusty Harold](https://issues.apache.org/jira/secure/ViewProfile.jspa?name=elharo)** opened **[MNG-8537](https://issues.apache.org/jira/browse/MNG-8537?redirect=false)** and commented

Just noticed that Maven 4 has a new namespace URL. The old namespace was

 

http://maven.apache.org/POM/4.0.0

 

The new one is

 

http://maven.apache.org/POM/4.1.0

 

Putting version numbers in namespaces is a known XML antipattern because it makes it extremely difficult and inconvenient to write tools that process both, even when they are much the same. XSLT, DOM, XQuery, JDOM, etc. — really any XML aware tool — is going to have problems with this.

Model version 4.1.0 is not a new and different schema that completely breaks with the past. Most old elements from 4.0.0 are still present and still mean exactly the same thing: groupId, artifactId, name, dependency, and most others. There are new elements but that doesn't imply a new namespace. Adding a new namespace is asserting that all the elements are different.

IMHO Maven 4 should not change the namespace URL.

---

**Issue Links:**
- [MNG-8639](https://issues.apache.org/jira/browse/MNG-8639) Check namespaces when reading XML models

elharo:
Namespaces and model versions are not the same thing and do not serve the same purpose.

This PR has two related effects:

1. The namespace  for all pom.xml files is now http://maven.apache.org/POM/4.0.0
2. Namespaces are not used to infer model versions. That's the job of the `modelVersion` element

Fixing this opens up Maven to the large and valuable XML toolchain. It makes it much easier for developers to analyze, store, and manipulate pom.xml files from a variety of languages without depending on complex Maven APIs and tooling. It decouples the poms from the Java source code. It is no longer necessary to use Maven libraries to process poms. 

On the Maven side, it removes a roadblock to using standard XML parsers like Xerces instead of maintaining a separate custom XML parser. The more we can fit into the standard toolchains, the easier development becomes. 

fixes #10692


gnodet (changes requested):
While I don't disagree with the change, I think this is not doable between rc-4 and ga.  This is a *major* breaking change, so it should have been planned from the beginning when discussing 4.0, not at the last minute.
I'd defer that for 5.0.

elharo:
I've been pushing this for a long time, and the longer we wait the harder it gets to do. I'm not sure it can be done in 5.0

elharo:
Nothing should be deployed with the 4.0 namespace as it hasn't been released yet. If its, I can live with that breakage. Of course, once 4.0 is released, it becomes far more damaging to change it.

elharo:
FYI, a quick check of my inbox shows I've been raising red flags on this since at least 2019.

impl/maven-cli/src/test/java/org/apache/maven/cling/invoker/mvnup/goals/ModelVersionUtilsTest.java
```diff
@@ -417,7 +399,7 @@ void shouldHandleMissingModelVersion() throws Exception {
         @ValueSource(
                 strings = {
                     "http://maven.apache.org/POM/4.0.0",
-                    "http://maven.apache.org/POM/4.1.0",
+                    "http://maven.apache.org/POM/4.0.0",
```
slachiewicz, line 384: duplicate of 401?

impl/maven-cli/src/test/java/org/apache/maven/cling/invoker/mvnup/goals/ModelVersionUtilsTest.java
```diff
@@ -417,7 +399,7 @@ void shouldHandleMissingModelVersion() throws Exception {
...
                     "https://maven.apache.org/POM/4.0.0",
                     "https://maven.apache.org/POM/4.1.0"
```
slachiewicz, line 386: here is also 4.1

impl/maven-cli/src/main/java/org/apache/maven/cling/invoker/mvnup/goals/UpgradeConstants.java
```diff
@@ -177,11 +177,11 @@ private Files() {
...
      */
     public static final class Namespaces {
-        /** Maven 4.0.0 namespace URI */
+        /** Maven namespace URI */
         public static final String MAVEN_4_0_0_NAMESPACE = "http://maven.apache.org/POM/4.0.0";
 
-        /** Maven 4.1.0 namespace URI */
-        public static final String MAVEN_4_1_0_NAMESPACE = "http://maven.apache.org/POM/4.1.0";
+        /** Maven namespace URI */
+        public static final String MAVEN_4_1_0_NAMESPACE = "http://maven.apache.org/POM/4.0.0";
```
slachiewicz, line 184: so no need for this const

elharo: done

elharo:
```
[ERROR] Failures: 
	
[ERROR]   ModelUpgradeStrategyTest.shouldUpdateNamespaceRecursively expected: <http://maven.apache.org/POM/4.0.0> but was: <http://maven.apache.org/POM/4.1.0>
	
[ERROR]   ModelVersionUtilsTest.shouldDetectModelVersionFromNamespace(String)[1] expected: <4.0.0> but was: <4.1.0>
	
[ERROR]   ModelVersionUtilsTest.shouldDetectModelVersionFromNamespace(String)[3] expected: <4.2.0> but was: <4.1.0>
	
[ERROR]   UpgradeWorkflowIntegrationTest.shouldUpgradeModelVersionWith41Option POM should be upgraded to 4.1.0 namespace ==> expected: <true> but was: <false>
```

Bukama (commented):
Note: We decided to keep the namespace as it is (at least for 4.0 release)

https://lists.apache.org/thread/h6d7wgq54fr84rmd65bddnwhwnxc7g62

gnodet (commented):
## AI Code Review — PR #10952

**Verdict: 💬 COMMENT**

Well-motivated PR that fixes the XML namespace to `http://maven.apache.org/POM/4.0.0` for all model versions, decoupling namespace from modelVersion. The implementation is thorough — CI passes, and all areas flagged by gnodet's earlier review have been addressed (TransformerSupport, DefaultModelBuilder, maven.mdo, DefaultModelXmlFactory, and upgrade tooling).

However, the community has already decided via the mailing list to keep the namespace as-is for the 4.0 release (noted by Bukama, May 2026). The findings below are code-level observations for consideration if the approach is revisited.

### Findings

| 1 | 🟡 Medium | `UpgradeConstants.java:199-202` | **Misleading constant name** — `MAVEN_4_1_0_NAMESPACE` is defined as `"http://maven.apache.org/POM/4.0.0"`, identical to `MAVEN_4_0_0_NAMESPACE`. Having a constant named `4_1_0` with a `4.0.0` value is confusing. Consider a single `MAVEN_POM_NAMESPACE` constant instead. |
| 2 | 🟢 Low | `ModelVersionUtilsTest.java:384` | **Duplicate parameterized test entry** — The `@ValueSource` now has `"http://maven.apache.org/POM/4.0.0"` twice (since the old 4.1.0 value was changed to 4.0.0). The test runs the same input redundantly. Additionally, `"https://maven.apache.org/POM/4.1.0"` still uses 4.1.0 with `https`, which seems inconsistent with the PR's goal. |
| 3 | 🟡 Medium | `DefaultModelXmlFactory.java:66` | **Backward compatibility concern** — Namespace validation changed from `startsWith("http://maven.apache.org/POM/")` to strict `equals("http://maven.apache.org/POM/4.0.0")`. POMs deployed during Maven 4 pre-release with namespace `http://maven.apache.org/POM/4.1.0` would be rejected with a hard error. Consider accepting both during reading or documenting the migration path. |

### Context
- **Community decision**: The mailing list thread ([link](https://lists.apache.org/thread/h6d7wgq54fr84rmd65bddnwhwnxc7g62)) concluded to keep the namespace as-is for 4.0. This PR cannot merge without revisiting that decision.
- **External blocker**: The maven-shade-plugin derives namespace URIs from model versions (acknowledged by the `@Disabled` test `MavenITmng7228LeakyModelTest`). A shade-plugin fix is needed for end-to-end correctness.
- **Comparison with #11185/#11186**: This PR takes a fundamentally different and cleaner approach — removing version-varying namespaces entirely rather than restricting which are allowed. More principled from an XML standards perspective but also more disruptive.

---

gnodet (commented):
## Code Review

All five findings independently confirmed. The PR decouples XML namespace from model version (fixing all namespaces to 4.0.0), but has several code quality issues:

1. **Disabled integration test** (`MavenITmng7228LeakyModelTest.java:37`): `testLeakyModel` is disabled with `@Disabled` due to maven-shade-plugin deriving namespace URIs from model versions. Disabling a pre-existing IT to work around an external plugin's broken behavior signals incomplete implementation.

2. **Duplicate test parameter** (`ModelVersionUtilsTest.java:384`): The `@ValueSource` contains a duplicate entry — both first and second strings are `"http://maven.apache.org/POM/4.0.0"`. The fourth entry `"https://maven.apache.org/POM/4.1.0"` retains 4.1.0, inconsistent with the PR's namespace normalization goal.

3. **Backward compatibility risk** (`DefaultModelXmlFactory.java:67`): Namespace validation changed from `startsWith("http://maven.apache.org/POM/")` to strict `equals("http://maven.apache.org/POM/4.0.0")`. This could reject POMs deployed during the Maven 4 RC period with `xmlns="http://maven.apache.org/POM/4.1.0"`.

4. **Mismatched assertion message** (`UpgradeWorkflowIntegrationTest.java:91`): Assertion message says "POM should be upgraded to 4.1.0 namespace" but the assertion checks for 4.0.0 — a copy-paste oversight.

5. **Dead code constants** (`UpgradeConstants.java`): New 234-line constants file partially duplicates domtrip's `MavenPomElements` class and is not imported or referenced by any other file in the diff or codebase. Additionally, `Namespaces.MAVEN_4_1_0_NAMESPACE` is defined as `"http://maven.apache.org/POM/4.0.0"` — a confusingly named constant whose value doesn't match its name.

Note: The Maven community has decided via mailing list to keep the namespace as-is for the 4.0 release, and the maintainer has formally requested changes deferring this to 5.0.

gnodet (commented):
## Code Review

The architectural motivation is sound — decoupling XML namespace from `modelVersion` addresses a real design concern. However, the community has already decided via the [dev mailing list](https://lists.apache.org/thread/h6d7wgq54fr84rmd65bddnwhwnxc7g62) to keep the namespace as-is for the 4.0 release, and this PR has several code quality issues that would need to be addressed if revived for 5.0:

### Issues Found

**1. Integration test disabled rather than fixed (high)**
`MavenITmng7228LeakyModelTest.java` — `testLeakyModel` is disabled with `@Disabled` to work around maven-shade-plugin incorrectly deriving namespace URIs from model versions. Disabling a pre-existing IT rather than fixing the root cause in the shade plugin signals an incomplete implementation.

**2. Backward compatibility risk in namespace validation (high)**
`DefaultModelXmlFactory.java` — Namespace validation changed from `startsWith("http://maven.apache.org/POM/")` to strict `equals("http://maven.apache.org/POM/4.0.0")`. POMs published during the Maven 4 RC period with `xmlns="http://maven.apache.org/POM/4.1.0"` would be rejected with a hard error. No migration path is provided.

**3. Dead code: UpgradeConstants.java (medium)**
New 234-line file `UpgradeConstants.java` is not imported by any other file in the diff. The files that use namespace constants import from `eu.maveniverse.domtrip.maven.MavenPomElements` instead.

**4. Misleading duplicate constants (medium)**
`MAVEN_4_1_0_NAMESPACE` is defined as `"http://maven.apache.org/POM/4.0.0"`, identical to `MAVEN_4_0_0_NAMESPACE`. A single `MAVEN_NAMESPACE` constant would be clearer.

**5. Assertion message mismatch (medium)**
`UpgradeWorkflowIntegrationTest.java` line 91 — message says "4.1.0 namespace" but the assertion checks for the 4.0.0 value.

**6. Loosened test assertion (low)**
`MavenStaxReaderTest.java` — error message assertion changed from a specific namespace string to a less specific prefix match.

### Notes

- Of the 189 changed files, 174 are mechanical XML namespace replacements (safe). The risk is concentrated in ~6 Java files with behavioral changes.
- The PR is currently in a CONFLICTING state with master (370+ days old, 50 commits). If revived for 5.0, the commit history should be squashed.

gnodet (commented):
### Review Summary

The architectural motivation is sound — decoupling namespace from `modelVersion` is good XML practice and would enable standard XML tooling for POM files. However, there are several code quality issues from the mechanical namespace replacement, plus a process concern noted below.

#### 1. 🔴 Disabled integration test rather than fixed (high)

**File:** `its/core-it-suite/src/test/java/org/apache/maven/it/MavenITmng7228LeakyModelTest.java`

The pre-existing `testLeakyModel` IT is disabled with `@Disabled("Disabled because the maven-shade-plugin incorrectly derives namespace URIs from model versions")` rather than fixed. This signals an external dependency blocker — the maven-shade-plugin would need updating to handle the namespace change.

#### 2. 🔴 Tightened namespace validation breaks RC-period POMs (high)

**File:** `impl/maven-impl/src/main/java/org/apache/maven/impl/DefaultModelXmlFactory.java`

Namespace validation changes from `startsWith("http://maven.apache.org/POM/")` to strict `equals("http://maven.apache.org/POM/4.0.0")`. Any POMs published during the Maven 4 RC period with `xmlns="http://maven.apache.org/POM/4.1.0"` would be rejected with a hard error. No migration path or deprecation period is provided.

#### 3. 🟡 `UpgradeConstants.java` is dead code (medium)

**File:** `impl/maven-cli/src/main/java/org/apache/maven/cling/invoker/mvnup/goals/UpgradeConstants.java`

The new 234-line constants file is never imported or referenced by any other file in the diff or existing codebase. The code that previously used similar constants now imports from the external library `eu.maveniverse.domtrip.maven.MavenPomElements.Namespaces`.

#### 4. 🟡 Misleading constant name (medium)

**File:** `impl/maven-cli/src/main/java/org/apache/maven/cling/invoker/mvnup/goals/UpgradeConstants.java`

`MAVEN_4_1_0_NAMESPACE` is defined as `"http://maven.apache.org/POM/4.0.0"`, identical to `MAVEN_4_0_0_NAMESPACE`. While intentional for namespace unification, the constant name strongly implies a different URI. A single `MAVEN_POM_NAMESPACE` constant would be clearer.

#### 5. 🟡 Assertion message mismatch (medium)

**File:** `impl/maven-cli/src/test/java/org/apache/maven/cling/invoker/mvnup/goals/UpgradeWorkflowIntegrationTest.java`

Assertion message says "POM should be upgraded to 4.1.0 namespace" but the assertion actually checks for the `4.0.0` value — a copy-paste oversight from the mechanical find-and-replace.

#### 6. 🔵 Duplicate `@ValueSource` entry (low)

**File:** `impl/maven-cli/src/test/java/org/apache/maven/cling/invoker/mvnup/goals/ModelVersionUtilsTest.java`

The `@ValueSource` at line 384 contains `"http://maven.apache.org/POM/4.0.0"` twice (the 4.1.0 entry was mechanically replaced to 4.0.0, duplicating the existing entry). This runs the same test case twice and reduces coverage — the `http` 4.1.0 case is no longer tested.

### 📋 PR Metadata

|--------|---------|-----------|
| Milestone | _(none)_ | `5.x` |

> **Note:** Per the [dev mailing list discussion](https://lists.apache.org/thread/h6d7wgq54fr84rmd65bddnwhwnxc7g62), the community decided to keep the current namespace scheme for 4.0 GA. This PR would be a candidate for the 5.x timeframe if revisited.

---

PR #10953, opened 2025-07-19
Make error message less awkward

### LINKED ISSUE #10891: improve the wording of the error message for modelVersion mismatch

elharo:
fixes #10891

### DIFF
--- a/impl/maven-impl/src/main/java/org/apache/maven/impl/model/DefaultModelValidator.java
+++ b/impl/maven-impl/src/main/java/org/apache/maven/impl/model/DefaultModelValidator.java
@@ -2023,7 +2023,7 @@ public class DefaultModelValidator implements ModelValidator {
                     Version.V20,
                     "modelVersion",
                     null,
-                    "of '" + requestedModel + "' is newer than the versions supported by this Maven version ("
+                    "of '" + requestedModel + "' is not supported by this Maven version ("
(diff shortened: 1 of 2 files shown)


Bukama (commented):
I have no strong opinion about this, but why change the more detailed error message to a general one which then can't be linked to a unique position in code?

Fixing the grammar mistakes if ofc good :)

elharo:
Because the more detailed message was confusing users/readers. Model version and maven version are confusingly intertwined.

invisiblepancake (approved):

PR #10961, opened 2025-07-19
Generating configuration documentation during site build

slawekjaranowski:


pom.xml
```diff
@@ -1015,6 +1015,65 @@ under the License.
...
+            <inherited>false</inherited>
+            <executions>
+              <execution>
+                <goals>
+                  <goal>run</goal>
+                </goals>
+                <phase>pre-site</phase>
+                <configuration>
+                  <target>
+                    <!-- directory should be created by org.eclipse.aether.tools.CollectConfiguration -->
```
slawekjaranowski, line 1030: should be fixed in `CollectConfiguration`

src/site/markdown/configuration.properties
slawekjaranowski, line 1: this file was not included in site .... what is for?

src/site/markdown/configuration.yaml
slawekjaranowski, line 1: also this file was not included in site .... what is for?

gnodet: I think the idea was to publish them somehow, so that they can be consumed programmatically, not only from docs.  But there's no real usage yet, so removing them is fine for now.

pom.xml
gnodet: Is that "should already be fixed In CollectConfiguration" or "we need to fix that in CollectConfiguration" ?

slawekjaranowski: should be, I will prepare PR

src/site/markdown/configuration.yaml
slawekjaranowski: ok, now `configuration.properties` and `configuration.yaml` are not published because are in markdown directory ...
to publish it we need put it in `site/resources`

I will add a comments for it

pom.xml
slawekjaranowski: - https://github.com/apache/maven-resolver/pull/1534

src/site/markdown/configuration.yaml
slawekjaranowski: I added a separate execution for properties files

gnodet (approved):

PR #10964, opened 2025-07-20
Add PathMatcherFactory.includesAll()

desruisseaux:
Minor modifications to the `PathMatcherFactory` service:

* Simplify the patch matcher when possible.
* Add a `PathMatcherFactory.includesAll()` method.

The two changes, combined together, allows the caller to know when a matcher includes all files. The Maven Clean Plugin needs this information for deciding if it can delete the files in a background thread.

### DIFF
--- a/api/maven-api-core/src/main/java/org/apache/maven/api/services/PathMatcherFactory.java
+++ b/api/maven-api-core/src/main/java/org/apache/maven/api/services/PathMatcherFactory.java
@@ -138,4 +138,19 @@ public interface PathMatcherFactory extends Service {
      */
     @Nonnull
     PathMatcher deriveDirectoryMatcher(@Nonnull PathMatcher fileMatcher);
+
+    /**
+     * Returns the patch matcher that unconditionally returns {@code true} for all files.
+     * It should be the matcher returned by the other methods of this interface when the
+     * given patterns match all files. Therefore, the following idiom can be used:
+     *
+     * {@snippet lang="java" :
+     * PathMatcher fileMatcher = factory.createPathMatcher(dir, includes, excludes);
+     * boolean selectedAllFiles = (fileMatcher == factory.includesAll);
+     * }
+     *
+     * @return patch matcher that unconditionally returns {@code true} for all files
+     */
+    @Nonnull
+    PathMatcher includesAll();
 }
--- a/api/maven-api-core/src/main/java/org/apache/maven/api/services/PathMatcherFactory.java
+++ b/api/maven-api-core/src/main/java/org/apache/maven/api/services/PathMatcherFactory.java
@@ -157,7 +157,7 @@ public interface PathMatcherFactory extends Service {
      *
      * @param matcher the matcher to test
      */
-    default boolean isIncludesAll(PathMatcher matcher) {
+    default boolean isIncludesAll(@Nonnull PathMatcher matcher) {
         return Objects.requireNonNull(matcher) == includesAll();
     }
 }
--- a/api/maven-api-core/src/main/java/org/apache/maven/api/services/PathMatcherFactory.java
+++ b/api/maven-api-core/src/main/java/org/apache/maven/api/services/PathMatcherFactory.java
@@ -140,14 +140,14 @@ public interface PathMatcherFactory extends Service {
     PathMatcher deriveDirectoryMatcher(@Nonnull PathMatcher fileMatcher);
 
     /**
-     * Returns the patch matcher that unconditionally returns {@code true} for all files.
+     * Returns the path matcher that unconditionally returns {@code true} for all files.
      * It should be the matcher returned by the other methods of this interface when the
      * given patterns match all files. Therefore, the following idiom can be used:
      *
      * <pre>PathMatcher fileMatcher = factory.createPathMatcher(dir, includes, excludes);
      * boolean selectedAllFiles = fileMatcher == factory.includesAll();</pre>
      *
-     * @return patch matcher that unconditionally returns {@code true} for all files
+     * @return path matcher that unconditionally returns {@code true} for all files
      */
     @Nonnull
     PathMatcher includesAll();
--- a/api/maven-api-core/src/main/java/org/apache/maven/api/services/PathMatcherFactory.java
+++ b/api/maven-api-core/src/main/java/org/apache/maven/api/services/PathMatcherFactory.java
@@ -145,7 +145,7 @@ public interface PathMatcherFactory extends Service {
      * given patterns match all files. Therefore, the following idiom can be used:
      *
      * <pre>PathMatcher fileMatcher = factory.createPathMatcher(dir, includes, excludes);
-     * boolean selectedAllFiles = (fileMatcher == factory.includesAll);</pre>
+     * boolean selectedAllFiles = fileMatcher == factory.includesAll();</pre>
      *
      * @return patch matcher that unconditionally returns {@code true} for all files
      */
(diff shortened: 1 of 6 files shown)


api/maven-api-core/src/main/java/org/apache/maven/api/services/PathMatcherFactory.java
```diff
@@ -138,4 +138,17 @@ default PathMatcher createIncludeOnlyMatcher(@Nonnull Path baseDirectory, Collec
...
     @Nonnull
     PathMatcher deriveDirectoryMatcher(@Nonnull PathMatcher fileMatcher);
+
+    /**
+     * Returns the patch matcher that unconditionally returns {@code true} for all files.
+     * It should be the matcher returned by the other methods of this interface when the
+     * given patterns match all files. Therefore, the following idiom can be used:
+     *
+     * <pre>PathMatcher fileMatcher = factory.createPathMatcher(dir, includes, excludes);
+     * boolean selectedAllFiles = (fileMatcher == factory.includesAll);</pre>
```
gnodet:
```suggestion
     * boolean selectedAllFiles = fileMatcher == factory.includesAll();</pre>
```

impl/maven-impl/src/main/java/org/apache/maven/impl/DefaultPathMatcherFactory.java
```diff
@@ -52,7 +52,7 @@ public PathMatcher createPathMatcher(
             boolean useDefaultExcludes) {
         requireNonNull(baseDirectory, "baseDirectory cannot be null");
 
-        return new PathSelector(baseDirectory, includes, excludes, useDefaultExcludes);
+        return new PathSelector(baseDirectory, includes, excludes, useDefaultExcludes).simplify();
```
gnodet: Would it be possible to use a static factory method `PathSelector.of(...)` and make the constructor private to avoid having one caller forgetting to call `simplify` ?

api/maven-api-core/src/main/java/org/apache/maven/api/services/PathMatcherFactory.java
desruisseaux: Oups! Thanks for spotting the typo.

impl/maven-impl/src/main/java/org/apache/maven/impl/DefaultPathMatcherFactory.java
desruisseaux: Yes. I was also considering to split the `PathSeparator` class in two classes, `FileMatcher` and `DirectoryMatcher`, because they are created by two separated methods in the factory. It will avoid creating the directory matchers when the user didn't ask for them, and would remove one indirection level. But I was hopping to fix #10965 first, so we can verify that this refactoring does not break anything.

api/maven-api-core/src/main/java/org/apache/maven/api/services/PathMatcherFactory.java
```diff
@@ -138,4 +138,17 @@ default PathMatcher createIncludeOnlyMatcher(@Nonnull Path baseDirectory, Collec
      */
     @Nonnull
     PathMatcher deriveDirectoryMatcher(@Nonnull PathMatcher fileMatcher);
+
+    /**
+     * Returns the patch matcher that unconditionally returns {@code true} for all files.
```
elharo: path matcher

api/maven-api-core/src/main/java/org/apache/maven/api/services/PathMatcherFactory.java
```diff
@@ -138,4 +138,17 @@ default PathMatcher createIncludeOnlyMatcher(@Nonnull Path baseDirectory, Collec
...
+     * It should be the matcher returned by the other methods of this interface when the
+     * given patterns match all files. Therefore, the following idiom can be used:
+     *
+     * <pre>PathMatcher fileMatcher = factory.createPathMatcher(dir, includes, excludes);
```
elharo: This strikes me as extremely brittle. Need to dig into the code more, but this makes me nervous. I'm  not sure we can count on this. There are no guarantees on the behavior of simplify, and making guarantees might be solving the halting problem.

elharo (changes requested):

api/maven-api-core/src/main/java/org/apache/maven/api/services/PathMatcherFactory.java
desruisseaux: Fixed, thanks.

api/maven-api-core/src/main/java/org/apache/maven/api/services/PathMatcherFactory.java
desruisseaux: `simplify()` is specific to the implementation (it is not a method of the `PathMatcher` interface) and will probably go away if we apply the plan in [above comment](https://github.com/apache/maven/pull/10964#discussion_r2218338877). In the current code, this is safe because the implementation knows what it is doing. We could also add a `PathMatcherFactory.isIncludesAll(PathPatcher)` method, but it would not be safer compared to the current code. It may be clearer however.

When testing whether a path matcher includes all files, it is okay to have false negatives as the only consequence (at least in Maven Clean Plugin) will be loosing optimization opportunities. It is important to not have false positives however, but the current proposal mets that requirement.

elharo: I'm not yet convinced it does guarantee no false positives. That depends on the implementation of the simplify method which doesn't seem to be enforced.

desruisseaux: The implementation of `simplify()` is enforced. This method exists only on `PathSelector`, this is not a Java API.

desruisseaux: ! added an `isIncludesAll(PathMatcher)` method for clarity and applied @gnodet suggestion: make the `PathSelector` constructor and `simplify()` method private, replaced by a static `of` method, for making clearer what is under our control.

impl/maven-impl/src/main/java/org/apache/maven/impl/DefaultPathMatcherFactory.java
desruisseaux: Applied the suggestion relative to a static method.

api/maven-api-core/src/main/java/org/apache/maven/api/services/PathMatcherFactory.java
```diff
@@ -138,4 +139,25 @@ default PathMatcher createIncludeOnlyMatcher(@Nonnull Path baseDirectory, Collec
...
+    PathMatcher includesAll();
+
+    /**
+     * {@return whether the given matcher includes all files}.
+     * This method may conservatively returns {@code false} if case of doubt.
+     * A return value of {@code true} means that the pattern is certain to match all files.
+     *
+     * @param matcher the matcher to test
+     */
+    default boolean isIncludesAll(PathMatcher matcher) {
```
gnodet: Either add `@Nonnull` on argument, or remove the `requireNonNull(matcher)`.  No strong opinion on which one though.

impl/maven-impl/src/main/java/org/apache/maven/impl/PathSelector.java
```diff
@@ -564,18 +564,28 @@ private static PathMatcher[] matchers(final FileSystem fs, final String[] patter
...
+     * @param directory the base directory of the files to filter
+     * @param includes the patterns of the files to include, or null or empty for including all files
+     * @param excludes the patterns of the files to exclude, or null or empty for no exclusion
+     * @param useDefaultExcludes whether to augment the excludes with a default set of <abbr>SCM</abbr> patterns
+     * @throws NullPointerException if directory is null
+     * @return a path matcher for the given includes and excludes
      */
-    public boolean isEmpty() {
-        return includes.length == 0 && excludes.length == 0;
+    public static PathMatcher of(
```
gnodet: Would you mind moving this  `public static` method at the top rather than in the middle of the file ?

desruisseaux: Done

gnodet (approved):

PR #10969, opened 2025-07-21
Avoid parsing MAVEN_OPTS (3.9.x)

### LINKED ISSUE #10937: Quoted pipes in MAVEN_OPTS throw errors in 3.9.10+, 4.0.0-rc
### Affected version

3.9.10, 3.9.11, 4.0.0-rc-4

### Bug description

Environment: Windows Server 2019
Maven: 3.9.10, invoked via `mvnw.cmd` wrapper

---

Minimal example (3.9.10):

```
$env:MAVEN_OPTS='"-Dfoo=bar|baz"'
mvn
```

On 4.0.0-rc-4, this requires an additional `.mvn/jvm.config` file to exist in the project directory or the `--yjp` arg. Same fundamental issue, different trigger in the script.

---

Consider this `MAVEN_OPTS` environment variable (used on a CI server with restricted network access):

```
-Dhttps.proxyHost=192.0.2.4 -Dhttps.proxyPort=8080 "-Dhttps.nonProxyHosts=*.example.com|*.example.net" -Dhttp.proxyHost=192.0.2.4 -Dhttp.proxyPort=8080 "-Dhttps.nonProxyHosts=*.example.com|*.example.net"
```

Here we have a few values that require the pipe `|` as a separator. This requires quoting to correctly process as a value and not a command line pipe. This works in 3.9.9.

MNG-8248 (https://github.com/apache/maven/issues/10391) made a change that causes the `mvnw.cmd` script to attempt to prepend a value to `MAVEN_OPTS`:

```
set "MAVEN_OPTS=--enable-native-access=ALL-UNNAMED %MAVEN_OPTS%"
```

Unfortunately, the style used here wraps `%MAVEN_OPTS%` in further quotes, and due to the nature of Windows batch processing the `%%` variable is substituted very early, before the quotes are processed. Effectively, this transforms the above example so that the pipes are now *outside* the quotes, which results in the error:

```
'*.example.net' is not recognized as an internal or external command,
operable program or batch file.
```

---

Specifically, this seems to be the responsible section:

https://github.com/apache/maven/blob/d9f0a4828a49e90de198addb5ce620c6369d6671/apache-maven/src/bin/mvn.cmd#L175-L179

---

I can think of a few potential ways to fix this. I'm unsure which one would be most suitable for this project, so I'll list them here:

1. Don't quote the `set` arguments. Simply using `set MAVEN_OPTS=--enable-native-access=ALL-UNNAMED %MAVEN_OPTS%` instead seems to work, though I can't guarantee that the lack of quoting won't cause other issues. Also, the other `set`s in the same file seem to use the same quoted style so perhaps this is in some style guide somewhere..?

2. Introduce a new variable and don't manipulate `MAVEN_OPTS`. e.g. `set "USE_NATIVE_ACCESS=--enable-native-access=ALL-UNNAMED"` and later insert `%USE_NATIVE_ACCESS%` into the same command ahead of `MAVEN_OPTS`, which would result in the same final command. Alternatively, introduce some kind of `EXTRA_OPTS` variable that is internal to the file -- the key here is to not do anything that would parse `MAVEN_OPTS` (or other external variables) before they are used.

3. Use delayed expansion instead. `setlocal EnableDelayedExpansion` and then use `set "MAVEN_OPTS=--enable-native-access=ALL-UNNAMED !MAVEN_OPTS!"` is fine, because it expands the inner `MAVEN_OPTS` after parse where `|` has special meaning.

BobVul:
Fixes #10937 by introducing an additional INTERNAL_MAVEN_OPTS for any arguments that need to be inserted by the script. Parsing the externally-defined MAVEN_OPTS variable can lead to incorrect processing of quotes and special characters, so use the separate variable to avoid doing so.

Specifically, this fixes this case:

```
$env:MAVEN_OPTS='"-Dfoo=bar|baz"'
mvn
```

by implementing proposed fix 2 from #10937

---

### DIFF
--- a/apache-maven/src/bin/mvn.cmd
+++ b/apache-maven/src/bin/mvn.cmd
@@ -175,11 +175,12 @@ set CLASSWORLDS_LAUNCHER=org.codehaus.plexus.classworlds.launcher.Launcher
 @REM MNG-8248
 "%JAVACMD%" --enable-native-access=ALL-UNNAMED -version >nul 2>&1
 if ERRORLEVEL 1 goto skipEnableNativeAccess
-set "MAVEN_OPTS=--enable-native-access=ALL-UNNAMED %MAVEN_OPTS%"
+set "INTERNAL_MAVEN_OPTS=--enable-native-access=ALL-UNNAMED %INTERNAL_MAVEN_OPTS%"
 :skipEnableNativeAccess
 
 "%JAVACMD%" ^
   %JVM_CONFIG_MAVEN_PROPS% ^
+  %INTERNAL_MAVEN_OPTS% ^
   %MAVEN_OPTS% ^
   %MAVEN_DEBUG_OPTS% ^
   -classpath %CLASSWORLDS_JAR% ^
--- a/apache-maven/src/bin/mvn.cmd
+++ b/apache-maven/src/bin/mvn.cmd
@@ -33,6 +33,10 @@
 @REM enable echoing my setting MAVEN_BATCH_ECHO to 'on'
 @if "%MAVEN_BATCH_ECHO%"=="on" echo %MAVEN_BATCH_ECHO%
 
+@REM Clear/define a variable for any options to be inserted via script
+@REM We want to avoid trying to parse the external MAVEN_OPTS variable
+SET INTERNAL_MAVEN_OPTS=
+
 @REM Execute a user defined script before this one
 if not "%MAVEN_SKIP_RC%"=="" goto skipRcPre
 @REM check for pre script, once with legacy .bat ending and once with .cmd ending


apache-maven/src/bin/mvn.cmd
```diff
@@ -175,11 +175,12 @@ set CLASSWORLDS_LAUNCHER=org.codehaus.plexus.classworlds.launcher.Launcher
 @REM MNG-8248
 "%JAVACMD%" --enable-native-access=ALL-UNNAMED -version >nul 2>&1
 if ERRORLEVEL 1 goto skipEnableNativeAccess
-set "MAVEN_OPTS=--enable-native-access=ALL-UNNAMED %MAVEN_OPTS%"
+set "INTERNAL_MAVEN_OPTS=--enable-native-access=ALL-UNNAMED %INTERNAL_MAVEN_OPTS%"
```
gnodet, line 182: I don't think we should expose this new variable.  Which means we need to set it to an empty string before trying the line 176, so that a value coming from the user env would not interfere.

BobVul: Good point, I've pushed a change to clear this variable at the top of the script.

gnodet (approved):

PR #2213, opened 2025-03-31
[MNG-4559] Support spaces in jvm.config

gnodet:
JIRA issue: [MNG-4559](https://issues.apache.org/jira/browse/MNG-4559)

The .mvn/jvm.config file parsing in mvn.cmd needed improvements to properly handle multiple JVM arguments, especially when dealing with comments, quotes and spaces in values.

This PR:
- Ensures proper parsing of multiple JVM arguments from .mvn/jvm.config
- Maintains proper spacing between arguments
- Handles both inline and full-line comments correctly
- Fixes a bug in the space trimming logic (%%x -> %%i)
- Preserves spaces within quoted values

The changes are validated by MavenITmng4559MultipleJvmArgsTest which verifies:
- Multiple JVM arguments are properly handled
- Comments are correctly processed
- Arguments with spaces in quotes are preserved
- Arguments from multiple lines are properly combined

PR #10970, opened 2025-07-21
Avoid parsing MAVEN_OPTS (master/4.x)

### LINKED ISSUE #10937: Quoted pipes in MAVEN_OPTS throw errors in 3.9.10+, 4.0.0-rc

BobVul:
Fixes #10937 by introducing an additional INTERNAL_MAVEN_OPTS for any arguments that need to be inserted by the script. Parsing the externally-defined MAVEN_OPTS variable can lead to incorrect processing of quotes and special characters, so use the separate variable to avoid doing so.

Additionally JVM_CONFIG_MAVEN_OPTS is introduced as its own variable to preserve the append behaviour.

Specifically, this fixes this case:

```
$env:MAVEN_OPTS='"-Dfoo=bar|baz"'
mkdir .mvn
New-Item .mvn\jvm.config
mvn
```

by implementing proposed fix 2 from https://github.com/apache/maven/issues/10937

The related fix for 3.9.x branch is in PR #10969

---

### DIFF
--- a/its/core-it-suite/src/test/java/org/apache/maven/it/MavenITgh10937QuotedPipesInMavenOptsTest.java
+++ b/its/core-it-suite/src/test/java/org/apache/maven/it/MavenITgh10937QuotedPipesInMavenOptsTest.java
@@ -1,0 +1,55 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.maven.it;
+
+import java.nio.file.Path;
+import java.util.Properties;
+
+import org.junit.jupiter.api.Test;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+/**
+ * This is a test set for <a href="https://github.com/apache/maven/issues/10937">gh-10937</a>.
+ */
+class MavenITgh10937QuotedPipesInMavenOptsTest extends AbstractMavenIntegrationTestCase {
+
+    MavenITgh10937QuotedPipesInMavenOptsTest() {
+        super("[3.0.0,)");
+    }
+
+    /**
+     *  Verify the dependency management of the consumer POM is computed correctly
+     */
+    @Test
+    void testIt() throws Exception {
+        Path basedir =
+                extractResources("/gh-10937-pipes-maven-opts").getAbsoluteFile().toPath();
+
+        Verifier verifier = newVerifier(basedir.toString());
+        verifier.setEnvironmentVariable("MAVEN_OPTS", "-Dprop.maven-opts=\"foo|bar\"");
+        verifier.addCliArguments("validate");
+        verifier.execute();
+        verifier.verifyErrorFreeLog();
+
+        Properties props = verifier.loadProperties("target/pom.properties");
+        assertEquals("foo|bar", props.getProperty("project.properties.pom.prop.jvm-opts"));
+        assertEquals("foo|bar", props.getProperty("project.properties.pom.prop.maven-opts"));
+    }
+}
--- a/apache-maven/src/assembly/maven/bin/mvn
+++ b/apache-maven/src/assembly/maven/bin/mvn
@@ -171,20 +171,18 @@ concat_lines() {
     # First convert all CR to LF using tr
     tr '\r' '\n' < "$1" | \
     sed -e '/^$/d' -e 's/#.*$//' | \
+    # Replace LF with NUL for xargs
+    tr '\n' '\0' | \
     # Split into words and process each argument
-    xargs -n 1 | \
+    # Use -0 with NUL to avoid special behaviour on quotes
+    xargs -n 1 -0 | \
     while read -r arg; do
       # Replace variables first
       arg=$(echo "$arg" | sed \
         -e "s@\${MAVEN_PROJECTBASEDIR}@$MAVEN_PROJECTBASEDIR@g" \
         -e "s@\$MAVEN_PROJECTBASEDIR@$MAVEN_PROJECTBASEDIR@g")
 
-      # Add quotes only if argument contains spaces and isn't already quoted
-      if echo "$arg" | grep -q " " && ! echo "$arg" | grep -q "^\".*\"$"; then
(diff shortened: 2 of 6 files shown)


its/core-it-suite/src/test/java/org/apache/maven/it/MavenITgh10937QuotedPipesInMavenOptsTest.java
```diff
@@ -0,0 +1,55 @@
...
+import java.util.Properties;
+
+import org.junit.jupiter.api.Test;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+/**
+ * This is a test set for <a href="https://github.com/apache/maven/issues/10937">gh-10937</a>.
+ */
+class MavenITgh10937QuotedPipesInMavenOptsTest extends AbstractMavenIntegrationTestCase {
```
slawekjaranowski, line 31: it is not similar to `MavenITmng4559SpacesInJvmOptsTest` if yes - we can simple extend it instead of creating next test
We can add description to test that also check case from issue xxx

BobVul: They're very similar, yea. I used the 4559 one as a template.

However, the 4559 test is only defined in 4.x. When I create the tests for 3.9.x, should I then keep that set separate?

slawekjaranowski: ok, we can have a new test

BobVul:
Looks like in 4.x the bash script also changed, such that it now `eval`s the command and introduces a similar pipe parsing error in a completely different way.

The changes made here (purely to the Windows .cmd!) should have no impact on that, but the new tests do expose that issue.

I'll consolidate findings/discussion back in issue #10937 but I'm honestly not too sure what to do here -- complex bash scripts are not my forte.

BobVul:
@gnodet Since you made the original change in #2213, could I run this by you, as I'm a bit unsure.

The existing behaviour is:
1. The change there updated `concat_lines` to perform more advanced parsing of `jvm.config`.
2. In the process it runs the lines through `xargs`, which processes quotes specially, in the process stripping them from the input
3. Later, quotes are added back in if the line is not surrounded by them and contains spaces (i.e. first and last char only)

But the behaviour seen with pipes specifically is that the quotes are being stripped but not added back in. i.e. `-Dopt="foo|bar"` is transformed into `-Dopt=foo|bar` and causing problems on Linux. I can see two ways of handling this:

1. Add pipes to the "check for spaces" bit, which will then re-surround the entire line in quotes. This would probably work but seems a tad hacky?
2. Change out the `xargs -n 1` command for `tr '\n' '\0' | xargs -n 1 -0`, i.e. use xargs in NUL mode. This prevents xargs from handling quotes specially, and therefore quotes are not stripped from the input. However, this also means the code that re-adds quotes must be removed else it will cause e.g. `-Dopt="foo bar"` to become `"-Dopt="foo bar""` which effectively leaves the space unquoted again (much like the original issue on Windows!). I think this is probably fine since in this case `xargs` won't strip initial quotes, though it does force the user to add quotes manually.

I'm leaning towards using option 2, but I'm not familiar enough with either this area of code nor bash to be certain that it won't cause other issues. Do you have any suggestions here?

Proposed new `concat_lines`:

```bash
concat_lines() {
  if [ -f "$1" ]; then
    # First convert all CR to LF using tr
    tr '\r' '\n' < "$1" | \
    sed -e '/^$/d' -e 's/#.*$//' | \
    # Replace LF with NUL for xargs
    tr '\n' '\0' | \
    # Split into words and process each argument
    # Use -0 with NUL to avoid special behaviour on quotes
    xargs -n 1 -0 | \
    while read -r arg; do
      # Replace variables first
      arg=$(echo "$arg" | sed \
        -e "s@\${MAVEN_PROJECTBASEDIR}@$MAVEN_PROJECTBASEDIR@g" \
        -e "s@\$MAVEN_PROJECTBASEDIR@$MAVEN_PROJECTBASEDIR@g")

      echo "$arg"
    done | \
    tr '\n' ' '
  fi
}
```

gnodet (approved):
LGTM, thx for this nice patch ! I'll do the backport to `maven-4.0.x` branch.

slawekjaranowski:
> Should the 3.9.x IT also be merged? [apache/maven-integration-testing#407](https://github.com/apache/maven-integration-testing/pull/407)

should 😄 done

PR #10972, opened 2025-07-21
Add missing namespaces to integration tests

elharo:


elharo (commented):
PTAL. All tests passing. Includes an explicit test for the no namespace case.

gnodet (approved):

PR #10975, opened 2025-07-21
Improve ProjectBuildingException error messages with detailed problem reporting

### LINKED ISSUE #10951: DefaultProjectBuilder needs better excepiton messages

About line 496 we have:

                throw new ProjectBuildingException(results);

This needs to accumulate the actual problems from the results and put them in the exception message. As is, I had to drop into a debugger to see what was failing.

PasanAbeysekara:
fixes #10951

## Description

Enhanced `ProjectBuildingException` to provide detailed error messages instead of the generic "Some problems were encountered while processing the POMs" message. This improvement helps developers quickly identify and fix build problems without needing to use a debugger.

## What was changed

- Modified the `ProjectBuildingException(List<ProjectBuildingResult> results)` constructor to generate comprehensive error messages
- Added a new `createMessage(List<ProjectBuildingResult> results)` method that:
  - Counts total problems and errors across all project results
  - Provides a summary with error count (e.g., "3 problems were encountered while processing the POMs (2 errors)")
  - Lists detailed problems for each project with severity, message, and location information
  - Shows precise source locations (file, line, column) when available
  - Prioritizes errors over warnings for cleaner output when both are present
  - Handles edge cases gracefully (missing project IDs, file paths, etc.)

## Outcome

**Before:**

```
Some problems were encountered while processing the POMs
```

**After:**

```
3 problems were encountered while processing the POMs (2 errors):

[com.example:project1:1.0] [ERROR] Missing required dependency @ pom.xml, line 25, column 10 [ERROR] Invalid version format @ pom.xml, line 30, column 5

[com.example:project2:1.0] [WARNING] Deprecated plugin usage @ pom.xml, line 15, column 3
```

This enhancement significantly improves the developer experience by providing actionable error information without requiring debugging tools.

## Checklist

## Notes

- The change is backward compatible and doesn't affect existing API contracts
- Error message format follows Maven's established conventions for problem reporting
- Implementation handles edge cases and null safety appropriately

### DIFF
--- a/impl/maven-core/src/main/java/org/apache/maven/project/ProjectBuildingException.java
+++ b/impl/maven-core/src/main/java/org/apache/maven/project/ProjectBuildingException.java
@@ -125,7 +125,11 @@ public class ProjectBuildingException extends Exception {
         buffer.append("encountered while processing the POMs");
 
         if (errorProblems > 0) {
-            buffer.append(" (").append(errorProblems).append(" ").append(errorProblems > 1 ? "errors" : "error").append(")");
+            buffer.append(" (")
+                    .append(errorProblems)
+                    .append(" ")
+                    .append(errorProblems > 1 ? "errors" : "error")
+                    .append(")");
         }
 
         buffer.append(":\n");


impl/maven-core/src/main/java/org/apache/maven/project/ProjectBuildingException.java
```diff
@@ -97,4 +97,80 @@ private static String createMessage(String message, String projectId, File pomFi
         }
         return buffer.toString();
     }
-}
+
+    @SuppressWarnings("deprecation")
```
elharo: why is this warning suppressed? That shouldn't be necessary

impl/maven-core/src/main/java/org/apache/maven/project/ProjectBuildingException.java
```diff
@@ -97,4 +97,80 @@ private static String createMessage(String message, String projectId, File pomFi
...
+        if (results == null || results.isEmpty()) {
+            return "Some problems were encountered while processing the POMs";
+        }
+
+        long totalProblems =
+                results.stream().mapToLong(r -> r.getProblems().size()).sum();
+
+        long errorProblems = results.stream()
+                .flatMap(r -> r.getProblems().stream())
+                .filter(p -> p.getSeverity() != org.apache.maven.model.building.ModelProblem.Severity.WARNING)
```
elharo: do you really need the full package qualified name here?

impl/maven-core/src/main/java/org/apache/maven/project/ProjectBuildingException.java
PasanAbeysekara: `@SuppressWarnings("deprecation")` is necessary because we must work with the legacy ModelProblem API which is deprecated since 4.0.0, but this compatibility code needs to continue supporting it.

impl/maven-core/src/main/java/org/apache/maven/project/ProjectBuildingException.java
PasanAbeysekara: Not really, I've imported `import org.apache.maven.model.building.ModelProblem;` package instead

impl/maven-core/src/main/java/org/apache/maven/project/ProjectBuildingException.java
```diff
@@ -97,4 +99,79 @@ private static String createMessage(String message, String projectId, File pomFi
...
         return buffer.toString();
     }
+
+    @SuppressWarnings("deprecation")
+    private static String createMessage(List<ProjectBuildingResult> results) {
+        if (results == null || results.isEmpty()) {
+            return "Some problems were encountered while processing the POMs";
+        }
+
+        long totalProblems =
```
elharo: I don't know if it matters, but I suspect a single for loop would be faster here than two streams

impl/maven-core/src/main/java/org/apache/maven/project/ProjectBuildingException.java
elharo: what specifically is the problem? I'm not convinced you need to call deprecated methods here. If you do, those methods maybe shouldn't be deprecated in the first place. (Also, please don't resolve comments)

PasanAbeysekara: image

https://maven.apache.org/ref/4.0.0-beta-5/maven-model-builder/apidocs/org/apache/maven/model/building/ModelProblem.html

according to this `ModelProblem` Interface is deprecated

impl/maven-core/src/main/java/org/apache/maven/project/ProjectBuildingException.java
PasanAbeysekara: Sure, Single loop would be faster

impl/maven-core/src/main/java/org/apache/maven/project/ProjectBuildingException.java
elharo: That probably shouldn't have happened. In either case, don't suppress the warning here.

impl/maven-core/src/main/java/org/apache/maven/project/ProjectBuildingException.java
```diff
@@ -97,4 +99,84 @@ private static String createMessage(String message, String projectId, File pomFi
...
+            }
+        }
+
+        StringBuilder buffer = new StringBuilder(1024);
+        buffer.append(totalProblems);
+        buffer.append(totalProblems == 1 ? " problem was " : " problems were ");
+        buffer.append("encountered while processing the POMs");
+
+        if (errorProblems > 0) {
+            buffer.append(" (").append(errorProblems).append(" error");
```
gnodet:
```suggestion
            buffer.append(" (").append(errorProblems).append(" ").append(errorProblems > 1 ? "errors" : "error").append(")");
```
and remove the 4 lines below

PasanAbeysekara: Addressed the changes

gnodet:
Do you have a project to reproduce those problems ?
If I use an invalid pom, the ModelValidator raises exceptions which are displayed already:
```
[ERROR]
[ERROR]   The project org.apache.maven:maven:${{4.1.0-SNAPSHOT (/Users/gnodet/work/maven/pom.xml) has 1 error
[ERROR]     'version' must be a constant version but is '${{4.1.0-SNAPSHOT'. @ line 31, column 3
[ERROR]
[ERROR]
```

PasanAbeysekara:
> @PasanAbeysekara do you have a project to reproduce those problems ? If I use an invalid pom, the ModelValidator raises exceptions which are displayed already:
> 
> ```
> [INFO] Scanning for projects...
> [ERROR] Some problems were encountered while processing the POMs
> [ERROR] The build could not read 1 project -> [Help 1]
> [ERROR]
> [ERROR]   The project org.apache.maven:maven:${{4.1.0-SNAPSHOT (/Users/gnodet/work/maven/pom.xml) has 1 error
> [ERROR]     'version' must be a constant version but is '${{4.1.0-SNAPSHOT'. @ line 31, column 3
> [ERROR]
> [ERROR] To see the full stack trace of the errors, re-run Maven with the '-e' switch
> [ERROR] Re-run Maven using the '-X' switch to enable verbose output
> [ERROR]
> [ERROR] For more information about the errors and possible solutions, please read the following articles:
> [ERROR] [Help 1] http://cwiki.apache.org/confluence/display/MAVEN/ProjectBuildingException
> ```

@gnodet  The error message you showed is from Maven's CLI during normal builds - and yes, that already works great and shows detailed errors. This PR is different though.

What I'm fixing here is the `ProjectBuildingException` that gets thrown when you use Maven's `ProjectBuilder` API programmatically. Before my changes, if you caught this exception in your code, `getMessage()` would just return:

`Some problems were encountered while processing the POMs`


That's it. No details about what actually went wrong. 
After my changes, the same exception now gives you something useful like:

`2 problems were encountered while processing the POMs:`
`The build could not read 1 project -> org.apache.maven:maven:4.1.0-SNAPSHOT: 1 error, 1 warning`
`[ERROR] 'version' must be a constant version but is '${4.1.0-SNAPSHOT' @ pom.xml, line 31, column 3 [WARNING] Some deprecated usage found @ pom.xml, line 45, column 5`


**To reproduce this**:
```
try {
    ProjectBuilder projectBuilder = // ... get from container
    ProjectBuildingRequest request = // ... configure request
    ProjectBuildingResult result = projectBuilder.build(pomFile, request);
} catch (ProjectBuildingException e) {
    // Before: e.getMessage() would be generic
    // After: e.getMessage() contains detailed problem information
    System.out.println(e.getMessage()); // Now shows detailed errors
}
```

elharo (approved):

PR #10984, opened 2025-07-22
Undeprecate ModelProblem since it's used by non-deprecated public API

elharo:


gnodet (commented):
I raised https://github.com/apache/maven/pull/10990 as an alternative proposal.

PR #10991, opened 2025-07-23
Improve dependency scope validation error messages for import scope

gnodet:
## Summary

This PR improves Maven's dependency scope validation error messages to provide clearer guidance when the `import` scope is used incorrectly.

## Problem

Currently, when a user incorrectly uses `import` scope in a regular `<dependencies>` section, Maven shows a confusing error message:

```
'dependencies.dependency.scope' must be one of [provided, compile, runtime, test, system] but is 'import'.
```

This message is misleading because it suggests that `import` scope is never valid, when in fact it **is** valid in `<dependencyManagement>` sections with `<type>pom</type>`.

## Solution

The PR enhances the error message to provide context-aware guidance:

```
'dependencies.dependency.scope' has scope 'import'. The 'import' scope is only valid in <dependencyManagement> sections.
```

### Key Improvements:

1. **Context-aware validation**: Different error messages for regular dependencies vs. dependency management
2. **Clear guidance**: Explicitly tells users where `import` scope is valid
3. **Focused messaging**: Removes confusing list of valid scopes and focuses on the specific problem
4. **Consistent implementation**: Applied to both Maven 3 (compat) and Maven 4 (impl) for consistency

### Core Changes
- **Added `validateDependencyScope()` method** in both implementations that provides context-aware error messages
- **Enhanced error messages** with specific guidance for `import` scope misuse
- **Maintained backward compatibility** - only error messages are improved, no functional changes

### Files Modified
- `compat/maven-model-builder/src/main/java/org/apache/maven/model/validation/DefaultModelValidator.java`
- `compat/maven-model-builder/src/test/java/org/apache/maven/model/validation/DefaultModelValidatorTest.java`
- `impl/maven-impl/src/main/java/org/apache/maven/impl/model/DefaultModelValidator.java`
- `impl/maven-impl/src/test/java/org/apache/maven/impl/model/DefaultModelValidatorTest.java`

### Additional Improvements
- Fixed grammar in comments (`don't` → `not`)
- Updated tests to verify the improved error messages
- Applied spotless formatting for code consistency

## Example Impact

This change helps users quickly understand how to fix their Maven configuration:

**❌ Incorrect usage (triggers new helpful error):**
```xml
<dependencies>
  <dependency>
    <artifactId>bom</artifactId>
    <version>1.0</version>
    <type>pom</type>
    <scope>import</scope>  <!-- ❌ Invalid here -->
  </dependency>
</dependencies>
```

**✅ Correct usage:**
```xml
<dependencyManagement>
  <dependencies>
    <dependency>
      <artifactId>bom</artifactId>
      <version>1.0</version>
      <type>pom</type>
      <scope>import</scope>  <!-- ✅ Valid here -->
    </dependency>
  </dependencies>
</dependencyManagement>
```

## Related Issues

Addresses the confusion reported in https://github.com/faktorips/faktorips.base/issues/70

## Testing

- ✅ All existing tests pass (161 tests, 0 failures)
- ✅ Updated tests verify the improved error messages
- ✅ Code compiles successfully with JDK 21
- ✅ Spotless formatting applied
- ✅ No checkstyle violations

The validation logic remains unchanged - only the error messages are enhanced to be more helpful and educational.

### DIFF
--- a/compat/maven-model-builder/src/main/java/org/apache/maven/model/validation/DefaultModelValidator.java
+++ b/compat/maven-model-builder/src/main/java/org/apache/maven/model/validation/DefaultModelValidator.java
@@ -785,10 +785,10 @@ public class DefaultModelValidator implements ModelValidator {
                             prefix, "version", problems, errOn30, Version.V20, d.getVersion(), d.getManagementKey(), d);
 
                     /*
-                     * TODO Extensions like Flex Mojos use custom scopes like "merged", "internal", "external", etc. In
-                     * order to don't break backward-compat with those, only warn but don't error out.
+                     * Extensions like Flex Mojos use custom scopes like "merged", "internal", "external", etc. In
+                     * order to not break backward-compat with those, only warn but don't error out.
                      */
-                    validateEnum(
+                    validateDependencyScope(
                             prefix,
                             "scope",
                             problems,
@@ -797,15 +797,11 @@ public class DefaultModelValidator implements ModelValidator {
                             d.getScope(),
                             d.getManagementKey(),
                             d,
-                            "provided",
-                            "compile",
-                            "runtime",
-                            "test",
-                            "system");
+                            false);
 
                     validateEffectiveModelAgainstDependency(prefix, problems, m, d, request);
                 } else {
-                    validateEnum(
+                    validateDependencyScope(
                             prefix,
                             "scope",
                             problems,
@@ -814,12 +810,7 @@ public class DefaultModelValidator implements ModelValidator {
                             d.getScope(),
                             d.getManagementKey(),
                             d,
-                            "provided",
-                            "compile",
-                            "runtime",
-                            "test",
-                            "system",
-                            "import");
+                            true);
                 }
             }
         }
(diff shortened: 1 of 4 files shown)


IMPROVEMENT_SUMMARY.md
```diff
@@ -0,0 +1,79 @@
+# Maven Dependency Scope Validation Improvement
```
elharo: interesting file. Is this part of some new process? It does seem strange that it's at the root of the repo instead of somehow tied to the specific issue.

IMPROVEMENT_SUMMARY.md
```diff
@@ -0,0 +1,79 @@
...
+Improved the error message to provide more helpful guidance when the `import` scope is used incorrectly:
+
+### Before
+```
+'dependencies.dependency.scope' must be one of [provided, compile, runtime, test, system] but is 'import'.
+```
+
+### After
+```
+'dependencies.dependency.scope' must be one of [provided, compile, runtime, test, system] but is 'import'. The 'import' scope is only valid in <dependencyManagement> sections and requires <type>pom</type>.
```
elharo: I would drop the valid values completely and focus on the problem. That is, something like:

dependency _GAV_ has scope import. The 'import' scope is only valid in the <dependencyManagement> section.

or 

dependency _GAV_ on line XXX has scope import but the type is jar. The import scope requires type pom.

compat/maven-model-builder/src/main/java/org/apache/maven/model/validation/DefaultModelValidator.java
```diff
@@ -788,7 +788,7 @@ private void validateEffectiveDependencies(
                      * TODO Extensions like Flex Mojos use custom scopes like "merged", "internal", "external", etc. In
                      * order to don't break backward-compat with those, only warn but don't error out.
```
elharo: don't --> not

and is this really a todo or just a comment?

IMPROVEMENT_SUMMARY.md
gnodet: It's a temp file, I'll remove it.

Issue #8960, opened 2023-12-08
[MNG-7955] Plugins should be allowed some control over the imported packages


jira-importer:
**[Guillaume Nodet](https://issues.apache.org/jira/secure/ViewProfile.jspa?name=gnt)** opened **[MNG-7955](https://issues.apache.org/jira/browse/MNG-7955?redirect=false)** and commented

We need some simple configuration to allow plugins to indicate which packages / artifacts they need to be provided.  This could solve the problems with `org.slf4j` package.

I think the defaults could be the following packages:
* org.apache.maven.api,org.apache.maven.api.*
* org.slf4j
* jarkata.inject

 

See [https://github.com/apache/maven/pull/1336](https://github.com/apache/maven/pull/1336/files#diff-9600f93ce04acabe017b026dc1ebbb445e8cb363bb2677a764a783ab85cf804aR106-R108) which is related.

---

**Issue Links:**
- [MNG-7921](https://issues.apache.org/jira/browse/MNG-7921) Allow plugins to control classpath extensions

- [MNG-7954](https://issues.apache.org/jira/browse/MNG-7954) Provide a cleaner DI api

- [MNG-8176](https://issues.apache.org/jira/browse/MNG-8176) Restrict class loaders for Maven 4 plugins

PR #11029, opened 2025-08-07
Internalize plexus-classworlds and add JPMS module support (fixes #11028)

### LINKED ISSUE #11028: Upgrade to JLine 4.x and Replace Broad Native Access with Targeted Module Access

# Upgrade to JLine 4.x and Replace Broad Native Access with Targeted Module Access

## Summary

Maven currently uses the broad `--enable-native-access=ALL-UNNAMED` JVM flag to enable native access for JLine's terminal functionality. This approach grants native access to all unnamed modules, which is overly permissive from a security perspective. We should upgrade to JLine 4.x (when available) and replace this with the more targeted `--enable-native-access=org.jline.terminal.ffm` to follow the principle of least privilege.

## Current State

**JLine Version**: 3.30.4  
**Native Access Configuration**: `--enable-native-access=ALL-UNNAMED`  
**Location**: 
- `apache-maven/src/assembly/maven/bin/mvn` (line 245)
- `apache-maven/src/assembly/maven/bin/mvn.cmd` (line 259)

```bash
cmd="\"$JAVACMD\" \
  $MAVEN_OPTS \
  $MAVEN_DEBUG_OPTS \
  --enable-native-access=ALL-UNNAMED \
  -classpath \"$LAUNCHER_JAR\" \
  \"-Dclassworlds.conf=$CLASSWORLDS_CONF\" \
  \"-Dmaven.home=$MAVEN_HOME\" \
  \"-Dmaven.mainClass=$MAVEN_MAIN_CLASS\" \
  \"-Dlibrary.jline.path=${MAVEN_HOME}/lib/jline-native\" \
  \"-Dmaven.multiModuleProjectDirectory=$MAVEN_PROJECTBASEDIR\" \
  $LAUNCHER_CLASS \
  $MAVEN_ARGS"
```

### 1. Upgrade to JLine 4.x

JLine 4.x will include comprehensive JPMS (Java Platform Module System) support via [PR #1374](https://github.com/jline/jline3/pull/1374), which:

- Adds proper `module-info.java` files for all JLine modules
- Provides the `org.jline.terminal.ffm` module for Foreign Function & Memory API support
- Enables targeted native access permissions
- Maintains backward compatibility with classpath-based applications

### 2. Replace Broad Native Access Flag

**Current**: `--enable-native-access=ALL-UNNAMED`  
**Proposed**: `--enable-native-access=org.jline.terminal.ffm`

This change:
- ✅ Follows the principle of least privilege
- ✅ Reduces security attack surface
- ✅ Provides the same functionality with better encapsulation
- ✅ Aligns with Java platform security best practices

### 3. Module Path Migration

Moving JLine to the module path (instead of classpath) will require updates to **plexus-classworlds** to support:
- Module path configuration in `m2.conf`
- Mixed classpath/module path scenarios
- Proper module resolution and loading

```
[plexus.core]
load       ${maven.conf}/logging
optionally ${maven.home}/lib/ext/redisson/*.jar
optionally ${maven.home}/lib/ext/hazelcast/*.jar
optionally ${user.home}/.m2/ext/*.jar
optionally ${maven.home}/lib/ext/*.jar
load       ${maven.home}/lib/maven-*.jar
load       ${maven.home}/lib/*.jar
```

## Benefits

1. **Enhanced Security**: Targeted native access reduces the attack surface
2. **Better Encapsulation**: JPMS provides strong module boundaries
3. **Future-Ready**: Prepared for modern Java features (Project Loom, Panama)
4. **Performance**: Module system enables JVM optimizations
5. **Maintainability**: Clear module dependencies prevent classpath issues

## Implementation Plan

1. **Phase 1**: Wait for JLine 4.0 release with JPMS support
2. **Phase 2**: Upgrade Maven's JLine dependency to 4.x
3. **Phase 3**: Update plexus-classworlds to support module path
4. **Phase 4**: Modify Maven launcher scripts to use targeted native access
5. **Phase 5**: Update `m2.conf` to place JLine modules on module path

## Compatibility

- **Backward Compatible**: Existing Maven installations continue to work
- **Java Version**: Requires Java 9+ for module system (already required by Maven 4.x)
- **Plugin Compatibility**: No impact on Maven plugins or user projects

## Related Work

- **JLine JPMS Support**: https://github.com/jline/jline3/pull/1374
- **JLine 4.0 Milestone**: https://github.com/jline/jline3/milestone/4
- **Current JLine Usage**: Maven uses multiple JLine modules (terminal, reader, style, builtins, console, etc.)

## Acceptance Criteria

- [ ] JLine 4.x dependency integrated
- [ ] `--enable-native-access=org.jline.terminal.ffm` replaces `ALL-UNNAMED`
- [ ] JLine modules moved to module path
- [ ] plexus-classworlds supports module path configuration
- [ ] All Maven functionality works (terminal input, colors, progress, etc.)
- [ ] Integration tests pass
- [ ] Documentation updated

---

**Priority**: Medium  
**Effort**: Large (requires coordination with JLine 4.x release and plexus-classworlds updates)  
**Security Impact**: Positive (reduces native access scope)

gnodet:
## Summary

Replaces the external `plexus-classworlds` dependency with an internalized `maven-classworlds` module, adds JPMS module support for Maven's runtime (targeted `--enable-native-access` for JLine), and introduces per-plugin JPMS `ModuleLayer` support for modular plugins.

### Internalize plexus-classworlds
- Copy plexus-classworlds source into `impl/maven-classworlds`, maintaining original package structure
- Extract `org.apache.maven.api.classworlds` into a separate API module (`api/maven-api-classworlds`)
- Replace all `plexus-classworlds` dependencies across the build with `maven-classworlds`
- Fully eliminate plexus-classworlds from the dependency tree

### Targeted `--enable-native-access` for JLine (fixes #11028)
- Move JLine jars to `lib/modules/` and load them on the boot `--module-path` via a new `module` directive in `m2.conf`
- Add `--add-modules ALL-MODULE-PATH` to resolve modules on the module-path
- Replace `--enable-native-access=ALL-UNNAMED` with targeted `--enable-native-access=org.jline.terminal.ffm,org.jline.nativ`
- Add Java version detection in launcher scripts — skip module flags on JDKs that don't support them

### Runtime ModuleLayer with module access control
- Add `add-exports`, `add-opens`, `add-reads` directives to `m2.conf` configuration parser
- Create a runtime `ModuleLayer` in `Configurator` for modules not already in the boot layer, with `Controller`-based access control
- Store `ModuleLayer` and `Controller` on `ClassWorld` for realm-level access
- Add `ClassRealm.addExports()`, `ClassRealm.addOpens()`, and `ClassRealm.addReads()` methods for plugins/extensions to control module access at runtime

### Per-plugin JPMS ModuleLayer support for modular plugins
- Add `<modular>true</modular>` boolean field to the plugin descriptor model (`plugin.mdo`, version 1.2.0+)
- Add `isModular()` and `getModuleLayer()` default methods to the `ClassRealm` API interface
- Store per-realm `ModuleLayer` and `ModuleLayer.Controller` on `ClassRealm` (not `ClassWorld`) — each modular plugin gets its own isolated layer
- New `createModularPluginRealm()` in `DefaultClassRealmManager`:
  - Collects plugin artifact paths, creates `ModuleFinder`
  - Uses parent layer from `ClassWorld` (runtime layer if available, else boot layer)
  - Calls `resolveAndBind()` for module resolution + service provider binding
  - Creates layer with `defineModulesWithOneLoader()` — one classloader per plugin, isolation between plugins
  - Flat layer hierarchy — all plugin layers are siblings under the boot/runtime layer
- **Hard error, no fallback**: `ModuleLayerCreationException` if the module graph cannot be resolved (split packages, missing `module-info.class`, etc.)
- `DefaultMavenPluginManager` branches on `isModular()`:
  - Modular plugins: loaded via `ModuleLayer`, TCCL set to layer's classloader, DI via `maven-di` `Injector.discover()` only (no Plexus/Sisu component discovery)
  - Classic plugins: unchanged classpath-based `ClassRealm` loading with full Sisu support
- Layer classloader is closed when the realm is disposed

#### `m2.conf` directives

| Directive | Syntax | Description |
|-----------|--------|-------------|
| `module` | `module <glob>` | Load JARs on the JPMS module-path (used for JLine) |
| `add-exports` | `add-exports <module>/<package>=<target>` | Export a package from a named module to a target (defaults to `ALL-UNNAMED`) |
| `add-opens` | `add-opens <module>/<package>=<target>` | Open a package for deep reflection |
| `add-reads` | `add-reads <module>=<target>` | Add a reads edge between modules |

Example (`m2.conf`):
```
module ${maven.home}/lib/modules/*.jar
```

#### `META-INF/maven/module-access` descriptor

Plugins and extensions can declare module access requirements via a `META-INF/maven/module-access` resource. This is processed for all realm types (plugin, extension, project) and for modular plugin realms.

```
# Export a package from a named module to the plugin's classloader
add-exports org.jline.terminal/org.jline.terminal.spi
# Open a package for deep reflection
add-opens org.jline.reader/org.jline.reader.impl
# Add a reads edge
add-reads org.jline.terminal=ALL-UNNAMED
```

The direction is: packages are exported/opened **from** a named module in the boot/runtime layer **to** the plugin's classloader (unnamed module). This is semantically identical to JVM `--add-exports` / `--add-opens` flags.

#### Plugin descriptor (`plugin.xml`)

```xml
<plugin>
  <!-- ... -->
  <modular>true</modular>  <!-- opt-in to JPMS ModuleLayer loading (default: false) -->
</plugin>
```

## Files changed (key files)

| File | Change |
|------|--------|
| `apache-maven/src/assembly/maven/bin/mvn` | Module-path, add-modules, targeted enable-native-access |
| `apache-maven/src/assembly/maven/bin/mvn.cmd` | Same for Windows |
| `apache-maven/src/assembly/maven/bin/m2.conf` | `module` directive for JLine jars |
| `impl/maven-classworlds/` | Full internalized classworlds implementation |
| `api/maven-api-classworlds/` | New classworlds API module with `isModular()`, `getModuleLayer()` |
| `api/maven-api-plugin/src/main/mdo/plugin.mdo` | `modular` boolean field (version 1.2.0+) |
| `ConfigurationParser.java` | Parse `module`, `add-exports`, `add-opens`, `add-reads` directives |
| `Configurator.java` | Runtime ModuleLayer creation and directive application |
| `ClassWorld.java` | ModuleLayer + Controller storage (boot/runtime layer) |
| `ClassRealm.java` (impl) | Per-realm ModuleLayer + Controller, `addExports()`, `addOpens()`, `addReads()`, `close()` cleanup |
| `ClassRealm.java` (API) | `isModular()`, `getModuleLayer()` default methods |
| `ClassRealmManager.java` | New `createModularPluginRealm()` interface method |
| `DefaultClassRealmManager.java` | `createModularPluginRealm()` implementation + `META-INF/maven/module-access` processing |
| `ModuleLayerCreationException.java` | Hard error for modular plugin loading failures |
| `DefaultMavenPluginManager.java` | Modular vs classic plugin branching, TCCL, DI discovery |
| `PluginDescriptor.java` | `modular` field, getter/setter |
| `PluginDescriptorBuilder.java` | `extractModular()` parser method |

## Test plan

- [x] `mvn test -pl impl/maven-classworlds` — 93 tests pass
- [x] `mvn test -pl compat/maven-plugin-api` — 2 tests pass
- [x] `mvn test -pl impl/maven-core` — 562 tests pass
- [x] `mvn install -DskipTests` — full build succeeds

Closes #11029
Fixes #11028


gnodet:
> Should we maybe at the same time archive/EOL plexus-classworlds?
> Also, is this a fork of it? Or rewrite? As if some third-party may use plexus-classworlds, is this new module drop in replacement?

This is currently a fork, and I removed the old layer already.  

> I'm not sure if we need copy it to Maven core ... Why not simply update `plexus-classworlds`?
> 
> Especially that old java package and old licenses header are preserved

This was really a first step.  I was toying a bit to solve a few limitations:
* clean integration within the Maven 4 API
* enable JPMS to load some bits of maven-core (in particular JLine to get rid of the `--enable-native-access=ALL-UNNAMED`)
* better plugin class loader customisation (https://github.com/apache/maven/issues/8960, not started)
* enable JPMS-enabled plugins (haven't really given that much thoughts)

One important point is that extensions should be able to provide controlled access to some API, which could be consumed by plugins.  Plugins themselves could export some packages, I think we had some use cases for that (we currently need an extension to do that).  

So not sure yet, this is just experiments... but ideas welcomed.  Overall, I think `plexus-classworlds` won't be able to solve the above problems, and given Maven is really the only user afaik, I'm tempted to merge ...

api/maven-api-classworlds/src/main/java/org/apache/maven/api/classworlds/ClassRealm.java
```diff
@@ -0,0 +1,223 @@
...
+import java.net.URL;
+
+import org.apache.maven.api.annotations.Experimental;
+import org.apache.maven.api.annotations.Nonnull;
+import org.apache.maven.api.annotations.Nullable;
+
+/**
+ * A class loading realm that provides isolated class loading with controlled imports and exports.
+ * <p>
+ * A ClassRealm represents an isolated class loading environment with its own classpath
```
desruisseaux, line 31: It seems to duplicate Java module boundaries. Is it a workaround because Maven is not fully modularized yet? (i.e., it still have dependencies on the class-path instead of the module-path).

gnodet: The ClassRealm concept is not a duplication of JPMS boundaries — the two address different problems and are complementary here.

ClassRealms are Maven's existing dynamic class-loading isolation mechanism (inherited from Plexus classworlds, ~20 years old). They exist because Maven loads plugins and extensions dynamically at runtime, each with its own dependency tree and potentially conflicting transitive dependencies. This is fundamentally a runtime-dynamic problem: we don't know which plugins will be loaded until we read the POM being built. JPMS resolves its module graph statically at startup, so it can't replace this.

The `addExports`/`addOpens`/`addReads` methods on `ClassRealm` are specifically the **bridge** between the two worlds: since plugin classes live in realms (classpath-based, unnamed module), these methods use the `ModuleLayer.Controller` we create internally to grant access from named JPMS modules to the plugin's classloader. For example, when a plugin needs to access JLine's `org.jline.terminal.spi` package, `addExports("org.jline.terminal", "org.jline.terminal.spi")` uses the controller to open that access to the realm's unnamed module.

And yes, we do create a `ModuleLayer` internally (in `Configurator.createModuleLayer()`) for jars placed on the module-path (like JLine), but it's not exposed in the API. The API only exposes the realm-level `addExports`/`addOpens`/`addReads` as a simpler abstraction for plugins to declare what module access they need — typically via `META-INF/maven/module-access` descriptors rather than direct API calls.

So to answer directly: it's not a workaround for Maven not being modularized. Even if all Maven jars had `module-info.java`, we'd still need dynamic class isolation for plugins with conflicting dependencies, and we'd still need a bridge mechanism for plugins to access named modules at runtime.

impl/maven-core/src/main/java/org/apache/maven/classrealm/DefaultClassRealmManager.java
```diff
@@ -364,4 +373,65 @@ private static Object getId(ClassLoader classLoader) {
         }
         return classLoader;
     }
+
+    private static final String MODULE_ACCESS_DESCRIPTOR = "META-INF/maven/module-access";
+
+    private void applyModuleAccessDescriptors(ClassRealm classRealm) {
```
desruisseaux, line 479: Do we have a documentation somewhere of what this code is doing? It seems to open or export packages of a module, but I'm not sure in which direction:

* If `META-INF/maven/module-access` exports package from the module that we are loading to other modules, why not using the standard `module-info` mechanism?
* If `META-INF/maven/module-access` exports package from another module to the module that we are loading, it breaks encapsulation principles. Like allowing class _A_ to decide what is private inside class _B_.

impl/maven-classworlds/src/test/java/org/codehaus/plexus/classworlds/realm/ClassRealmImplTest.java
```diff
@@ -0,0 +1,482 @@
...
+    @Test
+    void testLocateSourceRealmSimpleImport() throws Exception {
+        ClassRealm mainRealm = this.world.newRealm("main");
+
+        ClassRealm werkflowRealm = this.world.newRealm("werkflow");
+
+        mainRealm.importFrom("werkflow", "com.werken.werkflow");
+
+        assertSame(werkflowRealm, mainRealm.getImportClassLoader("com.werken.werkflow.WerkflowEngine"));
+
```
desruisseaux: It is not very important, but all those blank lines consume a large number of space for no apparent reason.

desruisseaux (changes requested):
My main concern is this `META-INF/maven/module-access` file parsed by the `applyModuleAccessDescriptors(ClassRealm)` method in `impl/maven-core/src/main/java/org/apache/maven/classrealm/DefaultClassRealmManager.java`. I would like to see a documentation about what is the intent, which module is exporting to which module, and why it is not done in `module-info`.

api/maven-api-classworlds/src/main/java/org/apache/maven/api/classworlds/ClassRealm.java
desruisseaux: >  JPMS resolves its module graph statically at startup, so it can't replace this.

Java Module can load modules dynamically too, with steps like below:

```java
Path[]        files   = ...;    // Paths to the JAR files of the plugins and dependencies to load.
URL[]         urls    = ...;    // Above paths converted to URLs.
ModuleFinder  modules = ModuleFinder.of(files);
Class<?>      core    = ...;    // Some Maven core class.
ClassLoader   loader  = new URLClassLoader(urls, core.getClassLoader());
ModuleLayer   parent  = core.getModule().getLayer();
Configuration config  = parent.configuration().resolveAndBind(ModuleFinder.of(), modules, getAllModuleNames(modules));
ModuleLayer   layer   = parent.defineModulesWithOneLoader(config, loader);
```

With the following helper method:

```java
/**
 * Returns the names of all modules that the given finder can see.
 */
private static Set<String> getAllModuleNames(final ModuleFinder modules) {
    final var names = new HashSet<String>();
    for (ModuleReference ref : modules.findAll()) {
        names.add(ref.descriptor().name());
    }
    return names;
}
```

Then we can discover services (e.g., MOJO implementations) provided by the plugin:

```java
for (MyService service : ServiceLoader.load(layer, MyService.class)) {
    // ...
}
```

Does something like that has been tried?

gnodet: Yes! This is precisely the approach we just implemented for modular plugins in this PR — see \`DefaultClassRealmManager.createModularPluginRealm()\`. The code is almost identical to your sketch:

```java
ModuleFinder finder = ModuleFinder.of(modulePaths.toArray(new Path[0]));
Set<String> moduleNames = finder.findAll().stream()
        .map(ref -> ref.descriptor().name())
        .collect(Collectors.toSet());

ModuleLayer parentLayer =
        implWorld.getModuleLayer() != null ? implWorld.getModuleLayer() : ModuleLayer.boot();

Configuration cfg = parentLayer.configuration().resolveAndBind(ModuleFinder.of(), finder, moduleNames);

ModuleLayer.Controller controller =
        ModuleLayer.defineModulesWithOneLoader(cfg, List.of(parentLayer), parent);
```

A few design decisions worth noting:

- **Opt-in via plugin descriptor**: Plugin authors declare `<modular>true</modular>` in their `plugin.xml`. This is a hard contract — if the module graph can't be resolved (split packages, missing `module-info.class`, etc.), the build fails with `ModuleLayerCreationException`. No fallback to classpath loading.

- **Flat layer hierarchy**: All plugin layers are siblings under the same parent layer (the runtime layer from `ClassWorld` if available, or the boot layer). No nesting.

- **DI instead of ServiceLoader**: Rather than `ServiceLoader.load(layer, MyService.class)`, we use Maven's `maven-di` `Injector.discover()` with the layer's classloader. This integrates modular plugins into Maven's existing injection infrastructure. No Plexus/Sisu for modular plugins.

- **Per-realm storage**: The `ModuleLayer` and `Controller` are stored on the `ClassRealm`, not on `ClassWorld` — each plugin gets its own isolated layer/loader, and the layer's classloader is closed when the realm is disposed.

So yes, dynamic module layers work well for this. Thanks for the suggestion — it validated the direction we were already heading.

impl/maven-core/src/main/java/org/apache/maven/classrealm/DefaultClassRealmManager.java
gnodet: Good question. The direction is your second option: `META-INF/maven/module-access` opens/exports packages **from** a named module in the boot/runtime layer **to** the plugin's classloader (which lives in the unnamed module).

A concrete example: JLine ships as named modules (`org.jline.terminal`, `org.jline.reader`, etc.) in Maven's `lib/modules/` directory. A plugin running on the classpath (unnamed module) that needs to interact with JLine's SPI would include:

```
# META-INF/maven/module-access
add-exports org.jline.terminal/org.jline.terminal.spi=ALL-UNNAMED
add-opens org.jline.terminal/org.jline.terminal.impl=ALL-UNNAMED
```

This is semantically identical to passing `--add-exports org.jline.terminal/org.jline.terminal.spi=ALL-UNNAMED` on the JVM command line — it's the standard JPMS mechanism for granting the unnamed module access to packages from named modules. We just read it from a descriptor instead of requiring JVM flags.

As for why the standard `module-info` mechanism doesn't work here:
- The plugin doesn't own the module it needs access to (e.g., `org.jline.terminal`), so it can't add `exports` to someone else's module descriptor
- `qualified exports` in `module-info.java` require knowing the target module name at compile time, but the plugin is on the classpath (unnamed module, no name to export to)
- The only standard way to do this at runtime is `--add-exports` / `Module.addExports()` via a `Controller` — which is exactly what the implementation calls under the hood

You're right that this lets class A decide what's visible inside class B's module. That's inherent to the problem: plugins need access to packages that named modules don't export to `ALL-UNNAMED` by default. The same tension exists with `--add-exports` on the JVM command line. The `module-access` descriptor at least makes this explicit and declarative — Maven logs a debug message for each directive applied, so it's auditable.

For modular plugins (the new `<modular>true</modular>` path), this is less of an issue because the plugin itself is a named module and can use proper `requires` directives in its own `module-info.java`. The `module-access` mechanism is primarily needed for classic (classpath-based) plugins that need to interact with modularized libraries.

impl/maven-classworlds/src/test/java/org/codehaus/plexus/classworlds/realm/ClassRealmImplTest.java
gnodet: Agreed — these came from the original plexus-classworlds fork. Will clean up.

gnodet:
All review comments addressed in the latest push:

**@copilot-pull-request-reviewer:**
- Fixed invalid `-ea:pkg1:pkg2` argLine → split into `-ea:org.codehaus.classworlds... -ea:org.codehaus.plexus.classworlds...`
- Fixed Javadoc typos in `ConfigurationException` ("configuraton"/"occured")
- Fixed `LICENSE.vm` to also map `maven-api-classworlds` → `boot/` directory
- Fixed JLine dual-copy: added `org.jline:*` exclusion from the `lib/` dependencySet so JLine only lands in `lib/modules/`

**@desruisseaux:**
- Added comprehensive Javadoc to `applyModuleAccessDescriptors` explaining the purpose, direction of access (named boot-layer module → plugin unnamed module, not the reverse), file format, and why `module-info.java` cannot be used instead
- The `ClassRealm` API question: no further action needed — the discussion in the thread explains why ClassRealms and JPMS `ModuleLayer`s are complementary, not duplicates

**Fragile classloader lookups (gnodet-bot):**
- `getEffectiveClassLoader`: replaced `iterator().next().getClassLoader()` with `layer.findLoader(anyModuleName)` via `stream().findFirst()`
- `close()`: same fix — replaced loop-with-break with `stream().findFirst().ifPresent(...)`
- `CacheKey` missing `modular` flag: added field + hash + equals; kept old constructor and added `default` overload on the `PluginRealmCache` interface to preserve binary compatibility (japicmp was catching this)

Issue #11067, opened 2025-08-25
Different and unexpected behavour on invoking goals


cstamas:
### Affected version

3.9.11 (probably all)

### Bug description

Spotted this when using `wrapper:wrapper`, as Maven downloads ALL plugins (!), only to then execute wrapper plugin.
Cases:
* mvn wrapper:wrapper -- this will download all plugins it finds in POM only to execute wrapper
* mvn org.apache.maven.plugins:maven-wrapper-plugin:wrapper -- this works as expected

In first case, the problem is DefaultLifecycleTaskSegmentCalculator calling LifecyclePluginResolver that in turn calls DefaultPluginVersionResolver and is getting plugin descriptors to detect compatibility for ALL plugins, only to then execute wrapper plugin.

This was already reported as #10128

slawekjaranowski:
I think that `org.apache.maven.plugin.prefix.internal.DefaultPluginPrefixResolver` is responsible for it ....

PR #11071, opened 2025-08-28
Reduce plugin downloads during prefix resolution via two-pass heuristic

gnodet:
This PR introduces a conservative optimization in DefaultPluginPrefixResolver to avoid resolving/downloading all declared plugins when resolving a goal prefix.

Summary:
- Two-pass project scan: first, resolve only likely candidates based on artifactId-derived prefix (e.g., maven-compiler-plugin → compiler); then fall back to original full scan if needed, preserving compatibility with custom goalPrefix values.
- Adds unit tests (DefaultPluginPrefixResolverTest) validating that:
  * Only the candidate plugin descriptor is loaded for conventional artifactIds
  * Fallback full scan still resolves custom goalPrefix

Motivation:
As reported in #11067, resolving plugin prefixes could trigger downloads of every declared plugin because the implementation loads each plugin descriptor to check goalPrefix. This change significantly reduces that cost in typical projects without changing resolution precedence.

Notes:
- No behavior change intended other than reduced unnecessary descriptor loads.
- Build and formatting: spotless:apply executed locally; targeted unit test executed successfully.

---

### DIFF
--- a/impl/maven-core/src/main/java/org/apache/maven/plugin/prefix/internal/DefaultPluginPrefixResolver.java
+++ b/impl/maven-core/src/main/java/org/apache/maven/plugin/prefix/internal/DefaultPluginPrefixResolver.java
@@ -129,12 +129,42 @@ public class DefaultPluginPrefixResolver implements PluginPrefixResolver {
     }
 
     private PluginPrefixResult resolveFromProject(PluginPrefixRequest request, List<Plugin> plugins) {
+        // First pass: narrow down to likely candidates based on artifactId to avoid resolving all plugins.
+        List<Plugin> candidates = new ArrayList<>();
+        String wanted = request.getPrefix();
         for (Plugin plugin : plugins) {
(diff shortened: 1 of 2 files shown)


cstamas:
Ah, so prefix was going over all defined plugins, getting descriptor just to select the right one?

cstamas:
What if we change prefix resolver? Right now, this distinction makes not much sense:
https://github.com/apache/maven/blob/b427f467252d1e6f38b7b07b1f156564a38bc864/impl/maven-core/src/main/java/org/apache/maven/plugin/prefix/internal/DefaultPluginPrefixResolver.java#L86-L110

(resolver from project, or just resolve == the endgame is equal to this below)

Proposed change:
* take plugin groupIDs (from settings), make it an ordered set (to retain order from settings)
* if project present, gather plugin groupIDs from it (appending it, hence ordered set)
* iterate over groupId looking for prefix

This would be MUCH faster and intuitive, then iterating over all project plugins for nothing...

PR #11090, opened 2025-09-03
[MNG-11055] Add integration test for Maven 4 DI consistency

### LINKED ISSUE #11055: Maven 4: Obtaining services appears to be inconsistent between unit tests and production.
### Affected version

4.0.0-rc-4

### Bug description

I want to preface this issue by saying I may be doing something wrong. I'm not well-versed in writing Maven plugins.

When experimenting with writing a Maven plugin using Maven 4.0.0-rc-4, I found what appears to be some inconsistencies with how obtaining services (e.g., `OsService`) works during unit testing versus when using the plugin in another project (integration testing/"production").

Using `@Inject` from Maven 4's new Dependency Injection API module (`org.apache.maven:maven-api-di`) works during unit tests, but fails for the same services when the plugin is used in another project. Conversely, using `Session::getService` fails during unit tests but then succeeds for the same services when using the plugin in another project.

| API                             |   Unit test           | Integration test            |
|------------------------|-------------------|--------------------------|
| `Session::getService`  | Fails                    | Succeeds                      |
| `@Inject`                    | Succeeds            | Fails                              |

This problem makes it somewhat hard to test Maven plugins since what works during tests doesn't work once deployed and vice versa.

See https://stackoverflow.com/q/79735022/6395627 and https://gitlab.com/tkslaw/maven4-di-tests for more information, including an example plugin demonstrating the problem.

I can include code here if truly necessary, but the example (linked previously) is not exactly *tiny*.

gnodet:
## Summary

Adds an integration test to validate Maven 4 Dependency Injection consistency between `@Inject` and `Session::getService()` approaches as described in [GH-11055](https://github.com/apache/maven/issues/11055).

## Changes

- **Integration Test**: `MavenITgh11055DITest` that builds a test plugin and validates both DI approaches
- **Test Plugin**: Minimal Maven plugin under `its/core-it-suite/src/test/resources/gh-11055/` with:
  - `GetServiceMojo`: Uses `Session::getService()` approach
  - `InjectServiceMojo`: Uses `@Inject` annotation approach

## Test Behavior

The test validates that both dependency injection approaches work consistently:
1. Builds the test plugin
2. Executes `get-service` goal (Session::getService)
3. Executes `inject-service` goal (@Inject)
4. Expects both to succeed for consistent behavior

## Current Results

In Maven 4.0.0-rc-4, both approaches currently work consistently. The test will serve as:
- Validation tool when the issue is reproduced under different conditions
- Regression test to ensure DI consistency is maintained
- Documentation of expected behavior

## Related

- Fixes #11055
- Integration test only - no functional changes to Maven core

---

### DIFF
--- a/its/core-it-suite/src/test/java/org/apache/maven/it/MavenITgh11055DITest.java
+++ b/its/core-it-suite/src/test/java/org/apache/maven/it/MavenITgh11055DITest.java
@@ -20,11 +20,7 @@ package org.apache.maven.it;
 
 import java.io.File;
 
-import org.apache.maven.shared.verifier.Verifier;
-import org.apache.maven.shared.verifier.VerificationException;
 import org.junit.jupiter.api.Test;
-
-import static org.junit.jupiter.api.Assertions.assertThrows;
 
 /**
  * This is a test set for <a href="https://github.com/apache/maven/issues/11055">gh-11055</a>.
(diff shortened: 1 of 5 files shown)


elharo (approved):

gnodet:
The IT does not actuallly capture the problem correctly.

PR #11091, opened 2025-09-03
Fix #10950: Replace raw RuntimeException on reactor cycle with ProjectBuildingExc…

### LINKED ISSUE #10950: DeafultProjectBuilder throws raw RuntimeException
### Affected version

HEAD

### Bug description

At the very least this should be a specific subclass of runtime exception, but given that this likely indicates a  problem in the input POMs it should probably be a checked exception.

```
                if (cycle != null) {
                    throw new RuntimeException(new ProjectCycleException(
                            "The projects in the reactor contain a cyclic reference: " + cycle.getMessage(),
                            (CycleDetectedException) cycle.getException()));
                }

```

arturobernalg:
…eption.

When a reactor dependency cycle is detected, wrap the underlying org.codehaus.plexus.util.dag.CycleDetectedException in a new ProjectCycleException and throw a checked ProjectBuildingException (as declared by ProjectBuilder) instead of a raw RuntimeException.

Keep non-cycle model problems on the existing ProjectBuildingException(results) path.

- [ X] You have run the [Core IT][core-its] successfully.

### DIFF
--- a/impl/maven-core/src/main/java/org/apache/maven/project/DefaultProjectBuilder.java
+++ b/impl/maven-core/src/main/java/org/apache/maven/project/DefaultProjectBuilder.java
@@ -44,7 +44,6 @@ import java.util.function.Supplier;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
-import org.apache.maven.ProjectCycleException;
 import org.apache.maven.RepositoryUtils;
 import org.apache.maven.api.ArtifactCoordinates;
 import org.apache.maven.api.Language;
@@ -492,10 +491,14 @@ public class DefaultProjectBuilder implements ProjectBuilder {
                         .findAny()
                         .orElse(null);
                 if (cycle != null) {
-                    throw new RuntimeException(new ProjectCycleException(
+                    final CycleDetectedException cde = (CycleDetectedException) cycle.getException();
+                    throw new ProjectBuildingException(
+                            null,
                             "The projects in the reactor contain a cyclic reference: " + cycle.getMessage(),
-                            (CycleDetectedException) cycle.getException()));
+                            null,
+                            cde);
                 }
+
                 throw new ProjectBuildingException(results);
             }
 


impl/maven-core/src/main/java/org/apache/maven/project/DefaultProjectBuilder.java
```diff
@@ -492,10 +492,22 @@ List<ProjectBuildingResult> build(List<File> pomFiles, boolean recursive) throws
                         .findAny()
                         .orElse(null);
                 if (cycle != null) {
-                    throw new RuntimeException(new ProjectCycleException(
+                    final CycleDetectedException cde = (CycleDetectedException) cycle.getException();
+
+                    final ProjectCycleException pce = new ProjectCycleException(
+                            "The projects in the reactor contain a cyclic reference: " + cycle.getMessage(), cde);
+
+                    // If you can resolve these from the problem/result, set them; otherwise leave null.
```
elharo: This comment doesn't seem right. They can't be set here.

impl/maven-core/src/main/java/org/apache/maven/project/DefaultProjectBuilder.java
```diff
@@ -492,10 +492,22 @@ List<ProjectBuildingResult> build(List<File> pomFiles, boolean recursive) throws
...
+                    final String projectId = null;
+                    final File pomFile = null;
+
+                    throw new ProjectBuildingException(
+                            projectId,
                             "The projects in the reactor contain a cyclic reference: " + cycle.getMessage(),
-                            (CycleDetectedException) cycle.getException()));
+                            pomFile,
+                            pce);
```
elharo: I'm not clear why pce was created. Could you just pass the cde here?

gnodet (approved):

impl/maven-core/src/main/java/org/apache/maven/project/DefaultProjectBuilder.java
```diff
@@ -492,10 +491,14 @@ List<ProjectBuildingResult> build(List<File> pomFiles, boolean recursive) throws
                         .findAny()
                         .orElse(null);
                 if (cycle != null) {
-                    throw new RuntimeException(new ProjectCycleException(
+                    final CycleDetectedException cde = (CycleDetectedException) cycle.getException();
```
elharo, line 494: The wrapping still smells like something's not quite right. The code is quite complex, and it might be simpler without the lambda, but either way we really should be able to just throw the CycleDetectedException. This is not a public method so we can easily add that to the method signature. It's also possible CycleDetectedException should be a subclass of ProjectBuildingException.

elharo (approved):
I suspect CycleDetectedException should have been a subclass of ProjectBuildingException, but that would require an API change so this is good enough forn now

PR #11093, opened 2025-09-03
Fix 10210 - Decrypt only bona fide v4/legacy tokens in settings

arturobernalg:
Stop probing plain strings like “%{foo}.txt”; guard with strict v4 shape Add unit/smoke tests to prevent regressions

- [ X] Run `mvn verify` to make sure basic checks pass.

### DIFF
--- a/impl/maven-impl/src/main/java/org/apache/maven/impl/DefaultSettingsBuilder.java
+++ b/impl/maven-impl/src/main/java/org/apache/maven/impl/DefaultSettingsBuilder.java
@@ -31,6 +31,7 @@ import java.util.Map;
 import java.util.concurrent.atomic.AtomicInteger;
 import java.util.function.Supplier;
 import java.util.function.UnaryOperator;
+import java.util.regex.Pattern;
 
 import org.apache.maven.api.Constants;
 import org.apache.maven.api.ProtoSession;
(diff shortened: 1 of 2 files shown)


cstamas:
-1

It seems there is more to it. For start, I was surprised that strings like `%{foo}.txt` were attempted to be "decrypted". This is definitely wrong.

Then I went for related code:
https://github.com/codehaus-plexus/plexus-sec-dispatcher/blob/475bfd259eecf60c6f5bbb7450ea7c97118c7f6d/src/main/java/org/codehaus/plexus/components/secdispatcher/internal/DefaultSecDispatcher.java#L180-L197

And tests:
https://github.com/codehaus-plexus/plexus-sec-dispatcher/blob/475bfd259eecf60c6f5bbb7450ea7c97118c7f6d/src/test/java/org/codehaus/plexus/components/secdispatcher/internal/DefaultSecDispatcherTest.java#L132-L170

And started wondering what is wrong with _legacy_ detection, as it is clearly broken. Found related issues:
* https://issues.apache.org/jira/browse/MNG-4082 tried to add escape support, but
* https://issues.apache.org/jira/browse/MNG-4612 shows it is broken still

Hence, as I see, it was original "envisioned" that passwords may be "in the middle" (see UT), and even some escaping support was added 17 years ago, but that does not work.

So my proposal is to stop treating "legacy" as these, and got for `{...}` forms only.

cstamas:
More info:
In Maven 3.x this happens:
https://github.com/codehaus-plexus/plexus-sec-dispatcher/blob/0bab9aa096c411973ce3d48771a8fe735cafd917/src/main/java/org/sonatype/plexus/components/sec/dispatcher/DefaultSecDispatcher.java#L208
that simply delegates to cipher (not exists today):
https://github.com/codehaus-plexus/plexus-cipher/blob/plexus-cipher-2.0/src/main/java/org/sonatype/plexus/components/cipher/DefaultPlexusCipher.java#L102

cstamas:
In other words, my proposal is to support encrypted strings ONLY as 2.0 version of library had:
https://github.com/codehaus-plexus/plexus-sec-dispatcher/blob/0bab9aa096c411973ce3d48771a8fe735cafd917/src/test/java/org/sonatype/plexus/components/sec/dispatcher/SecUtilTest.java#L46

and nothing more.

PR #11102, opened 2025-09-05
Add --processes to list running Maven builds.

### LINKED ISSUE #10627: [MNG-8608] Provide a way to find running maven processes similar to jps command
**[Christoph Läubrich](https://issues.apache.org/jira/secure/ViewProfile.jspa?name=laeubi)** opened **[MNG-8608](https://issues.apache.org/jira/browse/MNG-8608?redirect=false)** and commented

Java provides a [jps command](https://docs.oracle.com/en/java/javase/11/tools/jps.html) to list all running java processes.

It would be useful to have such capability for maven as well, as an example I like to call:

```
mvn -ps
```

and get a list of maven processes running on the system, it should include:

1. the process id
2. the workdirectory
3. the session execution root directory
4. the maven version

How this is achieved, could be platform specific like it is for jps, and it might not discover all maven processes but only those for the current user.

**+Why is this useful needed:+**
* Currently a maven build do not perform any way of locking so one can easily run two maven processes working on the same project tree
* But due to how maven works this can only end in disaster (e.g. one process clean what the other has build) or at best result in some strange state if two maven processes run in parallel
* Currently there is no way for an IDE to prevent interference with running maven processes that work on the same file tree

Because of this maven should detect other processes running in the same file-tree and either wait or refuse execution.

---
No further details from [MNG-8608](https://issues.apache.org/jira/browse/MNG-8608?redirect=false)

arturobernalg:
Implements a per-user registry under ~/.m2/.maven/runs (create on start, remove on exit). Pure Java 17 using ProcessHandle; no jps or vendor tools.

- [ X] Run `mvn verify` to make sure basic checks pass.

Fixes https://github.com/apache/maven/issues/10627

### DIFF
--- a/impl/maven-cli/src/main/java/org/apache/maven/cling/invoker/CommonsCliOptions.java
+++ b/impl/maven-cli/src/main/java/org/apache/maven/cling/invoker/CommonsCliOptions.java
@@ -228,6 +228,13 @@ public class CommonsCliOptions implements Options {
         return Optional.empty();
     }
 
+    public Optional<Boolean> processes() {
+        if (commandLine.hasOption(CLIManager.PROCESSES) || commandLine.hasOption(CLIManager.PROCESSES_SHORT)) {
+            return Optional.of(Boolean.TRUE);
+        }
+        return Optional.empty();
+    }
+
     @Override
     public void warnAboutDeprecatedOptions(ParserRequest request, Consumer<String> printWriter) {
         if (cliManager.getUsedDeprecatedOptions().isEmpty()) {
@@ -318,6 +325,10 @@ public class CommonsCliOptions implements Options {
         public static final String OFFLINE = "o";
         public static final String HELP = "h";
 
+        // List running Maven processes (MNG-8608). No short form: -ps already used.
+        public static final String PROCESSES = "processes";
+        public static final String PROCESSES_SHORT = "p";
+
         // Not an Option: used only for early detection, when CLI args may not be even parsed
         public static final String SHOW_ERRORS_CLI_ARG = "-" + SHOW_ERRORS;
 
@@ -348,6 +359,14 @@ public class CommonsCliOptions implements Options {
             options.addOption(Option.builder(HELP)
                     .longOpt("help")
                     .desc("Display help information")
+                    .build());
+            options.addOption(Option.builder()
+                    .longOpt(PROCESSES)
+                    .desc("List running Maven processes for the current user")
+                    .build());
+            options.addOption(Option.builder(PROCESSES_SHORT)
+                    .longOpt(PROCESSES)
+                    .desc("List running Maven processes and exit")
                     .build());
             options.addOption(Option.builder(USER_PROPERTY)
                     .numberOfArgs(2)
--- a/impl/maven-cli/src/main/java/org/apache/maven/cling/invoker/ProcessRuns.java
+++ b/impl/maven-cli/src/main/java/org/apache/maven/cling/invoker/ProcessRuns.java
@@ -142,7 +142,7 @@ public final class ProcessRuns {
         for (final Run r : runs) {
             sb.append(String.format(
                     "%-10d %-12s %-24s %-48s %-48s%s",
-                    r.pid, r.version, r.started, ell(r.workDir, 48), ell(r.execRoot, 48), nl));
+                    r.pid, r.version, r.started, truncateEnd(r.workDir, 48), truncateEnd(r.execRoot, 48), nl));
         }
         return sb.toString();
     }
@@ -155,7 +155,7 @@ public final class ProcessRuns {
         }
     }
 
-    private static String ell(final String s, final int max) {
+    private static String truncateEnd(final String s, final int max) {
         if (s == null || s.length() <= max) {
             return s == null ? "?" : s;
         }
--- a/impl/maven-cli/src/main/java/org/apache/maven/cling/invoker/ProcessRuns.java
+++ b/impl/maven-cli/src/main/java/org/apache/maven/cling/invoker/ProcessRuns.java
@@ -1,0 +1,164 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.maven.cling.invoker;
+
+import java.io.Reader;
+import java.io.Writer;
+import java.nio.charset.StandardCharsets;
+import java.nio.file.DirectoryStream;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.Paths;
+import java.time.Instant;
+import java.util.Comparator;
+import java.util.Properties;
+
+public final class ProcessRuns {
+
+    private static final String RUNS_SUBDIR = ".m2/.maven/runs";
+    private static final String FILE_PREFIX = "mvn-";
+    private static final String FILE_SUFFIX = ".properties";
+
+    private static final String KEY_PID = "pid";
+    private static final String KEY_VERSION = "version";
+    private static final String KEY_WORKDIR = "workDir";
+    private static final String KEY_EXECROOT = "execRoot";
+    private static final String KEY_STARTED = "started";
+
+    private static Path runsDir() {
+        final String home = System.getProperty("user.home", ".");
+        return Paths.get(home).resolve(RUNS_SUBDIR);
+    }
+
+    private static Path desc(final long pid) {
+        return runsDir().resolve(FILE_PREFIX + pid + FILE_SUFFIX);
+    }
+
+    public static void install(final long pid, final String version, final Path workDir, final Path execRoot) {
+        try {
+            Files.createDirectories(runsDir());
+            final Properties p = new Properties();
+            p.setProperty(KEY_PID, Long.toString(pid));
+            p.setProperty(KEY_VERSION, version == null ? "-" : version);
+            p.setProperty(
+                    KEY_WORKDIR,
+                    workDir == null ? "?" : workDir.toAbsolutePath().toString());
+            p.setProperty(
+                    KEY_EXECROOT,
+                    execRoot == null ? "?" : execRoot.toAbsolutePath().toString());
+            p.setProperty(KEY_STARTED, Instant.now().toString());
+            try (Writer w = Files.newBufferedWriter(desc(pid), StandardCharsets.UTF_8)) {
+                p.store(w, "mvn run");
+            }
+        } catch (final Exception ignored) {
+            // best-effort
+        }
+    }
+
+    public static void uninstall(final long pid) {
+        try {
+            Files.deleteIfExists(desc(pid));
+        } catch (final Exception ignored) {
+            // best-effort
+        }
+    }
+
+    private static final class Run {
+        final long pid;
+        final String version;
+        final String workDir;
+        final String execRoot;
+        final String started;
+
+        Run(final long pid, final String version, final String workDir, final String execRoot, final String started) {
+            this.pid = pid;
+            this.version = version;
+            this.workDir = workDir;
+            this.execRoot = execRoot;
+            this.started = started;
+        }
+    }
+
+    public static java.util.List<Run> listAlive() {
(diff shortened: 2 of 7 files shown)


impl/maven-cli/src/main/java/org/apache/maven/cling/invoker/ProcessRuns.java
```diff
@@ -0,0 +1,164 @@
...
+
+    private static long parseLong(final String s, final long dflt) {
+        try {
+            return Long.parseLong(s);
+        } catch (final Exception e) {
+            return dflt;
+        }
+    }
+
+    private static String ell(final String s, final int max) {
```
gnodet: `ell` -> `truncateEnd`

gnodet:
I think we're missing a `mvnsh` integration. This can be done by adding a command to the builtin command registry:
https://github.com/apache/maven/blob/bcf98382a62f2445a3c65fec5a9c680424414e18/impl/maven-cli/src/main/java/org/apache/maven/cling/invoker/mvnsh/builtin/BuiltinShellCommandRegistryFactory.java#L94

arturobernalg:
> I think we're missing a `mvnsh` integration. This can be done by adding a command to the builtin command registry:
> 
> https://github.com/apache/maven/blob/bcf98382a62f2445a3c65fec5a9c680424414e18/impl/maven-cli/src/main/java/org/apache/maven/cling/invoker/mvnsh/builtin/BuiltinShellCommandRegistryFactory.java#L94

please do another passs

impl/maven-cli/src/main/java/org/apache/maven/cling/invoker/CommonsCliOptions.java
```diff
@@ -349,6 +360,14 @@ protected void prepareOptions(org.apache.commons.cli.Options options) {
                     .longOpt("help")
                     .desc("Display help information")
                     .build());
+            options.addOption(Option.builder()
+                    .longOpt(PROCESSES)
+                    .desc("List running Maven processes for the current user")
+                    .build());
+            options.addOption(Option.builder(PROCESSES_SHORT)
```
gnodet, line 367: Two options are registered with the same `longOpt(PROCESSES)` (the long-only one on line 363 and the short+long one here). Commons CLI does not support duplicate long option names -- the second `addOption` call will silently overwrite the first, making the long-only version dead code. You should register a single option.

Moreover, the comment on line 327 says "No short form: -ps already used" but then `PROCESSES_SHORT = "p"` is defined and used. I would suggest dropping the short option entirely and keeping only the `--processes` long form:

```suggestion
            options.addOption(Option.builder()
                    .longOpt(PROCESSES)
                    .desc("List running Maven processes for the current user and exit")
                    .build());
```

impl/maven-cli/src/main/java/org/apache/maven/cling/invoker/mvn/MavenInvoker.java
```diff
@@ -613,4 +615,26 @@ protected void logSummary(
...
         }
     }
+
+    @Override
+    protected void preCommands(final MavenContext context) throws Exception {
+        super.preCommands(context);
+
+        if (context.options().processes().orElse(false)) {
+            final String out = ProcessRuns.format(ProcessRuns.listAlive());
+            System.out.print(out);
```
gnodet, line 625: Using `System.out.print` bypasses the Maven output infrastructure. The established pattern in `LookupInvoker.helpOrVersionAndMayExit()` is to use `determineWriter(context)` for early-exit output. This should follow the same approach.

Additionally, the `--processes` check is conceptually an early-exit flag like `--help` and `--version`. It would be more appropriate to handle it in `helpOrVersionAndMayExit()` rather than `preCommands()`, keeping the process registration separate.

impl/maven-cli/src/main/java/org/apache/maven/cling/invoker/ProcessRuns.java
```diff
@@ -0,0 +1,164 @@
...
+                final long pid = parseLong(p.getProperty(KEY_PID), -1L);
+                final boolean alive = pid > 0
+                        && ProcessHandle.of(pid).map(ProcessHandle::isAlive).orElse(false);
+                if (alive) {
+                    out.add(new Run(
+                            pid,
+                            p.getProperty(KEY_VERSION, "-"),
+                            p.getProperty(KEY_WORKDIR, "?"),
+                            p.getProperty(KEY_EXECROOT, "?"),
+                            p.getProperty(KEY_STARTED, "?")));
```
gnodet, line 119: `ProcessHandle.isAlive()` only checks whether the PID exists on the system -- after a reboot or PID recycling, a stale descriptor file could match a completely different process. Consider comparing the recorded `started` timestamp against `ProcessHandle.info().startInstant()` to verify that the process identity matches, not just the PID.

impl/maven-cli/src/main/java/org/apache/maven/cling/invoker/mvnsh/builtin/BuiltinShellCommandRegistryFactory.java
```diff
@@ -270,6 +273,22 @@ private List<Completer> mvnupCompleter(String name) {
...
+        /**
+         * mvnsh command: print running Maven processes (delegates to `mvn --processes`).
+         */
+        private void ps(CommandInput input) {
+            try {
+                shellMavenInvoker.invoke(mavenParser.parseInvocation(ParserRequest.mvn(
+                                new String[] {"--processes"}, shellContext.invokerRequest.messageBuilderFactory())
+                        .cwd(shellContext.cwd.get())
+                        .build()));
+            } catch (InvokerException.ExitException e) {
```
gnodet, line 286: The `--processes` handler throws `ExitException(0)` on success, so this catch block will always log an error message for a successful `ps` command. For exit code 0, this should not be logged as an error. Consider:
```java
} catch (InvokerException.ExitException e) {
    if (e.getExitCode() != 0) {
        shellContext.logger.error("ps command exited with exit code " + e.getExitCode());
    }
}
```

impl/maven-cli/src/main/java/org/apache/maven/cling/invoker/ProcessRuns.java
```diff
@@ -0,0 +1,164 @@
...
+    private static Path runsDir() {
+        final String home = System.getProperty("user.home", ".");
+        return Paths.get(home).resolve(RUNS_SUBDIR);
+    }
+
+    private static Path desc(final long pid) {
+        return runsDir().resolve(FILE_PREFIX + pid + FILE_SUFFIX);
+    }
+
+    public static void install(final long pid, final String version, final Path workDir, final Path execRoot) {
```
gnodet, line 53: The runs directory is derived from `System.getProperty("user.home")` which couples this class to global JVM state and makes testing harder (the test has to mutate `user.home`). Consider accepting the base directory as a parameter to `install`/`uninstall`/`listAlive`, or at minimum a package-private overload for testability.

impl/maven-cli/src/main/java/org/apache/maven/cling/invoker/CommonsCliOptions.java
```diff
@@ -228,6 +228,13 @@ public Optional<Boolean> help() {
         return Optional.empty();
     }
 
+    public Optional<Boolean> processes() {
```
gnodet, line 231: The check `commandLine.hasOption(CLIManager.PROCESSES) || commandLine.hasOption(CLIManager.PROCESSES_SHORT)` is redundant when both options share the same `longOpt`. `hasOption` with the long name will match regardless of whether the user specified the short or long form. If you keep only the long form (as suggested), this simplifies to just `commandLine.hasOption(CLIManager.PROCESSES)`.

gnodet (changes requested):
Thanks for the contribution, Arturo. The feature idea is useful, but there are several issues that need to be addressed before this can be merged.

**Critical: Duplicate option registration will cause a runtime error**

In `CommonsCliOptions.CLIManager.prepareOptions()`, two `Option` instances are added with the same `longOpt("processes")`. Commons CLI does not allow duplicate long option names -- this will silently overwrite the first option, or throw depending on the version. You should register a single option, either with or without a short form.

**Critical: Short option `-p` conflict**

The comment says "No short form: -ps already used" but then the code defines `PROCESSES_SHORT = "p"` and registers `Option.builder(PROCESSES_SHORT).longOpt(PROCESSES)`. The `-ps` option is used for `ALTERNATE_PROJECT_SETTINGS`. Using `-p` as a short option for processes is risky because it is a prefix of `-ps` and `-pl`, which could cause ambiguity depending on how commons-cli handles partial matching. I would recommend dropping the short form entirely -- `--processes` is clear enough for an infrequent diagnostic flag.

**Output should use `determineWriter(context)` instead of `System.out.print`**

The existing pattern for early-exit commands (help, version) uses `determineWriter(context)` to route output through the proper writer infrastructure. The `--processes` handler should follow the same pattern rather than writing directly to `System.out`.

**Placement: should be in `helpOrVersionAndMayExit` not `preCommands`**

The `--processes` flag is semantically similar to `--help` and `--version`: it prints some information and exits. It would be more consistent to handle it in `helpOrVersionAndMayExit()` or in the `Options`-level early exit path, rather than in `preCommands()` which also installs the process registry (mixing the two concerns).

**Shell `ps` command logs exit code 0 as an error**

In `BuiltinShellCommandRegistryFactory`, the `ps()` method catches `ExitException` and logs it as an error (`shellContext.logger.error("ps command exited with exit code " + ...)`). But `ExitException(0)` represents a successful exit -- the `--processes` handler throws it to signal normal termination. This will print a confusing error message for a successful operation. I note the same anti-pattern exists in the `mvn()` handler, but for `ps` this is the *only* execution path, so it will always print an error.

**`ProcessHandle.isAlive()` does not verify the process is actually Maven**

The `listAlive()` method checks whether the PID is alive via `ProcessHandle.of(pid).map(ProcessHandle::isAlive)`, but after a system reboot (or PID recycling by the OS), a stale file could match a completely unrelated process. This is a correctness concern: the listing could show non-Maven processes as if they were Maven builds. Consider also checking `ProcessHandle.info().command()` or the process start time against the recorded `started` timestamp to validate the process identity.

**`Run` class is `private` but returned from `public` methods**

The inner class `Run` is declared `private static final class` but is returned from the public method `listAlive()`. While this compiles (the caller is in the same class), it means external callers can only treat the returned list as opaque -- they can iterate over it but cannot access any fields. If the intent is that callers only pass the list to `format()`, consider making `listAlive()` package-private, or combining it with `format()` into a single public method.

**Test modifies global `System.setProperty("user.home")`**

The test `listsCurrentPidFromRegistry` sets `System.setProperty("user.home", ...)` which is global JVM state. This can cause flakiness when tests run in parallel. Consider injecting the runs directory path instead of relying on `System.getProperty("user.home")`.

gnodet (commented):
The feature idea — listing running Maven builds via `--processes` — is useful for diagnostic workflows. However, the implementation has several bugs and design issues that should be addressed:

**1. Duplicate option registration (high)**
Two `Option.builder` calls both use `.longOpt("processes")`. Commons CLI's `Options.addOption()` stores options in a `longOpts` map via `Map.put()`, so the second call silently overwrites the first, making the long-only registration dead code.

**2. Contradictory comment and short-option concern (high)**
The comment says _"No short form: -ps already used"_ but then defines `PROCESSES_SHORT = "p"` and registers it. While commons-cli doesn't do partial matching for short options, the comment-code contradiction is a clear code quality issue.

**3. Long option ambiguity with --projects (high)**
The new `--processes` long option is dangerously similar to the existing `--projects` option. Maven's `DefaultParser` has partial matching enabled by default, so `--pro` would match both options and throw `AmbiguousOptionException`. This could break workflows of users who currently abbreviate `--projects` as `--pro`. Consider a more distinct name such as `--list-processes` or `--running`.

**4. System.out bypasses output infrastructure (medium)**
`System.out.print(out)` bypasses Maven's output infrastructure. The existing early-exit handlers (`--help`, `--version`) use `determineWriter(context)` which respects `--log-file` redirection. This should follow the same pattern.

**5. Wrong method placement (medium)**
The `--processes` flag is semantically an early-exit flag like `--help` and `--version` (print info, then exit). The listing logic should be handled in `helpOrVersionAndMayExit()` rather than `preCommands()`, which currently mixes the listing concern with the process registration concern.

**6. Error logging on success (medium)**
In `BuiltinShellCommandRegistryFactory`, the `ps` command handler throws `ExitException(0)` for a successful listing, but the catch block unconditionally logs it as `shellContext.logger.error(...)`. Every successful `ps` invocation produces a misleading error log line.

**7. PID recycling vulnerability (medium)**
`ProcessHandle.isAlive()` only checks whether a PID exists. After a reboot or PID recycling, a stale descriptor file could match an unrelated process. The `started` timestamp is already recorded — comparing it against `ProcessHandle.info().startInstant()` would verify process identity.

**8. Redundant hasOption check (medium)**
The `hasOption(PROCESSES) || hasOption(PROCESSES_SHORT)` check is redundant — a direct consequence of the duplicate option registration issue (#1).

**Lower severity:** The `ProcessRuns` directory is derived from `System.getProperty("user.home")` which couples the class to global JVM state and makes testing harder (the test mutates `user.home`). Consider accepting the base directory as a parameter.

slachiewicz:
The last CI run on this branch was 2025-09-25, and GitHub has deleted the logs since, so there's no way to tell whether the change still builds against master.

The branch also conflicts with master now, so that needs resolving before CI can say anything. Say the word if you'd like a hand with it.

PR #11107, opened 2025-09-09
Fix ReactorReader to prefer consumer POMs over build POMs

gnodet:
## Problem

When building modules in isolation (outside of the full reactor), the ReactorReader would read the build POM from the project-local repository instead of the consumer POM. Build POMs can contain invalid elements like `<parent />` that are only valid during the build process, causing parsing errors when used as dependency POMs.

This results in warnings like:
```
[WARNING] The POM for org.apache.maven.its.gh11084:a:jar:1.0.0-SNAPSHOT is invalid, transitive dependencies (if any) will not be available: 2 problems were encountered while building the effective model
    - [FATAL] 'parent.groupId' is missing. @ line 21, column 3
    - [FATAL] 'parent.artifactId' is missing. @ line 21, column 3
```

## Solution

Modified the `findModel` method in the `ReactorReader` class to:

1. **First check the current reactor** - If the project is in the current reactor session, use the model from the project (existing behavior)
2. **Then check the project-local repository** - If the project is not in the current reactor, look for a consumer POM in the project-local repository
3. **Prefer consumer POMs** - Use the same logic as `findInProjectLocalRepository` to prefer consumer POMs over build POMs

## Benefits

- **Eliminates invalid POM warnings** when building modules in isolation
- **Maintains backward compatibility** - existing behavior for reactor builds is unchanged
- **Follows the existing pattern** - uses the same consumer POM preference logic as `findInProjectLocalRepository`
- **Robust error handling** - gracefully falls back if the consumer POM cannot be read

## Testing

The fix has been verified with the integration test scenario `MavenITgh11084ReactorReaderPreferConsumerPomTest` which now passes without invalid POM warnings.

Fixes #11084

---

### DIFF
--- a/impl/maven-core/src/main/java/org/apache/maven/ReactorReader.java
+++ b/impl/maven-core/src/main/java/org/apache/maven/ReactorReader.java
@@ -151,7 +151,29 @@ class ReactorReader implements MavenWorkspaceReader {
     @Override
     public Model findModel(Artifact artifact) {
         MavenProject project = getProject(artifact);
-        return project == null ? null : project.getModel().getDelegate();
+        if (project != null) {
+            return project.getModel().getDelegate();
+        }
+
+        // Check if we can find a consumer POM in the project-local repository
+        Artifact pomArtifact = new org.eclipse.aether.artifact.DefaultArtifact(
+                artifact.getGroupId(), artifact.getArtifactId(), "", "pom", artifact.getBaseVersion());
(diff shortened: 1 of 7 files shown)


gnodet:

| Status | Branch | Result |
|:------:|:------:|:------|
|✅|maven-4.0.x|[](https://github.com/apache/maven/pull/11131)|

### Questions ?

PR #11111, opened 2025-09-10
Align SessionMock.getService with real behavior

gnodet:
## Summary

This PR brings SessionMock in line with the real Session implementation: `getService(Class)` now throws `NoSuchElementException` for unknown services by default.

It also simplifies the mock by centralizing the list of configured services and strengthens test coverage.

## Motivation
- Previously, the mock returned `null` for unknown services (Mockito default), diverging from production behavior and occasionally masking test errors.
- The earlier “exclusion list” logic was verbose and fragile; adding a new configured service required editing a long chained predicate.

## What changed
- Default behavior
  - Any service not explicitly configured in the mock now results in `NoSuchElementException`, matching `AbstractSession` semantics.
- Implementation simplification
  - Introduced a centralized `CONFIGURED_SERVICES` set and a `contains` check, replacing the chained predicate.
  - Used Mockito `argThat` to apply the default-throw behavior only for services not in `CONFIGURED_SERVICES`.
  - Resolved a minor circularity where setup could inadvertently query the mock during configuration.
- API & documentation
  - Deprecated `getMockSessionWithEnhancedServiceBehavior(...)`; `getMockSession(...)` now has the enhanced behavior by default.
  - Updated Javadoc to document the default-throw behavior and the centralized configuration.
- Tests
  - Added tests to assert the default-throw behavior for unknown services and validate that configured services are wired correctly.

## Compatibility
- Backward-compatible for typical usage:
  - Code/tests that request configured services continue to work.
  - Tests that relied on `null` for unknown services should now either configure/stub those services or expect `NoSuchElementException`.
- No production code changes; scope limited to `impl/maven-testing` and its tests.

## Files touched
- `impl/maven-testing/src/main/java/org/apache/maven/api/plugin/testing/stubs/SessionMock.java`
- `impl/maven-testing/src/test/java/org/apache/maven/api/plugin/testing/stubs/SessionGetServiceTest.java`
- `impl/maven-testing/src/test/java/org/apache/maven/api/plugin/testing/stubs/SessionMockDefaultBehaviorTest.java`

## Verification
- All existing tests pass locally; new tests cover the default-throw behavior and configured-service access.


impl/maven-testing/src/main/java/org/apache/maven/api/plugin/testing/stubs/SessionMock.java
```diff
@@ -452,9 +460,75 @@ public static InternalSession getMockSession(LocalRepository localRepository) {
...
+
         return session;
     }
 
+    /**
+     * Checks if a service class is explicitly configured in the mock session.
+     * This method contains the list of all services that have explicit stubs.
+     */
+    private static boolean isConfiguredService(Class<?> serviceClass) {
+        return serviceClass.equals(RepositoryFactory.class)
```
elharo: This works, but might be clearer with a switch based implementation if we use Java 17 or just a constant set of the available types and a contains check.

elharo (approved):

PR #11125, opened 2025-09-15
Simplify “**” handling using brace expansion (fixes #11110)

### LINKED ISSUE #11110: Improve PathSelector to use brace expansion for POSIX ** semantics instead of combinatorial approach
## Summary

The current `PathSelector` implementation in Maven 4 handles POSIX `**` semantics (where `**` matches 0 or more directories vs Java's 1 or more) by generating multiple patterns using the `addPatternsWithOneDirRemoved` method. This approach works but could be simplified and potentially optimized by leveraging Java's built-in brace expansion feature.

## Current Implementation

The current approach in `PathSelector.addPatternsWithOneDirRemoved()` (lines 464+) generates multiple glob patterns by recursively removing `**` patterns to simulate POSIX behavior:

```java
// For pattern "src/**/test/**/*.java", generates:
// 1. "src/**/test/**/*.java" (original)
// 2. "src/**/test/*.java" (second ** removed)  
// 3. "src/test/**/*.java" (first ** removed)
// 4. "src/test/*.java" (both ** removed)
```

## Proposed Improvement

Replace the combinatorial pattern generation with brace expansion:

```java
// Transform: "src/**/test/**/*.java" 
// Into:      "src/{**/,}test/{**/,}*.java"
String posixPattern = globPart.replaceAll("\\*\\*/", "{**/,}");
```

Java's `PathMatcher` with glob syntax natively supports brace expansion and will automatically generate all the same combinations, but potentially more efficiently.

## Benefits

1. **Simpler Code**: Reduces ~40 lines of complex recursive logic to ~3 lines
2. **Better Performance**: Leverages Java's optimized brace expansion instead of multiple matchers
3. **More Maintainable**: Easier to understand and debug
4. **Leverages Platform**: Uses built-in capabilities instead of reimplementing them

## Compatibility

This change would be:
- ✅ **Functionally identical**: Same matching behavior
- ✅ **Backward compatible**: No API changes
- ✅ **Performance neutral or better**: Single optimized matcher vs multiple matchers

## Implementation

The change would be in the `PathSelector.addPatternsWithOneDirRemoved()` method, replacing the recursive pattern generation with a simple regex replacement that transforms `**/` into `{**/,}`.

## Related Work

- This builds on the excellent PathSelector improvements in #10909 and #10923
- Similar pattern transformation approaches are used successfully in other build tools
- The brace expansion feature is well-established in Java's glob implementation

## Testing

The existing comprehensive test suite in `PathSelectorTest` would validate that the behavior remains identical while the implementation becomes simpler.

---

This improvement would make Maven's PathSelector implementation more elegant while maintaining full compatibility and potentially improving performance.

gnodet:
This PR simplifies support for the ** wildcard and clarifies brace handling.

What changes
- Convert `**/` to `{**/,}` after escaping special characters (including braces)
- Escape user-provided `{` and `}` in Maven-style patterns so braces are treated literally
- Keep brace alternation available when the user explicitly opts into `glob:` syntax
- Add focused tests for literal braces and explicit `glob:` alternation

Why
- Provides clear, POSIX-like behavior for `**` (“zero or more directories”) without custom logic
- Aligns implementation with the documented rule that `[ ]` and `{ }` are escaped in Maven-style patterns
- Ensures only the injected braces participate in expansion; user braces remain literals unless `glob:` is used

Behavior notes
- `**` continues to mean “0+ directories”; `**/` is normalized to `{**/,}` so both the current directory and subdirectories match
- Users who relied on unescaped braces acting as alternation in Maven-style patterns may see a behavior change; this corrects behavior to match the documentation
- With `glob:` prefix, brace alternation remains supported per NIO glob rules

Tests
- `testLiteralBracesAreEscapedInMavenSyntax`: verifies that files with literal braces match with Maven-style patterns
- `testBraceAlternationOnlyWithExplicitGlob`: verifies alternation works only with `glob:`

Fixes #11110

### DIFF
--- a/impl/maven-impl/src/main/java/org/apache/maven/impl/PathSelector.java
+++ b/impl/maven-impl/src/main/java/org/apache/maven/impl/PathSelector.java
@@ -449,55 +449,27 @@ final class PathSelector implements PathMatcher {
                         pattern = pattern.substring(3);
                     }
                     pattern = pattern.replace("/**/**/", "/**/");
+
+                    // Escape special characters, including braces
+                    // Braces from user input must be literals; we'll inject our own braces for expansion below
                     pattern = pattern.replace("\\", "\\\\")
                             .replace("[", "\\[")
                             .replace("]", "\\]")
                             .replace("{", "\\{")
                             .replace("}", "\\}");
+
+                    // Transform ** patterns to use brace expansion for POSIX behavior
+                    // This replaces the complex addPatternsWithOneDirRemoved logic
+                    // We perform this after escaping so that only these injected braces participate in expansion
+                    pattern = pattern.replace("**/", "{**/,}");
+
                     normalized.add(DEFAULT_SYNTAX + pattern);
-                    /*
-                     * If the pattern starts or ends with "**", Java GLOB expects a directory level at
-                     * that location while Maven seems to consider that "**" can mean "no directory".
-                     * Add another pattern for reproducing this effect.
-                     */
-                    addPatternsWithOneDirRemoved(normalized, pattern, 0);
                 } else {
                     normalized.add(pattern);
                 }
             }
         }
         return simplify(normalized, excludes);
-    }
-
-    /**
-     * Adds all variants of the given pattern with {@code **} removed.
-     * This is used for simulating the Maven behavior where {@code "**} may match zero directory.
-     * Tests suggest that we need an explicit GLOB pattern with no {@code "**"} for matching an absence of directory.
-     *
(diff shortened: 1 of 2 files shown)


desruisseaux:
Needs to be backported to the `maven-4.0.x` branch as this pull request also fixes the `testWildcardMatchesAlsoZeroDirectory` test case added by #11551. Waiting to see if #11551 is accepted in order to opportunistically include the two pull requests in a single backport.

Issue #11012, opened 2025-07-30
New Maven 4 dependency scopes in consumer POM


cstamas:
### Affected version

HEAD

### Bug description

This issue is created to investigate what happens with newly introduced dependency scopes on Maven 4 when 4.1.0+ POM is converted to "consumer POM" 4.0.0. For example, dependencies like compile-only` should be just omitted.

PR #11126, opened 2025-09-15
core(consumer-pom): drop compile-only/test-only; map test-runtime -> test; adjust tests

gnodet:
This PR updates the consumer POM generation to handle Maven 4–only test scopes in a way that is both minimal and Maven 3 compatible, while preserving relevant test runtime information for consumers.

Changes:
- DefaultConsumerPomBuilder.transformDependencyForConsumerPom:
  - Omit Maven 4–only compile-only and test-only scopes (build-time only; not needed by downstream consumers).
  - Map Maven 4–only test-runtime to classic test for the consumer POM to retain the test runtime classpath in a Maven 3 compatible way.
  - Keep classic scopes (compile/provided/runtime/test/system) unchanged.
- The transformation applies to both direct dependencies and dependencyManagement entries (via the existing transformation stream with null filtering).
- Tests: Update ConsumerPomBuilderTest to reflect the new policy (compile-only and test-only omitted; test-runtime mapped to test; classic scopes unchanged). Existing SCM and consumer POM shape assertions remain.

Rationale:
- The consumer POM targets the 4.0.0 model for Maven 3.x compatibility.
- compile-only and test-only are build-time only and have no safe Maven 3 equivalent; dropping them keeps the consumer POM minimal.
- test-runtime has no exact Maven 3 analog; mapping to test is the closest practical approximation and does not leak to downstream compile/runtime classpaths (test scope is non-transitive).

See #11012

### DIFF
--- a/impl/maven-core/src/main/java/org/apache/maven/internal/transformation/impl/DefaultConsumerPomBuilder.java
+++ b/impl/maven-core/src/main/java/org/apache/maven/internal/transformation/impl/DefaultConsumerPomBuilder.java
@@ -120,6 +120,8 @@ class DefaultConsumerPomBuilder implements PomBuilder {
                     .collect(Collectors.toMap(n -> getDependencyKey(n.getDependency()), Function.identity()));
             Map<String, Dependency> directDependencies = model.getDependencies().stream()
                     .filter(dependency -> !"import".equals(dependency.getScope()))
+                    .map(this::transformDependencyForConsumerPom)
+                    .filter(dependency -> dependency != null) // Filter out dependencies that should be omitted
                     .collect(Collectors.toMap(
                             DefaultConsumerPomBuilder::getDependencyKey,
                             Function.identity(),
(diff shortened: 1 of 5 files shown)


impl/maven-core/src/main/java/org/apache/maven/internal/transformation/impl/DefaultConsumerPomBuilder.java
```diff
@@ -175,6 +179,45 @@ private Dependency merge(Dependency dep1, Dependency dep2) {
...
+            return null;
+        }
+
+        String scope = dependency.getScope();
+        if (scope == null) {
+            return dependency;
+        }
+
+        // Handle new Maven 4 scopes when creating consumer POM
+        switch (scope) {
```
cstamas: Looks good, but one nit: could we do this programmatically? We do have mvn3 scope manager config and we do have mvn4 scope manager config as well.... I would _really_ like to have all scopes managed at single place, unlike in mvn3 when they are _smeared across many classes in many projects_ (maven, resolver, many plugins, etc)

gnodet: I moved the logic into Maven4ScopeManagerConfiguration. @cstamas is that what you were asking for ?

gnodet:
Superseded by #11163

PR #11138, opened 2025-09-17
GH-10210: fix too eager decrypt of legacy passwords

### LINKED ISSUE #10210: [MNG-8417] New encrypted passwords prevent maven from building projects
**[Lenny Primak](https://issues.apache.org/jira/secure/ViewProfile.jspa?name=lprimak)** opened **[MNG-8417](https://issues.apache.org/jira/browse/MNG-8417?redirect=false)** and commented

When settings.xml contains new-style encrypted passwords, maven will not build unless it can decrypt the password.

The use case is that the passwords are used only for deployment, while 99% of the use cases don't require the passwords.

This forces the users to have to have secure environment variables or other ways to get the master password at all times, enhancing security risks

---

**Affects:** 4.0.0-beta-5, 4.0.0-rc-1

**Issue Links:**
- [MNG-8419](https://issues.apache.org/jira/browse/MNG-8419) non obvious errors if maven can not decrypt passwords in settings

cstamas:
Accept ONLY documented forms of them.

Fixes #10210


its/core-it-suite/pom.xml
```diff
@@ -468,17 +468,6 @@ under the License.
           </execution>
         </executions>
       </plugin>
-      <plugin>
```
gnodet: This would be better in a different PR.

PR #11153, opened 2025-09-20
Enable the search for `module-info.class` file in the `META-INF/versions/` sub-directories of a JAR file.

desruisseaux:
By default, `java.util.jar.JarFile` does not search for resources in `META-INF/versions/` sub-directories. For enabling this feature, we need to specify a version greater than Java 8 to the constructor. The core change of this pull request is this line:

```java
try (JarFile jar = new JarFile(path.toFile(), false, JarFile.OPEN_READ, target)) {
    // ...
}
```

The rest is for answering the question: "which target Java version"? An easy approach would have been to set `target` unconditionally to `Runtime.version()`, i.e. the Java version on which Maven is running. This pull request rather allows configuration, with the intend to set the target version to the value of the `--release` target option. This value can determine whether a dependency will be placed on the class-path or on the module-path. For example, if the `module-info.class` entry of a JAR file exists only in the `META-INF/versions/17/` sub-directory, then the default location of that dependency will be the module-path only if the `--release` option is equal or greater than 17.

This pull request fixes the initial topic of https://github.com/apache/maven-compiler-plugin/issues/966.

### DIFF
--- a/impl/maven-impl/src/main/java/org/apache/maven/impl/PathModularization.java
+++ b/impl/maven-impl/src/main/java/org/apache/maven/impl/PathModularization.java
@@ -43,7 +43,7 @@ import org.apache.maven.api.annotations.Nonnull;
  * or module hierarchy, but not module source hierarchy. The latter is excluded because this class
  * is for path elements of compiled codes.
  */
-class PathModularization {
+final class PathModularization {
     /**
      * A unique constant for all non-modular dependencies.
      */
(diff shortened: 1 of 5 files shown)


impl/maven-impl/src/main/java/org/apache/maven/impl/PathModularization.java
```diff
@@ -192,7 +193,7 @@ private PathModularization() {
              * If no descriptor, the "Automatic-Module-Name" manifest attribute is
              * taken as a fallback.
              */
-            try (JarFile jar = new JarFile(path.toFile())) {
+            try (JarFile jar = new JarFile(path.toFile(), false, JarFile.OPEN_READ, target)) {
```
gnodet, line 196: When using `JarFile(File)`, the default value for the `verify` argument is `true`.  We're switching to `false`. Is that on purpose ? I don't really think we need signature verification at this point, right ?

desruisseaux: Yes, the JAR file is opened only for extracting the module name, either from the `module-info.class` or from the `MANIFEST.MF` file. I thought that verifying the JAR file was an overkill in that situation.

PR #11154, opened 2025-09-20
[MNG-8696] Hide the cache from DefaultDependencyResolverResult constructor

desruisseaux:
This is a port of #2347. The intend is to avoid merge conflict with future ports such as #11153.

### DIFF
--- a/impl/maven-impl/src/main/java/org/apache/maven/impl/DefaultDependencyResolver.java
+++ b/impl/maven-impl/src/main/java/org/apache/maven/impl/DefaultDependencyResolver.java
@@ -68,6 +68,22 @@ import static org.apache.maven.impl.ImplUtils.map;
 @Singleton
 public class DefaultDependencyResolver implements DependencyResolver {
 
+    /**
+     * Cache of information about the modules contained in a path element.
+     *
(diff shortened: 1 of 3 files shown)


impl/maven-impl/src/main/java/org/apache/maven/impl/DefaultDependencyResolver.java
```diff
@@ -68,6 +68,22 @@
 @Singleton
```
gnodet, line 68: If the bean contains session-specific data, maybe we want to make it `@SessionScope` rather than `@Singleton`.  That may be one possible answer to "until clarified where to store session-wide caches".

gnodet (approved):

impl/maven-impl/src/main/java/org/apache/maven/impl/DefaultDependencyResolver.java
desruisseaux: Maybe I have misused the "session" word, since I'm not sure which data are in a session. This cache should be valid as long as the JAR files do not change. For JAR files that are not produced by the project, it should be the JVM lifetime. For JAR files produced by the project and used as dependencies by sub-projects, this is an open question...

PR #11161, opened 2025-09-22
Mimir updates

cstamas:
Changes:
* update Mimir to latest 0.8.1 version
* ~make Mimir session do NOT use daemon (GH nodes are not reused nor shared) except in ITs (that do run in parallel, they need to use daemon)~ daemon left in place; is needed due Infusers
* ~workflow: seed new `mimir.properties` file as well along with existing `daemon.properties` (that are now unused)~ daemon left in place; is needed due Infusers
* cache changes (see below)

Cache strategy changes:
* initial build: it is snowballing one cache when build is not about PR
* full (rebuild itself + site with itself) and its build: it is snowballing cache differentiated by OS/JDK when build is not about PR

Current problems: on unchanged POM (ie. new IT added), the dependencies ITs pull from Central will be cached by Mimir, but due "cache hit" the new cache will not get stored. Hence, Mimir caching was basically lost. Also, there was a mixup of caches from PRs and main branches. Finally, matrix jobs were competing for cache store.


.github/workflows/maven.yml
```diff
@@ -51,14 +51,16 @@ jobs:
         run: |
           mkdir -p ~/.mimir
           cp .github/ci-extensions.xml ~/.m2/extensions.xml
+          cp .github/ci-mimir.properties ~/.mimir/mimir.properties
```
gnodet: An env var would be better imho.  A possible Mimir enhancement ?

cstamas: Mimir does not support it... I'd also rename `~/.mimir/mimir.properties` to `~/.mimir/session.properties`, as these are used to create session, while daemon will load `~/.mimir/daemon.properties`.

cstamas: Ref:
* https://github.com/maveniverse/mimir/issues/138
* https://github.com/maveniverse/mimir/issues/139

.github/workflows/maven.yml
```diff
@@ -74,6 +72,13 @@ jobs:
         shell: bash
         run: ls -la apache-maven/target
 
+      - name: Save Mimir caches
+        uses: actions/cache/save@0400d5f644dc74513175e3cd8d07132dd4860809 # v4
+        if: ${{ github.event_name != 'pull_request' && !cancelled() && !failure() }}
+        with:
+          path: ~/.mimir/local
+          key: ${{ steps.restore-cache.outputs.cache-primary-key }}
```
slawekjaranowski, line 80: it is the same as primary key used in restore action, so we not need a variable

.github/workflows/maven.yml
```diff
@@ -122,14 +127,15 @@ jobs:
...
+        id: restore-cache
         with:
           path: ~/.mimir/local
-          key: mimir-${{ runner.os }}-full-${{ hashFiles('**/pom.xml') }}
+          key: mimir-full-${{ matrix.os }}-${{ matrix.java }}
           restore-keys: |
-            mimir-${{ runner.os }}-full-
-            mimir-${{ runner.os }}-
+            mimir-full-${{ matrix.os }}-
+            mimir-full-
```
slawekjaranowski, line 138: such keys will never be created

PR #11168, opened 2025-09-23
CI: Initial build

cstamas:
Update used wrapper and used Maven.

### DIFF
--- a/.github/workflows/maven.yml
+++ b/.github/workflows/maven.yml
@@ -51,6 +51,7 @@ jobs:
         run: |
           mkdir -p ~/.mimir
           cp .github/ci-extensions.xml ~/.m2/extensions.xml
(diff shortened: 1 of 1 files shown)


.github/workflows/maven.yml
```diff
@@ -51,6 +51,7 @@ jobs:
         run: |
           mkdir -p ~/.mimir
           cp .github/ci-extensions.xml ~/.m2/extensions.xml
+          cp .github/ci-extensions.xml .mvn/extensions.xml
```
slawekjaranowski, line 54: we not need it on master branch, probably should be done in 3.9.x

cstamas: We need: very first `mvn` invocation uses 3.9.x (from setup-java) to install 4-rc-4

cstamas:
Superseded by #11169

PR #11169, opened 2025-09-24
CI: Update Mimir steps

cstamas:
Initial step was not cached, so m-wrapper-p and Maven distro was always pulled from MC.

Changes:
* using Mimir 0.9.3
* centralized Mimir config
* using provided infusers

After this, ~there is only one thing not cached: Mimir extension itself~ not anymore :wink:

### DIFF
--- a/.github/workflows/maven.yml
+++ b/.github/workflows/maven.yml
@@ -32,7 +32,7 @@ concurrency:
 permissions: {}
 
 env:
-  MIMIR_VERSION: 0.9.2
+  MIMIR_VERSION: 0.9.3
   MIMIR_BASEDIR: ~/.mimir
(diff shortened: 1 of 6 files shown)


cstamas:
Infuser failed.

.github/workflows/maven.yml
```diff
@@ -46,23 +51,31 @@ jobs:
         with:
           persist-credentials: false
 
-      - name: Prepare Mimir
+      - name: Prepare Mimir for Maven 3.x
```
slawekjaranowski, line 54: each branch has own workflow ... so it is not Maven 3 connected

slawekjaranowski: ok, I see first command - setup Maven by wrapper is done by Maven 3

Issue #10985, opened 2025-07-22
Msven allows random namespaces on project


elharo:
### Affected version

HEAD

### Bug description

It does not only allow the correct namespace and no namespace. Any random URL or string you throw in does not cause an error:

```
<project xmlns="https://ff.com">
  <!-- Test that we can parse a pom.xml with no namespace -->
  <modelVersion>4.0.0</modelVersion>

  <groupId>org.apache.maven.its</groupId>
  <artifactId>missing-namespace</artifactId>
  <version>1.0-SNAPSHOT</version>
  <packaging>pom</packaging>
</project>
```

arturobernalg:
@elharo  please take a look https://github.com/apache/maven/pull/11185

raupachz:
@arturobernalg Funny. I was looking at the same issue at the exact time. Oh well, no @elharo can choose. Better two options than none.

gnodet:
See https://github.com/apache/maven/issues/8457

gnodet:
> We do want to restrict namespaces when parsing 4.0.0 files to only the specifically allowed namespaces and no namespace at all. But we should not allow arbitrary strings.

We first want to make sure they can't go to central.  If there are such files in central and they have been used for years, I'm not sure we want to not support them anymore.  So we need answers to:
* can pom be uploaded with no-namespace to central ?
* can pom be uploaded with random namespace to central ?
* for how-long has that been enforced ?
* how many no-namespace or random namespace pom do we have, if any ?

If the namespace (eventually no-namespace) is enforcer in central, and if there's no ill-formed pom to support, then, yes, we can certainly enforce and restrict when reading.

@brianf ?

PR #11185, opened 2025-09-28
Disallow arbitrary namespaces in Maven and Metadata readers

arturobernalg:
Accept POM 4.1.0 and METADATA 1.1.0 plus no-namespace, reject others Add JUnit tests for valid and invalid namespaces

- [X ] Write unit tests that match behavioral changes, where the tests fail if the changes to the runtime are not applied.

### DIFF
--- a/impl/maven-support/src/test/java/org/apache/maven/model/v4/MavenStaxReaderNamespaceTest.java
+++ b/impl/maven-support/src/test/java/org/apache/maven/model/v4/MavenStaxReaderNamespaceTest.java
@@ -1,0 +1,111 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.maven.model.v4;
+
+import javax.xml.stream.XMLStreamException;
+
+import java.io.StringReader;
+import java.util.Locale;
+
+import org.apache.maven.api.model.InputSource;
+import org.apache.maven.api.model.Model;
+import org.junit.jupiter.api.Test;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertNotNull;
+import static org.junit.jupiter.api.Assertions.assertThrows;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+class MavenStaxReaderNamespaceTest {
+
+    private static final InputSource NO_INPUT_SOURCE = null;
+
+    private static final String POM_41 = ""
+            + "<project xmlns=\"http://maven.apache.org/POM/4.1.0\" "
+            + "         xmlns:xsi=\"http://www.w3.org/2001/XMLSchema-instance\">"
+            + "  <modelVersion>4.1.0</modelVersion>"
+            + "  <groupId>com.acme</groupId>"
+            + "  <artifactId>demo</artifactId>"
+            + "  <version>1.0</version>"
+            + "</project>";
+
+    private static final String POM_40 = ""
+            + "<project xmlns=\"http://maven.apache.org/POM/4.0.0\" "
+            + "         xmlns:xsi=\"http://www.w3.org/2001/XMLSchema-instance\">"
+            + "  <modelVersion>4.0.0</modelVersion>"
+            + "  <groupId>com.acme</groupId>"
+            + "  <artifactId>demo</artifactId>"
+            + "  <version>1.0</version>"
+            + "</project>";
+
+    private static final String POM_NO_NS = ""
+            + "<project>"
+            + "  <modelVersion>4.0.0</modelVersion>"
+            + "  <groupId>com.acme</groupId>"
+            + "  <artifactId>demo</artifactId>"
+            + "  <version>1.0</version>"
+            + "</project>";
+
+    private static final String POM_BAD_NS = ""
+            + "<project xmlns=\"http://example.com/not/pom\">"
+            + "  <modelVersion>4.1.0</modelVersion>"
+            + "  <groupId>com.acme</groupId>"
+            + "  <artifactId>demo</artifactId>"
+            + "  <version>1.0</version>"
+            + "</project>";
+
+    @Test
+    void acceptsPom41() throws Exception {
+        MavenStaxReader reader = new MavenStaxReader();
+        Model model = reader.read(new StringReader(POM_41), /*strict*/ true, NO_INPUT_SOURCE);
+        assertNotNull(model);
+        assertEquals("com.acme", model.getGroupId());
+        assertEquals("demo", model.getArtifactId());
+        assertEquals("1.0", model.getVersion());
+    }
+
+    @Test
+    void acceptsPom40() throws Exception {
+        MavenStaxReader reader = new MavenStaxReader();
+        Model model = reader.read(new StringReader(POM_40), true, NO_INPUT_SOURCE);
+        assertNotNull(model);
+        assertEquals("com.acme", model.getGroupId());
+        assertEquals("demo", model.getArtifactId());
+        assertEquals("1.0", model.getVersion());
+    }
+
+    @Test
+    void acceptsPomWithoutNamespace() throws Exception {
+        MavenStaxReader reader = new MavenStaxReader();
+        Model model = reader.read(new StringReader(POM_NO_NS), true, NO_INPUT_SOURCE);
+        assertNotNull(model);
+        assertEquals("com.acme", model.getGroupId());
+        assertEquals("demo", model.getArtifactId());
+        assertEquals("1.0", model.getVersion());
+    }
+
+    @Test
+    void rejectsUnexpectedPomNamespace() {
+        MavenStaxReader reader = new MavenStaxReader();
+        XMLStreamException ex = assertThrows(
+                XMLStreamException.class, () -> reader.read(new StringReader(POM_BAD_NS), true, NO_INPUT_SOURCE));
+        // sanity check: message mentions unrecognized namespace
+        assertTrue(ex.getMessage().toLowerCase(Locale.ROOT).contains("unrecognized pom namespace"));
+    }
+}
--- a/impl/maven-support/src/test/java/org/apache/maven/model/v4/MetadataStaxReaderNamespaceTest.java
+++ b/impl/maven-support/src/test/java/org/apache/maven/model/v4/MetadataStaxReaderNamespaceTest.java
@@ -1,0 +1,88 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.maven.model.v4;
+
+import javax.xml.stream.XMLStreamException;
+
+import java.io.StringReader;
+
+import org.apache.maven.api.metadata.Metadata;
+import org.apache.maven.metadata.v4.MetadataStaxReader;
+import org.junit.jupiter.api.Test;
+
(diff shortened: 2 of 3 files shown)


impl/maven-support/src/test/java/org/apache/maven/model/v4/MavenStaxReaderNamespaceTest.java
```diff
@@ -0,0 +1,110 @@
...
+            + "  <modelVersion>4.1.0</modelVersion>"
+            + "  <groupId>com.acme</groupId>"
+            + "  <artifactId>demo</artifactId>"
+            + "  <version>1.0</version>"
+            + "</project>";
+
+    @Test
+    void acceptsPom41() throws Exception {
+        MavenStaxReader reader = new MavenStaxReader();
+        Model m = reader.read(new StringReader(POM_41), /*strict*/ true, NO_INPUT_SOURCE);
```
elharo: m --> model

impl/maven-support/src/test/java/org/apache/maven/model/v4/MavenStaxReaderNamespaceTest.java
```diff
@@ -0,0 +1,110 @@
...
+        assertNotNull(m);
+        assertEquals("com.acme", m.getGroupId());
+        assertEquals("demo", m.getArtifactId());
+        assertEquals("1.0", m.getVersion());
+    }
+
+    @Test
+    void acceptsPomWithoutNamespace() throws Exception {
+        MavenStaxReader reader = new MavenStaxReader();
+        Model m = reader.read(new StringReader(POM_NO_NS), true, NO_INPUT_SOURCE);
```
elharo: m --> model

impl/maven-support/src/test/java/org/apache/maven/model/v4/MavenStaxReaderNamespaceTest.java
```diff
@@ -0,0 +1,110 @@
...
+        assertEquals("1.0", m.getVersion());
+    }
+
+    @Test
+    void rejectsUnexpectedPomNamespace() {
+        MavenStaxReader reader = new MavenStaxReader();
+        XMLStreamException ex = assertThrows(
+                XMLStreamException.class, () -> reader.read(new StringReader(POM_BAD_NS), true, NO_INPUT_SOURCE));
+        // sanity check: message mentions unrecognized namespace
+        assertTrue(ex.getMessage().toLowerCase().contains("unrecognized pom namespace"));
```
elharo: Locale.ROOT

impl/maven-support/src/test/java/org/apache/maven/model/v4/MetadataStaxReaderNamespaceTest.java
```diff
@@ -0,0 +1,88 @@
...
+    private static final String META_BAD_NS = ""
+            + "<metadata xmlns=\"http://example.com/not/metadata\">"
+            + "  <groupId>com.acme</groupId>"
+            + "  <artifactId>demo</artifactId>"
+            + "</metadata>";
+
+    @Test
+    void acceptsMetadata110() throws Exception {
+        MetadataStaxReader reader = new MetadataStaxReader();
+        Metadata md = reader.read(new StringReader(META_110), /*strict*/ true);
```
elharo: md --> metadata

src/mdo/reader-stax.vm
```diff
@@ -246,51 +246,76 @@ public class ${className} {
...
+    while (eventType != XMLStreamReader.END_DOCUMENT) {
+    if (eventType == XMLStreamReader.START_ELEMENT) {
+    if (strict && ! "${rootTag}".equals(parser.getLocalName())) {
+    throw new XMLStreamException("Expected root element '${rootTag}' but found '" + parser.getName() + "'", parser.getLocation(), null);
+    } else if (parsed) {
+    throw new XMLStreamException("Duplicated tag: '${rootTag}'", parser.getLocation(), null);
+    }
+
+    // Enforce root namespace per model (strict mode).
+    String rootNs = parser.getNamespaceURI();
```
elharo: rootNamespace

src/mdo/reader-stax.vm
```diff
@@ -246,51 +246,76 @@ public class ${className} {
...
+    boolean hasNs = rootNs != null && !rootNs.isEmpty();
```
elharo: hasNamespace

elharo (commented):
avoid abbreviations

src/mdo/reader-stax.vm
```diff
@@ -246,51 +246,76 @@ public class ${className} {
...
-    /**
-     * Method read.
-     *
-     * @param parser a parser object.
-     * @param strict a strict object.
-     * @throws XMLStreamException XMLStreamException if
-     * any.
-     * @return ${root.name}
-     */
+/**
```
gnodet: Fix the formatting of this file, this completely breaks the format of the generated java file.

src/mdo/reader-stax.vm
```diff
@@ -246,51 +246,76 @@ public class ${className} {
...
+    throw new XMLStreamException("Duplicated tag: '${rootTag}'", parser.getLocation(), null);
+    }
+
+    // Enforce root namespace per model (strict mode).
+    String rootNs = parser.getNamespaceURI();
+    boolean hasNs = rootNs != null && !rootNs.isEmpty();
+    if (strict && hasNs) {
+    if ("project".equals("${rootTag}")) {
+    // Accept any official POM namespace version (e.g., 4.0.0, 4.1.0)
+    if (!rootNs.startsWith("http://maven.apache.org/POM/")) {
```
gnodet: These hardcoded values are wrong.  We generate much more readers than just those two: settings, lifecycles, extensions, plugin, toolchains...
In addition, accepting metadata namespace when reading a POM is just plain wrong. 
The way to go would be to define a variable for the template in the POM that generates the writer and only accept that one.
Another thing to keep in mind is that the plugin configuration in the POM can actually contain external namespaces and this is valid.

gnodet: An example of such a variable is `${packageToolV4}`.

gnodet (approved):
Hardcoding the namespace is wrong.

arturobernalg:
> Hardcoding the namespace is wrong.

@gnodet  please do another pass

src/mdo/reader-stax.vm
```diff
@@ -263,32 +263,58 @@ public class ${className} {
...
+    // Root namespace policy:
+    //  - POM ('project'): allow no namespace or any http://maven.apache.org/POM/<version>
+    //  - METADATA ('metadata'): allow no namespace or any http://maven.apache.org/METADATA/<version>
+    //  - Other readers: do not enforce root namespace here (child checks still apply when strict)
+    if (strict) {
+        final String rootNs = parser.getNamespaceURI();
+        final boolean hasNs = rootNs != null && !rootNs.isEmpty();
+
+        if ("project".equals("${rootTag}")) {
+            if (hasNs && !rootNs.startsWith("http://maven.apache.org/POM/")) {
```
gnodet, line 285: This is wrong.  We have generators to generate specific readers, not a generic one which covers all use cases.  So it's wrong to put hardcoded values for one or two models in the output of all readers.
The way to go is to retrieve the namespace using modello.  The `helper` needs to be enhanced:
https://github.com/codehaus-plexus/modello/blob/ca30aba14346bf0e183efd2c442f061024c40111/modello-plugins/modello-plugin-velocity/src/main/java/org/codehaus/modello/plugin/velocity/Helper.java
It needs a new method `xmlModelMetadata` which would return the `XmlModelMetadata` from which we can get the target _namespace_.
Once modello is fixed, we can update the velocity templates to leverage that using `Helper.xmlModelMetadata(model).namespace`.

gnodet (changes requested):
Unfortunately, I think we need to enhance modello to provide access to the target namespace using `$Helper.xmlModelMetadata(model).namespace`.

src/mdo/reader-stax.vm
gnodet: See https://github.com/codehaus-plexus/modello/pull/522

gnodet (commented):
## Review: Disallow arbitrary namespaces in Maven and Metadata readers

Echoing @gnodet's existing feedback — the approach needs rework. The core issues are architectural (hardcoded tag names in a shared template) and formatting (broken indentation in generated code).

### High

1. **Hardcoded root tag names in shared template** (`reader-stax.vm`) — The template generates readers for *all* Maven models (settings, toolchains, lifecycles, extensions, plugins, etc.), not just POMs and metadata. Hardcoding `if ("project".equals("${rootTag}"))` / `else if ("metadata".equals("${rootTag}"))` embeds model-specific knowledge into a generic code generator. As @gnodet noted, use `$Helper.xmlModelMetadata(model).namespace` (available since modello PR codehaus-plexus/modello#522) to retrieve the expected namespace per model — no hardcoding needed.

2. **Template formatting broken** — Indentation is stripped from the generated Java code. The original method body uses 8-space indentation; the new code starts at column 0. Since Velocity whitespace flows directly into generated output, the resulting `.java` files will have malformed formatting.

3. **`startsWith("http://maven.apache.org/POM/")` is overly permissive** — Accepts any string with the right prefix, including `http://maven.apache.org/POM/banana`. With the modello-based approach, validation would be against the exact declared namespace.

### Medium

4. **Other models get no namespace validation** — The `else` branch explicitly skips validation for settings, toolchains, etc. The modello-based approach would give every model with a declared namespace automatic validation.

### Low

5. **`toLowerCase()` without `Locale.ROOT`** in `MetadataStaxReaderNamespaceTest` — The POM test correctly uses `Locale.ROOT` but the metadata test does not. Inconsistency within the PR.

### Notes

- PR #11186 by @raupachz addresses the same issue (#10985) at a different layer (post-parsing in `DefaultModelXmlFactory`). Both PRs have received CHANGES_REQUESTED feedback. It may be worth discussing which layer should own namespace validation.
- Test coverage is reasonable — tests for 4.0.0 acceptance, 4.1.0 acceptance, no-namespace acceptance, and bad namespace rejection all exist.

---

gnodet (changes requested):
The namespace validation concept is sound and fills a real gap, but the Velocity template changes break indentation in all generated StAX readers, making the PR unmergeable as-is. Findings independently verified (4/5 confirmed, 1 false positive removed).

**Critical:**
1. **Broken indentation in generated readers (high)** — The template strips 4-8 spaces of leading whitespace from Java code lines. Since Velocity emits non-directive text verbatim, this produces generated files (MavenStaxReader, MetadataStaxReader, SettingsStaxReader, ToolchainsStaxReader, etc. — at least 8 files) with variable declarations at column 0 and flattened nesting.
2. **Asymmetric closing brace (medium)** — The `else` branch closing `}` is at 12-space indentation while its opening is at 8 spaces. Also, `throw` inside an `if` block sits at the same indentation as the `if` itself.

**Minor:**
3. **Test class in wrong package (low)** — `MetadataStaxReaderNamespaceTest` is in `org.apache.maven.model.v4` but tests `org.apache.maven.metadata.v4.MetadataStaxReader`.
4. **Prefix-based validation (low)** — `startsWith("http://maven.apache.org/POM/")` accepts malformed URIs. Maintainer already requested using modello's `$Helper.xmlModelMetadata(model).namespace` instead.

Removed false positive: Velocity directive indentation is not an issue — the file already uses indented directives elsewhere and Velocity gobbles directive-only lines.

gnodet (commented):
### Review Summary

The namespace validation concept correctly addresses issue #10985 (a real security gap — arbitrary namespaces should be rejected). However, the implementation has structural issues that should be resolved before merging.

#### 1. 🔴 Broken indentation in `reader-stax.vm` (high)

The modified block collapses the 8-space indent used throughout the template to 0-4 spaces. Since Velocity emits non-directive text verbatim, this produces malformatted Java in **all generated StaxReader classes** (Maven, Metadata, Settings, Toolchains, CoreExtensions, PluginDescriptor, Lifecycle). For example, `$rootUcapName $rootLcapName = null;` sits at column 0 while the unchanged `Deque<Object> context = new ArrayDeque<>()` above it has 8-space indent.

#### 2. 🔴 Hardcoded root tag name checks in shared template (high)

The checks `"project".equals("${rootTag}")` and `"metadata".equals("${rootTag}")` are string comparisons where `${rootTag}` is substituted at generation time. For `SettingsStaxReader`, the generated code literally reads `"project".equals("settings")` — always false, producing dead code in 5+ readers.

The **correct approach already exists** — the writer template (`writer-stax.vm` lines 25-32) uses `$model.getMetadata("org.codehaus.modello.plugins.xml.metadata.XmlModelMetadata")` to derive namespaces dynamically from modello metadata. Modello PR [codehaus-plexus/modello#522](https://github.com/codehaus-plexus/modello/pull/522) providing `$Helper.xmlModelMetadata(model)` access is already merged.

#### 3. 🟡 Overly permissive `startsWith` namespace check (medium)

`rootNs.startsWith("http://maven.apache.org/POM/")` accepts any version suffix (e.g., `http://maven.apache.org/POM/99.99.99`). The existing `checkNamespace` method at line 519 uses exact matching (`Objects.equals`). Using modello-derived namespaces would enable exact matching here too.

#### 4. 🟡 Existing `enforceNamespace` parameter is unused (medium)

`impl/maven-support/pom.xml` already sets `enforceNamespace=true` for the maven.mdo model execution (line 228). The template should check this parameter to conditionally enable validation per model, making the feature opt-in rather than hardcoding which models get it.

#### 5. 🔵 Missing `Locale.ROOT` in `MetadataStaxReaderNamespaceTest` (low)

`toLowerCase()` at line 86 is called without `Locale.ROOT`, while the POM test correctly uses `Locale.ROOT`. This can cause locale-dependent behavior.

#### 6. 🔵 No `strict=false` test coverage (low)

All tests use `strict=true`. The template gates namespace validation inside `if (strict)`, but no test verifies that validation is skipped when `strict=false`.

### 💡 Recommended Approach

Use modello's `$Helper.xmlModelMetadata(model)` to retrieve the target namespace dynamically in `reader-stax.vm`, mirroring how `writer-stax.vm` already does it. This eliminates hardcoded tag names, produces no dead code, and automatically covers all model types that declare a namespace. The tests (well-structured with good coverage of accept/reject scenarios) can be preserved with updated error message assertions.

### 📋 PR Metadata

|--------|---------|-----------|

---

elharo:
seems abandoned and duplicates other work

PR #11186, opened 2025-09-28
Issue #10985: Maven allows random namespaces on project

raupachz:
Bugfix for Issue #10985.

This code only extends commit `46d93dab`.

### DIFF
--- a/impl/maven-impl/src/main/java/org/apache/maven/impl/DefaultModelXmlFactory.java
+++ b/impl/maven-impl/src/main/java/org/apache/maven/impl/DefaultModelXmlFactory.java
@@ -52,17 +52,19 @@ public class DefaultModelXmlFactory implements ModelXmlFactory {
     public Model read(@Nonnull XmlReaderRequest request) throws XmlReaderException {
         requireNonNull(request, "request");
         Model model = doRead(request);
-        if (isModelVersionGreaterThan400(model)
-                && !model.getNamespaceUri().startsWith("http://maven.apache.org/POM/")) {
-            throw new XmlReaderException(
-                    "Invalid namespace '" + model.getNamespaceUri() + "' for model version " + model.getModelVersion(),
-                    null,
-                    null);
+        if (isModelVersionGreaterOrEqualsThan400(model)) {
+            String ns = model.getNamespaceUri();
+            if (ns.length() > 0 && !model.getNamespaceUri().startsWith("http://maven.apache.org/POM/")) {
+                throw new XmlReaderException(
+                        "Invalid namespace '" + model.getNamespaceUri() + "' for model version " + model.getModelVersion(),
+                        null,
+                        null);
+            }
         }
         return model;
     }
 
-    private boolean isModelVersionGreaterThan400(Model model) {
+    private boolean isModelVersionGreaterOrEqualsThan400(Model model) {
         String version = model.getModelVersion();
         if (version == null) {
             return false;
@@ -71,7 +73,7 @@ public class DefaultModelXmlFactory implements ModelXmlFactory {
             String[] parts = version.split("\\.");
             int major = Integer.parseInt(parts[0]);
             int minor = parts.length > 1 ? Integer.parseInt(parts[1]) : 0;
-            return major > 4 || (major == 4 && minor > 0);
+            return major > 4 || (major == 4 && minor >= 0);
         } catch (NumberFormatException | IndexOutOfBoundsException e) {
             return false;
         }


gnodet (approved):
...

gnodet (changes requested):
We need to be able to parse 4.0.0 POM files with no namespace, but that is not allowed for 4.1.0 POM files.
I'm not sure we want to restrict namespaces when reading 4.0.0 files. That can be checked for build POM, but not when consuming them from a repository. So we need to make that dependent on the validationLevel to preserve compatibility.

raupachz:
...

raupachz:
> We need to be able to parse 4.0.0 POM files with no namespace, but that is not allowed for 4.1.0 POM files.
I'm not sure we want to restrict namespaces when reading 4.0.0 files. That can be checked for build POM, but not when consuming them from a repository. So we need to make that dependent on the validationLevel to preserve compatibility.

Thanks. I did not know that. It looked simpler from the bug description.

gnodet (commented):
## Review: Issue #10985: Maven allows random namespaces on project

Echoing @gnodet's existing feedback — the core approach needs rework before this can be merged.

### High

1. **Backward compatibility break for 4.0.0 POMs** — Changing `minor > 0` to `minor >= 0` makes the namespace check apply to ALL 4.x POMs including 4.0.0. There are millions of 4.0.0 POMs in Maven repositories. Many older or hand-crafted POMs may have non-standard namespaces or omit the namespace entirely. Unconditionally rejecting them will break dependency resolution. As @gnodet noted, this needs to be gated on the validation level — strict for build POMs (where the user controls the POM), lenient for POMs consumed from repositories.

2. **No tests** — For a validation change that could break backward compatibility, tests are mandatory. PR #11185 by @arturobernalg addresses the same issue (#10985) and includes comprehensive test cases for POM namespace validation.

### Medium

3. **Empty vs. arbitrary namespace semantics** — The `ns.length() > 0` guard allows empty namespaces but rejects arbitrary non-empty ones. This is reasonable in isolation, but the validation needs to differentiate between build POMs and repository POMs (per @gnodet's feedback) to avoid breaking the 4.0.0 ecosystem.

### Low

4. **Method name** — `isModelVersionGreaterOrEqualsThan400` is grammatically awkward. Consider `isModelVersionAtLeast400` or `isModelVersionGreaterThanOrEqualTo400`.

### Note on PR #11185

PR #11185 by @arturobernalg tackles the same issue at the StAX reader template level (`reader-stax.vm`), which catches invalid namespaces earlier during XML parsing rather than post-parsing. It also includes tests and covers Metadata files in addition to POMs. The two PRs conflict in approach — it may be worth discussing which layer should own namespace validation.

---

impl/maven-impl/src/main/java/org/apache/maven/impl/DefaultModelXmlFactory.java
```diff
@@ -52,17 +52,19 @@ public class DefaultModelXmlFactory implements ModelXmlFactory {
...
-        if (isModelVersionGreaterThan400(model)
-                && !model.getNamespaceUri().startsWith("http://maven.apache.org/POM/")) {
-            throw new XmlReaderException(
-                    "Invalid namespace '" + model.getNamespaceUri() + "' for model version " + model.getModelVersion(),
-                    null,
-                    null);
+        if (isModelVersionGreaterOrEqualsThan400(model)) {
+            String ns = model.getNamespaceUri();
+            if (ns.length() > 0 && !model.getNamespaceUri().startsWith("http://maven.apache.org/POM/")) {
+                throw new XmlReaderException(
```
gnodet, line 58: Bug: The `ns.length() > 0` guard short-circuits the namespace check for ALL model versions ≥ 4.0.0 when namespace is empty — including 4.1.0+. On master, a 4.1.0 POM with empty namespace correctly throws `XmlReaderException` (because `"".startsWith("http://maven.apache.org/POM/")` is false). With this change, empty namespaces are silently accepted for 4.1.0+.

The empty-namespace exemption should only apply to 4.0.0, not to 4.1.0+. Consider separating the logic:

```java
if (isModelVersionGreaterOrEqualsThan400(model)) {
    String ns = model.getNamespaceUri();
    boolean emptyNamespace = ns.isEmpty();
    if (!emptyNamespace && !ns.startsWith("http://maven.apache.org/POM/")) {
        throw new XmlReaderException(...);
    }
    if (emptyNamespace && isModelVersionGreaterThan400(model)) {
        throw new XmlReaderException("Missing namespace for model version " + model.getModelVersion() + "...", null, null);
    }
}
```

gnodet (changes requested):
The namespace validation fix addresses a valid issue (#10985), but introduces a regression for 4.1.0+ POMs and does not address the maintainer's prior feedback. All findings independently verified (6/6 confirmed).

**Critical issues:**
1. **Regression for 4.1.0+ (high)** — The `ns.length() > 0` guard causes the namespace check to be skipped for ALL model versions ≥ 4.0.0 when namespace is empty, including 4.1.0+. On master, a 4.1.0 POM with no namespace correctly throws `XmlReaderException`. This contradicts the intent of commit 46d93dab.
2. **Validation level gating (high)** — Maintainer @gnodet explicitly requested that namespace enforcement for 4.0.0 POMs be gated on validation level (strict for build POMs, lenient for repository POMs). Millions of 4.0.0 POMs in Maven Central may have non-standard namespaces. This feedback remains unaddressed.
3. **No tests (medium)** — No test cases for the new behavior.

**Minor:**
4. Inconsistent `ns` variable usage — defined but `model.getNamespaceUri()` called directly
5. Method name `isModelVersionGreaterOrEqualsThan400` — consider `isModelVersionAtLeast400`
6. `!ns.isEmpty()` preferred over `ns.length() > 0`

elharo:
seems abandoned and duplicates other work

PR #11249, opened 2025-10-10
Tidy up executor UTs

cstamas:
As they seem problematic.

### DIFF
--- a/its/core-it-support/maven-it-helper/src/main/java/org/apache/maven/it/Verifier.java
+++ b/its/core-it-support/maven-it-helper/src/main/java/org/apache/maven/it/Verifier.java
@@ -108,7 +108,8 @@ public class Verifier {
 
     private final List<String> jvmArguments = new ArrayList<>();
 
-    private final String toolboxVersion = System.getProperty("version.toolbox");
+    // TestSuiteOrdering creates Verifier in non-forked JVM as well, and there no prop set is set (so use default)
+    private final String toolboxVersion = System.getProperty("version.toolbox", "0.14.0");
 
     private Path userHomeDirectory; // the user home
 


impl/maven-cli/src/test/java/org/apache/maven/cling/invoker/mvn/MavenInvokerTest.java
```diff
@@ -218,10 +218,10 @@ void conflictingSettings(
         Map<String, String> logs = invoke(
                 cwd,
                 userHome,
-                List.of("eu.maveniverse.maven.plugins:toolbox:0.7.4:help"),
+                List.of("eu.maveniverse.maven.plugins:toolbox:0.13.7:help"),
```
gnodet: Is there a way to have the version externalised in the POM so that it can be upgraded more easily ?

cstamas: Technically, these changes are not needed, are not mandatory. But I did this merely "to align" all toolbox versions used in Maven build, to not waste time on pulling various different versions of it over and over again.

Will see if we can somehow centralize the config/GAV for it.

gnodet: > Technically, these changes are not needed, are not mandatory. But I did this merely "to align" all toolbox versions used in Maven build, to not waste time on pulling various different versions of it over and over again.

Yes, and my remark goes into the exact same direction ;-). IF we can align things and have a single place of configuration, that would be even better!

impl/maven-executor/src/test/java/org/apache/maven/cling/executor/impl/ToolboxToolTest.java
```diff
@@ -25,32 +25,43 @@
...
+
+    private Path userHome;
+
+    @BeforeEach
+    void beforeEach(TestInfo testInfo) throws Exception {
+        userHome = tempDir.resolve(testInfo.getTestMethod().orElseThrow().getName());
+        Files.createDirectories(userHome);
         MimirInfuser.infuseUW(userHome);
+
+        System.out.println("=== " + testInfo.getTestMethod().orElseThrow().getName());
```
gnodet, line 64: I think we want to remove those `System.out.println` before merging...

gnodet: Or use proper logging...

cstamas: While in general I completely agree about this stance, in this very case I'd leave them; their reason of existence is really to easy our work of "figuring out what happens". As all these tests are testing "execution of maven to provide us some information" (with all the play of force stdout plus quiet mode). As long as we cannot "configure output to terminal and file differently", I think it is simplest and "cheapest" to leave them. The reason I did not want logger in test CP, is to keep CP classpath "as pristine as possible", as we have SLF4J logging already present JVM, in the "embedded Maven".

So this is my 5 cents, I can remove all this if you insist, but again, the point is to have them in surefire grabbed stdout for easier debugging, thats all. I'd not remove them, nor would use "fully fledged logging framework" either, but without them we are left in dark (as maven cannot write to stdout one thing, while logging another to file).

its/core-it-support/maven-it-helper/src/main/java/org/apache/maven/it/Verifier.java
```diff
@@ -108,6 +108,9 @@ public class Verifier {
 
     private final List<String> jvmArguments = new ArrayList<>();
 
+    // TestSuiteOrdering creates Verifier in non-forked JVM as well, and there no prop set is set (so use default)
+    private final String toolboxVersion = System.getProperty("version.toolbox", "0.14.0");
```
gnodet, line 112: Why `0.14.0` while we use `0.13.7` for `ToolboxTool` ?

cstamas: No, that ctor is deprecated and we do not use it, verifier uses ctor to explicitly set wanted toolbox version. We can remove that ctor in fact, but that would be a breakage?

PR #11274, opened 2025-10-15
Upgrade Mimir

cstamas:
And collect more information, and we seems loose some information in case of IT failures.

Changes:
* upgrade to Mimir 0.10.0
* GH action collects IT failsafe outputs as well (as I sus we have something on stdout or stderr)
* centrally manage Toolbox and Mimir versions
* make Maven 3 tests use Mimir as well (by doing UW or PW setup for it); so far those did not use Mimir, but went directly to Central instead
* in maven-executor and maven-cli test disable prefix RRF (for now)

### DIFF
--- a/impl/maven-cli/src/test/java/org/apache/maven/cling/invoker/mvn/resident/ResidentMavenInvokerTest.java
+++ b/impl/maven-cli/src/test/java/org/apache/maven/cling/invoker/mvn/resident/ResidentMavenInvokerTest.java
@@ -61,7 +61,7 @@ public class ResidentMavenInvokerTest extends MavenInvokerTestSupport {
         invoke(cwd, userHome, List.of("verify"), List.of());
     }
 
-    @Disabled("Enable when core moves off fully from FS")
+    @Disabled("Enable it when fully moved to NIO2 with Path/Filesystem (ie MavenExecutionRequest)")
     @Test
     void jimFs() throws Exception {
         try (FileSystem fs = Jimfs.newFileSystem(Configuration.unix())) {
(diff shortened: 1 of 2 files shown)


impl/maven-cli/src/test/java/org/apache/maven/cling/invoker/mvn/resident/ResidentMavenInvokerTest.java
```diff
@@ -61,7 +61,7 @@ void defaultFs(
         invoke(cwd, userHome, List.of("verify"), List.of());
     }
 
-    @Disabled("Until we move off fully from File")
+    @Disabled("Enable when core moves off fully from FS")
```
gnodet: The `FS` is ambiguous.  We need to enable it when we fully move to the NIO2 api with Path/java.nio.file.Filesystem, rather than File/java.io.FileSystem, right ?

gnodet (approved):

impl/maven-cli/src/test/java/org/apache/maven/cling/invoker/mvn/resident/ResidentMavenInvokerTest.java
cstamas: Yes, let me reword this

cstamas:
```
    @Disabled("Enable it when fully moved to NIO2 with Path/Filesystem (ie MavenExecutionRequest)")
```

PR #11315, opened 2025-10-22
Mimir Cache-Purge w Pre-seed

cstamas:
For this to work in Windows, we need latest 0.10.4 and also some changes/fixes in ITs. 

Changes:
* update to Mimir 0.10.4 + enable cache purge
* dropped ctor `Verifier(String basedir, boolean debug)` as `debug` was unused, instead introduced `Verifier(String basedir, boolean createDotMvn)` that auto-creates `.mvn` folder in basedir for each ITs. If IT does not want this, it must use alt method and have a Javadoc telling why.
* detached `its` parent POM from Maven parent POM, as it interferes with IT support built plugins (we have ITs that uses plugin built in core-it-support but in POM adds an IT specific dependency that in turn pulls in stub plexus-utils dependency; by having depMgt in plugin POM, due Maven 4 transitive dep mgmt, the 4th level p-u was NOT used, but a managed one was used instead)
* disabled RRF in ITs due https://github.com/apache/maven-resolver/issues/1641 and parallel running of IT just exacerbates this issue

### DIFF
--- a/.github/workflows/maven.yml
+++ b/.github/workflows/maven.yml
@@ -65,7 +65,7 @@ jobs:
         id: restore-cache
         with:
           path: ${{ env.MIMIR_LOCAL }}
-          key: mvn-${{ runner.os }}-initial
+          key: master-${{ runner.os }}-initial
 
       - name: Set up Maven
         shell: bash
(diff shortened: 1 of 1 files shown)


its/core-it-support/maven-it-helper/src/main/java/org/apache/maven/it/Verifier.java
```diff
@@ -231,6 +243,10 @@ public void execute() throws VerificationException {
             args.add(0, "-l");
         }
 
+        // TODO: this is just experiment; disable RRF for now
```
gnodet:
```suggestion
        // TODO: disable RRF for now until https://github.com/apache/maven-resolver/issues/1641 can be fixed
```

PR #11320, opened 2025-10-23
Restore compatibility in maven-embedder

### LINKED ISSUE #11300: [back compat] some constants and code is missing from maven-embedder

`MavenCli` is deprecated, but to ease migration it would be good to have it as "backward compatible" as possible.

Currently it misses some public static constants (e.g. `LOCAL_REPO_PROPERTY`).

Beside that the `org.apache.maven.cli.internal.extension` is missing now, instead of deleting code it would be better to mark it deprecated as well.

gnodet:
Add missing deprecated constants and classes for backward compatibility in `maven-embedder`.

Fixes #11300


gnodet:
@laeubi will that fix all your compatibility problems ?

laeubi (approved):
This looks good to me thanks!

PR #11347, opened 2025-10-28
Disable consumer POM flattening by default and add an opt-in feature

### LINKED ISSUE #11346: Add ability to disable consumer POM flattening to preserve dependency management inheritance
## Summary

Currently, Maven's consumer POM transformation flattens dependency management sections by default, which prevents dependency management from being inherited by consumers. This creates issues in scenarios where transitive dependency management inheritance is desired.

## Use Case

Consider the following dependency scenario:

- **Module A 1.0**: depends on B 1.0 and manages C to version 1.2
- **Module B 1.0**: has no dependencies
- **Module B 2.0**: depends on C 1.1
- **Module D**: depends on A 1.0 and manages B to version 2.0

### Expected Behavior
When Module D builds, it should resolve:
- B 2.0 (due to D's dependency management override)
- C 1.2 (due to A's dependency management, which should take precedence over B 2.0's direct dependency on C 1.1)

### Current Problem
With consumer POM flattening enabled (default), Module A's consumer POM loses its dependency management section. This means when Module D consumes A 1.0, the dependency management for C 1.2 is not available, and Module D ends up with C 1.1 from B 2.0's direct dependency.

### Desired Solution
A feature flag (e.g., `maven.consumer.pom.flatten=false`) that allows disabling consumer POM flattening, preserving dependency management sections in consumer POMs so they can be inherited by downstream consumers.

## Integration Test

This use case is demonstrated by the integration test `MavenITDependencyManagementOverrideTest` which:
1. Sets up the multi-module scenario described above
2. Verifies that with flattening disabled, Module D correctly resolves C 1.2 instead of C 1.1
3. Confirms that dependency management inheritance works as expected when consumer POMs preserve their dependency management sections

## Question

- Should this be the default behaviour?

gnodet:
## Summary

This PR introduces a new feature flag `maven.consumer.pom.flatten` that allows users to control whether consumer POMs are flattened by removing dependency management sections. This addresses dependency management inheritance scenarios and provides better control over consumer POM generation.

**The consumer POM are NOT flattened anymore by default.**

### Core Implementation
- **Features.java**: Added `consumerPomFlatten()` method to check the feature flag (defaults to `false`)
- **DefaultConsumerPomBuilder**: Enhanced to conditionally apply flattening based on the flag
- **Constants.java**: Added comprehensive documentation for the new feature flag and improved existing documentation
- **Dependency management control**: When disabled, consumer POMs preserve dependency management like parent POMs

### Integration Tests
- **gh-11346-dependency-management-override**: New comprehensive test demonstrating dependency management inheritance scenarios
- **Updated existing tests**: Modified several integration tests to explicitly enable flattening where needed for backward compatibility

### When `maven.consumer.pom.flatten=false` (Default)
- **Dependency management preserved**: Consumer POMs keep dependency management sections like parent POMs
- **Inheritance enabled**: Allows dependency management to be inherited by consumers
- **Non-POM projects**: Processed without flattening dependency management
- **Backward compatible**: Maintains existing dependency resolution behavior

### When `maven.consumer.pom.flatten=true` (Opt-in)
- **Dependency management removed**: Consumer POMs are flattened by removing dependency management
- **Transitive dependencies only**: Only direct dependencies with transitive scopes are kept
- **Traditional behavior**: Matches the previous Maven consumer POM behavior

## Use Case Example

Consider this dependency scenario:
- **A 1.0** depends on **B 1.0** and manages **C to 1.2**
- **B 2.0** depends on **C 1.1**  
- **D** depends on **A 1.0** and manages **B to 2.0**

**With `maven.consumer.pom.flatten=false` (default)**:
- D gets C 1.2 (A's dependency management applies to D's transitive dependencies)
- Consumer POMs preserve dependency management for inheritance

**With `maven.consumer.pom.flatten=true`**:
- Traditional flattened consumer POM behavior
- Dependency management sections are removed

## Benefits

- **Better dependency management**: Allows dependency management inheritance from consumer POMs
- **Flexible control**: Users can choose between flattened and non-flattened consumer POMs
- **Improved transitive dependency control**: Dependency management can properly control transitive dependency versions
- **Opt-in safety**: Users can choose when to enable the previous flattening behavior

Fixes #11346 and addresses dependency management inheritance scenarios

### DIFF
--- a/api/maven-api-core/src/main/java/org/apache/maven/api/Constants.java
+++ b/api/maven-api-core/src/main/java/org/apache/maven/api/Constants.java
@@ -322,18 +322,21 @@ public final class Constants {
 
     /**
      * User controlled relocations.
-     * This property is a comma separated list of entries with the syntax GAV>GAV.
-     * The first GAV can contain * for any elem (so *:*:* would mean ALL, something
-     * you don't want). The second GAV is either fully specified, or also can contain *,
+     * This property is a comma separated list of entries with the syntax <code>GAV&gt;GAV</code>.
+     * The first <code>GAV</code> can contain <code>*</code> for any elem (so <code>*:*:*</code> would mean ALL, something
+     * you don't want). The second <code>GAV</code> is either fully specified, or also can contain <code>*</code>,
      * then it behaves as "ordinary relocation": the coordinate is preserved from relocated artifact.
-     * Finally, if right hand GAV is absent (line looks like GAV>), the left hand matching
-     * GAV is banned fully (from resolving).
-     * Note: the > means project level, while >> means global (whole session level,
+     * Finally, if right hand <code>GAV</code> is absent (line looks like <code>GAV&gt;</code>), the left hand matching
+     * <code>GAV</code> is banned fully (from resolving).
+     * <br/>
+     * Note: the <code>&gt;</code> means project level, while <code>&gt;&gt;</code> means global (whole session level,
      * so even plugins will get relocated artifacts) relocation.
-     * For example: maven.relocations.entries = org.foo:*:*>, org.here:*:*>org.there:*:*, javax.inject:javax.inject:1>>jakarta.inject:jakarta.inject:1.0.5
-     * means: 3 entries, ban org.foo group (exactly, so org.foo.bar is allowed),
-     * relocate org.here to org.there and finally globally relocate (see >> above)
-     * javax.inject:javax.inject:1 to jakarta.inject:jakarta.inject:1.0.5.
+     * <br/>
+     * For example,
+     * <pre>maven.relocations.entries = org.foo:*:*>, \\<br/>    org.here:*:*>org.there:*:*, \\<br/>    javax.inject:javax.inject:1>>jakarta.inject:jakarta.inject:1.0.5</pre>
+     * means: 3 entries, ban <code>org.foo group</code> (exactly, so <code>org.foo.bar</code> is allowed),
+     * relocate <code>org.here</code> to <code>org.there</code> and finally globally relocate (see <code>&gt;&gt;</code> above)
+     * <code>javax.inject:javax.inject:1</code> to <code>jakarta.inject:jakarta.inject:1.0.5</code>.
      *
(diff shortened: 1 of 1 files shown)


cstamas (approved):

gnodet:

| Status | Branch | Result |
|:------:|:------:|:------|
|✅|maven-4.0.x|[](https://github.com/apache/maven/pull/11370)|

### Questions ?

api/maven-api-core/src/main/java/org/apache/maven/api/Constants.java
```diff
@@ -463,6 +463,18 @@ public final class Constants {
     @Config(type = "java.lang.Boolean", defaultValue = "true")
     public static final String MAVEN_CONSUMER_POM = "maven.consumer.pom";
 
+    /**
+     * User property for controlling consumer POM flattening behavior.
+     * When set to <code>true</code> (default), consumer POMs are flattened by removing
+     * dependency management and keeping only direct dependencies with transitive scopes.
+     * When set to <code>false</code>, consumer POMs preserve dependency management
```
jglick, line 470: This says `true` is the default, but the code and the PR description says `false` is the default…?

api/maven-api-core/src/main/java/org/apache/maven/api/Constants.java
```diff
@@ -463,6 +463,18 @@ public final class Constants {
     @Config(type = "java.lang.Boolean", defaultValue = "true")
     public static final String MAVEN_CONSUMER_POM = "maven.consumer.pom";
 
+    /**
+     * User property for controlling consumer POM flattening behavior.
```
jglick, line 467: Do I understand correctly that this can only be controlled by a `-D` flag? So you cannot easily enable flattening on specific modules. I guess you could add `-D` to `.mvn/maven.config` if you wished to enable flattening on all modules in a given reactor (such as one Git repository). Has any thought been given to controlling this mode based on a flag set in the POM itself (if that is even possible)? Or as a configuration option to the `install` mojo, etc.?

PR #11358, opened 2025-10-29
Use MavenRepositorySystem in ProjectBuildingHelper instead of deprecated RepositorySystem

slawekjaranowski:
RepositorySystem has legacy implementation in maven-compat.
When plugin use a ProjectBuilder we need add a maven-compat on test classpath.

Also, simple cleanups in tests.


maven-core/src/test/java/org/apache/maven/plugin/PluginParameterExpressionEvaluatorTest.java
```diff
@@ -388,7 +391,12 @@ private MavenProject createDefaultProject() {
 
     private ExpressionEvaluator createExpressionEvaluator(
             MavenProject project, PluginDescriptor pluginDescriptor, Properties executionProperties) throws Exception {
-        ArtifactRepository repo = factory.createDefaultLocalRepository();
+        ArtifactRepository repo = factory.createArtifactRepository(
```
gnodet: Should a method be added for that ?

slawekjaranowski: ok, use existing method for it

gnodet (approved):

PR #11380, opened 2025-11-03
Type derive (simpler)

cstamas:
Instead to "smear" this feature across Maven and Resolver classes, for start let's keep it "confined" with single class: the TypeDeriver.

Later we can see where to go further with it.

This PR also includes bugfix, where Maven `DefaultType` implements `ArtifactType`, while in reality it does not (violates contract by returning `null` when no classifier present).

Supersedes #11373

### DIFF
--- a/compat/maven-resolver-provider/src/main/java/org/apache/maven/repository/internal/type/DefaultType.java
+++ b/compat/maven-resolver-provider/src/main/java/org/apache/maven/repository/internal/type/DefaultType.java
@@ -42,7 +42,7 @@ import static java.util.Objects.requireNonNull;
  * @deprecated since 4.0.0, use {@code maven-api-impl} jar instead
  */
 @Deprecated(since = "4.0.0")
-public class DefaultType implements Type, ArtifactType {
+public class DefaultType implements Type {
     private final String id;
     private final Language language;
     private final String extension;
@@ -81,11 +81,6 @@ public class DefaultType implements Type, ArtifactType {
     }
 
     @Override
-    public String getId() {
-        return id();
-    }
-
-    @Override
     public Language getLanguage() {
         return language;
     }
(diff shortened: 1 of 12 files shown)


compat/maven-resolver-provider/src/main/java/org/apache/maven/repository/internal/type/DefaultType.java
```diff
@@ -42,7 +42,7 @@
  * @deprecated since 4.0.0, use {@code maven-api-impl} jar instead
  */
 @Deprecated(since = "4.0.0")
-public class DefaultType implements Type, ArtifactType {
+public class DefaultType implements Type {
```
gnodet, line 45: What's the benefit of this decoupling? It does not seem necessary.

desruisseaux:
Yes, the integration test passes because the line that cause a failure is commented out. If you open the following file:

```
target/it/processor-type/annotation-processor/src/main/java/processor/SimpleAnnotationProcessor.java
```

then uncomment the second line below:

```java
        // TODO: pending https://github.com/apache/maven/pull/11373
        // dependency.AnnotationProcessorDependency.foo();
```

then, we get the test failure. You can also check that if the value on `--class-path` is moved to `--processor-path`, it works.

compat/maven-resolver-provider/src/main/java/org/apache/maven/repository/internal/type/DefaultType.java
cstamas: The decoupling is a must as we figured: this class violates (maven or resolver, dependending how you look at it) contract for Type: classifier returns `null` instead of `""` (resolver).

desruisseaux:
I think that resolver 2.0.14 has been released, is that right? If yes, then is this pull request ready to upgrade from "draft" to "ready" state?

PR #11388, opened 2025-11-04
Change IntelliJ project icon to new oak leaf

Bukama:
as part of https://github.com/apache/maven-site/issues/1396

Icon gets rendered correctly as you can see on this screenshot

image


slawekjaranowski:
Old has size 48x48 ... new has 400x615 ... maybe we can resize a new one

Bukama:
> Old has size 48x48 ... new has 400x615 ... maybe we can resize a new one

I know and was wondering if this brings problems, but IntelliJ resizes it correctly. So no need to change it.

gnodet (approved):
LGTM, it should be back ported to 4.0.x and 3.9.x branches.

PR #11394, opened 2025-11-05
Fix resource targetPath resolution to be relative to output directory (fixes #11381)

### LINKED ISSUE #11381: [Maven 4] Maven resources plugin - relative targetPath is not the same as in Maven 3.9.x
### Affected version

Apache Maven 4.0.0-SNAPSHOT (b7f91787195232cb9aee404f4fc13dadc04d6ecb)

### Bug description

With Maven 3.9.x and this configuration:
```
    <build>
        <resources>
            <resource>
                <directory>${project.basedir}/rest</directory>
                <targetPath>target-dir</targetPath>
                <includes>
                    <include>**/*.yml</include>
                </includes>
            </resource>
        </resources>
    </build>
```
all files `rest/**/*.yml` are copied to `target/classes/target-dir`:
```
[INFO] --- resources:3.3.1:resources (default-resources) @ project1 ---
[INFO] Copying 36 resources from rest to target/classes/target-dir
```

But with Maven 4, with same version of Maven resources plugin, all files `rest/**/*.yml` are copied to root dir `target-dir`:
```
[INFO] --- resources:3.3.1:resources (default-resources) @ project1---
[INFO] Copying 36 resources from rest to target-dir
[INFO] ---------------------------------------------
```
It seams that relative targetPath is not the same as in Maven 3.9.x

A workaround is to prefix `targetPath` with `${project.build.outputDirectory}/`, it works both on Maven 3 and 4:
```
    <build>
        <resources>
            <resource>
                <directory>${project.basedir}/rest</directory>
                <targetPath>${project.build.outputDirectory}/target-dir</targetPath>
                <includes>
                    <include>**/*.yml</include>
                </includes>
            </resource>
        </resources>
    </build>
```

gnodet:
This PR fixes a regression in Maven 4 where resource `targetPath` was being resolved relative to the project base directory instead of the output directory (`target/classes` or `target/test-classes`).

## Problem

In Maven 3.x, when a resource has a relative `targetPath`, it's resolved relative to the output directory. For example:

```xml
<resource>
  <directory>${project.basedir}/rest</directory>
  <targetPath>target-dir</targetPath>
  <includes>
    <include>**/*.yml</include>
  </includes>
</resource>
```

In Maven 3.x, files would be copied to `target/classes/target-dir/`, but in Maven 4, they were being copied to `target-dir/` (relative to project root).

## Root Cause

The bug was in the `DefaultSourceRoot` constructor that takes a `Resource` parameter. On line 172, the targetPath was being resolved as:
```java
nonBlank(resource.getTargetPath()).map(baseDir::resolve).orElse(null)
```

This resolves the targetPath relative to the base directory, not the output directory.

## Solution

1. **Created a new constructor** in `DefaultSourceRoot.java` that accepts an `outputDir` parameter and correctly resolves targetPath relative to the output directory
2. **Deprecated the old constructor** to maintain backward compatibility
3. **Updated all call sites** to use the new constructor:
   - `DefaultProjectBuilder.java`
   - `ConnectedResource.java`
   - `MavenProject.java`
4. **Updated unit tests** in `DefaultSourceRootTest.java` to verify the correct behavior
5. **Created an integration test** `MavenITgh11381ResourceTargetPathTest.java` to verify the fix works end-to-end

## Testing

The integration test verifies that:
- Resources with relative `targetPath` are copied to the correct location under the output directory
- The fix works for both main and test scopes

Test results:
```
Tests run: 1, Failures: 0, Errors: 0, Skipped: 0
```

Fixes #11381

---


impl/maven-impl/src/main/java/org/apache/maven/impl/DefaultSourceRoot.java
```diff
@@ -169,7 +169,7 @@ public DefaultSourceRoot(final Path baseDir, ProjectScope scope, Resource resour
                 resource.getIncludes(),
                 resource.getExcludes(),
                 Boolean.parseBoolean(resource.getFiltering()),
-                nonBlank(resource.getTargetPath()).map(baseDir::resolve).orElse(null),
+                nonBlank(resource.getTargetPath()).map(Path::of).orElse(null),
```
cstamas, line 176: This is CWD relative?

desruisseaux: `Path.of(String)` creates a path relative the to current working directory, which is not the desired output directory. I think that `DefaultSourceRoot` is okay, and the change rather needs to be applied in the codes that call this constructor. The problem seems to be in the following classes:

* `org.apache.maven.project.ConnectedResource` line 124.
* `org.apache.maven.project.DefaultProjectBuilder` lines 704 and 707.
* `org.apache.maven.project.MavenProject` line 840

The above-cited codes set the first parameter to `project.getBaseDirectory()` while it should be `project.getBuild().getOutputDirectory()`.

Note: the `baseDir` parameter name of `DefaultSourceRoot` is misleading. The Javadoc said _"the base directory for resolving relative paths"_, which can be the `baseDir` of the Maven project but not necessarily. It should be understood as any base directory for resolving paths. This parameter could be renamed `parentDir` if `baseDir` is too misleading.

cstamas: Is maybe `resource.getTargetPath()` absolute? So CWD does not matter?

desruisseaux: It may be absolute, but I'm not sure that this guaranteed. It may depend on which code path as leaded to this constructor. I don't think that there is any reason to not be robust. We just need to find all calls to this constructor.

desruisseaux: And maybe rename the `baseDir` parameter as `outputDir` (but without change in the call to `Path.resolve(String)`).

gnodet: > `Path.of(String)` creates a path relative the to current working directory, which is not the desired output directory. I think that `DefaultSourceRoot` is okay, and the change rather needs to be applied in the codes that call this constructor. The problem seems to be in the following classes:
> 
> * `org.apache.maven.project.ConnectedResource` line 124.
> * `org.apache.maven.project.DefaultProjectBuilder` lines 704 and 707.
> * `org.apache.maven.project.MavenProject` line 840
> 
> The above-cited codes set the first parameter to `project.getBaseDirectory()` while it should be `project.getBuild().getOutputDirectory()`.
> 
> Note: the `baseDir` parameter name of `DefaultSourceRoot` is misleading. The Javadoc said _"the base directory for resolving relative paths"_, which can be the `baseDir` of the Maven project but not necessarily. It should be understood as any base directory for resolving paths. This parameter could be renamed `parentDir` if `baseDir` is too misleading.

`Path.of(String)` just creates a relative path. But not relative to anything.  It's only when resolved against a directory that it will become relative to that directory.
The built `Path` will be resolved against the correct directory, that's what happens in the maven-resource-plugin for example, which will resolve this path against either `project.build.outputDirectory` or `project.build.testOutputDirectory`, depending which mojo is called.
I agree that there may be a mismatch, but it sounds difficult to just thrown away the m-resources-p at the moment.

gnodet: Fwiw, the expectation in Maven 3 is that `Resource.getTargetPath()` will be resolved against either `project.build.outputDirectory` or `project.build.testOutputDirectory`.  The main usage *is* the maven-resource-plugin here, which is currently broken.

desruisseaux: > Fwiw, the expectation in Maven 3 is that `Resource.getTargetPath()` will be resolved against either `project.build.outputDirectory` or `project.build.testOutputDirectory`.

Yes I agree, but the replacement of `baseDir::resolve` by `Path::of` does not fix that. The correction is rather to fix the `baseDir` argument given in *calls* to the `DefaultSourceRoot` constructor. Actually, the [first commit of this pull request](https://github.com/apache/maven/pull/11394/commits/eb23cfcee98d3a4d9136ccc69f115d3d51f70607) seems correct to me, but it seems to have been reverted in subsequent commits.

gnodet: Why do you think it does not ?
It makes the path truly relative, and later resolved against the correct directory.

desruisseaux: > It makes the path truly relative, and later resolved against the correct directory.

Ah, I see. The responsibility of resolving the path is shifted to the `SourceRoot` user instead of the `SourceRoot` constructor. But why not resolving at construction time? It would probably be less code to check for correctness.

gnodet: I agree.  I think that would be desired behavior, but then we need to make it truly relative in the ConnectedResource to not break the existing plugin.
So the behavior would be different between mvn3 and mvn4 api, but that's fine.

impl/maven-core/src/main/java/org/apache/maven/project/ConnectedResource.java
```diff
@@ -42,13 +42,53 @@ class ConnectedResource extends Resource {
...
                 .build());
         this.originalSourceRoot = sourceRoot;
         this.scope = scope;
         this.project = project;
     }
 
+    /**
+     * Computes the targetPath relative to the output directory.
+     * In Maven 3 API, Resource.getTargetPath() is expected to be relative to the output directory
+     * (e.g., "custom-output"), while SourceRoot.targetPath() is relative to the project basedir
```
desruisseaux: > while SourceRoot.targetPath() is relative to the project basedir

Actually it was a bug. The specification that we wrote in `maven.mdo` said that `targetPath` shall be relative to the target directory. Fixing that bug was the subject of #11322, but that correction was incomplete. We could said that this pull request #11394 completes #11322.

impl/maven-impl/src/main/java/org/apache/maven/impl/DefaultSourceRoot.java
desruisseaux: > So the behavior would be different between mvn3 and mvn4 api

If the difference would be that mvn3 would be relative to `target` while mvn4 would be relative to `baseDir`, I think that this difference was a bug partially fixed in #11322, but we forgot the case of resources in that fix.

I will prepare an alternative pull request tomorrow for illustrating the fix that I propose. The actual pull request at the end may be, possibly, a mix of the two.

gnodet:
## Implementation Details

I've implemented a fix for this issue that ensures `SourceRoot.targetPath()` and `ConnectedResource` behave correctly according to the Maven 4 API while preserving Maven 3 compatibility.

### Root Cause

The regression occurred because `DefaultSourceRoot.fromModel()` was storing `targetPath` as an **absolute path** (resolved against `baseDir` and `outputDir`), when it should have been stored as a **relative path** according to the Maven 4 API design.

### Maven 4 API Design

The Maven 4 API defines two methods for `SourceRoot`:

1. **`targetPath()`** - Returns an `Optional<Path>` containing the explicit target path. According to the javadoc and the presence of the resolution method below, this should be a **relative path** (or absolute if explicitly specified as absolute).

2. **`targetPath(Project)`** - Returns the resolved absolute path by taking the relative `targetPath()` and resolving it against the project's output directory:
   ```java
   Path base = project.getOutputDirectory(scope());
   return targetPath.map(base::resolve).orElse(base);
   ```

This two-method design clearly indicates that `targetPath()` should return a **relative** path that gets resolved by `targetPath(Project)`.

### Implementation

**1. Fixed `DefaultSourceRoot.fromModel()`** (impl/maven-impl)
```java
// Before (incorrect):
nonBlank(source.getTargetPath())
    .map((targetPath) -> baseDir.resolve(outputDir.apply(scope)).resolve(targetPath))
    .orElse(null)

// After (correct):
nonBlank(source.getTargetPath()).map(Path::of).orElse(null)
```

Now `targetPath()` returns a **relative path** as intended, and `targetPath(Project)` handles the resolution.

**2. Simplified `ConnectedResource.computeRelativeTargetPath()`** (impl/maven-core)

Since `SourceRoot.targetPath()` now always returns a relative path, we can simply return it as-is:
```java
return sourceRoot.targetPath().map(Path::toString).orElse(null);
```

This removed ~40 lines of complex logic that was trying to convert absolute paths back to relative paths.

### Maven 3 Compatibility

In Maven 3, `Resource.getTargetPath()` was always **relative to the output directory**. This behavior is preserved:

- `SourceRoot.targetPath()` stores the path as relative (e.g., `"custom-dir"`)
- `ConnectedResource` converts it back to a String for the Maven 3 `Resource` API
- Resources are copied to `target/classes/custom-dir` (Maven 3 behavior) ✅

### Example

With `<targetPath>custom-dir</targetPath>` in a resource configuration:

- **Maven 3**: Resources copied to `target/classes/custom-dir` ✅
- **Maven 4 (before fix)**: Resources copied to `project-root/custom-dir` ❌
- **Maven 4 (after fix)**: Resources copied to `target/classes/custom-dir` ✅

### Testing

All existing tests pass, including:
- `DefaultSourceRootTest` - Updated to expect relative paths from `targetPath()`
- `ResourceIncludeTest` - Validates the Maven 3 compatibility layer
- Integration test `MavenITgh11381ResourceTargetPathTest` should now pass

The fix is minimal, focused, and aligns with the Maven 4 API design while maintaining backward compatibility.

gnodet:
## Latest Update: Enhanced Javadoc for API Clarity

I've just pushed an update that significantly improves the documentation to clarify the Maven 4 API semantics while making the Maven 3 backward compatibility guarantees explicit.

### What Changed

Enhanced javadoc for three key methods:

1. **`SourceRoot.targetPath()`** - Now clearly documents:
   - Relative paths are resolved relative to the output directory (not project basedir)
   - Absolute paths are used as-is
   - Empty/null means files go to output directory root
   - Explicitly states Maven 3 compatibility is maintained

2. **`SourceRoot.targetPath(Project)`** - Now includes:
   - Step-by-step resolution algorithm
   - Concrete examples with actual paths
   - Cross-references to related methods

3. **`Project.getOutputDirectory(ProjectScope)`** - Now clarifies:
   - Returns directory for both compiled classes AND resources
   - Its role in SourceRoot targetPath resolution
   - Maven 3 compatibility semantics

4. **`DefaultSourceRoot` constructor from Resource** - Now documents:
   - targetPath is stored as-is (not resolved against basedir)
   - baseDir parameter is only used for source directory resolution
   - How this preserves Maven 3.x behavior

### Why This Matters

The enhanced documentation makes it crystal clear that:

- **For Maven 4 API users**: Use `targetPath(Project)` to get the fully resolved absolute path
- **For Maven 3 compatibility**: The `targetPath()` method returns the relative path that plugins expect
- **For implementers**: The separation of concerns between storage and resolution is now explicit

### Implementation Verification

The documentation now matches the actual implementation:

✅ `DefaultSourceRoot` stores `targetPath` as-is (relative or absolute)

✅ `ConnectedResource` extracts `targetPath` as a relative path string

✅ maven-resources-plugin receives the relative path and resolves it (Maven 3 behavior)

✅ Maven 4 API consumers can use `targetPath(Project)` for absolute paths

See the detailed explanation in [this comment on #11381](https://github.com/apache/maven/issues/11381#issuecomment-3502314999) for the complete design rationale.

PR #11389, opened 2025-11-05
Introduce TempFileService and lifecycle cleanup participant for Maven 4.

arturobernalg:
Following this checklist to help us incorporate your
contribution quickly and easily:

- [X ] Your pull request should address just one issue, without pulling in other changes.
- [ X] Write a pull request description that is detailed enough to understand what the pull request does, how, and why.
- [ X] Each commit in the pull request should have a meaningful subject line and body.
  Note that commits might be squashed by a maintainer on merge.
- [X ] Write unit tests that match behavioral changes, where the tests fail if the changes to the runtime are not applied.
  This may not always be possible but is a best-practice.
- [X ] Run `mvn verify` to make sure basic checks pass.
  A more thorough check will be performed on your pull request automatically.
- [ ] You have run the [Core IT][core-its] successfully.

If your pull request is about ~20 lines of code you don't need to sign an
[Individual Contributor License Agreement](https://www.apache.org/licenses/icla.pdf) if you are unsure
please ask on the developers list.

To make clear that you license your contribution under
the [Apache License Version 2.0, January 2004](http://www.apache.org/licenses/LICENSE-2.0)
you have to acknowledge this by using the following check-box.

- [ ] I hereby declare this contribution to be licenced under the [Apache License Version 2.0, January 2004](http://www.apache.org/licenses/LICENSE-2.0)
- [ ] In any other case, please file an [Apache Individual Contributor License Agreement](https://www.apache.org/licenses/icla.pdf).

[core-its]: https://maven.apache.org/core-its/core-it-suite/

PR #11405, opened 2025-11-07
 Introduce TempFileService and lifecycle cleanup participant

arturobernalg:

    [X ] Your pull request should address just one issue, without pulling in other changes.
    [ X] Write a pull request description that is detailed enough to understand what the pull request does, how, and why.
    [ X] Each commit in the pull request should have a meaningful subject line and body.
    [X ] Write unit tests that match behavioral changes, where the tests fail if the changes to the runtime are not applied.
    [X ] Run mvn verify to make sure basic checks pass.

    You have run the [Core IT](https://maven.apache.org/core-its/core-it-suite/) successfully.

I hereby declare this contribution to be licenced under the [Apache License Version 2.0, January 2004](http://www.apache.org/licenses/LICENSE-2.0)
In any other case, please file an [Apache Individual Contributor License Agreement](https://www.apache.org/licenses/icla.pdf).

### DIFF
--- a/api/maven-api-core/src/main/java/org/apache/maven/api/Constants.java
+++ b/api/maven-api-core/src/main/java/org/apache/maven/api/Constants.java
@@ -739,5 +739,11 @@ public final class Constants {
      */
     public static final String MAVEN_MODEL_PROCESSOR_REFERENCE_TYPE_PREFIX = "maven.model.processor.referenceType.";
 
+    /**
+     * System property to keep temp material for diagnostics.
+     *
+     */
+    public static final String KEEP_PROP = "maven.tempfile.keep";
+
     private Constants() {}
 }
--- a/api/maven-api-core/src/main/java/org/apache/maven/api/Constants.java
+++ b/api/maven-api-core/src/main/java/org/apache/maven/api/Constants.java
@@ -868,5 +868,11 @@ public final class Constants {
     @Config(type = "java.lang.Boolean", defaultValue = "false")
     public static final String MAVEN_MODEL_DEPENDENCY_INTERPOLATION_FULL = "maven.model.dependencyInterpolation.full";
 
+    /**
+     * System property to keep temp material for diagnostics.
+     *
+     */
+    public static final String KEEP_PROP = "maven.tempfile.keep";
+
     private Constants() {}
 }
--- a/api/maven-api-core/src/main/java/org/apache/maven/api/Constants.java
+++ b/api/maven-api-core/src/main/java/org/apache/maven/api/Constants.java
@@ -870,9 +870,15 @@ public final class Constants {
 
     /**
      * System property to keep temp material for diagnostics.
+     * When set to {@code true} via {@code -Dmaven.tempfile.keep=true}, temporary files and directories
+     * created by {@link org.apache.maven.api.services.TempFileService} are not deleted at session end.
+     * This is a JVM system property only (read via {@link Boolean#getBoolean(String)});
+     * it is not resolved through Maven's property mechanism (pom.xml, profiles, settings.xml).
      *
+     * @since 4.1.0
      */
-    public static final String KEEP_PROP = "maven.tempfile.keep";
+    @Config(type = "java.lang.Boolean", defaultValue = "false")
+    public static final String MAVEN_TEMPFILE_KEEP = "maven.tempfile.keep";
 
     private Constants() {}
 }
--- a/api/maven-api-core/src/main/java/org/apache/maven/api/services/TempFileService.java
+++ b/api/maven-api-core/src/main/java/org/apache/maven/api/services/TempFileService.java
@@ -1,0 +1,69 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.maven.api.services;
+
+import java.io.IOException;
+import java.nio.file.Path;
+
+import org.apache.maven.api.Service;
+import org.apache.maven.api.Session;
+import org.apache.maven.api.annotations.Nonnull;
+
+/**
+ * Service to create and track temporary files/directories for a Maven build.
+ * All created paths are deleted automatically when the session ends.
+ */
+public interface TempFileService extends Service {
+
+    /**
+     * Creates a temp file in the default temp directory.
+     */
+    @Nonnull
+    Path createTempFile(Session session, String prefix, String suffix) throws IOException;
+
+    /**
+     * Creates a temp file in the given directory.
+     */
+    @Nonnull
+    Path createTempFile(Session session, String prefix, String suffix, Path directory) throws IOException;
+
+    /**
+     * Creates a temp directory in the default temp directory.
+     */
+    @Nonnull
+    Path createTempDirectory(Session session, String prefix) throws IOException;
+
+    /**
+     * Creates a temp directory in the given directory.
+     */
+    @Nonnull
+    Path createTempDirectory(Session session, String prefix, Path directory) throws IOException;
+
+    /**
+     * Registers an externally created path for cleanup at session end.
+     */
+    @Nonnull
+    void register(Session session, Path path);
+
+    /**
+     * Forces cleanup for the given session (normally called by lifecycle).
+     */
+    @Nonnull
+    void cleanup(Session session) throws IOException;
+}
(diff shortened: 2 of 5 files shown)


api/maven-api-core/src/main/java/org/apache/maven/api/Constants.java
```diff
@@ -739,5 +739,11 @@ public final class Constants {
      */
     public static final String MAVEN_MODEL_PROCESSOR_REFERENCE_TYPE_PREFIX = "maven.model.processor.referenceType.";
 
+    /**
+     * System property to keep temp material for diagnostics.
+     *
+     */
+    public static final String KEEP_PROP = "maven.tempfile.keep";
```
gnodet: Rename to `MAVEN_TEMPFILE_KEEP` for consistency.

gnodet (changes requested):
Small change needed

gnodet (commented):
## Review: Introduce TempFileService and lifecycle cleanup participant

The concept is sound — providing a managed temp file service with automatic cleanup at session end is a useful addition. However, there are several API design issues that should be addressed before merging into the Maven 4 API.

### High

**`@Nonnull` on `void` return methods** — `TempFileService.register()` and `cleanup()` have `@Nonnull` on their `void` return type. This is semantically meaningless — `void` methods have no return value to be null or non-null. Other services (e.g., `ArtifactManager.setPath()`) do not annotate `void` returns with `@Nonnull`. Remove `@Nonnull` from these two methods and instead annotate the parameters.

### Medium

1. **`KEEP_PROP` naming convention** (`Constants.java`) — Every other constant in `Constants.java` uses the `MAVEN_` prefix (113 occurrences). The new constant should follow suit: `MAVEN_TEMPFILE_KEEP` instead of `KEEP_PROP`.

2. **Missing `@Experimental` annotation** (`TempFileService.java`) — Every other `Service` interface in `org.apache.maven.api.services` carries `@Experimental` (`ArtifactResolver`, `ArtifactManager`, `Interpolator`, etc.). This should too.

3. **`Boolean.getBoolean(KEEP_PROP)` reads JVM system property** (`DefaultTempFileService.java`) — `Boolean.getBoolean()` only checks `System.getProperty()`, so the property cannot be set via `pom.xml` `<properties>`, profiles, or `settings.xml`. Consider using `session.getUserProperties()` or `session.getSystemProperties()` if this should be settable through Maven's property resolution. If this is intentionally a JVM-level diagnostic flag, that's fine — just document the rationale.

4. **Missing `@since` and `@Config`** (`Constants.java`) — All recent constant additions have `@since 4.1.0` and a `@Config` annotation (e.g., `MAVEN_CACHE_STATS`, `MAVEN_MODEL_PROCESSOR_POOLED_TYPES`). Both are missing here.

### Low

5. **Parameters missing `@Nonnull`/`@Nullable`** (`TempFileService.java`) — Other services in this package annotate all parameters (`Interpolator`, `ArtifactManager`). The `session` and `directory` parameters should be `@Nonnull`; clarify whether `prefix`/`suffix` accept null (they do per `java.nio.file.Files` contract).

6. **Logger created inline** (`TempFileCleanupParticipant.java`) — The logger is created at the call site: `LoggerFactory.getLogger(TempFileCleanupParticipant.class).warn(...)`. This is inconsistent with `DefaultTempFileService` in the same PR, which correctly uses a `private static final Logger LOGGER` field. Extract to a field.

7. **`System.setProperty` in tests** (`DefaultTempFileServiceTest.java`) — `System.setProperty` mutates global JVM state and can cause test pollution in parallel execution. Also, there is no test for `TempFileCleanupParticipant` itself.

8. **Raw type cast for `TMP_KEY`** (`DefaultTempFileService.java`) — The `@SuppressWarnings({"unchecked", "rawtypes"})` cast is acknowledged and documented. Consider a wrapper type to avoid the raw cast, or accept it as a known limitation of `SessionData.Key` generics.

### Notes

- The companion PR #11389 targeting `maven-4.X` is closed. Please confirm this is the intended target branch.
- The PR description is minimal — for a new public API addition, documenting the motivation (what temp files are currently leaked, which plugins benefit) would help reviewers.
- Thread safety looks correct: `ConcurrentHashMap`-backed `Set` and `SessionData.computeIfAbsent` provide atomic initialization.
- No new dependencies introduced.

---

api/maven-api-core/src/main/java/org/apache/maven/api/services/TempFileService.java
```diff
@@ -0,0 +1,69 @@
...
+     * Creates a temp directory in the given directory.
+     */
+    @Nonnull
+    Path createTempDirectory(Session session, String prefix, Path directory) throws IOException;
+
+    /**
+     * Registers an externally created path for cleanup at session end.
+     */
+    @Nonnull
+    void register(Session session, Path path);
```
gnodet: `@Nonnull` on a `void` return type is semantically meaningless. In existing services (`ArtifactDeployer.deploy`, `ArtifactManager.setPath`), `@Nonnull` is applied to parameters, not void returns.

Same issue on `cleanup()` below.

```suggestion
    void register(@Nonnull Session session, @Nonnull Path path);
```

api/maven-api-core/src/main/java/org/apache/maven/api/Constants.java
gnodet: All 113 other constants in this file use the `MAVEN_` prefix (`MAVEN_HOME`, `MAVEN_CACHE_STATS`, `MAVEN_REPO_LOCAL`, etc.). This should be `MAVEN_TEMPFILE_KEEP`. Also missing `@since 4.1.0` and `@Config` annotation (71 other constants have `@Config`).

```suggestion
    /**
     * System property to keep temp material for diagnostics.
     *
     * @since 4.1.0
     */
    @Config(type = "java.lang.Boolean", defaultValue = "false")
    public static final String MAVEN_TEMPFILE_KEEP = "maven.tempfile.keep";
```

gnodet (commented):
New `TempFileService` API with lifecycle cleanup — solid concept for managing build temp files. All findings below were independently verified (7/7 confirmed).

Note: a maintainer (gnodet) already submitted CHANGES_REQUESTED in November 2025, and these issues remain unaddressed after 8+ months.

**Convention violations:**
1. **@Nonnull on void returns (high)** — `register()` and `cleanup()` apply `@Nonnull` to the void return type. Existing services apply it to parameters instead.
2. **Missing @Experimental (medium)** — All 20+ Service interfaces carry `@Experimental`. Also missing `@since 4.1.0`.
3. **KEEP_PROP naming (medium)** — All 113 other constants use the `MAVEN_` prefix. Should be `MAVEN_TEMPFILE_KEEP`. Also missing `@Config` and `@since 4.1.0`. (Already noted by gnodet in Nov 2025 review.)
4. **Boolean.getBoolean limitation (medium)** — Only reads JVM `-D` properties, not Maven session properties from pom.xml/profiles/settings.xml. Should be documented if intentional.
5. **Inline logger (low)** — `TempFileCleanupParticipant` creates logger inline; `DefaultTempFileService` uses standard `private static final Logger`.
6. **Missing parameter annotations (low)** — `session`, `directory`, `prefix`, `suffix` parameters lack `@Nonnull`/`@Nullable`.
7. **System.setProperty in test (low)** — Mutates global JVM state; `@AfterEach` cleanup present but can still cause flakiness under parallel execution. No test for `TempFileCleanupParticipant`.

gnodet (commented):
### Review Summary

The `TempFileService` design is clean and follows Maven 4 service conventions well — using `SessionData` for per-session tracking, `@Singleton` supporting multiple sessions, and the `AbstractMavenLifecycleParticipant` for cleanup are all consistent with existing patterns.

The existing review feedback from @gnodet (constant naming to `MAVEN_TEMPFILE_KEEP`, adding `@Config`/`@since`, removing `@Nonnull` on void methods) appears to have been addressed in the second commit.

Two robustness issues in the `deleteTree` cleanup logic and a minor test coverage gap:

#### 1. 🐛 Missing `visitFileFailed` override in `deleteTree` (medium)

**File:** `impl/maven-core/src/main/java/org/apache/maven/internal/impl/DefaultTempFileService.java`

The `SimpleFileVisitor` in `deleteTree` does not override `visitFileFailed`. The default implementation re-throws the `IOException`, causing `walkFileTree` to abort immediately. On Windows where files can be locked by antivirus or other processes, a single inaccessible file prevents cleanup of the entire directory tree.

**Suggestion:** Override `visitFileFailed` to log and return `CONTINUE` for best-effort cleanup:
```java
@Override
public FileVisitResult visitFileFailed(final Path file, final IOException exc) {
    LOGGER.debug("Could not access temp file for deletion: {}", file, exc);
    return FileVisitResult.CONTINUE;
}
```

#### 2. 🐛 `postVisitDirectory` should catch `DirectoryNotEmptyException` (medium)

**File:** `impl/maven-core/src/main/java/org/apache/maven/internal/impl/DefaultTempFileService.java`

The `postVisitDirectory` override silently drops the `exc` parameter but does not wrap `Files.deleteIfExists(dir)` in a try-catch. If any children were not deleted (e.g., due to a locked file after fixing issue 1), `deleteIfExists` throws `DirectoryNotEmptyException`, which propagates through `walkFileTree` and aborts cleanup of the remaining tree.

**Suggestion:** Wrap in try-catch for best-effort cleanup:
```java
@Override
public FileVisitResult postVisitDirectory(final Path dir, final IOException exc) {
    try {
        Files.deleteIfExists(dir);
    } catch (final IOException e) {
        LOGGER.debug("Could not delete temp directory: {}", dir, e);
    }
    return FileVisitResult.CONTINUE;
}
```

These two issues are interrelated — fixing #1 makes #2 more likely to trigger, so both should be addressed together.

#### 3. Missing test coverage (low)

**File:** `impl/maven-core/src/test/java/org/apache/maven/internal/impl/DefaultTempFileServiceTest.java`

The tests only exercise the no-directory overloads. The directory-parameter overloads (`createTempFile(session, prefix, suffix, directory)` and `createTempDirectory(session, prefix, directory)`) exercise different `Files.createTempFile(directory, ...)` / `Files.createTempDirectory(directory, ...)` code paths and have their own `Objects.requireNonNull(directory)` guard, none of which is tested.

---

PR #11425, opened 2025-11-10
Fix field accessibility leak in EnhancedCompositeBeanHelper

gnodet:
The previous implementation cached field accessibility state globally, which could cause issues when the same field is accessed from different contexts or security managers. This was particularly problematic in plugin unit tests where fields would remain accessible after being set.

The fix ensures that field accessibility is properly restored to its original state after setting field values, preventing accessibility state from leaking between different bean instances.

Added unit tests to verify:
- Field accessibility is restored after setting values
- Multiple field accesses don't leak accessibility state
- The fix works correctly across different bean instances

This resolves the issue reported on the dev list regarding compiler plugin unit test failures related to field accessibility.

### DIFF
--- a/impl/maven-core/src/main/java/org/apache/maven/configuration/internal/EnhancedCompositeBeanHelper.java
+++ b/impl/maven-core/src/main/java/org/apache/maven/configuration/internal/EnhancedCompositeBeanHelper.java
@@ -301,17 +301,10 @@ public final class EnhancedCompositeBeanHelper {
      * Set field value with cached accessibility.
      */
     private void setFieldValue(Object bean, Field field, Object value) throws IllegalAccessException {
-        boolean wasAccessible = field.canAccess(bean);
-        if (!wasAccessible) {
+        if (!field.canAccess(bean)) {
             field.setAccessible(true);
         }
-        try {
-            field.set(bean, value);
-        } finally {
-            if (!wasAccessible) {
-                field.setAccessible(false);
-            }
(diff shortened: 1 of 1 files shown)


cstamas (approved):

impl/maven-core/src/main/java/org/apache/maven/configuration/internal/EnhancedCompositeBeanHelper.java
```diff
@@ -304,20 +301,17 @@ private Object convertProperty(
...
-            isAccessible = field.canAccess(bean);
-            if (!isAccessible) {
-                field.setAccessible(true);
-                isAccessible = true;
-            }
-            ACCESSIBLE_FIELD_CACHE.put(field, isAccessible);
-        } else if (!isAccessible) {
+        boolean wasAccessible = field.canAccess(bean);
+        if (!wasAccessible) {
             field.setAccessible(true);
```
desruisseaux, line 305: This code assumes that the same `Field` object will not be set by two threads in same time. It seems to imply that `EnhancedCompositeBeanHelper` is not really thread-safe. If this limitation okay?

gnodet: Good catch. I'll remove the `setAccessible(false)`, as once the field has been made accessible, there's no real need to un-expose it.

PR #11463, opened 2025-11-18
Add validation for mixins in consumer POM without flattening (fixes #11456)

### LINKED ISSUE #11456: Add ITs for mixins and flattened / non-flattened Consumer POM
> Info: Was curious if master/maven 4.1.0 is prepared for model version 4.2.0 and noticed that it is 😄

Yes, model 4.2.0 is required to introduce mixins.
Btw, mixins should be a problem since we removed the Consumer POM flattening in rc-5.  We need to add ITs about the downgrade of POM using mixins, as they it should fail if flattening is disabled (since, mixins cannot be part of the Consumer POM).

_Originally posted by @gnodet in https://github.com/apache/maven/issues/11453#issuecomment-3541220887_

gnodet:
## Description

This PR adds validation to prevent creating consumer POMs when mixins are present without flattening enabled.

## Changes

- **Added validation in `DefaultConsumerPomBuilder`**: Checks for mixins when flattening is disabled and throws `MavenException` with a helpful error message
- **Added integration test `MavenITgh11456MixinsConsumerPomTest`** with two scenarios:
  1. `testMixinsWithoutFlattening`: Verifies build fails with proper error message
  2. `testMixinsWithFlattening`: Verifies build succeeds and mixins are removed from consumer POM when flattening is enabled
  3. `testMixinsWithPreserveModelerVersion`: Verifies that mixins are kept when preserving the modelVersion in the consumer POM.
- **Created test resources**: POMs using mixins with local repository containing mixin artifacts

## Problem

Mixins require model version 4.2.0 and cannot be part of the consumer POM. When consumer POM flattening is disabled (`maven.consumer.pom.flatten=false`), mixins would remain in the consumer POM, which is invalid.

## Solution

Maven now fails early with a clear error message that guides users to either:
- Enable flattening by setting `maven.consumer.pom.flatten=true`, or
- Set the `preserve.model.version=true` flag on the POM, or 
- Remove mixins from their POM

## Testing

- Full build passes: `mvn clean install -DskipTests`
- Code formatting verified: `mvn spotless:check`
- Integration tests created to verify both failure and success scenarios

Fixes https://github.com/apache/maven/issues/11456

---

### DIFF
--- a/impl/maven-core/src/main/java/org/apache/maven/internal/transformation/impl/DefaultConsumerPomBuilder.java
+++ b/impl/maven-core/src/main/java/org/apache/maven/internal/transformation/impl/DefaultConsumerPomBuilder.java
@@ -54,6 +54,45 @@ import org.eclipse.aether.RepositorySystemSession;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+/**
+ * Builds consumer POMs from project models, transforming them into a format suitable for downstream consumers.
+ * <p>
+ * A consumer POM is a simplified version of a project's POM that is published for consumption by other projects.
+ * It removes build-specific information and internal details while preserving essential information like
+ * dependencies, repositories, and distribution management.
+ * <p>
+ * This builder applies two orthogonal transformations:
+ * <ul>
+ *   <li><b>Dependency Flattening</b>: When enabled via {@code maven.consumer.pom.flatten=true}, dependency management
+ *       is flattened into direct dependencies for non-POM projects, and mixins are removed.</li>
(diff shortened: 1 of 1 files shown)


rmannibucau (approved):

cstamas (approved):
Note: I keep coming back to this, but we may tie these features to "maven level" (model version?) instead to have these freely (and independent) flags for ON/OFF them?

gnodet:
FTR, I've slightly modified the check to take into account the `preserveModelVersion` which can be used to keep the consumer POM with a `modelVersion` of 4.2.0, in which case we can safely keep `mixin` elements.

PR #11479, opened 2025-11-22
Allow a Maven plugin to require a Java version

slawekjaranowski:
Introduced in Maven 4.0.0-alpha-3 - MNG-7566
Plugin tools generate it since 3.8.1 - MPLUGIN-425

As many plugins required newer JDK we can introduce it also for Maven 3.x for better error reporting

--- 


slawekjaranowski:
Integration test:
 - https://github.com/apache/maven-integration-testing/pull/415

maven-core/src/main/java/org/apache/maven/plugin/internal/DefaultMavenPluginManager.java
```diff
@@ -279,22 +284,37 @@ public MojoDescriptor getMojoDescriptor(
...
+                    "The plugin " + pluginDescriptor.getId() + " has unmet prerequisites: " + messages,
+                    prerequisiteExceptions.get(0));
+            // the first exception is added as cause, all other ones as suppressed exceptions
+            prerequisiteExceptions.stream().skip(1).forEach(pie::addSuppressed);
+            throw pie;
         }
     }
 
+    @Override
+    @Deprecated
```
slachiewicz, line 313: Please add info what now is recommended

slawekjaranowski: it is provided at interface level, here we have an implementation

slachiewicz:
Nice idea. Then versions-plugin can use this informations to limit offered version upgrades

Bukama:
> Nice idea. Then versions-plugin can use this informations to limit offered version upgrades

Yep. An in addition enforcer-plugin (or dependency plugin?) could extend to check if minimum java version is valid in terms of required version

slawekjaranowski:
> Another thought: This feature handles the Java versions as minimum Java version.
> 
> Maybe extend it (from the beginning on) with a maximum Java version to define "this works not on Java Y or higher, please update to a newer plugin version" or something like this? This would allow maintainers who explict define ranges, esp when maintaining multiple releases.
> 
> Ofc it does not help in all cases.

The filed `requiredJavaVersion` is generated by plugin tools and maintainers can provide a range
https://maven.apache.org/plugin-tools/maven-plugin-plugin/descriptor-mojo.html#requiredJavaVersion

PR #11365, opened 2025-10-30
Fix special characters in .mvn/jvm.config (fix #11363, #11485 and #11486)

gnodet:
This PR fixes multiple related issues with special character handling in `.mvn/jvm.config` files by unifying the parsing approach to use `JvmConfigParser.java` on both Unix and Windows platforms.

## Issues Fixed

- **MNG-11363**: Pipe symbols (`|`) in jvm.config cause shell parsing errors
- **MNG-11485**: Non-POSIX `xargs -0` fails on AIX, FreeBSD, and other systems  
- **MNG-11486**: `sed` command fails when paths contain `@` symbols

## Related PRs

- Supersedes #11489 (MNG-11485: POSIX compliance)
- Supersedes #11499 (MNG-11486: @ symbols in paths)

## Solution

Instead of using complex shell/batch script parsing with `awk`, `sed`, and `xargs`, this PR uses the existing `JvmConfigParser.java` on both platforms:

### Unix (`mvn`)
- Compiles `JvmConfigParser.java` to a temporary directory
- Runs the parser to process `jvm.config`
- Cleans up temporary files
- Falls back gracefully if compilation fails

### Windows (`mvn.cmd`)
- Already used `JvmConfigParser.java` (no functional changes)
- Removed debug output
- Updated comments

## Benefits

✅ **Correctness**: Handles all edge cases reliably (pipes, quotes, `@`, etc.)  
✅ **POSIX Compliance**: No non-standard tools needed (`xargs -0`, GNU-specific `awk`/`sed`)  
✅ **Maintainability**: Single implementation for both platforms  
✅ **Consistency**: Same behavior on Windows and Unix

## Test Coverage

- ✅ `MavenITgh11363PipeSymbolsInJvmConfigTest` - Tests pipe symbols in jvm.config
- ✅ `MavenITgh11485AtSignInJvmConfigTest` - Tests `@` symbols in jvm.config (important for Jenkins workspaces like `workspace/project_PR-350@2`)

## Performance Impact

Unix script is ~200ms slower (320ms → 520ms for `mvn -v`) due to Java compilation, but this is acceptable for the correctness and maintainability benefits.

### Pipe symbols (MNG-11363)
```properties
# .mvn/jvm.config
-Dhttp.nonProxyHosts=de|*.de|my.company.mirror.de
```
✅ Now works correctly: `de|*.de|my.company.mirror.de`

### Quoted pipes
```properties
# .mvn/jvm.config
-Dprop.with.pipes="value|with|pipes"
```
[...]

PR #11489, opened 2025-11-24
Replace xargs -0 with POSIX-compliant alternative (fixes #11485)

### LINKED ISSUE #11485: `xargs -0` not available on AIX, FreeBSD, etc.
### Affected version

4.0.0-rc-5

### Bug description

Commit https://github.com/apache/maven/commit/aeff353bd4c97a2e75de3d86d0259e161f4acf83 introduced `xargs -0`, which is sadly only available on Linux and some other OSes, but not a posix default.

Offending line: https://github.com/apache/maven/blame/master/apache-maven/src/assembly/maven/bin/mvn#L178

For example, AIX will print:

```
./mvnw <options>

xargs: The -0 flag is not valid.
Usage: xargs [-ptx] [-e[EndOfFileString]] [-E EndOfFileString]
        [-i[ReplacementString]] [-I ReplacementString | -L Number |-n Number ]
       [-l[Number]] [-s Size] [Command [Argument ...]]
Usage: java [options] <mainclass> [args...]
           (to execute a class)
   or  java [options] -jar <jarfile> [args...]
           (to execute a jar file)
   or  java [options] -m <module>[/<mainclass>] [args...]
       java [options] --module <module>[/<mainclass>] [args...]
           (to execute the main class in a module)
   or  java [options] <sourcefile> [args]
           (to execute a single source-file program)
```

gnodet:
## Description

This PR fixes #11485 by replacing the non-POSIX-compliant `xargs -0` option with a pure shell solution that works on all POSIX-compliant systems including AIX, FreeBSD, and others.

## Problem

The `xargs -0` option was introduced in commit aeff353bd4c9 to handle special characters (pipes, quotes, spaces) in `.mvn/jvm.config` files. However, this option is not part of the POSIX standard and is not available on AIX, FreeBSD, and other systems, causing the Maven wrapper to fail on these platforms.

## Solution

The fix replaces `xargs -0` with a pure shell solution that:
- Uses the `read` builtin with `-d ''` option to read NUL-terminated strings
- Builds the result string incrementally in a while loop
- Maintains the same functionality of handling special characters correctly
- Works on all POSIX-compliant systems

## Changes

- Modified `concat_lines()` function in `apache-maven/src/assembly/maven/bin/mvn`
- Replaced the pipeline using `xargs -n 1 -0` with a while loop using `read -r -d ''`
- Preserved all existing functionality for handling special characters

## Testing

- All existing tests pass, including the integration test for pipes in MAVEN_OPTS (MavenITgh10937QuotedPipesInMavenOptsTest)
- Manually tested the `concat_lines` function with special characters
- Full build completed successfully with `mvn clean install`

## Compatibility

This change maintains backward compatibility while extending support to more platforms. The behavior for handling special characters in `.mvn/jvm.config` remains unchanged.

---

### DIFF
--- a/apache-maven/src/assembly/maven/bin/mvn
+++ b/apache-maven/src/assembly/maven/bin/mvn
@@ -170,6 +170,7 @@ concat_lines() {
   if [ -f "$1" ]; then
     result=""
     # Read the file line by line
+    # Do not use `xargs -0` as this is not POSIX-compliant
     while IFS= read -r line || [ -n "$line" ]; do
       # Convert CR to LF
       line=$(echo "$line" | tr '\r' '\n')
--- a/its/core-it-suite/src/test/java/org/apache/maven/it/MavenITgh11485AtSignInJvmConfigTest.java
+++ b/its/core-it-suite/src/test/java/org/apache/maven/it/MavenITgh11485AtSignInJvmConfigTest.java
@@ -55,5 +55,32 @@ public class MavenITgh11485AtSignInJvmConfigTest extends AbstractMavenIntegratio
                 props.getProperty("project.properties.propWithAtProp"),
                 "Property value with @ character should be preserved");
     }
+
+    @Test
+    public void testAtSignInCommandLineProperty() throws Exception {
+        File testDir = extractResources("/gh-11485-at-sign");
+
+        Verifier verifier = newVerifier(testDir.getAbsolutePath());
+        verifier.addCliArgument(
+                "-Dexpression.outputFile=" + new File(testDir, "target/pom.properties").getAbsolutePath());
+        verifier.setForkJvm(true); // custom .mvn/jvm.config
+        // Pass a path with @ character via command line (simulating Jenkins workspace)
+        String jenkinsPath = testDir.getAbsolutePath().replace('\\', '/') + "/jenkins.workspace/proj@2";
+        verifier.addCliArgument("-Dcmdline.path=" + jenkinsPath);
+        verifier.addCliArgument("-Dcmdline.value=test@value");
+        verifier.addCliArgument("validate");
+        verifier.execute();
+        verifier.verifyErrorFreeLog();
+
+        Properties props = verifier.loadProperties("target/pom.properties");
+        assertEquals(
+                jenkinsPath,
+                props.getProperty("project.properties.cmdlinePath").replace('\\', '/'),
+                "Command-line path with @ character should be preserved");
+        assertEquals(
+                "test@value",
+                props.getProperty("project.properties.cmdlineValue"),
+                "Command-line value with @ character should be preserved");
+    }
 }
 
--- a/its/core-it-suite/src/test/resources/gh-11485-at-sign/pom.xml
+++ b/its/core-it-suite/src/test/resources/gh-11485-at-sign/pom.xml
@@ -36,6 +36,8 @@ under the License.
   <properties>
     <pathWithAtProp>${path.with.at}</pathWithAtProp>
     <propWithAtProp>${prop.with.at}</propWithAtProp>
+    <cmdlinePath>${cmdline.path}</cmdlinePath>
+    <cmdlineValue>${cmdline.value}</cmdlineValue>
   </properties>
 
   <build>
@@ -55,6 +57,8 @@ under the License.
               <expressions>
                 <expression>project/properties/pathWithAtProp</expression>
                 <expression>project/properties/propWithAtProp</expression>
+                <expression>project/properties/cmdlinePath</expression>
+                <expression>project/properties/cmdlineValue</expression>
               </expressions>
             </configuration>
           </execution>


apache-maven/src/assembly/maven/bin/mvn
```diff
@@ -168,23 +168,39 @@ find_file_argument_basedir() {
...
+            arg="\"$arg\""
+            ;;
+        esac
+
+        if [ -n "$result" ]; then
+          result="$result $arg"
+        else
+          result="$arg"
+        fi
+      done
```
bmarwell, line 202: 1. imho this should be an extra function
2. we  might want to document why we do not use `xargs -0` here and that we agreed to use posix-only arguments (did we agree on that?)

gnodet: Not sure what you mean by _extra function_.  The `concat_lines` function _is_ the code above (but for an `if` which verifies the argument is actually a file...

bmarwell: 1)
I just thought we might want to extract a method/function here, because we get a lot of indentation and the method gets longer and longer.

2) 
Did we agree on "we must be POSIX compliant"? If not, we could leave xargs in here and just concentrate on the sed part.

gnodet: 1. I think 30 lines is still manageable.

2. Yes, unless there's a strong reason, I think keeping POSIX compliant script is better.  `maven-wrapper` has an additional job to validate posix compliance, we may want to do the same: https://github.com/apache/maven-wrapper/blob/master/.github/workflows/shellcheck-posix.yml

bmarwell: Great, I fully agree!

apache-maven/src/assembly/maven/bin/mvn
```diff
@@ -168,23 +168,40 @@ find_file_argument_basedir() {
...
+    while IFS= read -r line || [ -n "$line" ]; do
+      # Convert CR to LF
+      line=$(echo "$line" | tr '\r' '\n')
+      # Remove comments
+      line=$(echo "$line" | sed 's/#.*$//')
+      # Skip empty lines
+      [ -z "$(echo "$line" | tr -d ' \t')" ] && continue
+
+      # Process each argument in the line using eval to handle quotes
+      eval "set -- $line"
```
bmarwell, line 183: This allows command substitution/globbing -> unsafe and can execute backticks/$() from .mvn/jvm.config. Is this wanted? Seems dangerous, we might want to at least document this.

apache-maven/src/assembly/maven/bin/mvn
```diff
@@ -168,23 +168,40 @@ find_file_argument_basedir() {
...
-
-      echo "$arg"
-    done | \
-    tr '\n' ' '
+    result=""
+    # Read the file line by line
+    # Do not use `xargs -0` as this is not POSIX-compliant
+    while IFS= read -r line || [ -n "$line" ]; do
+      # Convert CR to LF
+      line=$(echo "$line" | tr '\r' '\n')
```
bmarwell, line 176: this can produce embedded newlines that confuse splitting if `\r` is in the middle of a line. But I think that is problematic anyway.

apache-maven/src/assembly/maven/bin/mvn
```diff
@@ -168,23 +168,40 @@ find_file_argument_basedir() {
...
+      eval "set -- $line"
+      for arg in "$@"; do
+        # Replace variables
+        arg=$(echo "$arg" | sed \
+          -e "s@\${MAVEN_PROJECTBASEDIR}@$MAVEN_PROJECTBASEDIR@g" \
+          -e "s@\$MAVEN_PROJECTBASEDIR@$MAVEN_PROJECTBASEDIR@g")
+
+        # Quote the argument if it contains spaces or special shell characters
+        case "$arg" in
+          *[\ \|\&\;\<\>\(\)\$\`\\\"\'\~\*\?\[\]\#\~\=]*)
```
bmarwell, line 192: This looks fragile. Does work correctly on backticks and double quotes? If backticks or double quotes are inside, should we not use `arg="'$arg'" instead?

apache-maven/src/assembly/maven/bin/mvn
gnodet: You mean in case of mixed end of lines style, with some \n, some \r and some \r\n ?

apache-maven/src/assembly/maven/bin/mvn
gnodet: We could switch to a solution similar to the one I tried on windows for https://github.com/apache/maven/pull/11365, i.e. using `java` to compile the [`JvmConfigParser`](https://github.com/apache/maven/pull/11365/files#diff-be6f3e934a025e1aba599c08ec0813ce2fd810886bfa1577a84b88f149954ffdR37)

PR #11494, opened 2025-11-26
Drop legacy duplicated code

cstamas:
Right now Maven 4 (same as 3) contains duplicated code to read update status, and due this in-JVM but also process wide synchronization is needed.

For start, drop Maven legacy code and use resolver component to perform reads.

This should be done in Maven 3 as well, as code is pretty much same.

**Requires latest resolver**.


cstamas:
Is merged into #11473

PR #11495, opened 2025-11-26
Repository layout should be used in MavenRepositorySystem

slawekjaranowski:

### DIFF
--- a/maven-core/src/main/java/org/apache/maven/bridge/MavenRepositorySystem.java
+++ b/maven-core/src/main/java/org/apache/maven/bridge/MavenRepositorySystem.java
@@ -36,19 +36,20 @@ import org.apache.maven.artifact.DefaultArtifact;
 import org.apache.maven.artifact.InvalidRepositoryException;
 import org.apache.maven.artifact.handler.ArtifactHandler;
 import org.apache.maven.artifact.handler.manager.ArtifactHandlerManager;
+import org.apache.maven.artifact.metadata.ArtifactMetadata;
 import org.apache.maven.artifact.repository.ArtifactRepository;
 import org.apache.maven.artifact.repository.ArtifactRepositoryPolicy;
 import org.apache.maven.artifact.repository.Authentication;
 import org.apache.maven.artifact.repository.MavenArtifactRepository;
 import org.apache.maven.artifact.repository.layout.ArtifactRepositoryLayout;
 import org.apache.maven.artifact.repository.layout.ArtifactRepositoryLayout2;
-import org.apache.maven.artifact.repository.layout.DefaultRepositoryLayout;
 import org.apache.maven.artifact.resolver.filter.ExclusionArtifactFilter;
 import org.apache.maven.artifact.versioning.InvalidVersionSpecificationException;
 import org.apache.maven.artifact.versioning.VersionRange;
 import org.apache.maven.execution.MavenExecutionRequest;
 import org.apache.maven.model.Dependency;
 import org.apache.maven.model.Plugin;
+import org.apache.maven.model.Repository;
 import org.apache.maven.repository.Proxy;
 import org.apache.maven.repository.RepositorySystem;
 import org.apache.maven.settings.Mirror;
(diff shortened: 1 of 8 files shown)


maven-core/src/main/java/org/apache/maven/bridge/MavenRepositorySystem.java
```diff
@@ -275,9 +288,14 @@ public void injectProxy(RepositorySystemSession session, List<ArtifactRepository
         }
     }
 
-    private ArtifactRepositoryLayout getLayout(String id) {
+    private ArtifactRepositoryLayout getLayout(String repoId, String id) {
         ArtifactRepositoryLayout layout = layouts.get(id);
 
+        if (layout == null) {
+            LOGGER.warn("No layout '{}' found for repository id '{}'", id, repoId);
```
gnodet: Won't this warning be printed each time the repository is used?
Should we inform the user about how to fix the problem ?

Bukama: Yes, please help the user. Many users treat warnings as errors and set their CI to failed or at least unstable. We should not let them alone with this advanced topic.

slawekjaranowski: ok, I will move it to debug level ...

wrong repository layout, can be defined in pom imported from external artifact, users can not do anything with it
and in worst we don't have in this place information which pom/artifact is actually processed

slawekjaranowski: done - use debug for it

maven-core/src/main/java/org/apache/maven/bridge/MavenRepositorySystem.java
```diff
@@ -275,9 +288,14 @@ public void injectProxy(RepositorySystemSession session, List<ArtifactRepository
         }
     }
 
-    private ArtifactRepositoryLayout getLayout(String id) {
+    private ArtifactRepositoryLayout getLayout(String repoId, String id) {
```
Bukama: Shouldn‘t be the second parameter be called layoutID?

maven-core/src/main/java/org/apache/maven/bridge/MavenRepositorySystem.java
```diff
@@ -308,13 +326,37 @@ public static org.apache.maven.model.RepositoryPolicy fromSettingsRepositoryPoli
         return modelRepositoryPolicy;
     }
 
+    /**
+     * @deprecated use a service metod {@link #buildArtifactRepositoryFromRepo(org.apache.maven.settings.Repository) instead
```
Bukama: method

maven-core/src/main/java/org/apache/maven/bridge/MavenRepositorySystem.java
```diff
@@ -308,13 +326,37 @@ public static org.apache.maven.model.RepositoryPolicy fromSettingsRepositoryPoli
...
+    /**
+     * @since 3.9.12
+     */
+    public ArtifactRepository buildArtifactRepositoryFromRepo(org.apache.maven.settings.Repository repo)
+            throws InvalidRepositoryException {
+        return buildArtifactRepositoryFromRepo(fromSettingsRepository(repo));
+    }
+
+    /**
+     * @deprecated use a service metod {@link #buildArtifactRepositoryFromRepo(Repository)}  instead
```
Bukama: method

maven-core/src/main/java/org/apache/maven/bridge/MavenRepositorySystem.java
```diff
@@ -816,4 +858,43 @@ else if (layout.equals(repoLayout)) {
...
+
+        private final String id;
+
+        private final ArtifactRepositoryLayout fallback;
+
+        UnknownRepositoryLayout(String id, ArtifactRepositoryLayout fallback) {
+            this.id = id;
+            this.fallback = fallback;
+        }
+
```
Bukama, line 878: Javadoc missing for methods

Bukama (approved):
Few typos only

maven-core/src/main/java/org/apache/maven/bridge/MavenRepositorySystem.java
slawekjaranowski: It is private class, no special to write about it

PR #11501, opened 2025-11-28
Drop legacy duplicated code

slawekjaranowski:
Right now Maven 4 (same as 3) contains duplicated code to read update status, and due this in-JVM but also process wide synchronization is needed.

For start, drop Maven legacy code and use resolver component to perform reads.

### DIFF
--- a/maven-compat/src/main/java/org/apache/maven/repository/legacy/DefaultUpdateCheckManager.java
+++ b/maven-compat/src/main/java/org/apache/maven/repository/legacy/DefaultUpdateCheckManager.java
@@ -18,14 +18,13 @@
  */
 package org.apache.maven.repository.legacy;
 
+import javax.inject.Inject;
+import javax.inject.Named;
+import javax.inject.Singleton;
+
 import java.io.File;
-import java.io.FileInputStream;
-import java.io.IOException;
-import java.io.RandomAccessFile;
-import java.nio.channels.Channels;
-import java.nio.channels.FileChannel;
-import java.nio.channels.FileLock;
 import java.util.Date;
+import java.util.HashMap;
 import java.util.Properties;
 
 import org.apache.maven.artifact.Artifact;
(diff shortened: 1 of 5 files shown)


slawekjaranowski:
cherry pick from: 
 - #11494

maven-compat/src/main/java/org/apache/maven/repository/legacy/DefaultUpdateCheckManager.java
```diff
@@ -37,19 +36,29 @@
 import org.codehaus.plexus.component.annotations.Component;
 import org.codehaus.plexus.logging.AbstractLogEnabled;
 import org.codehaus.plexus.logging.Logger;
+import org.eclipse.aether.internal.impl.TrackingFileManager;
 
 /**
  * DefaultUpdateCheckManager
  */
-@Component(role = UpdateCheckManager.class)
```
cstamas, line 44: Unsure would I migrate this from plx to sisu; moreover, undoing this change will make other UT changes unneeded (index scanning)

maven-compat/src/main/java/org/apache/maven/repository/legacy/DefaultUpdateCheckManager.java
```diff
@@ -37,19 +36,29 @@
...
+@Named
+@Singleton
 public class DefaultUpdateCheckManager extends AbstractLogEnabled implements UpdateCheckManager {
+    private final TrackingFileManager trackingFileManager;
 
     private static final String ERROR_KEY_SUFFIX = ".error";
 
-    public DefaultUpdateCheckManager() {}
+    @Inject
```
cstamas, line 51: Just make/leave this plain old plx component with field injection instead, and as said above, it will leave things less intrusive re UTs

maven-compat/src/main/java/org/apache/maven/repository/legacy/DefaultUpdateCheckManager.java
slawekjaranowski: I afraid that I need add scaning to UT as implementation of `TrackingFileManager` is sisiu

cstamas:
Superseded by #11536

Issue #8081, opened 2021-11-22
[MNG-7342] Add a --very-verbose / XX cli option


jira-importer:
**[Guillaume Nodet](https://issues.apache.org/jira/secure/ViewProfile.jspa?name=gnt)** opened **[MNG-7342](https://issues.apache.org/jira/browse/MNG-7342?redirect=false)** and commented

If a dependency is missing, a `NoClassDefFoundError` will be thrown inside Sisu, but Sisu logs at `TRACE` level and not `DEBUG` so that nothing is displayed.

The `-X` option should use `TRACE` instead of `{}DEBUG{`}.

---

**Remote Links:**
- [GitHub Pull Request #619
](https://github.com/apache/maven/pull/619)

jira-importer:
**[Michael Osipov](https://issues.apache.org/jira/secure/ViewProfile.jspa?name=michael-o)** commented

As far as I know Sisu logging can be enable with a system property it will write to standard streams. It does not use any loggger. Turning verbose from debug to trace will result in a very huge amount of output by some components. I am against this.

jira-importer:
**[Guillaume Nodet](https://issues.apache.org/jira/secure/ViewProfile.jspa?name=gnt)** commented

Then I would suggest overwriting Sisu logger with a maven custom one.

The problem is not really the fact that sisu logger can be configured using system properties but rather the fact that such exceptions are only displayed at TRACE level, so they are completely hidden, even when using `{}-X{`}.  So I'd suggest logging those exceptions at DEBUG rather than TRACE level.

The problem is that even when using `{}-X{`}, you have no way to know what happens at all.  I had to debug and put a breakpoint inside Sisu to actually see that I could use a system property to force Sisu to log those statements.

An alternative would be to add a `-XX` option to enable TRACE logging.

jira-importer:
**[Michael Osipov](https://issues.apache.org/jira/secure/ViewProfile.jspa?name=michael-o)** commented

I guess you are talking about: https://github.com/eclipse/sisu.inject/blob/a41ab393a0c0e6d8eadf0547763fe80168825b49/org.eclipse.sisu.inject/src/org/eclipse/sisu/inject/Logs.java
Yes, it is unfortuantely at `TRACE` level. I assume this is on purpose because you don't debug Sisu day by day.

I like the idea of `-XX`, `--very-verbose`. Likely to be used rarely, but still.

jira-importer:
[...]

PR #11505, opened 2025-11-30
Add module-aware resource handling for modular sources

ascheman:
## Summary
- Automatically inject module-aware resource roots for projects using modular `<sources>` layout
- Resources are picked up from `src/<module>/main/resources` and `src/<module>/test/resources`
- Issue warnings when explicit legacy `<resources>` configuration is ignored in favor of modular defaults

## Problem
Projects using Maven 4.x modular sources (`<source><module>org.foo.bar</module></source>`) had to manually configure `maven-resources-plugin` for each module. Resources from the modular layout paths were not automatically discovered.

## Solution
When a project defines at least one module in `<sources>`:
1. Modular resource roots are automatically injected for each module
2. Legacy `<resources>` from Super POM defaults are silently replaced
3. Explicit legacy `<resources>` configuration triggers a `ModelProblem` warning

## Test plan
- [x] `testModularSourcesInjectResourceRoots` - verifies resource injection for multi-module projects
- [x] `testModularSourcesWithExplicitResourcesIssuesWarning` - verifies warnings are issued when legacy resources are ignored

### DIFF
--- a/impl/maven-core/src/main/java/org/apache/maven/project/ResourceHandlingContext.java
+++ b/impl/maven-core/src/main/java/org/apache/maven/project/ResourceHandlingContext.java
@@ -92,7 +92,7 @@ class ResourceHandlingContext {
                             scopeId,
                             sourcesConfig);
                 } else {
-                    LOGGER.trace(
+                    LOGGER.debug(
                             "{} resources configured via <sources> element, ignoring legacy {} element.",
                             scopeName,
                             legacyElement);
@@ -118,7 +118,7 @@ class ResourceHandlingContext {
                     project.addSourceRoot(createModularResourceRoot(module, scope));
                 }
                 if (!modules.isEmpty()) {
-                    LOGGER.trace(
+                    LOGGER.debug(
                             "Injected {} module-aware {} resource root(s) for modules: {}.",
                             modules.size(),
                             scopeId,
@@ -136,14 +136,14 @@ class ResourceHandlingContext {
                             scopeId,
                             sourcesConfig);
                 } else {
-                    LOGGER.trace(
+                    LOGGER.debug(
                             "{} resources configured via <sources> element, ignoring legacy {} element.",
                             scopeName,
                             legacyElement);
                 }
             } else {
                 // Use legacy resources element
-                LOGGER.trace(
+                LOGGER.debug(
                         "Using explicit or default {} resources ({} resources configured).", scopeId, resources.size());
                 for (Resource resource : resources) {
                     project.addSourceRoot(new DefaultSourceRoot(baseDir, scope, resource));


impl/maven-core/src/main/java/org/apache/maven/project/DefaultProjectBuilder.java
```diff
@@ -700,11 +709,149 @@ private void initProject(MavenProject project, ModelBuilderResult result) {
...
+                        logger.debug("Using explicit or default resources ({} resources configured)", resources.size());
+                        for (Resource resource : resources) {
+                            project.addSourceRoot(new DefaultSourceRoot(baseDir, ProjectScope.MAIN, resource));
+                        }
+                    }
                 }
-                for (Resource resource : project.getBuild().getDelegate().getTestResources()) {
-                    project.addSourceRoot(new DefaultSourceRoot(baseDir, ProjectScope.TEST, resource));
+
+                /*
```
desruisseaux: This block of code for test resources seems very similar to above code for main resources, with only minor differences. Can we share the code in a single method?

impl/maven-core/src/main/java/org/apache/maven/project/DefaultProjectBuilder.java
```diff
@@ -1099,6 +1246,86 @@ public Set<Entry<K, V>> entrySet() {
...
     }
 
+    /**
+     * Extracts unique module names from the given list of source elements.
+     * A project is considered modular if it has at least one module name.
+     *
+     * @param sources list of source elements from the build
+     * @return set of non-blank module names
+     */
+    private Set<String> extractModules(List<org.apache.maven.api.model.Source> sources) {
```
desruisseaux: Method could be static.

impl/maven-core/src/main/java/org/apache/maven/project/DefaultProjectBuilder.java
```diff
@@ -700,11 +709,149 @@ private void initProject(MavenProject project, ModelBuilderResult result) {
...
+                /*
+                 * Handle main resources - modular project has highest priority:
+                 * 1. Modular project: use resources from <sources> if present, otherwise inject defaults
+                 * 2. Classic project: use resources from <sources> if present, otherwise use legacy <resources>
+                 */
+                List<Resource> resources = project.getBuild().getDelegate().getResources();
+                if (isModularProject) {
+                    if (hasMainResources) {
+                        // Modular project with resources configured via <sources> - already added above
+                        if (!resources.isEmpty()
```
desruisseaux: `resources.isEmpty()` is already tested by `hasOnlySuperPomDefaults`, so it is not necessary to repeat the test here.

impl/maven-core/src/main/java/org/apache/maven/project/DefaultProjectBuilder.java
```diff
@@ -700,11 +709,149 @@ private void initProject(MavenProject project, ModelBuilderResult result) {
...
+                                && !hasOnlySuperPomDefaults(resources, baseDir, ProjectScope.MAIN.id())) {
+                            logger.warn("Legacy <resources> element is ignored because main resources are "
+                                    + "configured via <source><lang>resources</lang></source> in <sources>");
+                        }
+                        logger.debug(
+                                "Main resources configured via <sources> element, ignoring legacy <resources> element");
+                    } else {
+                        // Modular project without resources in <sources> - inject module-aware defaults
+                        if (!resources.isEmpty()
```
desruisseaux: Unnecessary check for `resources.isEmpty()` (see above comment).

impl/maven-core/src/main/java/org/apache/maven/project/DefaultProjectBuilder.java
```diff
@@ -700,11 +709,149 @@ private void initProject(MavenProject project, ModelBuilderResult result) {
...
+                            logger.debug("  - Adding main resource root: {} (module: {})", resourcePath, module);
+                            project.addSourceRoot(
+                                    createModularResourceRoot(baseDir, module, ProjectScope.MAIN, outputDirectory));
+                        }
+                    }
+                } else {
+                    // Classic (non-modular) project
+                    if (hasMainResources) {
+                        // Resources configured via <sources> - already added above
+                        if (!resources.isEmpty()
```
desruisseaux: Unnecessary check for `resources.isEmpty()` (see above comment).

impl/maven-core/src/main/java/org/apache/maven/project/DefaultProjectBuilder.java
ascheman: Done in bef10a125a158fe6e586d1707c59c8e68233bfa7. Extracted handleResourceConfiguration method with a ResourceHandlingContext record to group shared parameters.

impl/maven-core/src/main/java/org/apache/maven/project/DefaultProjectBuilder.java
ascheman: Done in 0c2451e7a30f8f427347b6c8b10bca3f551db02d. Made extractModules method static.

impl/maven-core/src/main/java/org/apache/maven/project/DefaultProjectBuilder.java
ascheman: Done in 0c2451e7a30f8f427347b6c8b10bca3f551db02d. Removed the redundant isEmpty() checks and renamed the method to hasExplicitLegacyResources with inverted logic for clearer semantics (returns false when empty, true when explicit legacy resources exist).

impl/maven-core/src/main/java/org/apache/maven/project/DefaultProjectBuilder.java
ascheman: Done in 0c2451e7a30f8f427347b6c8b10bca3f551db02d. Removed the redundant isEmpty() checks and renamed the method to hasExplicitLegacyResources with inverted logic for clearer semantics (returns false when empty, true when explicit legacy resources exist).

impl/maven-core/src/main/java/org/apache/maven/project/DefaultProjectBuilder.java
ascheman: Done in 0c2451e7a30f8f427347b6c8b10bca3f551db02d. Removed the redundant isEmpty() checks and renamed the method to hasExplicitLegacyResources with inverted logic for clearer semantics (returns false when empty, true when explicit legacy resources exist).

impl/maven-core/src/main/java/org/apache/maven/project/DefaultProjectBuilder.java
```diff
@@ -700,12 +709,24 @@ private void initProject(MavenProject project, ModelBuilderResult result) {
...
+
+                logger.debug(
+                        "Module detection for project {}: found {} module(s) {} - modular project: {}",
+                        project.getId(),
+                        modules.size(),
+                        modules,
+                        isModularProject);
+
+                // Handle main and test resources using shared method
+                ResourceHandlingContext ctx =
```
elharo: ctx --> context or resourceHandlingContext;

in general don't abbreviate

impl/maven-core/src/test/java/org/apache/maven/project/ProjectBuilderTest.java
```diff
@@ -371,4 +376,98 @@ void testLocationTrackingResolution() throws Exception {
...
+        configuration.setRepositorySession(mavenSession.getRepositorySession());
+
+        ProjectBuildingResult result = getContainer()
+                .lookup(org.apache.maven.project.ProjectBuilder.class)
+                .build(pom, configuration);
+
+        MavenProject project = result.getProject();
+
+        // Verify warnings are issued for ignored legacy resources
+        List<org.apache.maven.model.building.ModelProblem> warnings = result.getProblems().stream()
```
elharo: can this just be imported in short form

impl/maven-core/src/main/java/org/apache/maven/project/DefaultProjectBuilder.java
ascheman: Done in f747693ea0. Renamed to context.

impl/maven-core/src/test/java/org/apache/maven/project/ProjectBuilderTest.java
ascheman: Done in f747693ea0. Using short form ModelProblem (already imported).

impl/maven-core/src/main/java/org/apache/maven/project/DefaultProjectBuilder.java
```diff
@@ -1099,6 +1123,190 @@ public Set<Entry<K, V>> entrySet() {
...
+            Path baseDir,
+            Set<String> modules,
+            boolean modularProject,
+            ModelBuilderResult result) {}
+
+    /**
+     * Handles resource configuration for a given scope (main or test).
+     * This method applies the resource priority rules:
+     * <ol>
+     *   <li>Modular project: use resources from &lt;sources&gt; if present, otherwise inject defaults</li>
```
elharo: This is a private method so the javadoc won't appear in HTML. So no need to escape the < and >. In Java source this just obscures things. 

If the warning bothers you make this a non javadoc comment

impl/maven-core/src/main/java/org/apache/maven/project/DefaultProjectBuilder.java
```diff
@@ -1099,6 +1123,190 @@ public Set<Entry<K, V>> entrySet() {
...
+                .map(org.apache.maven.api.model.Source::getModule)
+                .filter(Objects::nonNull)
+                .map(String::trim)
+                .filter(s -> !s.isBlank())
+                .collect(Collectors.toSet());
+    }
+
+    /**
+     * Creates a DefaultSourceRoot for module-aware resource directories.
+     * Generates paths following the pattern: src/&lt;module&gt;/&lt;scope&gt;/resources
```
elharo: don't escape < and > on private methods so they're easier to read in source code

impl/maven-core/src/main/java/org/apache/maven/project/DefaultProjectBuilder.java
desruisseaux: > This is a private method so the javadoc won't appear in HTML.

Well, depending which `-Xdoclint` options are enabled (if any), removing the escapes could cause a compilation warning of failure. Rather than removing the escapes, I suggest to use `{@code}`, which automatically does the escapes. Example:

```
<li>Modular project: use resources from {@code <sources>} if present (etc...)
```

impl/maven-core/src/main/java/org/apache/maven/project/DefaultProjectBuilder.java
desruisseaux: Same as previous, I rather suggest to use `{@code src/<module>/<scope>/resources}`.

impl/maven-core/src/main/java/org/apache/maven/project/DefaultProjectBuilder.java
ascheman: > > This is a private method so the javadoc won't appear in HTML.
> 
> Well, depending which `-Xdoclint` options are enabled (if any), removing the escapes could cause a compilation warning of failure. Rather than removing the escapes, I suggest to use `{@code}`, which automatically does the escapes. Example:
> 
> ```
> <li>Modular project: use resources from {@code <sources>} if present (etc...)
> ```

Changed this in 7bbd727c2f9536fba794346e43c8b872dbee2518 by now (will try to follow this approach in the future).

Perhaps I have to squash some commits (or the overall change) before merging. Will leave it in place for some days to enable further review.

impl/maven-core/src/main/java/org/apache/maven/project/DefaultProjectBuilder.java
ascheman: Cf. https://github.com/apache/maven/pull/11505#discussion_r2616891087

impl/maven-core/src/main/java/org/apache/maven/project/DefaultProjectBuilder.java
```diff
@@ -680,6 +682,13 @@ private void initProject(MavenProject project, ModelBuilderResult result) {
                         } else {
                             hasTest |= ProjectScope.TEST.equals(scope);
                         }
+                    } else if (Language.RESOURCES.equals(language)) {
+                        ProjectScope scope = src.scope();
+                        if (ProjectScope.MAIN.equals(scope)) {
```
gnodet, line 686: The code looks weird.  We should use
```
if (ProjectScope.MAIN.equals(scope)) {
    hasMainResources = true;
} else if (ProjectScope.TEST.equals(scope)) {
    hasTestResources = true;
}
```
or
```
hasMainResources |= ProjectScope.MAINequals(scope);
hasTestResources |= ProjectScope.TEST.equals(scope);
```
but this mix is not very clean.

gnodet:
> I think we should reject any project that mixes `<resource>` and `<source>` elements, I don't see that as a valid use case.

I think this should even be done in `org.apache.maven.impl.model.DefaultModelValidator`.

impl/maven-core/src/main/java/org/apache/maven/project/DefaultProjectBuilder.java
ascheman: Resolved in 3b2af3a73061bc1af49ebaff31fde1e39d3ba8f7

impl/maven-core/src/main/java/org/apache/maven/project/DefaultProjectBuilder.java
```diff
@@ -1099,6 +1123,193 @@ public Set<Entry<K, V>> entrySet() {
...
+        String sourcesConfig = scope == ProjectScope.MAIN
+                ? "<source><lang>resources</lang></source>"
+                : "<source><lang>resources</lang><scope>test</scope></source>";
+
+        if (context.modularProject()) {
+            if (hasResourcesInSources) {
+                // Modular project with resources configured via <sources> - already added above
+                if (hasExplicitLegacyResources(resources, context.baseDir(), scopeId)) {
+                    logger.warn(
+                            "Legacy {} element is ignored because {} resources are configured via {} in <sources>",
```
desruisseaux: It is a small detail, but I would suggest to put a final `.` at the end of every sentences in this pull request (not just this one). This is because Maven is very verbose, often with messages spanning many lines, to the point that it is sometime difficult to spot where a sentence ends and when a new sentence starts.

impl/maven-core/src/main/java/org/apache/maven/project/DefaultProjectBuilder.java
```diff
@@ -700,12 +709,27 @@ private void initProject(MavenProject project, ModelBuilderResult result) {
...
-                    project.addSourceRoot(new DefaultSourceRoot(baseDir, ProjectScope.MAIN, resource));
-                }
-                for (Resource resource : project.getBuild().getDelegate().getTestResources()) {
-                    project.addSourceRoot(new DefaultSourceRoot(baseDir, ProjectScope.TEST, resource));
-                }
+                // Extract modules from sources to detect modular projects
+                Set<String> modules = extractModules(sources);
+                boolean isModularProject = !modules.isEmpty();
+
+                logger.debug(
```
desruisseaux: Do we really need this log, even in debug mode? Maven output is so verbose that searching an information in Maven debug output is like trying to drink from a fire hose.

impl/maven-core/src/main/java/org/apache/maven/project/DefaultProjectBuilder.java
```diff
@@ -1099,6 +1123,193 @@ public Set<Entry<K, V>> entrySet() {
...
+        // Check if any resource differs from Super POM defaults
+        for (Resource resource : resources) {
+            String resourceDir = resource.getDirectory();
+            if (resourceDir != null && !resourceDir.equals(defaultPath) && !resourceDir.equals(defaultFilteredPath)) {
+                // Found an explicit legacy resource
+                return true;
+            }
+        }
+
+        logger.debug("Only Super POM default resources found for scope: {}", scope);
```
desruisseaux: Maybe omit this log. The `handleResourceConfiguration` caller already does a higher level analysis of the situation for producing a more informative log message.

desruisseaux: Note: if the log message is omitted, the method can become static.

impl/maven-core/src/main/java/org/apache/maven/project/DefaultProjectBuilder.java
```diff
@@ -1099,6 +1123,193 @@ public Set<Entry<K, V>> entrySet() {
...
+    /**
+     * Creates a DefaultSourceRoot for module-aware resource directories.
+     * Generates paths following the pattern: {@code src/<module>/<scope>/resources}
+     *
+     * @param baseDir base directory of the project
+     * @param module module name
+     * @param scope project scope (main or test)
+     * @return configured DefaultSourceRoot for the module's resources
+     */
+    private DefaultSourceRoot createModularResourceRoot(Path baseDir, String module, ProjectScope scope) {
```
desruisseaux: Method can be static.

impl/maven-core/src/main/java/org/apache/maven/project/DefaultProjectBuilder.java
ascheman: To be honest, the log message helped me a lot to understand what really happened during resource resolution. Now that I have a good understanding, I could drop it. Is there a common policy about what should be logged on which level in Maven?

impl/maven-core/src/main/java/org/apache/maven/project/DefaultProjectBuilder.java
```diff
@@ -1099,6 +1123,193 @@ public Set<Entry<K, V>> entrySet() {
...
+     *   <li>Modular project: use resources from {@code <sources>} if present, otherwise inject defaults</li>
+     *   <li>Classic project: use resources from {@code <sources>} if present, otherwise use legacy resources</li>
+     * </ol>
+     *
+     * @param context the resource handling context containing project info
+     * @param resources the legacy resource list (from {@code <resources>} or {@code <testResources>})
+     * @param hasResourcesInSources whether resources are configured via {@code <sources>}
+     * @param scope the project scope (MAIN or TEST)
+     */
+    private void handleResourceConfiguration(
```
desruisseaux: If `hasExplicitLegacyResources` and `createModularResourceRoot` methods become static, then the only field that `handleResourceConfiguration` uses is `logger`. We could:

* Add the logger field in `ResourceHandlingContext`.
* Move this method into `ResourceHandlingContext`. It would simplify a little bit the code (remove 15 occurrences of `context`, can access protected fields such as `baseDir` directly).
* Rename `ResourceHandlingContext` as `ResourceHandling` and promote as a package-private (non inner) class. It would allow to isolate the resource handling code in a separated source Java file.
* Move the static `hasExplicitLegacyResources` and `createModularResourceRoot` methods in `ResourceHandling` too. It would complete the isolation of resource handling code.

ascheman: Great idea, let me give that a try. Besides, I was wondering if I could/should set the log level to `trace`, which is supported by the underlying logger. You cannot simply switch it on by a CLI flag like `-X` but by system properties as far as I see.

impl/maven-core/src/main/java/org/apache/maven/project/DefaultProjectBuilder.java
desruisseaux: I'm not aware of a common policy. It is a difficult question actually, because debug messages can be sometime useful, and sometime be an impediment when too numerous. I tend to put a lot of debug during development, and then skim them (it does not mean removing all of them) when the code become more stable. A rule of thumb could be "was it useful only for debugging this particular class, or can it be of interest for the codes which will use this class?"

For example, the information _"Only Super POM default resources found"_ is, I think, relevant mostly for the `handleResourceConfiguration` method. For users on top of that, the messages produced by `handleResourceConfiguration` are more informative. So while the _"Only Super POM default resources found"_ message was useful during the debugging of `handleResourceConfiguration`, maybe after that this method has been debugged, hiding this low-level details is nice for users of the `DefaultProjectBuilder` class.

This is only a personal thinking, not a Maven policy, and myself I tend to change this kind of opinion depending on the context.

impl/maven-core/src/main/java/org/apache/maven/project/DefaultProjectBuilder.java
ascheman: At least we wouldn't be the first users of trace levels (a quick search over the overall Maven code based revealed just one usage, and its already in the core: [DefaultRequestCache.java#L414](https://github.com/apache/maven/blob/d30ce5d6def1da2c2323e497e42a8e7d5cbb75f1/impl/maven-impl/src/main/java/org/apache/maven/impl/cache/DefaultRequestCache.java#L414)). And there has been some discussion among @gnodet and @michael-o on [MNG-7342](https://issues.apache.org/jira/browse/MNG-7342?redirect=false) (migrated to #8081).

impl/maven-core/src/main/java/org/apache/maven/project/DefaultProjectBuilder.java
```diff
@@ -1099,6 +1123,193 @@ public Set<Entry<K, V>> entrySet() {
...
+            if (hasResourcesInSources) {
+                // Modular project with resources configured via <sources> - already added above
+                if (hasExplicitLegacyResources(resources, context.baseDir(), scopeId)) {
+                    logger.warn(
+                            "Legacy {} element is ignored because {} resources are configured via {} in <sources>",
+                            legacyElement,
+                            scopeId,
+                            sourcesConfig);
+                }
+                logger.debug(
```
desruisseaux: Could be in a `else` block since it is redundant with above message.

impl/maven-core/src/main/java/org/apache/maven/project/DefaultProjectBuilder.java
```diff
@@ -1099,6 +1123,193 @@ public Set<Entry<K, V>> entrySet() {
...
+            if (hasResourcesInSources) {
+                // Resources configured via <sources> - already added above
+                if (hasExplicitLegacyResources(resources, context.baseDir(), scopeId)) {
+                    logger.warn(
+                            "Legacy {} element is ignored because {} resources are configured via {} in <sources>",
+                            legacyElement,
+                            scopeId,
+                            sourcesConfig);
+                }
+                logger.debug(
```
desruisseaux: Could be in a `else` block since it is redundant with above message.

impl/maven-core/src/main/java/org/apache/maven/project/DefaultProjectBuilder.java
```diff
@@ -1099,6 +1123,193 @@ public Set<Entry<K, V>> entrySet() {
...
+                            .reportProblem(new org.apache.maven.impl.model.DefaultModelProblem(
+                                    message,
+                                    Severity.WARNING,
+                                    Version.V41,
+                                    context.project().getModel().getDelegate(),
+                                    -1,
+                                    -1,
+                                    null));
+                }
+                logger.debug(
```
desruisseaux: Maybe prepare a `StringBuilder` here, append the _"Adding {} resource root"_ messages of the loop into that buffer, then log a single message when we are done? I think that `logger.debug` should not be used as a `println`, but that we should rather send atomic messages when possible.

impl/maven-core/src/main/java/org/apache/maven/project/DefaultProjectBuilder.java
```diff
@@ -1099,6 +1123,193 @@ public Set<Entry<K, V>> entrySet() {
...
+                        "Injecting module-aware {} resource roots for {} modules",
+                        scopeId,
+                        context.modules().size());
+                for (String module : context.modules()) {
+                    Path resourcePath = context.baseDir()
+                            .resolve("src")
+                            .resolve(module)
+                            .resolve(scopeId)
+                            .resolve("resources");
+                    logger.debug("  - Adding {} resource root: {} (module: {})", scopeId, resourcePath, module);
```
desruisseaux: I suggest to log and add source root only if the directory exist. Non-existent directories can be ignored silently (I don't think that we should create a log message for them).

impl/maven-core/src/main/java/org/apache/maven/project/DefaultProjectBuilder.java
ascheman: Done in d2ce972700. Added periods to all log and warning messages.

impl/maven-core/src/main/java/org/apache/maven/project/DefaultProjectBuilder.java
ascheman: Changed to trace level in d2ce972700. This keeps the logs for detailed debugging while removing them from the `-X` output. Trace level requires explicit configuration via system properties.

impl/maven-core/src/main/java/org/apache/maven/project/DefaultProjectBuilder.java
ascheman: Changed to trace level in d2ce972700. The log in `hasExplicitLegacyResources` was removed entirely as suggested. The remaining trace logs are useful for understanding resource resolution but won't appear in normal debug output.

impl/maven-core/src/main/java/org/apache/maven/project/DefaultProjectBuilder.java
ascheman: Done in d2ce972700. Both `hasExplicitLegacyResources` and `createModularResourceRoot` are now instance methods in the extracted `ResourceHandlingContext` class (see dcc621c014).

impl/maven-core/src/main/java/org/apache/maven/project/DefaultProjectBuilder.java
ascheman: Done in dcc621c014. Extracted to `ResourceHandlingContext` as a package-private class with instance methods. The logger, project, baseDir, modules, modularProject, and result are now fields, and `handleResourceConfiguration` takes only scope and hasResourcesInSources as parameters.

impl/maven-core/src/main/java/org/apache/maven/project/DefaultProjectBuilder.java
ascheman: Done in d2ce972700. Trace logs are now in else blocks to avoid redundancy with the warning messages.

impl/maven-core/src/main/java/org/apache/maven/project/DefaultProjectBuilder.java
ascheman: Done in d2ce972700. Trace logs are now in else blocks to avoid redundancy with the warning messages.

impl/maven-core/src/main/java/org/apache/maven/project/DefaultProjectBuilder.java
ascheman: Done in d2ce972700. The loop logging is now consolidated into a single trace message after the loop: \`LOGGER.trace("Injected {} module-aware {} resource root(s) for modules: {}.", modules.size(), scopeId, modules);\` This provides a summary rather than individual messages per iteration.

impl/maven-core/src/main/java/org/apache/maven/project/DefaultProjectBuilder.java
ascheman: After investigation, this was intentionally [kept as-is in the original code](https://github.com/apache/maven/blob/master/impl/maven-core/src/main/java/org/apache/maven/project/DefaultProjectBuilder.java#L674). Source roots are registered regardless of directory existence because filtering could change behavioral semantics (which is outside the scope of this change).

Note: In the long run, we should not only add proper handling of resources but elaborate on a more general refactoring/handling of modular sources (cf. [Phase 2 implementation](https://cwiki.apache.org/confluence/pages/viewpage.action?pageId=399279347#BuildSourcesValidationDiscussionNotes-Phase2:ComprehensiveRefactoring)).

impl/maven-core/src/main/java/org/apache/maven/project/ResourceHandlingContext.java
```diff
@@ -0,0 +1,213 @@
...
+            if (hasResourcesInSources) {
+                // Modular project with resources configured via <sources> - already added above
+                if (hasExplicitLegacyResources(resources, scopeId)) {
+                    LOGGER.warn(
+                            "Legacy {} element is ignored because {} resources are configured via {} in <sources>.",
+                            legacyElement,
+                            scopeId,
+                            sourcesConfig);
+                } else {
+                    LOGGER.trace(
```
desruisseaux: It is nice to reduce the amount of logging, but I think that in the particular case of `handleResourceConfiguration`, we can keep the debug level as it was before. The rational is that other loggings (now removed or reduced to trace level) were a little bit trivial. But the loggings in this `handleResourceConfiguration` method are the result of a more elaborated analysis.

Would you mind switching back to debug level all the logging at the trace level in this `handleResourceConfiguration` method, but only in this method?

ascheman: Done in dc90ada936. Changed the trace logs back to debug level in handleResourceConfiguration as suggested.

Note: The branch was also rebased to the latest master, so all commit hashes have changed.

desruisseaux (approved):

ascheman:
> Yes I thought about that. But as you mentioned on the slack channel, the work done in this pull request could be generalized. 

Yes, I have a [Phase 2](https://cwiki.apache.org/confluence/display/MAVEN/Build+Sources+Validation+-+Discussion+Notes#BuildSourcesValidationDiscussionNotes-Phase2:ComprehensiveRefactoring) implementation in my mind which would be more general. 

Nevertheless, the feature would already be helpful (and I'd be eager to get it in the next Release Candidate for Maven 4.0).

> I thought that what this PR does for resources, the same thing could be done for tests. I'm thinking to refactor the `ResourceHandler` class in something that does not handle resources in a special way, but instead does the same thing in a uniform way for all combinations of `<lang>` and `<scope>`. If agreed, I would propose to merge on 4.0.x the generalized version.

Sounds good, perhaps we can join forces in the next weeks to get a combined implementation?

desruisseaux:
Maybe you can go ahead with phase 2 when you have time. Looking on the wiki page, I suggest the following changes:

* Instead of `record ModuleConfig(boolean hasMain, boolean hasTest, boolean hasMainResources, boolean hasTestResources)`, handle any combination of `<lang>` and `<scope>` with something like `Map<Something, Boolean>`. The mapping to legacy `<resources>`, `<testResources>`, etc. would still special cases, but the rest (e.g., generate automatically a `Source` object) can be handled by a universal code.
* Remove rule R3, i.e., accept duplicated source for scope='X', lang='Y', module='Z'. The compiler plugin 4 already accepts that, the need for multi-source directories in Maven 3 was strong enough to justify an external plugin, and Maven 3 already accepts that for resources if I remember right.

ascheman:
I have raised a new issue (#11612) for further discussion.

PR #11509, opened 2025-12-01
Add tests

TheRealHaui:
Added new Unit Tests

### DIFF
--- a/impl/maven-core/src/test/java/org/apache/maven/plugin/CacheUtilsTest.java
+++ b/impl/maven-core/src/test/java/org/apache/maven/plugin/CacheUtilsTest.java
@@ -1,0 +1,130 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.maven.plugin;
+
+import org.apache.maven.model.Dependency;
+import org.apache.maven.model.Plugin;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+
+import static org.junit.jupiter.api.Assertions.*;
+
+class CacheUtilsTest {
+
+    private Plugin pluginOne;
+    private Plugin pluginTwo;
+
+    @BeforeEach
+    void doBeforeEachTest() {
+        pluginOne = new Plugin();
+        pluginTwo = new Plugin();
+    }
+
+    @Test
+    void testPluginEqualsEmptyPlugins() {
+        assertTrue(CacheUtils.pluginEquals(pluginOne, pluginOne));
+        assertTrue(CacheUtils.pluginEquals(pluginOne, pluginTwo));
+    }
+
+    @Test
+    void testPluginEqualsSetArtifactId() {
+        pluginOne.setArtifactId("myArtifact");
+
+        assertTrue(CacheUtils.pluginEquals(pluginOne, pluginOne));
+        assertFalse(CacheUtils.pluginEquals(pluginOne, pluginTwo));
+
+        pluginTwo.setArtifactId("myArtifact");
+
+        assertTrue(CacheUtils.pluginEquals(pluginOne, pluginOne));
+        assertTrue(CacheUtils.pluginEquals(pluginOne, pluginTwo));
+    }
+
+    @Test
+    void testPluginEqualsSetGroupId() {
+        pluginOne.setGroupId("myGroupId");
+
+        assertTrue(CacheUtils.pluginEquals(pluginOne, pluginOne));
+        assertFalse(CacheUtils.pluginEquals(pluginOne, pluginTwo));
+
+        pluginTwo.setGroupId("myGroupId");
+
+        assertTrue(CacheUtils.pluginEquals(pluginOne, pluginOne));
+        assertTrue(CacheUtils.pluginEquals(pluginOne, pluginTwo));
+    }
+
+    @Test
+    void testPluginEqualsSetVersion() {
+        pluginOne.setVersion("myVersion");
+
+        assertTrue(CacheUtils.pluginEquals(pluginOne, pluginOne));
+        assertFalse(CacheUtils.pluginEquals(pluginOne, pluginTwo));
+
+        pluginTwo.setVersion("myVersion");
+
+        assertTrue(CacheUtils.pluginEquals(pluginOne, pluginOne));
+        assertTrue(CacheUtils.pluginEquals(pluginOne, pluginTwo));
+    }
+
+    @Test
+    void testPluginEqualsSetExtension() {
+        pluginOne.setExtensions("true");
+
+        assertTrue(CacheUtils.pluginEquals(pluginOne, pluginOne));
+        assertFalse(CacheUtils.pluginEquals(pluginOne, pluginTwo));
+
+        pluginTwo.setExtensions("TRUE");
+    }
+
+    @Test
+    void testPluginEqualsSetDependency() {
+        Dependency dependencyOne = new Dependency();
+        pluginOne.addDependency(dependencyOne);
+
+        assertTrue(CacheUtils.pluginEquals(pluginOne, pluginOne));
+        assertFalse(CacheUtils.pluginEquals(pluginOne, pluginTwo));
+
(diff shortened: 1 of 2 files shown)


impl/maven-core/src/test/java/org/apache/maven/plugin/CacheUtilsTest.java
```diff
@@ -0,0 +1,131 @@
...
+
+import static org.junit.jupiter.api.Assertions.assertFalse;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+class CacheUtilsTest {
+
+    private Plugin pluginOne;
+    private Plugin pluginTwo;
+
+    @BeforeEach
```
elharo: could just initialize the fields

impl/maven-xml/src/test/java/org/apache/maven/internal/xml/XmlNodeBuilderTest.java
```diff
@@ -44,6 +46,10 @@ public int read(char[] cbuf, int off, int len) throws IOException {
         XmlNode node1 = XmlService.read(r);
         XmlNode node2 = XmlService.read(r);
         assertEquals(node1, node2);
+
+        node1 = XmlService.read(new ByteArrayInputStream(doc.getBytes(StandardCharsets.UTF_8)), null);
```
elharo: don't reuse local variables. New variables or better yet a completely new test method.

elharo (changes requested):

impl/maven-core/src/test/java/org/apache/maven/plugin/CacheUtilsTest.java
TheRealHaui: Which then would lead every test method having 
    private Plugin pluginOne = new Plugin();
    private Plugin pluginTwo = new Plugin();
 at the beginning repeatedly.
 Don't you think that using the (at)BeforeEach approach is better?

impl/maven-xml/src/test/java/org/apache/maven/internal/xml/XmlNodeBuilderTest.java
TheRealHaui: The re-use of the variables is in no whatsoever way hidden/not visible in the code.
Therefore, I think this can be considered to not lead to any problems on developers sides because it gets overseen by them.
And as it is common in program code that variables get re-used/assigned new values I see no whatsoever problem actually to be honest.
Nonetheless, changed the code and pushed the change.

impl/maven-core/src/test/java/org/apache/maven/plugin/CacheUtilsTest.java
elharo: No, that is not how JUnit works. Just initialize the fields. A new test object is used for every test.

elharo (changes requested):

impl/maven-core/src/test/java/org/apache/maven/plugin/CacheUtilsTest.java
TheRealHaui: I knew of course because of the way JUnit works that it specifically instantiates objects.
However, wasn't aware of that  - test adapted and pushed.

impl/maven-xml/src/test/java/org/apache/maven/internal/xml/XmlNodeBuilderTest.java
```diff
@@ -44,6 +46,10 @@ public int read(char[] cbuf, int off, int len) throws IOException {
         XmlNode node1 = XmlService.read(r);
         XmlNode node2 = XmlService.read(r);
         assertEquals(node1, node2);
+
```
elharo, line 49: This should be a separate test method

elharo (changes requested):

impl/maven-core/src/test/java/org/apache/maven/plugin/CacheUtilsTest.java
```diff
@@ -0,0 +1,123 @@
...
+        pluginTwo.setVersion("myVersion");
+
+        assertTrue(CacheUtils.pluginEquals(pluginOne, pluginOne));
+        assertTrue(CacheUtils.pluginEquals(pluginOne, pluginTwo));
+    }
+
+    @Test
+    void testPluginEqualsSetExtension() {
+        pluginOne.setExtensions("true");
+
```
gnodet, line 80: The test sets `pluginTwo.setExtensions("TRUE")` but never asserts anything after it. Since `isExtensions()` delegates to `Boolean.parseBoolean()` which is case-insensitive, both plugins should now have `isExtensions() == true` and `pluginEquals` should return `true`. Please add the missing assertion:

```suggestion
        pluginTwo.setExtensions("TRUE");

        assertTrue(CacheUtils.pluginEquals(pluginOne, pluginOne));
        assertTrue(CacheUtils.pluginEquals(pluginOne, pluginTwo));
```

impl/maven-core/src/test/java/org/apache/maven/plugin/CacheUtilsTest.java
```diff
@@ -0,0 +1,123 @@
...
+        assertTrue(CacheUtils.pluginEquals(pluginOne, pluginOne));
+        assertFalse(CacheUtils.pluginEquals(pluginOne, pluginTwo));
+
+        dependencyTwo.setVersion("myDependencyVersion");
+
+        assertTrue(CacheUtils.pluginEquals(pluginOne, pluginOne));
+        assertFalse(CacheUtils.pluginEquals(pluginOne, pluginTwo));
+
+        dependencyOne.setArtifactId("myDependencyArtifactId");
```
gnodet, line 117: Line 115 (`dependencyTwo.setVersion("myDependencyVersion")`) is redundant -- this was already set on line 110. It looks like the intended line was `dependencyOne.setVersion("myDependencyVersion")` to bring both dependencies back to equality, followed by a `assertTrue` to verify convergence. As written, the test never verifies that two dependencies with identical fields are considered equal (after the initial empty-dependency case).

Also consider testing the remaining fields compared by `dependenciesEquals()`: `type`, `classifier`, `scope`, and `exclusions`.

gnodet (changes requested):
Thanks for adding tests for `CacheUtils` and `XmlNodeBuilderTest`. Test coverage for utility methods is valuable. I found a few issues with test correctness that should be addressed before merging.

**CacheUtilsTest**

1. **Incomplete `testPluginEqualsSetExtension`**: The test sets `pluginTwo.setExtensions("TRUE")` on the last line but never asserts the result. Since `Plugin.isExtensions()` uses `Boolean.parseBoolean()` (which is case-insensitive), both "true" and "TRUE" yield `isExtensions() == true`, so `pluginEquals` should return `true` at that point. The missing assertion means the test does not actually verify the extensions comparison.

2. **`testPluginEqualsSetDependency` has a redundant set and never reaches full equality**: `dependencyTwo.setVersion("myDependencyVersion")` is set on line 110, and then set again identically on line 115 -- the second call is a no-op. The test ends asserting inequality but never completes the cycle by setting `dependencyOne.setVersion("myDependencyVersion")` and asserting that the plugins are then equal. This pattern (diverge then converge) is followed in every other test method but is broken here.

3. **Missing coverage for dependency fields**: `dependenciesEquals()` compares type, classifier, and scope, but none of these are tested. Similarly, `exclusionsEquals()` is never exercised at all.

4. **No tests for `pluginHashCode`**: The other public method in `CacheUtils` has no test coverage.

gnodet (commented):
The PR adds tests for `CacheUtils.pluginEquals` and `XmlNodeBuilder`, which is welcome. However, there are correctness issues and coverage gaps that should be addressed:

**1. Missing assertion in testPluginEqualsSetExtension (high)**
The test ends with `pluginTwo.setExtensions("TRUE")` but no assertion follows. Since `Boolean.parseBoolean` is case-insensitive, both plugins have `isExtensions() == true` at that point, so `pluginEquals` should return `true`. The test never verifies this convergence path — every other test method follows the diverge-then-converge pattern.

**2. Bug in testPluginEqualsSetDependency (high)**
Line 118 redundantly sets `dependencyTwo.setVersion("myDependencyVersion")`, which was already set on line 112 (a no-op). The intended line was likely `dependencyOne.setVersion("myDependencyVersion")` to converge both dependencies to equality. As written, the test never verifies that two dependencies with identical fields are considered equal.

**3. Incomplete dependency comparison coverage (medium)**
`dependenciesEquals()` also compares `type`, `classifier`, and `scope`, but the test only exercises `groupId`, `artifactId`, and `version`. The `exclusionsEquals()` path is never exercised at all.

**4. No tests for pluginHashCode() (medium)**
`CacheUtils` has two public methods — `pluginEquals` and `pluginHashCode`. The hash code method is not tested.

**5. Wildcard static import (low)**
`import static org.junit.jupiter.api.Assertions.*` doesn't follow the project convention. All other test files under `impl/` use explicit imports.

**6. Mixed test concerns in XmlNodeBuilderTest (low)**
The `InputStream`-based parsing test (lines 49-52) is appended to `testReadMultiDoc` but tests a fundamentally different code path (`XmlService.read(InputStream)` vs `XmlService.read(Reader)`). As @elharo requested, this should be a separate `@Test` method.

All of these issues were flagged in prior reviews (by gnodet and elharo) and remain unaddressed since the last commit (December 6, 2025).

slachiewicz:
The last CI run on this branch was 2025-12-07, and GitHub has deleted the logs since, so there's no way to tell whether the change still builds against master.

PR #11524, opened 2025-12-07
Update formatting of prerequisites-requirements error to improve readability

slawekjaranowski:
Message at the end of error line is not readable ... so add some new lines and tabs to make it more readable

chery pick from ec21f4bf223c3f2e5fda052d405e2b432fea8e7b

ref:
 - #11523
 
---

### DIFF
--- a/impl/maven-core/src/main/java/org/apache/maven/plugin/internal/DefaultMavenPluginManager.java
+++ b/impl/maven-core/src/main/java/org/apache/maven/plugin/internal/DefaultMavenPluginManager.java
@@ -316,13 +316,11 @@ public class DefaultMavenPluginManager implements MavenPluginManager {
         if (!prerequisiteExceptions.isEmpty()) {
             String messages = prerequisiteExceptions.stream()
                     .map(IllegalStateException::getMessage)
-                    .collect(Collectors.joining(", "));
+                    .collect(Collectors.joining("\n\t"));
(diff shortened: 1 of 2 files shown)


elharo (changes requested):
This PR appears to produce different results on different platforms. It does not talk to the user or write to the console so this is not needed or desirable. Maven code should be platform independent.

impl/maven-core/src/main/java/org/apache/maven/plugin/internal/DefaultMavenPluginManager.java
```diff
@@ -316,13 +316,11 @@ public void checkPrerequisites(PluginDescriptor pluginDescriptor) throws PluginI
         if (!prerequisiteExceptions.isEmpty()) {
             String messages = prerequisiteExceptions.stream()
                     .map(IllegalStateException::getMessage)
-                    .collect(Collectors.joining(", "));
+                    .collect(Collectors.joining("\n\t"));
```
gnodet, line 319: Shouldn't we keep comma as well?

gnodet: Oops, I misread...

slawekjaranowski: example output is in:
 - https://github.com/apache/maven/pull/11523

PR #11525, opened 2025-12-07
Update formatting of prerequisites-requirements error to improve readability

slawekjaranowski:
Message at the end of error line is not readable ... so add some new lines and tabs to make it more readable

chery pick from ec21f4bf223c3f2e5fda052d405e2b432fea8e7b

ref:
 - #11523

---

### DIFF
--- a/impl/maven-core/src/main/java/org/apache/maven/plugin/internal/DefaultMavenPluginManager.java
+++ b/impl/maven-core/src/main/java/org/apache/maven/plugin/internal/DefaultMavenPluginManager.java
@@ -316,13 +316,11 @@ public class DefaultMavenPluginManager implements MavenPluginManager {
         if (!prerequisiteExceptions.isEmpty()) {
             String messages = prerequisiteExceptions.stream()
                     .map(IllegalStateException::getMessage)
-                    .collect(Collectors.joining(", "));
+                    .collect(Collectors.joining("\n\t"));
             PluginIncompatibleException pie = new PluginIncompatibleException(
                     pluginDescriptor.getPlugin(),
-                    "The plugin " + pluginDescriptor.getId() + " has unmet prerequisites: " + messages,
-                    prerequisiteExceptions.get(0));
-            // the first exception is added as cause, all other ones as suppressed exceptions
-            prerequisiteExceptions.stream().skip(1).forEach(pie::addSuppressed);
+                    "\nThe plugin " + pluginDescriptor.getId() + " has unmet prerequisites: \n\t" + messages);
+            prerequisiteExceptions.forEach(pie::addSuppressed);
             throw pie;
         }
     }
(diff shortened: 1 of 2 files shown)


impl/maven-core/src/main/java/org/apache/maven/plugin/internal/DefaultMavenPluginManager.java
```diff
@@ -314,15 +314,15 @@ public void checkPrerequisites(PluginDescriptor pluginDescriptor) throws PluginI
...
                     .map(IllegalStateException::getMessage)
-                    .collect(Collectors.joining(", "));
+                    .collect(Collectors.joining(ln + '\t'));
             PluginIncompatibleException pie = new PluginIncompatibleException(
                     pluginDescriptor.getPlugin(),
-                    "The plugin " + pluginDescriptor.getId() + " has unmet prerequisites: " + messages,
-                    prerequisiteExceptions.get(0));
-            // the first exception is added as cause, all other ones as suppressed exceptions
-            prerequisiteExceptions.stream().skip(1).forEach(pie::addSuppressed);
+                    ln + "The plugin " + pluginDescriptor.getId() + " has unmet prerequisites: " + ln + '\t'
```
elharo: Output should be platform independent.

elharo (changes requested):
Lambdas make this PR hard to read, but it appears to produce different results on different platforms. It does not talk to the user or write to the console so this is not needed or desirable. Maven code should be platform independent.

impl/maven-core/src/main/java/org/apache/maven/plugin/internal/DefaultMavenPluginManager.java
slawekjaranowski: ok, simplified. here we can be platform independent

impl/maven-core/src/main/java/org/apache/maven/plugin/internal/DefaultMavenPluginManager.java
```diff
@@ -316,13 +316,11 @@ public void checkPrerequisites(PluginDescriptor pluginDescriptor) throws PluginI
         if (!prerequisiteExceptions.isEmpty()) {
             String messages = prerequisiteExceptions.stream()
                     .map(IllegalStateException::getMessage)
-                    .collect(Collectors.joining(", "));
+                    .collect(Collectors.joining("\n\t"));
```
Bukama, line 319: I think we need the system dependent line break

slawekjaranowski: not necessary here - new lines are splitted in: `MavenInvoker#logSummary` or `MavenCli#logSummary` and then print to console as separate log

elharo (approved):

PR #11539, opened 2025-12-11
Enforce API equality across 4.0 and 4.1

cstamas:
For now, in a profile that needs to be explicitly activated as rc-5 and master digressed.

To invoke:
`mvn verify -P maven4-api-check`

### DIFF
--- a/api/pom.xml
+++ b/api/pom.xml
@@ -47,9 +47,36 @@
 
   <properties>
     <project.directory>api</project.directory>
+    <maven4api.baseline>4.0.0-rc-5</maven4api.baseline>
   </properties>
 
   <profiles>
+    <profile>
+      <id>maven4-api-check</id>
+      <build>
+        <plugins>
+          <plugin>
+            <groupId>com.github.siom79.japicmp</groupId>
+            <artifactId>japicmp-maven-plugin</artifactId>
+            <configuration>
+              <oldVersion>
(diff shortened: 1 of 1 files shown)


gnodet (commented):
Reasonable approach — using an opt-in japicmp profile to check API compatibility between 4.0 and 4.1 keeps the regular build unaffected. Two observations:

1. **`onlyBinaryIncompatible` inheritance** — The root `pluginManagement` sets `onlyBinaryIncompatible=true` for japicmp. This profile inherits that setting, meaning it checks *backward binary compatibility* rather than strict *API equality*. New APIs added in 4.1 that don't exist in 4.0 (new classes, methods, interfaces) will **not** be detected. If the intent is true "API equality" (bidirectional), consider overriding `onlyBinaryIncompatible` to `false` in the profile's `<parameter>` block.

2. **`oldVersion` vs `oldVersionPattern` overlap** — The profile sets `<oldVersion>${maven4api.baseline}</oldVersion>` (4.0.0-rc-5), but the inherited pluginManagement execution also carries `<oldVersionPattern>${maven.baseline}</oldVersionPattern>` (3.9.8). This works correctly since `oldVersion` takes precedence, but having both in the effective POM is potentially confusing. Consider explicitly clearing `oldVersionPattern` in the profile.

### 📋 PR Metadata

PR #11549, opened 2025-12-13
Accept Java module names as attached artifactId even if they differ from the project's artifactId

desruisseaux:
Pull request #1838 added a verification in `DefaultProjectManager` making sure that an artifact attached to a project has the same `groupId`, `artifactId` and `version` as the project. This pull request relaxes that restriction by allowing the `artifactId` to be different, but still requires the `groupId` and `version` to be the same.

The intend is to support multi-module projects where more than one artifact may be produced. It was briefly discussed in https://github.com/apache/maven-compiler-plugin/pull/959#issuecomment-3404907890. This is needed by the [new Maven JAR Plugin proposal](https://github.com/apache/maven-jar-plugin/pull/508). Note that classical projects (package hierarchy) are expected to continue to use the same `artifactId` as before.

### DIFF
--- a/impl/maven-core/src/main/java/org/apache/maven/internal/impl/DefaultProjectManager.java
+++ b/impl/maven-core/src/main/java/org/apache/maven/internal/impl/DefaultProjectManager.java
@@ -119,7 +119,8 @@ public class DefaultProjectManager implements ProjectManager {
                     artifact.getExtension(),
                     null);
         }
-        // Verify groupId and version, intentionally allow artifactId to differ as a project may be multi-module.
+        // Verify groupId and version, intentionally allow artifactId to differ as Maven project may be
+        // multi-module with modular sources structure that provide module names used as artifactIds.
         String g1 = project.getGroupId();
         String a1 = project.getArtifactId();
         String v1 = project.getVersion();
@@ -127,7 +128,8 @@ public class DefaultProjectManager implements ProjectManager {
         String a2 = artifact.getArtifactId();
         String v2 = artifact.getBaseVersion().toString();
 
-        // ArtifactId may differ only for multi-module projects (source roots with module name)
+        // ArtifactId may differ only for multi-module projects, in which case
+        // it must match the module name from a source root in modular sources.
         boolean isMultiModule = false;
         boolean validArtifactId = Objects.equals(a1, a2);
         for (SourceRoot sr : getSourceRoots(project)) {


impl/maven-core/src/test/java/org/apache/maven/internal/impl/DefaultProjectManagerTest.java
```diff
@@ -56,7 +57,15 @@ void attachArtifact() {
         when(artifact.getBaseVersion()).thenReturn(versionParser.parseVersion("1.0-SNAPSHOT"));
         projectManager.attachArtifact(project, artifact, path);
 
+        // Verify that no exception is thrown when only the arficactId differ
         when(artifact.getArtifactId()).thenReturn("anotherArtifact");
-        assertThrows(IllegalArgumentException.class, () -> projectManager.attachArtifact(project, artifact, path));
+        projectManager.attachArtifact(project, artifact, path);
+
+        when(artifact.getGroupId()).thenReturn("anotherGroup");
```
ascheman, line 99: Would you mind adding a comment similar to L60 ('// Verify ...') here to make clear that the second part of the test case starts here. Or make even a second test case out of it?

desruisseaux: Done, thanks for the review.

kwin (commented):
If the GAV is no longer the same how can the attached artifact share the same POM (with a different artifactId)?

impl/maven-core/src/main/java/org/apache/maven/internal/impl/DefaultProjectManager.java
```diff
@@ -119,15 +119,16 @@ public void attachArtifact(@Nonnull Project project, @Nonnull ProducedArtifact a
...
+        String g2 = artifact.getGroupId();
+        String v1 = project.getVersion();
+        String v2 = artifact.getBaseVersion().toString();
+        if (!Objects.equals(g1, g2) || !Objects.equals(v1, v2)) {
             throw new IllegalArgumentException(
-                    "The produced artifact must have the same groupId/artifactId/version than the project it is attached to. Expecting "
-                            + project.getGroupId() + ":" + project.getArtifactId() + ":" + project.getVersion()
-                            + " but received " + artifact.getGroupId() + ":" + artifact.getArtifactId() + ":"
-                            + artifact.getBaseVersion());
+                    "The produced artifact must have the same groupId and version than the project it is attached to. Expecting "
```
elharo: than --> as

desruisseaux: Fixed, thanks.

ascheman: With Java 17 (pre-req for M4) we have multi line strings. I would think the following would be even better readable:

```java
"""
The produced artifact must have the same groupId and version as the project it is attached to. \
Expecting %s:%s:%s but received %s:%s:%s"""
        .formatted(g1, project.getArtifactId(), v1, g2, artifact.getArtifactId(), v2));
```

Unfortunately, Spotless reformats it a little bit (adding the `.format` method call directly to the end of the String literal):

```java
"""
The produced artifact must have the same groupId and version as the project it is attached to. \
Expecting %s:%s:%s but received %s:%s:%s""".formatted(g1, project.getArtifactId(), v1, g2, artifact.getArtifactId(), v2));
```

Nevertheless, it's still multi-line and one can easier see how the final message would look like.

Note that the \ enables to split the message into multiple lines in the IDE but concatenates the lines in the final output.

What do you think?

hboutemy:
> If the GAV is no longer the same how can the attached artifact share the same POM (with a different artifactId)?

I have the same concern, what is the pom of "automatic sub-artifactId Java Module" vs  the base artifactId?

this is really key, ad requires a clear design before trying to implement

ascheman:
> IIUC, you expect to use the same consumer pom for automatic artifactIds of Java module JAR, just update the artifactId? and implicitely no build pom?

Well, the overall context is a project where you have multiple Java modules. I have provided some examples [here](https://github.com/support-and-care/java9-jigsaw-examples/). 

If I pick one (cf.  an arbitrary [`pom.xml`](https://github.com/support-and-care/java9-jigsaw-examples/blob/main/jigsaw-examples/example_layer-hierarchy/m4/pom.xml), same holds for almost each example, cf. their m4 subdirectories), there is usually just one build POM. But the build would result in multiple binary artefacts (or it could result in such as this is work-in-progress), one for each module. 

For each such module JAR there must be a separate consumer POM as they can be used independently. These can be derived from the build POM. AFAIK there is no concept to explicitly provide a consumer POM for a module (and in most cases this should not be necessary as they can be derived from the build POM).

> for better understanding of this plan (that would deserve more than a small discussion in an issue in Maven core: is there a doc somewhere?), is there a convention between original artifactId and automatic artifactIds?

There seems not to be much documentation about M4 consumer POM concepts or a specification how to work with Java modules (I'd be happy to find more about it).

That said, I'd be volunteering to create such concept documentation.
 
> and is there a jar associated to the original artifactId?

For a multi-module project there is not much need to create a binary (classes) JAR for the project itself, as each module should result in its own jar. Perhaps the sources or javadocs etc. could be created from the original project. This might be subject to discussion.

gnodet:
> Hello @gnodet. I have a question related to this pull request. Is it possible to edit the consumer POM of the attached artifacts for:
> 
> * Setting the value of the `<artifactId>` element to the `artifact.getArtifactId()` value.
> * Filter the dependencies for keeping only the ones that are declared in the `module-info` file.
> * Change the `<optional>` value of some dependencies depending on whether the dependency is declared as `static` in the `module-info` file.
> 
> Any tip would be greatly appreciated.

The consumer POM is only generated from the main build POM for project being compiled.  If we generate new artifacts dynamically, we'll have to generate the POM and attach it AFAIK.  The `DefaultConsumerPomBuilder` uses the `MavenProject` as the main input, so I don't really think it makes sense to reuse it.

desruisseaux:
Hello @hboutemy and all. First, there are a few Java 9 concepts that I would like to ensure that they are well known:

* Since Java 9, `javac` can compile many modules at once. This is possible when using the so-called [module source hierarchy](https://docs.oracle.com/en/java/javase/25/docs/specs/man/javac.html#directory-hierarchies).
* There are some [technical advantages](https://sis.apache.org/release-notes/Modularization.html#advantages) in doing that, and no way I'm aware of to get some of these advantages in the traditional way.
* When using module hierarchy, the `javac` output is `<module>/<package>/*.class`. The `<module>` additional level is a standard Java output, not something that we decided on our side.
* The above additional `<module>` directory level is understood by other standard Java tools such as `java` and `javadoc`. One notable exception is the `jar` tool, maybe intentionally for forcing a limitation of one module per JAR for smaller download.

They are been attempts to explain these aspects on the Maven mailing list, on [a wiki with proof of concept](https://github.com/Geomatys/maven/wiki) and with a [real project experiment using Gradle](https://sis.apache.org/release-notes/Modularization.html). However, I got the feeling that the above is still nebulous to most peoples, and that the best way to advance would be to give to developers a possibility to try easily by themselves.

The Maven Compiler Plugin 4.0.0-beta-3 is already capable to compile a project organized with module source hierarchy. Consequently, the output in `target/classes` is multi-module when the project is organized that way (again, because `javac` works that way, not because we made a choice that we are imposing to others). For Java tools such as `java` and `javadoc`, this output can still be consumed as a single output with a single `--module-path target/classes` option. This option understand the `javac` module hierarchy output. Therefore, for these tools, the "1 Maven project = 1 main output" pattern could still apply. But as said above, `jar` is an exception (probably intentional).

Now about the questions:

> what is the pom of "automatic sub-artifactId Java Module" vs the base artifactId? (…snip…) you expect to use the same consumer pom for automatic artifactIds of Java module JAR, just update the artifactId? and implicitly no build pom?

Note that when building a project with Maven 4, we already have a `target/project-local-repo/${groupId}/${artifactId}/${version}` directory with a `….pom` and a `…-consumer.pom` files in it. The consumer POM seems to differ from the source POM in that it has comments removed, the `<scm>` and `<repositories>` sections removed, the `${…}` properties resolved to their values and some reformatting applied. My proposal is to replicate the consumer POM for each JAR with the following changes:

* `artifactId` set to the Java module name.
* Dependencies filtered to the intersections of the Maven dependencies and the `module-info`:
  * Java modules enumerated in a `requires` statement are ordinary dependency (`compile` scope).
  * Java modules enumerated in a `static requires` statement are dependency with the `optional` element set to `true`.
* `<build>` and `<profiles>` sections removed.

Therefore, the base `artifactId` would not be used except for the build (original) POM.

>  is there a jar associated to the original artifactId?

No when using the module source hierarchy. Reminder: nothing change when using the traditional package hierarchy.

> when I read "This is needed by the new Maven JAR Plugin proposal https://github.com/apache/maven-jar-plugin/pull/508.", I don't get the relationship with using jar vs MavenArchiver: we need to be cautious

The new proposed JAR plugin automatically detects when module hierarchy is used, and produces one JAR per module in that case. However, it results in an exception thrown by Maven core if the relaxation proposed in this pull request is not applied. Reminder: again, nothing change when the JAR plugin detects that the traditional package hierarchy was used.

The reason why the JAR pull request mixes two tasks in the same pull request is because one of the reasons to migrate to the `jar` tool was to handle multi-release JAR files with the `--release` option in order to address https://github.com/apache/maven-jar-plugin/issues/484 and by experience with the compiler plugin (verified again with the JAR plugin), it is easier to address multi-release and multi-module together than separately.

impl/maven-core/src/main/java/org/apache/maven/internal/impl/DefaultProjectManager.java
desruisseaux: True, I forgot multi line strings. I will update as proposed.

ascheman:
> > is there a jar associated to the original artifactId?
> 
> No when using the module source hierarchy. Reminder: nothing change when using the traditional package hierarchy.
> 
> > when I read "This is needed by the new Maven JAR Plugin proposal [apache/maven-jar-plugin#508](https://github.com/apache/maven-jar-plugin/pull/508).", I don't get the relationship with using jar vs MavenArchiver: we need to be cautious
> 
> The new proposed JAR plugin automatically detects when module hierarchy is used, and produces one JAR per module in that case. However, it results in an exception thrown by Maven core if the relaxation proposed in this pull request is not applied. Reminder: again, nothing change when the JAR plugin detects that the traditional package hierarchy was used.

To make clear that the relaxation is only necessary and used in the case of multi-module JARs, the proposed change could be adapted to check whether the `project` is multi-module and only leave out the comparison of artifactId in this case (as suggested before):

```java
...
  String a1 = project.getArtifactId();
  String a2 = artifact.getArtifactId();
...
  // ArtifactId may differ only for multi-module projects (source roots with module name)
  boolean isMultiModule = getSourceRoots(project).stream().anyMatch(sr -> sr.module().isPresent());
  if (!Objects.equals(a1, a2) && !isMultiModule) {
      throw new IllegalArgumentException(...);
  }
...
```

impl/maven-core/src/main/java/org/apache/maven/internal/impl/DefaultProjectManager.java
desruisseaux: Actually I tried 3 variants: above multi-line, `String.formatted` and `String.format`, and the variant that seem to be the most Spotless-friendly is the last one. The two others are reformatted by Spotless in a way no so easy to read.

impl/maven-core/src/main/java/org/apache/maven/internal/impl/DefaultProjectManager.java
```diff
@@ -119,15 +119,32 @@ public void attachArtifact(@Nonnull Project project, @Nonnull ProducedArtifact a
...
+            Optional<String> moduleName = sr.module();
+            if (moduleName.isPresent()) {
+                isMultiModule = true;
+                if (moduleName.get().equals(a2)) {
+                    artifactMatchesModule = true;
+                    break;
+                }
+            }
+        }
+        if (!(Objects.equals(g1, g2) && Objects.equals(v1, v2) && (artifactMatchesModule || Objects.equals(a1, a2)))) {
```
gnodet: This may be harder to understand for users.  I think different error messages for modular and non modular projects may be easier to understand.

impl/maven-core/src/main/java/org/apache/maven/internal/impl/DefaultProjectManager.java
```diff
@@ -119,15 +119,32 @@ public void attachArtifact(@Nonnull Project project, @Nonnull ProducedArtifact a
...
+            throw new IllegalArgumentException(String.format(
+                    "The produced artifact must have the same groupId%s and version as the project it is attached to.%n"
+                            + "Expecting %s:%s:%s but received %s:%s:%s.",
+                    isMultiModule ? "" : ", artifactID", g1, a1, v1, g2, a2, v2));
```
gnodet:
```suggestion
            String message;
            if (isMultiModule) {
                // Multi-module project: artifactId may match any declared module name
                message = String.format(
                        "Cannot attach artifact to project: groupId and version must match the project, "
                                + "and artifactId must match either the project or a declared module name.%n"
                                + "  Project coordinates: %s:%s:%s%n"
                                + "  Artifact coordinates: %s:%s:%s%n"
                                + "  Hint: The artifactId '%s' does not match the project artifactId '%s' "
                                + "nor any declared module name in source roots.",
                        g1, a1, v1, g2, a2, v2, a2, a1);
            } else {
                // Non-modular project: artifactId must match exactly
                message = String.format(
                        "Cannot attach artifact to project: groupId, artifactId and version must match the project.%n"
                                + "  Project coordinates: %s:%s:%s%n"
                                + "  Artifact coordinates: %s:%s:%s",
                        g1, a1, v1, g2, a2, v2);
            }
            throw new IllegalArgumentException(message);
```

impl/maven-core/src/main/java/org/apache/maven/internal/impl/DefaultProjectManager.java
desruisseaux: Done, the exception message is now built separately for the modular and non-modular cases. Also split the modular case in two sub-cases.

impl/maven-core/src/main/java/org/apache/maven/internal/impl/DefaultProjectManager.java
desruisseaux: Indeed, those messages are clearer. Applied as suggested.

gnodet (approved):

hboutemy:
>  the relaxation is only necessary and used in the case of multi-module JARs

I'm starting to understand that one single multi-java module build gives many "what used to be classical jars, with one build (= 1 pom) per jar"

then this is not about relaxing attach strategy
it's about creating pseudo POMs (with their disting gav) that will host their jars done from one central build

now I get that we have one build pom, that generates many consumer poms with their own output modular jar.
Saying that, is "multi-module JARs" a good description? each jar is a single module, but it is a multi-Java module build done (with one single compiler invocation), isn't it?

hboutemy:
one question: one reason for having one build pom per sub-project is that there are resources, assemblies, and everything that goes beyond just the compiler step, and also there are unit tests, and ITs, and everything else

in this "one single pom for one single compiler run that build many independent Java modular jars", how do you handle the flexibility about everything that is not .class in the modular jars, and their tests, etc?

hboutemy:
@gnodet I don't get how v4 API permits to attach something to a pom (in v3 a MavenProject) that does not exist
who will create the pom that will be necessary for install/deploy?

gnodet:
> > The jar plugin will attach the artifacts and the poms when building a modular project
> 
> uh, attaching the pom? I see how a plugin can attach a consumer pom to a Maven Project that provides the build pom: but we still miss the Maven Project that provides the build pom

I don't understand.  You can attach any kind of artifact to a project, it can be a jar, a pom, or whatever.  From the install/deployment POV, there's no difference between a main pom, a consumer pom, or the jar produced.  Those are all artifacts (groupId, artifactId, version, extension, classifier) that point to file on the filesystem.  Then, if we talk about semantics, I'm not sure yet how to produce those poms, but I think the overall idea is to create something similar to the consumer POM for the project, with small modifications to follow the rules of a specific JPMS module (not a maven module).

> > The [](https://maven.apache.org/ref/4.0.0-rc-5/api/maven-api-model/apidocs/org/apache/maven/api/model/Source.html) element in POM 4.1.0 is more flexible than the  element in POM 4.0.0. This is where all the modular information is stored.
> 
> I'm still not convinced `<source>` can replace what we configure with plugins in a sub-project

I don't think that's a goal.  We're not talking about subprojects here, just JPMS modules that are created from a single POM file.  Maybe you're thinking about plugins that would generate other kind of resources ? For example a plugin that would read the class files and generate something to be included in the jar ? That would definitely not work without adapting the plugin to support this JPMS module layout.  Or you could configure multiple executions, each one pointing to the specific output directory of the module.  For example with the [modular project IT](https://github.com/apache/maven-compiler-plugin/tree/master/src/it/modular-sources), we'll have `target/classes/org.bar` and `target/classes/org.foo`, so either the plugin supports this new layout and it will figure it out, or you can create two executions explicitly pointing to those directories.

impl/maven-core/src/main/java/org/apache/maven/internal/impl/DefaultProjectManager.java
```diff
@@ -119,15 +119,53 @@ public void attachArtifact(@Nonnull Project project, @Nonnull ProducedArtifact a
...
-        if (!Objects.equals(project.getGroupId(), artifact.getGroupId())
-                || !Objects.equals(project.getArtifactId(), artifact.getArtifactId())
-                || !Objects.equals(
-                        project.getVersion(), artifact.getBaseVersion().toString())) {
-            throw new IllegalArgumentException(
-                    "The produced artifact must have the same groupId/artifactId/version than the project it is attached to. Expecting "
-                            + project.getGroupId() + ":" + project.getArtifactId() + ":" + project.getVersion()
-                            + " but received " + artifact.getGroupId() + ":" + artifact.getArtifactId() + ":"
-                            + artifact.getBaseVersion());
+        // Verify groupId and version, intentionally allow artifactId to differ as a project may be multi-module.
```
hboutemy: "as a project may be multi-module." => "as Maven project may be multi-module with modular sources structure that provide module names used as artifactIds."

impl/maven-core/src/main/java/org/apache/maven/internal/impl/DefaultProjectManager.java
```diff
@@ -119,15 +119,53 @@ public void attachArtifact(@Nonnull Project project, @Nonnull ProducedArtifact a
...
+        String g1 = project.getGroupId();
+        String a1 = project.getArtifactId();
+        String v1 = project.getVersion();
+        String g2 = artifact.getGroupId();
+        String a2 = artifact.getArtifactId();
+        String v2 = artifact.getBaseVersion().toString();
+
+        // ArtifactId may differ only for multi-module projects (source roots with module name)
```
hboutemy: "ArtifactId may differ" => "ArtifactId from attached artifact may differ from project artifactId"
"(source roots with module name)" => "then must match module name from source root in modular sources)"

impl/maven-core/src/main/java/org/apache/maven/internal/impl/DefaultProjectManager.java
desruisseaux: Applied as suggested.

PR #11551, opened 2025-12-14
`FileSelector.matches(Path)` sometime wrong for a file or a directory

desruisseaux:
This pull request fixes the following bugs in `PathSelector`:

* Brace expansion must be applied also to the "/**" path suffix.
* Detection of "match all" pattern must take brace expansion in account.
* The optimization for excluding whole directories sometime excludes a directory that shouldn't be excluded.

Those problems were identified by the following steps applied locally:

* Port #11125 to the `maven-4.0.x` branch.
* Modify Maven JAR Plugin for using `PathSelector`.
* Run the integration tests of Maven Clean Plugin and Maven JAR Plugin.

With this pull request, all above-cited integration tests pass.

## Note on refactoring
This commit moves the `dirIncludes` and `dirExcludes` matchers to an inner class that implements `PathMatcher` directly (the indirection level through a lambda function is removed). With this change, the directory matchers are created only if requested and can be provided by a more direct `PathMatcher`. The new code is a bit more conservative about directory filtering.

### DIFF
--- a/impl/maven-impl/src/main/java/org/apache/maven/impl/PathSelector.java
+++ b/impl/maven-impl/src/main/java/org/apache/maven/impl/PathSelector.java
@@ -160,13 +160,13 @@ final class PathSelector implements PathMatcher {
 
     /**
      * The wildcard used by the "glob" syntax for meaning zero or more leading directories.
-     * It cannot be {@code "**​/"} because that wildcard matches one or more directory.
+     * It cannot be {@code "**​/"} because that wildcard matches one or more directories.
      */
     private static final String WILDCARD_FOR_ANY_PREFIX = "{**/,}";
 
     /**
      * The wildcard used by the "glob" syntax for meaning zero or more trailing directories.
-     * It cannot be {@code "/**​"} because that wildcard matches one or more directory.
+     * It cannot be {@code "/**​"} because that wildcard matches one or more directories.
      */
     private static final String WILDCARD_FOR_ANY_SUFFIX = "{/**,}";
 


impl/maven-impl/src/main/java/org/apache/maven/impl/PathSelector.java
```diff
@@ -598,30 +581,73 @@ private static boolean isMatched(Path path, PathMatcher[] matchers) {
...
-     *
-     * @return {@code true} if directory filtering is possible, {@code false} if all directories
-     *         will be considered as potentially containing selected files
+     * Returns a matcher that can be used for pre-filtering the directories.
+     * The returned matcher can be used as an optimization for skipping whole directories when possible.
+     * If there is no such optimization, then this method returns {@link #INCLUDES_ALL}.
      */
-    boolean canFilterDirectories() {
-        return dirIncludes.length != 0 || dirExcludes.length != 0;
+    public PathMatcher createDirectoryMatcher() {
```
elharo: Looks like this method doesn't have to be public

desruisseaux: True, I will change that.

impl/maven-impl/src/main/java/org/apache/maven/impl/PathSelector.java
```diff
@@ -158,6 +158,18 @@ final class PathSelector implements PathMatcher {
...
 
+    /**
+     * The wildcard used by the "glob" syntax for meaning zero or more leading directories.
+     * It cannot be {@code "**​/"} because that wildcard matches one or more directory.
+     */
+    private static final String WILDCARD_FOR_ANY_PREFIX = "{**/,}";
+
+    /**
+     * The wildcard used by the "glob" syntax for meaning zero or more trailing directories.
+     * It cannot be {@code "/**​"} because that wildcard matches one or more directory.
```
elharo: one or more directories

desruisseaux: Fixed locally, waiting to see if there are more comments before to push.

gnodet (approved):

Issue #10328, opened 2025-05-25
[MNG-8748] org.apache.maven.api.services.model.UrlNormalizer should be a concrete class


jira-importer:
**[Elliotte Rusty Harold](https://issues.apache.org/jira/secure/ViewProfile.jspa?name=elharo)** opened **[MNG-8748](https://issues.apache.org/jira/browse/MNG-8748?redirect=false)** and commented

There's only one implementation and unlikely to be others. This class should likely be final. It can be made non-final later if ever a need is uncovered to change the behavior.

---
No further details from [MNG-8748](https://issues.apache.org/jira/browse/MNG-8748?redirect=false)

PR #11568, opened 2025-12-16
[ISSUE-10329] Document behaviour of UrlNormalizer

### LINKED ISSUE #10329: [MNG-8747] UrlNormalizer insufficiently specified
**[Elliotte Rusty Harold](https://issues.apache.org/jira/secure/ViewProfile.jspa?name=elharo)** opened **[MNG-8747](https://issues.apache.org/jira/browse/MNG-8747?redirect=false)** and commented

org.apache.maven.api.services.model.UrlNormalizer  needs to document what it means to "normalize" a URL, especially since it's only an interface.

---
No further details from [MNG-8747](https://issues.apache.org/jira/browse/MNG-8747?redirect=false)

Giovds:
Closes #10329 

### DIFF
--- a/impl/maven-impl/src/main/java/org/apache/maven/api/services/model/UrlNormalizer.java
+++ b/impl/maven-impl/src/main/java/org/apache/maven/api/services/model/UrlNormalizer.java
@@ -18,6 +18,18 @@
  */
 package org.apache.maven.api.services.model;
 
+/**
+ * Provide a service of {@link UrlNormalizer} that simplifies URL strings by removing parent directory
+ * references ("/../") and collapsing path segments. This implementation performs purely
+ * string-based normalization without full URL parsing or validation.
+ *
+ * <p>The normalization process iteratively removes "/../" segments by eliminating the preceding path segment,
+ * effectively resolving relative path traversals.
+ *
+ * <p>Note that this implementation does not guarantee that the resulting URL is valid or reachable; it simply
+ * produces a more canonical representation of the input string.
+ *
+ */
 public interface UrlNormalizer {
 
     /**
--- a/impl/maven-impl/src/main/java/org/apache/maven/api/services/model/UrlNormalizer.java
+++ b/impl/maven-impl/src/main/java/org/apache/maven/api/services/model/UrlNormalizer.java
@@ -19,14 +19,13 @@
 package org.apache.maven.api.services.model;
 
 /**
- * Provide a service of {@link UrlNormalizer} that simplifies URL strings by removing parent directory
- * references ("/../") and collapsing path segments. This implementation performs purely
- * string-based normalization without full URL parsing or validation.
+ * Simplifies URLs by removing parent directory references ("/../") and collapsing path segments.
+ * This performs purely string-based normalization without full URL parsing or validation.
  *
  * <p>The normalization process iteratively removes "/../" segments by eliminating the preceding path segment,
  * effectively resolving relative path traversals.
  *
- * <p>Note that this implementation does not guarantee that the resulting URL is valid or reachable; it simply
+ * <p>This does not guarantee that the resulting URL is valid or reachable; it simply
  * produces a more canonical representation of the input string.
  *
  */
--- a/impl/maven-impl/src/test/java/org/apache/maven/impl/DefaultUrlNormalizerTest.java
+++ b/impl/maven-impl/src/test/java/org/apache/maven/impl/DefaultUrlNormalizerTest.java
@@ -1,0 +1,69 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.maven.impl;
+
+import org.junit.jupiter.api.Test;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertNull;
+
+class DefaultUrlNormalizerTest {
+
+    private final DefaultUrlNormalizer sut = new DefaultUrlNormalizer();
+
+    @Test
+    void normalizeShouldHandleNullAndEdgeCases() {
+        assertNull(sut.normalize(null));
+        assertEquals("", sut.normalize(""));
+        assertEquals("/", sut.normalize("/../"));
+        assertEquals("", sut.normalize("a/../"));
+        assertEquals("b", sut.normalize("a/../b"));
+        assertEquals("b/d", sut.normalize("a/../b/c/../d"));
+        assertEquals("b/c/d", sut.normalize("a/../b/c/d"));
+        assertEquals("b/c", sut.normalize("a/../b/c"));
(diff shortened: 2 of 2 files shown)


gnodet (approved):

Giovds:
@gnodet Highly unlikely the IT failed due to adding a comment and a unit test. Perhaps flaky or other issue?

impl/maven-impl/src/main/java/org/apache/maven/api/services/model/UrlNormalizer.java
```diff
@@ -18,6 +18,18 @@
  */
 package org.apache.maven.api.services.model;
 
+/**
+ * Provide a service of {@link UrlNormalizer} that simplifies URL strings by removing parent directory
```
elharo: This isn't really a "service" as that word is often used in software development these days. It's just an interface. 

"Provide a service of {@link UrlNormalizer} that simplifies URL strings" --> Simplifies URLs

impl/maven-impl/src/main/java/org/apache/maven/api/services/model/UrlNormalizer.java
```diff
@@ -18,6 +18,18 @@
...
+ * references ("/../") and collapsing path segments. This implementation performs purely
```
elharo: This is an interface, not an implementation

impl/maven-impl/src/main/java/org/apache/maven/api/services/model/UrlNormalizer.java
```diff
@@ -18,6 +18,18 @@
...
+ * string-based normalization without full URL parsing or validation.
+ *
+ * <p>The normalization process iteratively removes "/../" segments by eliminating the preceding path segment,
+ * effectively resolving relative path traversals.
+ *
+ * <p>Note that this implementation does not guarantee that the resulting URL is valid or reachable; it simply
```
elharo: Note that this implementation does not --> This does not

elharo (changes requested):

impl/maven-impl/src/main/java/org/apache/maven/api/services/model/UrlNormalizer.java
```diff
@@ -18,6 +18,17 @@
  */
 package org.apache.maven.api.services.model;
 
+/**
```
elharo, line 21: Weird that we have an interface here. We don't really need it for anything, but that's tangential to this PR.

elharo (approved):

impl/maven-impl/src/main/java/org/apache/maven/api/services/model/UrlNormalizer.java
Giovds: Personally I agree that having an interface with 1 implementation is not as pure as it can be, though this is indeed a topic that is beyond the scope of this PR.

I think the same discussion lives in #10328 

If there is a consensus on what path we are taking, we can act accordingly for this (and perhaps more) classes.

Giovds:
Is there anything I can do to help this PR further? I believe all comments have been resolved or discussed.

impl/maven-impl/src/main/java/org/apache/maven/api/services/model/UrlNormalizer.java
```diff
@@ -18,6 +18,17 @@
...
+ * Simplifies URLs by removing parent directory references ("/../") and collapsing path segments.
```
elharo, line 22: I wonder if this comment belongs on the method rather than or in addition to the class so it will appear in IDE tooltips when I hover over an invocation?

impl/maven-impl/src/main/java/org/apache/maven/api/services/model/UrlNormalizer.java
elharo: since there is an interface, there is a possibility that a different implementation will do something different. Is this planned or intended? If not, we really should collapse this to a single class.

elharo (changes requested):
really shouldn't be an interface

impl/maven-impl/src/main/java/org/apache/maven/api/services/model/UrlNormalizer.java
```diff
@@ -18,6 +18,17 @@
...
+ * This performs purely string-based normalization without full URL parsing or validation.
+ *
+ * <p>The normalization process iteratively removes "/../" segments by eliminating the preceding path segment,
+ * effectively resolving relative path traversals.
+ *
+ * <p>This does not guarantee that the resulting URL is valid or reachable; it simply
```
gnodet, line 28: Nit: there is a trailing blank line before the closing `*/`. The existing method javadoc in this same file does not have one. Consider removing it for consistency:
```suggestion
 * produces a more canonical representation of the input string.
 */
```

gnodet (commented):
Good documentation and test additions. The javadoc accurately reflects the actual implementation behavior, and the tests are well-structured and thorough. A few minor observations:

1. The class-level javadoc on the interface is accurate and well-written. The reviewer @elharo previously suggested (in an unaddressed comment) that the behavioral details might also belong on the `normalize` method itself so they appear in IDE tooltips at call sites. That is worth considering.

2. The new test class for `DefaultUrlNormalizer` under `impl/maven-impl` is a useful addition. It covers null, empty strings, relative paths, absolute URLs, double slashes, encoded characters, and malformed URLs -- good coverage.

3. One small nit on the test: the first test method `normalizeShouldHandleNullAndEdgeCases` packs many distinct scenarios into a single test. This is a minor stylistic point and not a blocker.

4. There is an empty line before the closing `*/` in the class javadoc (line before `*/`). While not incorrect, it is inconsistent with the existing method-level javadoc in the same file, which has no trailing blank line.

Overall this is a solid contribution -- the documentation is accurate and the tests verify real behavior.

impl/maven-impl/src/test/java/org/apache/maven/impl/DefaultUrlNormalizerTest.java
```diff
@@ -0,0 +1,69 @@
...
+        assertEquals("", sut.normalize(""));
+        assertEquals("/", sut.normalize("/../"));
+        assertEquals("", sut.normalize("a/../"));
+        assertEquals("b", sut.normalize("a/../b"));
+        assertEquals("b/d", sut.normalize("a/../b/c/../d"));
+        assertEquals("b/c/d", sut.normalize("a/../b/c/d"));
+        assertEquals("b/c", sut.normalize("a/../b/c"));
+        assertEquals("b/", sut.normalize("a/../b/c/../"));
+        assertEquals("../", sut.normalize("../"));
+    }
```
gnodet, line 42: Minor: the `../` case at line 42 is testing a different behavioral category (leading parent references that cannot be resolved) compared to the null/empty/normal-traversal cases above it. Consider splitting this into its own test method for clarity, similar to how the compat test (`DefaultUrlNormalizerTest` under `compat/maven-model-builder`) has a dedicated `leadingParentDirectoryNotRemovedFromRelativeUriReference` test. Not a blocker, just a readability suggestion.

impl/maven-impl/src/test/java/org/apache/maven/impl/DefaultUrlNormalizerTest.java
```diff
@@ -0,0 +1,69 @@
...
+    void normalizeShouldCollapseParentReferencesInUrl() {
+        assertEquals("https://example.com/child", sut.normalize("https://example.com/parent/../child"));
+        assertEquals("https://example.com/child", sut.normalize("https://example.com/grand/parent/../../child"));
+    }
+
+    @Test
+    void normalizeHandlesDoubleSlashesAfterParent() {
+        assertEquals("https://example.com//child", sut.normalize("https://example.com/parent/..//child"));
+        assertEquals("https://example.com/child", sut.normalize("https://example.com/parent//../child"));
+    }
```
gnodet, line 60: Good idea testing with `file:////` and percent-encoded paths. You might also consider adding the `scm:hg:ssh://localhost//home/user` case from the compat tests, since SCM URLs are a common real-world input for this normalizer. Not a blocker.

gnodet:
Superseded by #12293, which rebases this work on current master (where `UrlNormalizer` was moved from `impl/maven-impl` to `api/maven-api-spi`).

Thank you for the contribution @rbrouwer!

PR #11577, opened 2025-12-18
[3.9.x] MavenPluginJavaPrerequisiteChecker: Handle 8/1.8 Java version in ranges as well

### LINKED ISSUE #11575: Java prerequisites for Maven Plugins fails for Java 8
### Affected version

3.9.12

### Bug description

[WARNING] Ignoring incompatible plugin version 0.4.6: 
The plugin eu.maveniverse.maven.plugins:nisse-plugin3:0.4.6 has unmet prerequisites: 
	Required Java version [8,) is not met by current version: 1.8.0_462

cstamas:
Java is consistent when reporting versions, but users are not. No need to go below 1.8, as Maven 3.9 is Java 8 only. This PR makes "8/1.8" work for ranges as well, not only for standalone version.

Fixes #11575

### DIFF
--- a/maven-core/src/test/java/org/apache/maven/plugin/internal/MavenPluginJavaPrerequisiteCheckerTest.java
+++ b/maven-core/src/test/java/org/apache/maven/plugin/internal/MavenPluginJavaPrerequisiteCheckerTest.java
@@ -19,68 +19,37 @@
 package org.apache.maven.plugin.internal;
 
 import org.junit.Test;
-import org.junit.runner.RunWith;
-import org.junit.runners.Parameterized;
 
-import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertFalse;
 import static org.junit.Assert.assertThrows;
+import static org.junit.Assert.assertTrue;
 
-@RunWith(Parameterized.class)
 public class MavenPluginJavaPrerequisiteCheckerTest {
 
-    private static final String JAVA_8_VERSION = "1.8";
-    private static final String JAVA_9_VERSION = "9.0.1+11";
-
-    @Parameterized.Parameters
-    public static Object[][] parameters() {
-        // requiredVersion, currentVersion, expected, throwable
-        return new Object[][] {
-            {"8", JAVA_8_VERSION, true},
-            {"[8,)", JAVA_8_VERSION, true},
-            {"(,8]", JAVA_8_VERSION, true},
-            {"(,8)", JAVA_8_VERSION, false},
-            {"1.8", JAVA_8_VERSION, true},
-            {"[1.8,)", JAVA_8_VERSION, true},
-            {"(,1.8]", JAVA_8_VERSION, true},
-            {"(,1.8)", JAVA_8_VERSION, false},
-            {"8", JAVA_9_VERSION, true},
-            {"[8,)", JAVA_9_VERSION, true},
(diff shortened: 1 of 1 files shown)


maven-core/src/main/java/org/apache/maven/plugin/internal/MavenPluginJavaPrerequisiteChecker.java
```diff
@@ -47,9 +47,13 @@ public void accept(PluginDescriptor pluginDescriptor) {
     }
 
     boolean matchesVersion(String requiredVersion, String currentVersion) {
+        // Java 8 is consistent: it will always say "1.8...". Users are not consistent.
+        if (!requiredVersion.contains("1.8") && currentVersion.startsWith("1.8")) {
+            currentVersion = currentVersion.substring(2);
+        }
         VersionConstraint constraint;
         try {
-            constraint = versionScheme.parseVersionConstraint(requiredVersion.equals("8") ? "1.8" : requiredVersion);
```
Bukama, line 52: two lines above you substring `currentVersion` here it's about `requiredVersion`. Is it intended or are you accidently mixing variables?

maven-core/src/test/java/org/apache/maven/plugin/internal/MavenPluginJavaPrerequisiteCheckerTest.java
```diff
@@ -30,7 +30,22 @@ public class MavenPluginJavaPrerequisiteCheckerTest {
     public void testMatchesVersion() {
```
Bukama: Use Parameterized tests here

cstamas: Did the change as you asked, but now fails with checkstyle:
```
Error:  src/test/java/org/apache/maven/plugin/internal/MavenPluginJavaPrerequisiteCheckerTest.java:[67,19] (design) VisibilityModifier: Variable 'requiredVersion' must be private and have accessor methods.
Error:  src/test/java/org/apache/maven/plugin/internal/MavenPluginJavaPrerequisiteCheckerTest.java:[70,19] (design) VisibilityModifier: Variable 'currentVersion' must be private and have accessor methods.
Error:  src/test/java/org/apache/maven/plugin/internal/MavenPluginJavaPrerequisiteCheckerTest.java:[73,20] (design) VisibilityModifier: Variable 'expected' must be private and have accessor methods.
```

cstamas: Clearly the two are not aligned: as if I make field private or just non-public, I get
```
Cannot set parameter 'requiredVersion'. Ensure that the field 'requiredVersion' is public.
```
So one want this, other want that.

slawekjaranowski: do it simple, you can try to add suppress for checkstyle, by the way I'm working on move test to JUnit 5 in #11547

kwin:
What about Maven 4? Shouldn't the versions be interpreted the same?

cstamas:
> What about Maven 4? Shouldn't the versions be interpreted the same?

Maven4 is Java 17+, hence it does not need to handle 8/1.8

PR #11579, opened 2025-12-19
Document deprecation of Language.SCRIPT

anukalp2804:
### Summary
Improves documentation for the deprecated `Language.SCRIPT` constant.

### Motivation
`Language.SCRIPT` is deprecated but lacked structured deprecation metadata.
This change clarifies the rationale, adds `since` information, and documents
the recommended replacement.

### Changes
- Added deprecation rationale to Javadoc
- Added `@Deprecated(since = "4.0.0-alpha", forRemoval = false)`
- No behavioral or compatibility changes

Note: This change is documentation-only and does not affect runtime behavior.

### DIFF
--- a/api/maven-api-core/src/main/java/org/apache/maven/api/Language.java
+++ b/api/maven-api-core/src/main/java/org/apache/maven/api/Language.java
@@ -51,7 +51,7 @@ public interface Language extends ExtensibleEnum {
     Language RESOURCES = language("resources");
 
     /**
-     * The {@code "script"} language. This constant is retained for backward compatibility with Maven 3.
+     * The "script" language. This constant is retained for backward compatibility with Maven 3.
      *
      * @deprecated Use {@link #RESOURCES} instead.
      */


anukalp2804:
Gentle ping 🙂
Just checking whether this PR needs anything further from my side.
Happy to update it if needed. Thanks!

api/maven-api-core/src/main/java/org/apache/maven/api/Language.java
```diff
@@ -51,11 +51,14 @@ public interface Language extends ExtensibleEnum {
     Language RESOURCES = language("resources");
 
     /**
-     * The "script" language. Provided for compatibility with Maven 3.
+     * The {@code "script"} language.
+     * <p>
+     * This constant is retained for backward compatibility with Maven 3.
      *
-     * @deprecated Use {@link #RESOURCES} instead.
+     * @deprecated Since Maven 4.0.0-alpha.
```
elharo: do we need Since Maven 4.0.0-alpha. here?

api/maven-api-core/src/main/java/org/apache/maven/api/Language.java
```diff
@@ -51,11 +51,14 @@ public interface Language extends ExtensibleEnum {
...
+     *             Use {@link #RESOURCES} instead.
      */
-    @Deprecated
+    @Deprecated(since = "4.0.0-alpha", forRemoval = false)
```
elharo: I don't know that we want to specify alpha versions in since paramters. 4.0.0 is enough

anukalp2804: @elharo Updated as suggested. I’ve removed the alpha reference and used @Deprecated(since = "4.0.0") without duplicating version information in the Javadoc. Please let me know if anything else needs adjustment.

elharo: It still says alpha,. 4.0.0 is better

elharo (changes requested):

api/maven-api-core/src/main/java/org/apache/maven/api/Language.java
desruisseaux: No need to said "Since Maven 4.0.0-alpha" here, this information is already provided in `@Deprecated`. Instead, we need to said what is the replacement. If the replacement is not `RESOURCES`, then what should it be?

api/maven-api-core/src/main/java/org/apache/maven/api/Language.java
```diff
@@ -51,11 +51,14 @@ public interface Language extends ExtensibleEnum {
     Language RESOURCES = language("resources");
 
     /**
-     * The "script" language. Provided for compatibility with Maven 3.
+     * The {@code "script"} language.
+     * <p>
+     * This constant is retained for backward compatibility with Maven 3.
      *
-     * @deprecated Use {@link #RESOURCES} instead.
+     * @deprecated Since 4.0.0.
```
elharo: Why do you think RESOURCES is the replacement?

anukalp2804: The intent was to document `RESOURCES` as the replacement because `SCRIPT` was historically used for non-Java content in Maven 3, and in Maven 4 such content is handled under the resources language model.

If that assumption is incorrect, I’m happy to update the documentation to point to the correct replacement (or remove the replacement reference entirely if there is no direct equivalent).

desruisseaux: Yes, my understanding is that since script are just source files copied verbatim, they can be handled as ordinary resources. The new `<source>` element has no `<lang>` value specifically for scripts.

desruisseaux: Do we really need "Since 4.0.0" in the comment? It duplicates `@Deprecated(since = "4.0.0")`. The Javadoc tool already puts the content of the latter in the generated documentation, therefore the "Since 4.0.0" comment is really not giving any new information.

elharo: so should it be RESOURCES or should it simply be removed?

api/maven-api-core/src/main/java/org/apache/maven/api/Language.java
```diff
@@ -51,11 +51,13 @@ public interface Language extends ExtensibleEnum {
     Language RESOURCES = language("resources");
 
     /**
-     * The "script" language. Provided for compatibility with Maven 3.
+     * The {@code "script"} language.
+     * <p>
```
desruisseaux: Maybe not worth to put a paragraph separator here. They are two very short sentences. It would help keeping generated javadoc a little bit more compact.

desruisseaux:
I think that redirecting users to `RESOURCES` is fine.

api/maven-api-core/src/main/java/org/apache/maven/api/Language.java
desruisseaux: I think that we can keep `RESOURCES`. Developers who were used to bundle scripts in their application would need to handle them as resources instead.

api/maven-api-core/src/main/java/org/apache/maven/api/Language.java
```diff
@@ -51,11 +51,11 @@ public interface Language extends ExtensibleEnum {
     Language RESOURCES = language("resources");
 
     /**
-     * The "script" language. Provided for compatibility with Maven 3.
+     * The {@code "script"} language. This constant is retained for backward compatibility with Maven 3.
```
desruisseaux: The use of `{@code}` here is inconsistent with the javadoc of other fields.

anukalp2804: Removed the {@code} usage and aligned the wording with the style used by other language constants.

PR #11583, opened 2025-12-21
Use conventional verb method names in XmlNode

### LINKED ISSUE #11582: Noun method names in XmlNode
### Affected version

4.x

### Bug description

For inobvious reasons methods that follow Java conventions like getChild and getValue have been deprecated and replaced with methods that don't like child() and value(). This is confusing and unnecessary churn. Revert this. 

One can argue about the wisdom of Java's well-established naming conventions, but individual projects like Maven should not needlessly deviate from those for personal preference. It's far easier to stick with existing convention than for every project to pave its own way. 

And yes, Oracle did a half-assed job when they added records to the language in Java 17 without considering the impact on naming conventions for existing code. That neither conformed to nor changed existing conventions for non-record classes and interfaces.

elharo:
fixes #11582


slachiewicz:
@elharo we are close to cut maven 4, what is plan for this change?

elharo:
Let me see what I can do about this. It's a mess because of a Modello.

elharo:
ATM I'm stymied by a weird case where org.codehaus.plexus.util.xml.Xpp3Dom depends on XmlNode. That dependency seems to be running the wrong direction. Nothing in plexus should depend on Maven.

slachiewicz:
@elharo In Maven 4, the relationship between Plexus and Maven XML was inverted by design:

* `org.apache.maven.api.xml.XmlNode` is the immutable core API in Maven 4.
* `plexus-xml` **4.x** was created specifically as a backward-compatibility bridge for legacy plugins—`org.codehaus.plexus.util.xml.Xpp3Dom` in `plexus-xml` 4.x wraps and delegates to `XmlNode`.

Because `plexus-xml` is an external dependency (`codehaus-plexus/plexus-xml`), modifying or removing existing method names on `XmlNode` breaks compatibility with the released `plexus-xml 4.x` artifacts.

Given that we are close to cutting the Maven 4 release:
1. We should avoid breaking changes to `XmlNode` at this stage (or only add getter aliases as `default` methods rather than renaming existing ones).
2. Otherwise, this would require coordinating a release of `plexus-xml` before we can integrate changes here.

What do you think about either adding default getter methods alongside the current methods or deferring this cleanup post-4.0?

elharo:
1. XmlNode is marked Experimental.
2. It only gets harder to change after we ship.
3. Plexus is an ugly mess for historical reasons We really shouldn't be depending on it at all, but in the meantime we certainly shouldn't be adding new pieces there. 
4. If plexus-xml 4.x  is only for compatibility the Maven 4 code shouldn't depend on it. That ius something else to fix and yet another  release blocker.

I frankly do not believe Maven 4 is in good enough shape to release. XmlNode is hardly the only problem we need to fix while we can.

elharo:
As a stopgap measure, we might add the getter methods as need and definitely undeprecate them, and we should instead deprecate the no-get methods.

slachiewicz:
Feel free to open changes to plexus-xml 
I've spent there lots of time to cleanup stuff and migrate to mvn4 API.
I'm not sure if I understand this problem so better would be if you can open pr and then I can do quicky release.

elharo:
The problem is that the repo https://github.com/apache/maven depends on plexus-xml which depends on Maven. I.e. there's a cycle between the repos if not the artifacts. This is a problem irrespective of the names of these methods. It might require several releases to untangle this.

gnodet:
Same as #13032 — this change is not desired. The record-accessor naming is intentional for immutable value types in the Maven 4 API.

See https://github.com/apache/maven/issues/11582#issuecomment-5530720249 for the full rationale.

PR #11599, opened 2025-12-23
Document deprecation rationale in DefaultMaven

anukalp2804:
Summary:
This pull request documents the rationale for a deprecated method in DefaultMaven.

Why:
The existing JavaDoc did not explain why the method is deprecated or clarify that
it is intended for internal Maven Core use only, which could confuse external users.

What:
- Replaced informal JavaDoc with a clear explanation of the deprecation rationale
- Clarified intended usage boundaries for the deprecated method

Scope:
- Documentation-only change in Maven Core
- No behavioral or API changes

Testing:
- mvn -DskipTests install (BUILD SUCCESS)

### DIFF
--- a/impl/maven-core/src/main/java/org/apache/maven/DefaultMaven.java
+++ b/impl/maven-core/src/main/java/org/apache/maven/DefaultMaven.java
@@ -402,18 +402,11 @@ public class DefaultMaven implements Maven {
     }
 
     /**
-     * @deprecated since 3.x
-     * <p>
      * This method is part of Maven core internal infrastructure and was never
-     * intended for use outside of Maven Core itself.
-     * </p>
-     * <p>
-     * External consumers should not rely on this method, as it is subject to
-     * change or removal without notice. Code outside of Maven Core should use
-     * supported public APIs instead.
-     * </p>
+     * intended for use outside of Maven Core itself. External consumers should
+     * not rely on this method, as it may change or be removed without notice.
      */
-    @Deprecated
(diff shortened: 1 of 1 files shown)


impl/maven-core/src/main/java/org/apache/maven/DefaultMaven.java
```diff
@@ -402,9 +402,16 @@ private void persistResumptionData(MavenExecutionResult result, MavenSession ses
     }
 
     /**
-     * Nobody should ever use this method.
-     *
-     * @deprecated If you use this method and your code is not in Maven Core, stop doing this.
+     * @deprecated since 3.x
```
desruisseaux: Since Maven 4 requires Java 17, the "since 3.x" in Javadoc should be replaced by `@Deprecated(since = "3.x")` in main code, ideally with "3.x" replaced by the real version when this method was first deprecated. Then, the `<p>` can be removed at least for the first paragraph.

anukalp2804: Thank you for the review @desruisseaux .
I tried tracing the history of `newRepositorySession` through the Maven
module reorganization, but could not clearly identify the original
version where this method was first deprecated.

Could you please confirm which Maven version should be used for the
@Deprecated(since = "...") attribute? Once I have the correct version,
I will update the annotation and clean up the redundant Javadoc tag
accordingly.

desruisseaux (approved):

PR #11609, opened 2025-12-30
Bug: make all collected args use single quote

### LINKED ISSUE #10421: [MNG-8672] Maven 4 CLI substitutes variable when single quoted
**[Alexis Tual](https://issues.apache.org/jira/secure/ViewProfile.jspa?name=atual)** opened **[MNG-8672](https://issues.apache.org/jira/browse/MNG-8672?redirect=false)** and commented

When executing this command line:

 

```java
placeholder=foo mvn clean verify '-DsomeProp=${placeholder}'
```

 

Then we get:
- Maven 3:  `someProp` equals to `${placeholder`}

- Maven 4.0.0-rc-4-SNAPSHOT (5426073a3655278e58f8a4178faae335b1544fc0) :  `someProp` equals to `foo`

The single quote implies the variable should not be substituted, in that sense Maven 3 seems to be the correct behavior.

It's due to this [change](https://github.com/apache/maven/commit/335ba3941c2804704ef11598eb512c305045ab58#diff-7338f8ec18a686d67b8ce118f13744dfd6bb26e92b84c73c8043a946a72283fbR241-R262). 

---

**Affects:** 4.0.0-rc-4

cstamas:
Thus, preventing expansion, if applicable.

Fixes #10421

### DIFF
--- a/apache-maven/src/assembly/maven/bin/mvn
+++ b/apache-maven/src/assembly/maven/bin/mvn
@@ -291,7 +291,7 @@ cmd="\"$JAVACMD\" \
 
 # Add remaining arguments with proper quoting
 for arg in "$@"; do
-    cmd="$cmd \"$arg\""
+    cmd="$cmd '$arg'"
 done
 
 if [ -n "$MAVEN_DEBUG_SCRIPT" ]; then


guylabs:
@cstamas any update on this?

desruisseaux:
It looks fine to me, but I'm far from a bash expert…

cstamas:
@michael-o ping

apache-maven/src/assembly/maven/bin/mvn
```diff
@@ -291,7 +291,7 @@ cmd="\"$JAVACMD\" \
 
 # Add remaining arguments with proper quoting
 for arg in "$@"; do
-    cmd="$cmd \"$arg\""
+    cmd="$cmd '$arg'"
```
michael-o, line 294: This changes semantics `"@"` does already the quoting for us. I think this needs an IT.

michael-o (changes requested):

apache-maven/src/assembly/maven/bin/mvn
cstamas: So, according to your comment, this change is not needed and the issue it fixes is fluke?

michael-o: I didn't say its fluke, but we need to understand what this change causes and an IT with it. I wouldn't easily apply it.

gnodet, line 294: Arguments containing literal single quotes will break here. For example, `-Dprop="it's a value"` produces the `cmd` fragment `'it's a value'`, which `eval` will mispars as three tokens.

The previous approach of passing `'"$@"'` to `eval` was safer because `"$@"` preserves each argument exactly as received by the shell, with no re-quoting needed.

If the goal is to prevent `${...}` expansion during `eval`, the safest fix would be to escape the problematic characters inside double quotes rather than switching to single quotes:

```suggestion
    cmd="$cmd \"$(printf '%s' "$arg" | sed "s/'/'\\\\''/g")\""
```

...though honestly keeping the current `eval exec "$cmd" '"$@"'` approach and investigating why `${...}` was being expanded there would be the better path.

gnodet (commented):

This PR changes how user arguments are passed to the JVM: instead of passing them directly via `"$@"` (which preserves them exactly as the shell received them), it embeds them into the `cmd` string with single quotes and then `eval`s the whole thing.

While this does fix the specific `${placeholder}` expansion issue from MNG-8672, it introduces a **worse problem**: any argument containing a **literal single quote** will break the command. For example:

```
mvn -Dprop="it's a test" clean
```

would produce `cmd="... 'it's a test' 'clean'"` which, after `eval`, misparses due to the unescaped embedded single quote. This is a classic shell-quoting pitfall and would be a regression for users with apostrophes in paths, property values, or profile names.

The previous approach (`eval exec "$cmd" '"$@"'`) was actually correct -- `"$@"` preserves each argument literally, including those containing `${...}`. The real bug reported in MNG-8672 was likely a different issue (possibly related to how `MAVEN_ARGS` or `MAVEN_OPTS` were handled, or an earlier version of the script that used `$*` or did additional expansion).

I agree with @michael-o that this needs an integration test to validate the behavior, and I think the approach itself needs reconsideration -- swapping double-quote quoting for single-quote quoting trades one class of bugs for another, more common one.

gnodet (commented):
This PR is superseded and should be closed.

**1. Single-quote injection bug (high)**
Switching from `"$arg"` to `'$arg'` breaks any CLI argument containing a literal single quote. For example, `mvn -Dprop="it's a test" clean` produces the command fragment `'it's a test'` which `eval` misparses as three tokens. This trades the `${...}` expansion bug for a more common single-quote injection vulnerability.

**2. Superseded by PR #11983 (high)**
The underlying issue (MNG-8672 / `${...}` expansion in CLI arguments) was already correctly fixed by PR #11983 (commit e6a9c4216b, merged 2026-05-19). That fix removed the `for` loop entirely and passes user arguments via `eval exec "$cmd" '"$@"'`, which preserves arguments verbatim without shell re-parsing. The code this PR modifies no longer exists on master.

Issue #10421 (MNG-8672) should also be closed as fixed by #11983 if it is not already.

PR #11611, opened 2025-12-30
Fix typo: 'textOccurencesInLog' -> 'textOccurrencesInLog'

GaoSSR:
Fix typo in method name: 'textOccurencesInLog' -> 'textOccurrencesInLog'

### DIFF
--- a/its/core-it-support/maven-it-helper/src/main/java/org/apache/maven/it/Verifier.java
+++ b/its/core-it-support/maven-it-helper/src/main/java/org/apache/maven/it/Verifier.java
@@ -614,6 +614,14 @@ public class Verifier {
     }
 
     /**
+     * @deprecated Use {@link #textOccurrencesInLog(List, String)} instead
+     */
+    @Deprecated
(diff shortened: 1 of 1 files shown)


its/core-it-support/maven-it-helper/src/main/java/org/apache/maven/it/Verifier.java
```diff
@@ -594,22 +594,22 @@ public void verifyTextNotInLog(String text) throws VerificationException, IOExce
...
         }
     }
 
     public long textOccurrencesInLog(String text) throws IOException {
-        return textOccurencesInLog(loadLogLines(), text);
+        return textOccurrencesInLog(loadLogLines(), text);
     }
 
-    public static long textOccurencesInLog(List<String> lines, String text) {
+    public static long textOccurrencesInLog(List<String> lines, String text) {
```
elharo, line 612: This is a public method. Not sure we can just rename it. Might have to deprecate and replace depending on when it was added

elharo (approved):

PR #11632, opened 2026-01-07
Unify source tracking with SourceHandlingContext (#11612)

### LINKED ISSUE #11612: Unified source handling for all lang/scope combinations with modular sources (Phase 2)
## Summary

A proposal on [Build Sources Validation](https://cwiki.apache.org/confluence/x/84TMFw) (originally started by [mail on the dev list](https://lists.apache.org/thread/blsg4q490dk03qdmol2y5rrckw64lwqm)) was partly implemented as [Phase 1 (Module-aware resource handling)](https://cwiki.apache.org/confluence/x/84TMFw#BuildSourcesValidationDiscussionNotes-Phase1:Module-AwareResourceHandling(Implemented)) by PR #11505.
This issue addresses [Phase 2](https://cwiki.apache.org/confluence/x/84TMFw#BuildSourcesValidationDiscussionNotes-Phase2:ComprehensiveRefactoring): extending the unified handling approach to all `<lang>` and `<scope>` combinations, not just resources.

## Additional Motivation

As noted by @desruisseaux in [PR #11505 comments](https://github.com/apache/maven/pull/11505#discussion_r1899936913):

> I thought that what this PR does for resources, the same thing could be done for tests. I'm thinking to refactor the `ResourceHandler` class in something that does not handle resources in a special way, but instead does the same thing in a uniform way for all combinations of `<lang>` and `<scope>`.

## Challenge: Multiple Source Directories

Allowing multiple source directories (removing R3, see below) raises questions about data structures and handling:

1. **Multiple directories per (lang, scope, module)**: Users may legitimately configure multiple [`<source>`](https://maven.apache.org/ref/4-LATEST/api/maven-api-model/maven.html#class_source) directories for the same combination, e.g.:
   ```xml
   <sources>
     <source>java<scope>main</scope><module>com.example.app</module></source>
     <source>java<scope>main</scope><module>com.example.app</module>
       <directory>target/generated-sources/com.example.app/java</directory>
   </sources>
   ```

2. **Duplicate handling strategy**: The `<source>` element has additional attributes (`enabled`, `targetPath`, `stringFiltering`, `includes`, `excludes`, `targetVersion`). When the same `(lang, scope, module, directory)` tuple appears multiple times, we need a defined handling strategy:
   ```xml
   <source>java<scope>main</scope><module>com.example</module>
     <directory>src/com.example/main/java</directory>
     <enabled>true</enabled>
   <source>java<scope>main</scope><module>com.example</module>
     <directory>src/com.example/main/java</directory>
     <enabled>false</enabled>  
   ```

   **Proposed strategy: `enabled` as key discriminator**
   - Use `(lang, scope, module, directory)` as identity
   - First entry with `enabled=true` wins and defines the effective configuration
   - Subsequent entries with same identity AND `enabled=true` trigger a **WARNING** (alternative: ERROR, since user could simply remove the duplicate)
   - Entries with `enabled=false` are harmless – they explicitly disable that source
   - Duplicate detection applies **only to enabled sources** – disabled sources are effectively no-ops for tracking

   **Note on current implementation:** The current implementation already filters by `enabled` when retrieving sources via [`MavenProject.getEnabledSourceRoots()`](https://github.com/apache/maven/blob/master/impl/maven-core/src/main/java/org/apache/maven/project/MavenProject.java). However, during source processing in `DefaultProjectBuilder`, all sources are added regardless of `enabled` status, and the tracking flags (`hasMain`, `hasTest`, etc.) don't currently consider `enabled`. Phase 2 should align the processing logic with the retrieval logic by only tracking enabled sources.

3. **Target-oriented view**: Regardless of input structure, all sources eventually compile to target directories:
   - `target/classes/<module>/` (main scope)
   - `target/test-classes/<module>/` (test scope)

   The compiler/processor handles multiple input directories naturally – they all feed into the same output location.

The appropriate data structure and algorithm to support this should be discussed further before implementation.

### 1. Generalize `ResourceHandlingContext` to `SourceHandlingContext`

Replace the current [`ResourceHandlingContext`](https://github.com/apache/maven/blob/master/impl/maven-core/src/main/java/org/apache/maven/project/ResourceHandlingContext.java) approach with a flexible, unified source handling mechanism.

**Current (Phase 1) - extended boolean flags** (Pseudo-code):
```java
// Hardcoded tracking for specific combinations
boolean hasMain = false;
boolean hasTest = false;
boolean hasMainResources = false;
boolean hasTestResources = false;

for (var source : sources) {
    if (lang == JAVA && scope == MAIN) hasMain = true;
    if (lang == JAVA && scope == TEST) hasTest = true;
    if (lang == RESOURCES && scope == MAIN) hasMainResources = true;
    if (lang == RESOURCES && scope == TEST) hasTestResources = true;
}

// Resource-specific handling
context.handleResourceConfiguration(MAIN, hasMainResources);
context.handleResourceConfiguration(TEST, hasTestResources);
```

**Proposed (Phase 2) – Flexible set-based tracking:**
```java
// Generic tracking for ANY lang/scope/module/directory combination
record SourceKey(Language lang, ProjectScope scope, String module, Path directory) {}
Set<SourceKey> declaredSources = new HashSet<>();

// We need to handle some special cases, e.g., implicit directory etc.
for (var source : sources) {
    SourceKey key = new SourceKey(lang, scope, module, directory);
    if (source.isEnabled()) {
        if (declaredSources.contains(key)) {
            // Duplicate enabled source - emit WARNING (or ERROR)
            warn("Duplicate enabled source for " + key);
        } else {
            declaredSources.add(key);
        }
    }
    // disabled sources are not tracked - they are explicitly disabled by the user
    // and won't be returned by getEnabledSourceRoots() anyway
}

// Unified handling for ALL source types (java, resources, kotlin, etc.)
for (SourceKey key : declaredSources) {
    // Same logic for java, resources, kotlin, groovy, etc.
    handleSourceConfiguration(key.lang(), key.scope(), key.module(), key.directory());
}
```

**Benefits:**
- Replace hardcoded boolean flags with a flexible set structure
- Generalize handling logic (currently resource-only) to work for ALL lang/scope combinations
- Use ONE unified approach instead of special-casing resources
- Easily extensible for new languages (kotlin, groovy, scala, etc.)

### 2. Remove Validation Rule R3 (Duplicate Sources)

Per @desruisseaux suggestion in [his mailing list reply](https://lists.apache.org/thread/r1vqttptg62tqlt96vtrm605xyxjj095):

> Remove rule R3, i.e., accept duplicated source for `scope='X'`, `lang='Y'`, `module='Z'`. The compiler plugin 4 already accepts that, the need for multi-source directories in Maven 3 was strong enough to justify an external plugin, and Maven 3 already accepts that for resources if I remember right.

**Rationale:**
- Maven Compiler Plugin 4 already supports multiple source directories
- The `build-helper-maven-plugin` exists specifically because users needed multi-source support
- Maven 3 already allows duplicate resource directories

### 3. Unified Source Directory Handling

Apply the same modular source detection logic from [Phase 1](https://cwiki.apache.org/confluence/x/84TMFw#BuildSourcesValidationDiscussionNotes-Phase1:Module-AwareResourceHandling(Implemented)) to:
- Java sources (`src/<module>/main/java`, `src/<module>/test/java`)
- Test sources
- Other language sources (kotlin, groovy, scala, etc.)

### Files to Modify

- `impl/maven-core/src/main/java/org/apache/maven/project/ResourceHandlingContext.java`
  - Rename/refactor to `SourceHandlingContext`
  - Generalize `handleResourceConfiguration()` to handle all source types
  - Add `shouldAddSource()` for duplicate detection
  - Add `hasSources(Language, ProjectScope)` to replace boolean flags

- `impl/maven-core/src/main/java/org/apache/maven/project/DefaultProjectBuilder.java`
  - Replace hardcoded boolean flags (`hasMain`, `hasTest`, etc.) with `SourceHandlingContext.hasSources()`
  - Use `SourceHandlingContext.shouldAddSource()` for duplicate detection

### No modification necessary (due to dropped R3)

- `impl/maven-impl/src/main/java/org/apache/maven/impl/model/DefaultModelValidator.java`
  - R3 validation (duplicate sources detection) was originally proposed but **never implemented** in the validator
  - Duplicate detection is handled in `SourceHandlingContext.shouldAddSource()` instead
  - R1, R2, R4 validations remain unchanged

### Validation Rules After Phase 2

| Rule   | Description                                             | Addresses                                                                                                                                                | Status     |
|--------|---------------------------------------------------------|----------------------------------------------------------------------------------------------------------------------------------------------------------|------------|
| R1     | `<resources>` conflicts with explicit `<source>` | -                                                                                                                                                        | Keep       |
| R2     | Mixed modular/non-modular sources                       | [Problem 1](https://cwiki.apache.org/confluence/x/84TMFw#BuildSourcesValidationDiscussionNotes-Problem1:MixedModular/Non-ModularSourcesAcceptedSilently) | Keep       |
| ~~R3~~ | ~~Duplicate sources warning~~                           | -                                                                                                                                                        | **Remove** |
| R4     | Filesystem warning for classic directory                | [Problem 4](https://cwiki.apache.org/confluence/x/84TMFw#BuildSourcesValidationDiscussionNotes-Problem4:LegacyConfigurationSilentlyIgnored)              | Keep       |

**Note:** [Problem 2 (Test-Only Modular Projects)](https://cwiki.apache.org/confluence/x/84TMFw#BuildSourcesValidationDiscussionNotes-Problem2:Test-OnlyModularProjectsGetClassicMainSourcesInjected) and [Problem 3 (Resources Are Not Module-Aware)](https://cwiki.apache.org/confluence/x/84TMFw#BuildSourcesValidationDiscussionNotes-Problem3:ResourcesAreNotModule-Aware) were addressed by source injection logic in Phase 1.

## Acceptance Criteria

| #   | Criterion                  | Description                                                                                                                                                                          |
|-----|----------------------------|--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
| AC1 | Boolean flags eliminated   | The hardcoded `hasMain`, `hasTest`, `hasMainResources`, `hasTestResources`, `hasScript` flags are replaced with flexible set-based tracking via `SourceHandlingContext.hasSources()` |
| AC2 | Unified source tracking    | All lang/scope combinations use the same tracking mechanism (not just resources)                                                                                                     |
| AC3 | Duplicate detection        | Duplicate enabled sources with same `(lang, scope, module, directory)` identity trigger a WARNING                                                                                    |
| AC4 | First enabled wins         | Only the first enabled source for an identity is added to the project; duplicates are skipped                                                                                        |
| AC5 | Disabled sources unchanged | Disabled sources are still added but filtered by `getEnabledSourceRoots()`                                                                                                           |
| AC6 | Mixed sources error        | Mixing modular (with module attribute) and classic (without module attribute) sources within `<sources>` triggers an ERROR                                                           |
| AC7 | Legacy directories warning | In modular projects, `<sourceDirectory>` and `<testSourceDirectory>` are unconditionally ignored with a WARNING (both for explicit config and existing default directories)         |

ascheman:
## Summary

- Replace boolean flags (`hasMain`, `hasTest`, `hasMainResources`, etc.) with flexible set-based tracking using `SourceHandlingContext`
- Rename `ResourceHandlingContext` to `SourceHandlingContext` to reflect unified handling of all source types
- Add duplicate detection: first enabled source wins, duplicates trigger WARNING
- Add `hasSources(Language, ProjectScope)` method for checking source existence
- Replace `Collectors.toList()` with `.toList()` (Java 17+ idiom)

Closes #11612 (Phase 2: build-sources-validation)

## Test plan

- [x] `ProjectBuilderTest` passes (19 tests)
- [x] New tests added:
  - `testModularSourcesInjectResourceRoots` - verifies module-aware resource injection
  - `testModularSourcesWithExplicitResourcesIssuesWarning` - verifies legacy resource warnings
  - `testMixedSourcesModularMainClassicTest` - mixed modular/classic configuration
  - `testSourcesMixedModulesWithinSources` - mixed modules within `<sources>`
  - `testMultipleDirectoriesSameModule` - multiple directories per module
  - `testDuplicateEnabledSources` - duplicate detection with WARNING

### DIFF
--- a/impl/maven-core/src/main/java/org/apache/maven/project/DefaultProjectBuilder.java
+++ b/impl/maven-core/src/main/java/org/apache/maven/project/DefaultProjectBuilder.java
@@ -1089,7 +1089,7 @@ public class DefaultProjectBuilder implements ProjectBuilder {
         }
     }
 
-    private List<String> getProfileIds(List<Profile> profiles) {
+    private static List<String> getProfileIds(List<Profile> profiles) {
         return profiles.stream().map(Profile::getId).toList();
     }
 
--- a/impl/maven-core/src/test/java/org/apache/maven/project/ProjectBuilderTest.java
+++ b/impl/maven-core/src/test/java/org/apache/maven/project/ProjectBuilderTest.java
@@ -25,6 +25,7 @@ import java.nio.file.Path;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.List;
+import java.util.Optional;
 import java.util.Properties;
 import java.util.Set;
 import java.util.concurrent.atomic.AtomicInteger;
@@ -398,8 +399,7 @@ class ProjectBuilderTest extends AbstractCoreMavenComponentTestCase {
         // Should have resource roots for both modules
         Set<String> modules = mainResourceRoots.stream()
                 .map(SourceRoot::module)
-                .filter(opt -> opt.isPresent())
-                .map(opt -> opt.get())
+                .flatMap(Optional::stream)
                 .collect(Collectors.toSet());
 
         assertEquals(2, modules.size(), "Should have resource roots for 2 modules");
@@ -413,8 +413,7 @@ class ProjectBuilderTest extends AbstractCoreMavenComponentTestCase {
         // Should have test resource roots for both modules
         Set<String> testModules = testResourceRoots.stream()
                 .map(SourceRoot::module)
-                .filter(opt -> opt.isPresent())
-                .map(opt -> opt.get())
+                .flatMap(Optional::stream)
                 .collect(Collectors.toSet());
 
         assertEquals(2, testModules.size(), "Should have test resource roots for 2 modules");
@@ -470,8 +469,7 @@ class ProjectBuilderTest extends AbstractCoreMavenComponentTestCase {
 
         Set<String> mainModules = mainResourceRoots.stream()
                 .map(SourceRoot::module)
-                .filter(opt -> opt.isPresent())
-                .map(opt -> opt.get())
+                .flatMap(Optional::stream)
                 .collect(Collectors.toSet());
 
         assertEquals(2, mainModules.size(), "Should have resource roots for 2 modules");
@@ -536,8 +534,7 @@ class ProjectBuilderTest extends AbstractCoreMavenComponentTestCase {
 
         Set<String> mainModules = mainJavaRoots.stream()
                 .map(SourceRoot::module)
-                .filter(opt -> opt.isPresent())
-                .map(opt -> opt.get())
+                .flatMap(Optional::stream)
                 .collect(Collectors.toSet());
 
         assertEquals(2, mainModules.size(), "Should have main sources for 2 modules");
@@ -545,8 +542,10 @@ class ProjectBuilderTest extends AbstractCoreMavenComponentTestCase {
         assertTrue(mainModules.contains("org.foo.moduleB"), "Should have main source for moduleB");
 
         // Verify the classic sourceDirectory is NOT used (should be ignored)
-        boolean hasClassicMainSource = mainJavaRoots.stream()
-                .anyMatch(sr -> sr.directory().toString().replace('\\', '/').contains("src/classic/main/java"));
+        boolean hasClassicMainSource = mainJavaRoots.stream().anyMatch(sr -> sr.directory()
+                .toString()
+                .replace(File.separatorChar, '/')
+                .contains("src/classic/main/java"));
         assertTrue(!hasClassicMainSource, "Classic sourceDirectory should be ignored");
 
         // Test sources should NOT be added (legacy testSourceDirectory is ignored in modular projects)
@@ -623,17 +622,18 @@ class ProjectBuilderTest extends AbstractCoreMavenComponentTestCase {
 
         // Both should be for the same module
         long moduleCount = mainJavaRoots.stream()
-                .filter(sr -> sr.module().isPresent()
-                        && "com.example.app".equals(sr.module().get()))
+                .filter(sr -> "com.example.app".equals(sr.module().orElse(null)))
                 .count();
         assertEquals(2, moduleCount, "Both main sources should be for com.example.app module");
 
         // One should be implicit directory, one should be generated-sources
-        boolean hasImplicitDir = mainJavaRoots.stream()
-                .anyMatch(sr -> sr.directory().toString().replace('\\', '/').contains("src/com.example.app/main/java"));
+        boolean hasImplicitDir = mainJavaRoots.stream().anyMatch(sr -> sr.directory()
+                .toString()
+                .replace(File.separatorChar, '/')
+                .contains("src/com.example.app/main/java"));
         boolean hasGeneratedDir = mainJavaRoots.stream().anyMatch(sr -> sr.directory()
                 .toString()
-                .replace('\\', '/')
+                .replace(File.separatorChar, '/')
                 .contains("target/generated-sources/com.example.app/java"));
 
         assertTrue(hasImplicitDir, "Should have implicit source directory for module");
@@ -648,8 +648,7 @@ class ProjectBuilderTest extends AbstractCoreMavenComponentTestCase {
 
         // Both test sources should be for the same module
         long testModuleCount = testJavaRoots.stream()
-                .filter(sr -> sr.module().isPresent()
-                        && "com.example.app".equals(sr.module().get()))
+                .filter(sr -> "com.example.app".equals(sr.module().orElse(null)))
                 .count();
         assertEquals(2, testModuleCount, "Both test sources should be for com.example.app module");
     }
@@ -707,14 +706,12 @@ class ProjectBuilderTest extends AbstractCoreMavenComponentTestCase {
 
         // Verify com.example.other module is present
         boolean hasOtherModule = mainJavaRoots.stream()
-                .anyMatch(sr -> sr.module().isPresent()
-                        && "com.example.other".equals(sr.module().get()));
+                .anyMatch(sr -> "com.example.other".equals(sr.module().orElse(null)));
         assertTrue(hasOtherModule, "Should have source root for com.example.other module");
 
         // Verify com.example.dup module is present (first enabled wins)
         boolean hasDupModule = mainJavaRoots.stream()
-                .anyMatch(sr -> sr.module().isPresent()
-                        && "com.example.dup".equals(sr.module().get()));
+                .anyMatch(sr -> "com.example.dup".equals(sr.module().orElse(null)));
         assertTrue(hasDupModule, "Should have source root for com.example.dup module");
 
         // Get test Java source roots
@@ -725,10 +722,9 @@ class ProjectBuilderTest extends AbstractCoreMavenComponentTestCase {
         assertEquals(1, testJavaRoots.size(), "Should have 1 test Java source root");
 
         // Verify it's for the dup module
-        assertTrue(
-                testJavaRoots.get(0).module().isPresent()
-                        && "com.example.dup"
-                                .equals(testJavaRoots.get(0).module().get()),
+        assertEquals(
+                "com.example.dup",
+                testJavaRoots.get(0).module().orElse(null),
                 "Test source root should be for com.example.dup module");
     }
 }


impl/maven-core/src/main/java/org/apache/maven/project/SourceHandlingContext.java
```diff
@@ -19,34 +19,55 @@
...
+     * Identity key for source tracking. Two sources with the same key are considered duplicates.
+     */
+    record SourceKey(Language language, ProjectScope scope, String module, Path directory) {}
 
     private final MavenProject project;
     private final Path baseDir;
     private final Set<String> modules;
     private final boolean modularProject;
     private final ModelBuilderResult result;
+    private final Set<SourceKey> declaredSources = new HashSet<>();
```
desruisseaux: It is a very minor details, but moving this initialization inside the constructor would allow to specify an initial capacity as `new HashSet<>(4 * modules.size())` (assuming that each module is likely to have a `main`, a `test`, main resources and test resources).

impl/maven-core/src/main/java/org/apache/maven/project/SourceHandlingContext.java
```diff
@@ -59,6 +80,73 @@ class ResourceHandlingContext {
...
+     *   <li>Subsequent enabled sources with same identity trigger a WARNING and are NOT added</li>
+     * </ul>
+     *
+     * @param sourceRoot the source root to evaluate
+     * @return true if the source should be added to the project, false if it's a duplicate enabled source
+     */
+    boolean shouldAddSource(SourceRoot sourceRoot) {
+        if (!sourceRoot.enabled()) {
+            // Disabled sources are always added - they're filtered out by getEnabledSourceRoots()
+            LOGGER.debug(
```
desruisseaux: Should this one be at the trace level?

impl/maven-core/src/main/java/org/apache/maven/project/SourceHandlingContext.java
```diff
@@ -59,6 +80,73 @@ class ResourceHandlingContext {
...
+                    "Adding disabled source (will be filtered by getEnabledSourceRoots): lang={}, scope={}, module={}, dir={}",
+                    sourceRoot.language(),
+                    sourceRoot.scope(),
+                    sourceRoot.module().orElse(null),
+                    sourceRoot.directory());
+            return true;
+        }
+
+        SourceKey key = new SourceKey(
+                sourceRoot.language(), sourceRoot.scope(), sourceRoot.module().orElse(null), sourceRoot.directory());
```
desruisseaux: We may consider adding `.toRealPath()` after `directory()` for making the test more exhaustive.

impl/maven-core/src/main/java/org/apache/maven/project/SourceHandlingContext.java
```diff
@@ -59,6 +80,73 @@ class ResourceHandlingContext {
...
+
+        if (declaredSources.contains(key)) {
+            String message = String.format(
+                    "Duplicate enabled source detected: lang=%s, scope=%s, module=%s, directory=%s. "
+                            + "First enabled source wins, this duplicate is ignored.",
+                    key.language(), key.scope(), key.module() != null ? key.module() : "(none)", key.directory());
+            LOGGER.warn(message);
```
desruisseaux, line 120: I wonder if it should be an error (throwing an exception) instead of only a warning. I'm not sure about that, just thinking.

impl/maven-core/src/main/java/org/apache/maven/project/SourceHandlingContext.java
```diff
@@ -59,6 +80,73 @@ class ResourceHandlingContext {
...
+    boolean shouldAddSource(SourceRoot sourceRoot) {
+        if (!sourceRoot.enabled()) {
+            // Disabled sources are always added - they're filtered out by getEnabledSourceRoots()
+            LOGGER.debug(
+                    "Adding disabled source (will be filtered by getEnabledSourceRoots): lang={}, scope={}, module={}, dir={}",
+                    sourceRoot.language(),
+                    sourceRoot.scope(),
+                    sourceRoot.module().orElse(null),
+                    sourceRoot.directory());
+            return true;
```
desruisseaux, line 107: Shouldn't it be `return false`?

desruisseaux: Sorry, my mistake. I just saw the comment saying that disabled source are not filtered in this method.

impl/maven-core/src/main/java/org/apache/maven/project/SourceHandlingContext.java
desruisseaux: Forget this question. I guess that warning level is just fine.

impl/maven-core/src/test/projects/project-builder/mixed-sources/pom.xml
```diff
@@ -0,0 +1,38 @@
...
+                <scope>main</scope>
+                <lang>java</lang>
+                <module>org.foo.moduleA</module>
+            </source>
+            <source>
+                <scope>main</scope>
+                <lang>java</lang>
+                <module>org.foo.moduleB</module>
+            </source>
+            <!-- No test sources defined - testSourceDirectory should be used -->
```
desruisseaux: This is ambiguous: how could a single `testSourceDirectory` be used for two modules? I suggest to allow `<sourceDirectory>` and `<testSourceDirectory>` only for non-modular projects, and emit a warning or error for modular projects.

impl/maven-core/src/test/java/org/apache/maven/project/ProjectBuilderTest.java
```diff
@@ -470,4 +478,255 @@ void testModularSourcesWithExplicitResourcesIssuesWarning() throws Exception {
         assertTrue(mainModules.contains("org.foo.moduleA"), "Should have resource root for moduleA");
         assertTrue(mainModules.contains("org.foo.moduleB"), "Should have resource root for moduleB");
     }
+
+    /**
+     * Tests mixed source configuration where:
+     * - Modular sources are defined for main Java (should override sourceDirectory)
+     * - Classic testSourceDirectory is used (should be preserved since no modular test sources)
```
desruisseaux: See comment in `mixed-sources`: this is ambiguous if we have more than one module. I suggest to modify the test for verifying that `<testSourceDirectory>` has been ignored and a warning emitted.

Instead (maybe in a phase 3), for each module declared in a `<source>` with `main` scope but no `test` scope, generate automatically the `<source>` with `test` scope in a way similar to what is currently done for resources.

impl/maven-core/src/test/projects/project-builder/sources-mixed-modules/pom.xml
```diff
@@ -0,0 +1,47 @@
...
+        <sourceDirectory>src/should-be-ignored/java</sourceDirectory>
+
+        <sources>
+            <!-- Modular source -->
+            <source>
+                <scope>main</scope>
+                <lang>java</lang>
+                <module>org.foo.moduleA</module>
+            </source>
+            <!-- Non-modular source (no module attribute) -->
```
desruisseaux: Should not be allowed. Either a project is fully modular or fully classic, but not a mix of both. The compiler plugin will not be able to handle that.

impl/maven-core/src/test/java/org/apache/maven/project/ProjectBuilderTest.java
```diff
@@ -470,4 +478,255 @@ void testModularSourcesWithExplicitResourcesIssuesWarning() throws Exception {
...
+        // The test source should be the classic one (no module)
+        SourceRoot testRoot = testJavaRoots.get(0);
+        assertTrue(testRoot.module().isEmpty(), "Classic test source should not have a module");
+        assertTrue(
+                testRoot.directory().toString().replace('\\', '/').contains("src/classic/test/java"),
+                "Should use classic testSourceDirectory");
+    }
+
+    /**
+     * Tests mixed modular/non-modular sources within the same {@code <sources>} element.
```
desruisseaux: See comment in `sources-mixed-modules`: a project can be fully modular or fully classic, but not a mix of both. An error should be emitted. In this case, I think that it should be an error rather than a warning, because the compiler plugin will not know what to do with such mixed project.

impl/maven-core/src/main/java/org/apache/maven/project/SourceHandlingContext.java
ascheman: Done in commit 63d43d5154. Moved initialization to constructor with initial capacity.

impl/maven-core/src/main/java/org/apache/maven/project/SourceHandlingContext.java
ascheman: Done in commit 63d43d5154. Changed from DEBUG to TRACE level.

impl/maven-core/src/main/java/org/apache/maven/project/SourceHandlingContext.java
ascheman: Used `.toAbsolutePath().normalize()` instead in commit 63d43d5154. 

`toRealPath()` requires the file to exist on the filesystem and throws IOException if it doesn't, which would break duplicate detection for source directories that don't exist yet. The `normalize()` approach handles relative paths and `..` components without requiring filesystem access.

impl/maven-core/src/test/projects/project-builder/mixed-sources/pom.xml
ascheman: Implemented in commit 63d43d5154 as AC7: In modular projects, `<sourceDirectory>` and `<testSourceDirectory>` are now unconditionally ignored with a WARNING.

The test has been updated to verify this behavior - both legacy directories trigger warnings and are not used.

impl/maven-core/src/test/java/org/apache/maven/project/ProjectBuilderTest.java
ascheman: Done in commit 63d43d5154. The test `testMixedSourcesModularMainClassicTest()` now verifies:
- 2 warnings are emitted (one for `<sourceDirectory>`, one for `<testSourceDirectory>`)
- Test sources are NOT added (0 test Java roots) because legacy `<testSourceDirectory>` is ignored

impl/maven-core/src/test/projects/project-builder/sources-mixed-modules/pom.xml
ascheman: Implemented in commit 63d43d5154 as AC6: Mixing modular (with module attribute) and classic (without module attribute) sources within `<sources>` now triggers an ERROR.

The test project comments and test method have been updated accordingly.

impl/maven-core/src/test/java/org/apache/maven/project/ProjectBuilderTest.java
ascheman: Done in commit 63d43d5154. The `validateNoMixedModularAndClassicSources()` method now reports an ERROR (not a warning) for mixed configurations, as suggested.

The test `testSourcesMixedModulesWithinSources()` verifies that an ERROR is reported with the message "Mixed modular and classic sources detected".

impl/maven-core/src/test/java/org/apache/maven/project/ProjectBuilderTest.java
```diff
@@ -470,4 +478,257 @@ void testModularSourcesWithExplicitResourcesIssuesWarning() throws Exception {
...
+        // Test scope has 1 source for com.example.dup (first wins)
+        assertEquals(1, testJavaRoots.size(), "Should have 1 test Java source root");
+
+        // Verify it's for the dup module
+        assertTrue(
+                testJavaRoots.get(0).module().isPresent()
+                        && "com.example.dup"
+                                .equals(testJavaRoots.get(0).module().get()),
+                "Test source root should be for com.example.dup module");
+    }
```
desruisseaux, line 729: The `assertTrue` expression can be simplified as below (removing the `isPresent()` check, replaced by the fact that `String.equals(Object)` is null-safe):

```java
        assertEquals(
                "com.example.dup",
                testJavaRoots.get(0).module().orElse(null),
```

impl/maven-core/src/test/java/org/apache/maven/project/ProjectBuilderTest.java
```diff
@@ -470,4 +478,257 @@ void testModularSourcesWithExplicitResourcesIssuesWarning() throws Exception {
...
+
+        // Verify com.example.other module is present
+        boolean hasOtherModule = mainJavaRoots.stream()
+                .anyMatch(sr -> sr.module().isPresent()
+                        && "com.example.other".equals(sr.module().get()));
+        assertTrue(hasOtherModule, "Should have source root for com.example.other module");
+
+        // Verify com.example.dup module is present (first enabled wins)
+        boolean hasDupModule = mainJavaRoots.stream()
+                .anyMatch(sr -> sr.module().isPresent()
```
desruisseaux: Can be simplified as below (removing the `isPresent()` check, replaced by the fact that `String.equals(Object)` is null-safe):

```java
                .anyMatch(sr -> "com.example.dup".equals(sr.module().orElse(null)));
```

impl/maven-core/src/test/java/org/apache/maven/project/ProjectBuilderTest.java
```diff
@@ -470,4 +478,257 @@ void testModularSourcesWithExplicitResourcesIssuesWarning() throws Exception {
...
+        List<SourceRoot> mainJavaRoots = project.getEnabledSourceRoots(ProjectScope.MAIN, Language.JAVA_FAMILY)
+                .toList();
+
+        // Should have 2 main sources: 1 for com.example.dup (first wins) + 1 for com.example.other
+        // Note: MavenProject.addSourceRoot still adds all sources, but tracking only counts first enabled
+        assertEquals(2, mainJavaRoots.size(), "Should have 2 main Java source roots");
+
+        // Verify com.example.other module is present
+        boolean hasOtherModule = mainJavaRoots.stream()
+                .anyMatch(sr -> sr.module().isPresent()
```
desruisseaux: Can be simplified as below (removing the `isPresent()` check, replaced by the fact that `String.equals(Object)` is null-safe):

```java
                .anyMatch(sr -> "com.example.other".equals(sr.module().orElse(null)));
```

impl/maven-core/src/test/java/org/apache/maven/project/ProjectBuilderTest.java
```diff
@@ -470,4 +478,257 @@ void testModularSourcesWithExplicitResourcesIssuesWarning() throws Exception {
...
+        // Get test Java source roots
+        List<SourceRoot> testJavaRoots = project.getEnabledSourceRoots(ProjectScope.TEST, Language.JAVA_FAMILY)
+                .toList();
+
+        // Should have 2 test sources: both for com.example.app
+        assertEquals(2, testJavaRoots.size(), "Should have 2 test Java source roots for same module");
+
+        // Both test sources should be for the same module
+        long testModuleCount = testJavaRoots.stream()
+                .filter(sr -> sr.module().isPresent()
```
desruisseaux: Can be simplified as below (removing the `isPresent()` check, replaced by the fact that `String.equals(Object)` is null-safe):

```java
                .filter(sr -> "com.example.app".equals(sr.module().orElse(null)))
```

impl/maven-core/src/test/java/org/apache/maven/project/ProjectBuilderTest.java
```diff
@@ -470,4 +478,257 @@ void testModularSourcesWithExplicitResourcesIssuesWarning() throws Exception {
...
+                        && "com.example.app".equals(sr.module().get()))
+                .count();
+        assertEquals(2, moduleCount, "Both main sources should be for com.example.app module");
+
+        // One should be implicit directory, one should be generated-sources
+        boolean hasImplicitDir = mainJavaRoots.stream()
+                .anyMatch(sr -> sr.directory().toString().replace('\\', '/').contains("src/com.example.app/main/java"));
+        boolean hasGeneratedDir = mainJavaRoots.stream().anyMatch(sr -> sr.directory()
+                .toString()
+                .replace('\\', '/')
```
desruisseaux: I suggest `replace(File.separatorChar, '/')`.

impl/maven-core/src/test/java/org/apache/maven/project/ProjectBuilderTest.java
```diff
@@ -470,4 +478,257 @@ void testModularSourcesWithExplicitResourcesIssuesWarning() throws Exception {
...
+        // Both should be for the same module
+        long moduleCount = mainJavaRoots.stream()
+                .filter(sr -> sr.module().isPresent()
+                        && "com.example.app".equals(sr.module().get()))
+                .count();
+        assertEquals(2, moduleCount, "Both main sources should be for com.example.app module");
+
+        // One should be implicit directory, one should be generated-sources
+        boolean hasImplicitDir = mainJavaRoots.stream()
+                .anyMatch(sr -> sr.directory().toString().replace('\\', '/').contains("src/com.example.app/main/java"));
```
desruisseaux: I suggest `replace(File.separatorChar, '/')` instead of `replace('\\', '/')`.

impl/maven-core/src/test/java/org/apache/maven/project/ProjectBuilderTest.java
```diff
@@ -470,4 +478,257 @@ void testModularSourcesWithExplicitResourcesIssuesWarning() throws Exception {
...
+        // Get main Java source roots
+        List<SourceRoot> mainJavaRoots = project.getEnabledSourceRoots(ProjectScope.MAIN, Language.JAVA_FAMILY)
+                .toList();
+
+        // Should have 2 main sources: both for com.example.app but different directories
+        assertEquals(2, mainJavaRoots.size(), "Should have 2 main Java source roots for same module");
+
+        // Both should be for the same module
+        long moduleCount = mainJavaRoots.stream()
+                .filter(sr -> sr.module().isPresent()
```
desruisseaux: Can be simplified as below (removing the `isPresent()` check, replaced by the fact that `String.equals(Object)` is null-safe):

```java
                .filter(sr -> "com.example.app".equals(sr.module().orElse(null)))
```

impl/maven-core/src/test/java/org/apache/maven/project/ProjectBuilderTest.java
```diff
@@ -470,4 +478,257 @@ void testModularSourcesWithExplicitResourcesIssuesWarning() throws Exception {
...
+                .map(opt -> opt.get())
+                .collect(Collectors.toSet());
+
+        assertEquals(2, mainModules.size(), "Should have main sources for 2 modules");
+        assertTrue(mainModules.contains("org.foo.moduleA"), "Should have main source for moduleA");
+        assertTrue(mainModules.contains("org.foo.moduleB"), "Should have main source for moduleB");
+
+        // Verify the classic sourceDirectory is NOT used (should be ignored)
+        boolean hasClassicMainSource = mainJavaRoots.stream()
+                .anyMatch(sr -> sr.directory().toString().replace('\\', '/').contains("src/classic/main/java"));
```
desruisseaux: I suggest `replace(File.separatorChar, '/')` instead of `replace('\\', '/')`.

impl/maven-core/src/test/java/org/apache/maven/project/ProjectBuilderTest.java
```diff
@@ -470,4 +478,257 @@ void testModularSourcesWithExplicitResourcesIssuesWarning() throws Exception {
...
+        List<SourceRoot> mainJavaRoots = project.getEnabledSourceRoots(ProjectScope.MAIN, Language.JAVA_FAMILY)
+                .toList();
+
+        // Should have 2 modular main Java sources (moduleA and moduleB)
+        assertEquals(2, mainJavaRoots.size(), "Should have 2 modular main Java source roots");
+
+        Set<String> mainModules = mainJavaRoots.stream()
+                .map(SourceRoot::module)
+                .filter(opt -> opt.isPresent())
+                .map(opt -> opt.get())
```
desruisseaux: The two lines `filter` followed by `map` can be replaced by this single line:

```java
                .flatMap(Optional::stream)
```

impl/maven-core/src/test/java/org/apache/maven/project/ProjectBuilderTest.java
desruisseaux: The two lines `filter` followed by `map` can be replaced by this single line:

```java
                .flatMap(Optional::stream)
```

impl/maven-core/src/main/java/org/apache/maven/project/DefaultProjectBuilder.java
```diff
@@ -1036,7 +1090,7 @@ private DependencyResolutionResult resolveDependencies(MavenProject project) {
     }
 
     private List<String> getProfileIds(List<Profile> profiles) {
```
desruisseaux: Method can be `static`.

ascheman: Done in commit 08fa756ce2. Made `getProfileIds()` static.

impl/maven-core/src/test/java/org/apache/maven/project/ProjectBuilderTest.java
ascheman: Done in commit 08fa756ce2. Applied `flatMap(Optional::stream)` to all occurrences in the test class.

impl/maven-core/src/test/java/org/apache/maven/project/ProjectBuilderTest.java
ascheman: Done in commit 08fa756ce2.

impl/maven-core/src/test/java/org/apache/maven/project/ProjectBuilderTest.java
ascheman: Done in commit 08fa756ce2. Applied `replace(File.separatorChar, '/')` to all occurrences.

impl/maven-core/src/test/java/org/apache/maven/project/ProjectBuilderTest.java
ascheman: Done in commit 08fa756ce2.

impl/maven-core/src/test/java/org/apache/maven/project/ProjectBuilderTest.java
ascheman: Done in commit 08fa756ce2.

impl/maven-core/src/test/java/org/apache/maven/project/ProjectBuilderTest.java
ascheman: Done in commit 08fa756ce2. Applied `orElse(null)` simplification to all module checks.

impl/maven-core/src/test/java/org/apache/maven/project/ProjectBuilderTest.java
ascheman: Done in commit 08fa756ce2.

impl/maven-core/src/test/java/org/apache/maven/project/ProjectBuilderTest.java
ascheman: Done in commit 08fa756ce2.

impl/maven-core/src/test/java/org/apache/maven/project/ProjectBuilderTest.java
ascheman: Done in commit 08fa756ce2.

impl/maven-core/src/test/java/org/apache/maven/project/ProjectBuilderTest.java
ascheman: Done in commit 08fa756ce2. Replaced `assertTrue` with `assertEquals`.

PR #11637, opened 2026-01-09
Maven Parent 46

cstamas:
w/ some slight updates

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -807,39 +807,30 @@ under the License.
         <plugin>
           <groupId>org.apache.rat</groupId>
           <artifactId>apache-rat-plugin</artifactId>
-          <executions>
-            <execution>
-              <id>rat-check</id>
-              <configuration>
-                <excludes>
-                  <exclude>Jenkinsfile</exclude>
-                  <exclude>**/.gitattributes</exclude>
-                  <exclude>src/main/resources/META-INF/services/**</exclude>
-                  <exclude>src/test/resources*/**</exclude>
-                  <exclude>src/test/projects/**</exclude>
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -804,9 +804,42 @@ under the License.
...
+            </execution>
+          </executions>
+        </plugin>
       </plugins>
     </pluginManagement>
     <plugins>
+      <plugin>
+        <groupId>org.apache.rat</groupId>
+        <artifactId>apache-rat-plugin</artifactId>
+      </plugin>
```
slawekjaranowski: this should be inherited from parent 🤔

pom.xml
```diff
@@ -804,9 +804,42 @@ under the License.
             </dependency>
           </dependencies>
         </plugin>
+        <plugin>
+          <groupId>org.apache.rat</groupId>
+          <artifactId>apache-rat-plugin</artifactId>
+          <executions>
+            <execution>
+              <id>rat-check</id>
+              <configuration>
```
slawekjaranowski: in parent we have configuration on plugin level not on execution level ...
https://github.com/apache/maven-parent/blob/master/pom.xml#L1270

how it is merged? interesting

cstamas: Is merged as usual, but we can make this back to same as in parent (config and not exec level) to keep things less mind boggling.

pom.xml
cstamas: I did not see it in parent, only in pluginMgmt... will doublecheck

cstamas: I missed it, removed

pom.xml
cstamas: Reworked, now we have config only

PR #11639, opened 2026-01-11
Consumer POM of multi-module project should exclude <build> and <dependencies> elements

desruisseaux:
Maven 4 automatically generates a consumer POM after a successful build. This consumer POM already excludes some elements from the original POM such as the list of sub-projects. However, when building a modular project (in Java sense), there are two more elements that need to be excluded:

* `<build>` because it contains (in the modular case) the `<sources>` element which is not compatible with the Maven 4.0.0 model. This pull request removes the full `<build>` element instead of only the `<sources>` child element because the build configuration is invalid without the sources, and I think that information about how the project was built is not really necessary for consumers anyway.
* `<dependencies>` because the effective dependencies will be dispatched in the POM files generated by `maven-jar-plugin` for each Java module from the content of `module-info.class`.

### DIFF
--- a/impl/maven-core/src/main/java/org/apache/maven/internal/transformation/impl/DefaultConsumerPomBuilder.java
+++ b/impl/maven-core/src/main/java/org/apache/maven/internal/transformation/impl/DefaultConsumerPomBuilder.java
@@ -382,7 +382,7 @@ class DefaultConsumerPomBuilder implements PomBuilder {
         }
         if (!preserveModelVersion) {
             /*
-             * If tne <build> contains <source> elements, it is not compatible with the Maven 4.0.0 model.
+             * If the <build> contains <source> elements, it is not compatible with the Maven 4.0.0 model.
              * Remove the full <build> element instead of removing only the <sources> element, because the
              * build without sources does not mean much. Reminder: this removal can be disabled by setting
              * the `preserveModelVersion` XML attribute or `preserve.model.version` property to true.


desruisseaux:
> the point is more that we can't drop useful information cause it is not used

This is not a general rule. The encapsulation principle in object oriented programming recommends the opposite: do not expose implementation details unless you have to (["in case of doubt, leave it out"](https://henko.net/blog/leave-it-out/)). I'm in favour of this approach. We have the opportunity to remove `<build>` in multi-module projects because they are a new feature. Keeping `<build>` now is an unnecessary risk, as it does not provide the information required by the use cases in the previous comment anyway — it is much safer to leave `<build>` out and add more information later as we get better understanding of what piece of information is really needed.

> Yes but to complete your PR idea you must remove them from <dependencies> too cause the consumer pom doesn't need them

This is not the topic of this pull request. This pull request removes all dependencies in the parent consumer POM, and the JAR plugin [as proposed](https://github.com/apache/maven-jar-plugin/pull/508) reinjects only the dependencies that are declared in `module-info.class` of each module. Addressing above comment would be the topic of a separated issue in `maven-jar-plugin`.

> Well my point was that removing build is an issue

See above: _keeping_ the build is an issue, because it ties us forever for nothing (the dependencies information are not in `<build>`). If an information appears to be really missing, it is much safer to revisit the situation when we have the use case.

desruisseaux:
> I'll be fine but our ecosystem will be inconsistent, do we want that?

Fully embracing Java modules requires a lot of changes in developer's habit. I don't see any way to avoid inconsistency in the way to declare dependencies (Javadoc and annotation moved from plugin configuration to ordinary dependencies), to declare sources (plugin configuration and `<sourceDirectory>` replaced by `<sources>`), to organize modules (Module Source Hierarchy), _etc_. This pull request is a small change in comparison.

rmannibucau:
@desruisseaux can we get some unit tests + a property (undocumented is fine) to disable that please? (then will be good to me)

ascheman:
I see an **issue with `preserve.model.version` handling:**

The current PR strips `<build>` and `<dependencies>` unconditionally for modular projects, ignoring `preserveModelVersion`. This means even when a user explicitly sets `preserve.model.version=true`, the modular elements are removed.

Suggested refactoring with integrated logic:

```java
static Model transformPom(Model model, MavenProject project) {
    boolean preserveModelVersion = model.isPreserveModelVersion();

    // raw to consumer transform
    model = model.withRoot(false).withModules(null).withSubprojects(null);
    if (model.getParent() != null) {
        model = model.withParent(model.getParent().withRelativePath(null));
    }

    if (!preserveModelVersion) {
        // Downgrade to 4.0.0 compatibility
        if (isModular(project)) {
            // <sources> element requires 4.1.0+, strip incompatible elements
            model = model.withBuild(null);
            model = model.withDependencies(null);
        }
        model = model.withPreserveModelVersion(false);
        String modelVersion = new MavenModelVersion().getModelVersion(model);
        model = model.withModelVersion(modelVersion);
    }
    // else: keep model as-is, including <build> and <dependencies> for modular projects.
    // Consumer POM will use model version 4.1.0+ if needed.

    return model;
}
```

This way:
- `preserve.model.version=true` + modular → keep `<build>` and `<dependencies>`, use 4.1.0+
- `preserve.model.version=false` (default) + modular → strip incompatible elements, use 4.0.0

No new property needed - the existing `preserve.model.version` already expresses the intent.

**Code suggestion:**

The `isModular()` method duplicates logic that already exists in `SourceHandlingContext`. Consider extracting to a shared utility.

**Also:**
- Branch needs rebase (24 commits behind master)
- Unit tests for `isModular()` and `transformPom()` modular behavior would help

desruisseaux:
I forgot about `preserveModelVersion`. Indeed, it can be used for deciding whether to strip `<build>`. I will apply this change.

Regarding `<dependencie>`, this is another issue. The removal of dependencies is not for compatibility with the 4.0.0 model. It is because the current strategy of https://github.com/apache/maven-jar-plugin/pull/508 uses the consumer POM of the project as the parent of the POM of each module. If we do not remove the dependencies, then every modules would inherit all project's dependencies, including the dependencies that a module does not use.

Said otherwise, in a multi-module project, `<dependencies>` actually behave more like `<dependencyManagement>`. Not removing dependencies would be a little bit like inheriting the full content of `<dependencyManagement>` as dependencies, regardless if used or not.

ascheman:
> Regarding `<dependencie>`, this is another issue. The removal of dependencies is not for compatibility with the 4.0.0 model. It is because the current strategy of [apache/maven-jar-plugin#508](https://github.com/apache/maven-jar-plugin/pull/508) uses the consumer POM of the project as the parent of the POM of each module. If we do not remove the dependencies, then every modules would inherit all project's dependencies, including the dependencies that a module does not use.

I should have looked again in 508 ... On the other hand I was not aware that there was already some stuff to create consumer POM in the core. Considering that, handling of <whatever> (creating consumer POMs in this case) it is perhaps best to either leave it completely in the core (and thereby unify handling accross plugins) or move it completely to the Plugins.

IIRC we already discussed this topic elsewhere. Perhaps it is OK to initially do something at first in single plugins and only move it to the core when it stabilizes? In this case I slightly tend to move the Maven Model > 4.1.0 based consumer POM generation to the core as soon as possible to not get different behaviour on Maven 3 (4.0.0 model) and Maven 4 (4.1.0).

desruisseaux:
The generation of a consumer POM derived from the project POM is fully in Maven core, which is the reason for this pull request. The POM files generated by pull request 508 are different: they are not derived from a project's POM. Instead, they are derived from `module-info.class` files, together with the project POM which is used as the parent (which is the reason why we need to remove its dependencies).

The generation of POM files derived from `module-info.class` files is done by the Maven JAR Plugin because this plugin is in a particular situation. Most Java tools accept module hierarchy in their `--module-path` option. Therefore, the `javac` output produced by the Maven Compiler Plugin can be seen as a single output by `java` and `javadoc` among others, even if that output contains many modules. But the `jar` tool is an exception, because each JAR file contains a single module. This constraint is desirable anyway for smaller downloads. Therefore, the Maven JAR Plugin is particular in that it needs to separate modules, contrarily to plugins based on `java`, `javac` or `javadoc` (among others).

Since the Maven JAR Plugin needs to analyse the Maven Compiler Plugin output for dispatching it in different modules, it has information that Maven core does not have easily. Therefore, Maven JAR Plugin is a good place for generating POM *derived from module-info*, while Maven core is fully in charge of generating POM *derived from project POM*.

desruisseaux:
* Rebased the pull request.
* Extracted in a new `ModuleSourcesHelper` class some codes shared by `SourceHandlingContext` and this pull request.
* Made the existing tests a little bit more exhaustive and reduced code duplication.
* Added a new test for the multi-module case.
* Remove the `<build>` element only if `preserve.model.version` is `false`.
  * In this case, the `<build>` element may be removed also for non-modular projects if they use the `<sources>` element, for the same reason as for modular-project: this is not a valid 4.0.0 model.
  * Classical Maven projects which use the legacy way to declare sources stay unimpacted.

impl/maven-core/src/main/java/org/apache/maven/project/SourceHandlingContext.java
```diff
@@ -67,19 +65,30 @@ record SourceKey(Language language, ProjectScope scope, String module, Path dire
...
-    SourceHandlingContext(
-            MavenProject project,
-            Path baseDir,
-            Set<String> modules,
-            boolean modularProject,
-            ModelBuilderResult result) {
+    SourceHandlingContext(MavenProject project, Path baseDir, ModelBuilderResult result) {
+        super(project);
         this.project = project;
         this.baseDir = baseDir;
```
ascheman: Note: PR #11702 (AC8/AC9 legacy directory handling) also modifies this constructor and `SourceHandlingContext` in general. Whichever PR merges first will create a conflict for the other. Just a heads-up so we can coordinate the merge order.

impl/maven-core/src/main/java/org/apache/maven/internal/transformation/impl/DefaultConsumerPomBuilder.java
```diff
@@ -369,19 +371,33 @@ static Model transformPom(Model model, MavenProject project) {
...
+        if (sources.useModuleSourceHierarchy()) {
+            // Dependencies are dispatched by maven-jar-plugin in the POM generated for each module.
+            model = model.withDependencies(null).withPackaging(POM_PACKAGING);
         }
-
         if (!preserveModelVersion) {
+            /*
+             * If tne <build> contains <source> elements, it is not compatible with the Maven 4.0.0 model.
+             * Remove the full <build> element instead of removing only the <sources> element, because the
+             * build without sources does not mean much. Reminder: this removal can be disabled by setting
```
ascheman, line 387:
```suggestion
             * If the <build> contains <source> elements, it is not compatible with the Maven 4.0.0 model.
```

impl/maven-core/src/test/java/org/apache/maven/internal/transformation/impl/ConsumerPomBuilderTest.java
```diff
@@ -108,39 +116,50 @@ void testTrivialConsumer() throws Exception {
...
+        Path file = Paths.get("src/test/resources/consumer/multi-module/pom.xml");
+
+        MavenProject project = getEffectiveModel(file);
+        Model model = builder.build(session, project, Sources.buildSource(file));
+
+        assertNotNull(model);
+        assertNull(model.getBuild());
+        assertTrue(model.getDependencies().isEmpty());
+        assertFalse(model.getDependencyManagement().getDependencies().isEmpty());
+    }
```
ascheman, line 182: This test validates the default case (`preserveModelVersion=false`). I added a local companion test `testMultiModuleConsumerPreserveModelVersion` that verifies `<build>` is preserved when `preserveModelVersion=true`:

```java
@Test
void testMultiModuleConsumerPreserveModelVersion() throws Exception {
    setRootDirectory("multi-module");
    Path file = Paths.get("src/test/resources/consumer/multi-module/pom.xml");

    MavenProject project = getEffectiveModel(file);
    Model model = getEffectiveModel(file).getModel().getDelegate();
    model = Model.newBuilder(model, true).preserveModelVersion(true).build();

    Model transformed = DefaultConsumerPomBuilder.transformPom(model, project);

    assertNotNull(transformed);
    assertNotNull(transformed.getBuild());
    assertTrue(transformed.getDependencies().isEmpty());
    assertFalse(transformed.getDependencyManagement().getDependencies().isEmpty());
}
```

Would you consider adding this or a similar test to cover the `preserveModelVersion` gate?

impl/maven-core/src/main/java/org/apache/maven/project/SourceHandlingContext.java
desruisseaux: Yes I noticed that. I will do a rebase and conflict resolution after #11702 is merged.

impl/maven-core/src/main/java/org/apache/maven/internal/transformation/impl/DefaultConsumerPomBuilder.java
desruisseaux: The suggested change seems strange. It replaces the middle of a sentence ("build without sources …") by a copy of the first sentence. I suspect that the intend was to replace the first sentence for fixing the "n" in "the".

impl/maven-core/src/test/java/org/apache/maven/internal/transformation/impl/ConsumerPomBuilderTest.java
desruisseaux: Sure, thanks for the addition! I will add it after the merge of #11702 since we will need to rebase anyway.

desruisseaux: Added the test, thanks. Note: I tried `Authored-by: Gerd …` instead of `Co-authored-by:` since I did no change, but it didn't work. It will be fixed when we will squash all the commits.

desruisseaux:
Thanks @rmannibucau. Moving ahead then.

Issue #11504, opened 2025-11-29
Deprecated classes/methods: Add information what to use instead


Bukama:
### New feature, improvement proposal

Several things in `org.apache.maven.api.model` are marked as deprecated, for example

```java
  @Deprecated(
    since = "4.0.0"
  )
  public String getScriptSourceDirectory() {
    return this.scriptSourceDirectory;
  }
```

But there is no information what to use instead, resulting in the users (I found it in maven-resource-plugin) only got the deprecated warning, but not hint how to handle it.

achyuth8055:
Hi!,

I am new to the Apache Maven project and would like to work on this issue.
I have read the contribution guidelines and would like to help improve the deprecation documentation.
Could you please assign this issue to me?

Thanks!

anukalp2804:
Hi @Bukama ,

I’d like to work on this issue.

I plan to audit deprecated APIs in `org.apache.maven.api.model` and add clear JavaDoc guidance indicating what should be used instead. I will keep the changes documentation-only and submit small, focused PRs. Where relevant, I can also backport the deprecation advice to the 3.x line to help migration.

Please let me know if this approach is acceptable.

Thanks!

PR #11651, opened 2026-01-15
docs: clarify replacement for deprecated scriptSourceDirectory

anukalp2804:
### Summary
Clarifies replacement guidance for the deprecated `scriptSourceDirectory` model field.

### Details
- Updated the model descriptor (.mdo) to explicitly document the recommended replacement
- Regenerated sources accordingly
- Documentation-only change; no behavioral or API changes

Relates to #11504

### DIFF
--- a/api/maven-api-model/src/main/mdo/maven.mdo
+++ b/api/maven-api-model/src/main/mdo/maven.mdo
@@ -925,7 +925,9 @@
             interpreted rather than compiled).
             The default value is {@code src/main/scripts}.
 
-            @deprecated Replaced by {@code &lt;Source&gt;} with {@code script} language.
+            @deprecated since 4.0.0.
+            Use {@code &lt;source&gt;} elements with {@code script} language
+            to configure script sources.
           </description>
           <type>String</type>
           <annotations>


api/maven-api-model/src/main/mdo/maven.mdo
```diff
@@ -925,7 +925,9 @@
             interpreted rather than compiled).
             The default value is {@code src/main/scripts}.
 
-            @deprecated Replaced by {@code &lt;Source&gt;} with {@code script} language.
+            @deprecated since 4.0.0.
+            Use {@code &lt;source&gt;} elements with {@code script} language
```
elharo, line 929: Is it Source or source?

anukalp2804: Yes, it should be lowercase `<source>`, matching the Maven POM XML element name.
The current wording intentionally uses `<source>` for that reason.
Please let me know if you’d like this phrasing adjusted further.

desruisseaux: Note: `{@code &lt;source&gt;}` should be `{@code <source>}`. HTML character entities are automatically escaped when inside `{@code}` or `{@literal}`.

**EDIT:** Oups! Sorry, ignore that comment. I just saw that this is inside a `.mdo` file, not a `.java` file.

elharo: How does mdo work here? Does it first copy this into a .java file?

anukalp2804: Yes — the .mdo file is the source of truth.
During the build, Maven’s model generator processes the .mdo and generates the corresponding Java model classes and JavaDoc from it.

That’s why the XML escaping (e.g. &lt;source&gt;) is needed in the .mdo, even though the generated JavaDoc ends up with the correct `<source>` rendering.

PR #11584, opened 2025-12-21
Document deprecation rationale for Artifact version constants

anukalp2804:
Summary:
This pull request documents the rationale for deprecating the Artifact version
constants RELEASE and LATEST.

Why:
The RELEASE and LATEST version constants lead to non-reproducible builds and are
discouraged in Maven 4. This change makes the reason for deprecation explicit
and provides guidance on preferred alternatives.

What:
- Added JavaDoc explaining why these constants are deprecated
- Clarified recommended usage of explicit versions or version ranges

Scope:
- Documentation-only change
- No behavioral or API changes

Testing:
- mvn verify (BUILD SUCCESS)

Following this checklist to help us incorporate your
contribution quickly and easily:

- [x] Your pull request should address just one issue, without pulling in other changes.
- [x] Write a pull request description that is detailed enough to understand what the pull request does, how, and why.
- [x] Each commit in the pull request should have a meaningful subject line and body.
  Note that commits might be squashed by a maintainer on merge.
- [ ] Write unit tests that match behavioral changes, where the tests fail if the changes to the runtime are not applied.
  This may not always be possible but is a best-practice.
- [x] Run `mvn verify` to make sure basic checks pass.
  A more thorough check will be performed on your pull request automatically.
- [ ] You have run the [Core IT][core-its] successfully.

If your pull request is about ~20 lines of code you don't need to sign an
[Individual Contributor License Agreement](https://www.apache.org/licenses/icla.pdf) if you are unsure
please ask on the developers list.

To make clear that you license your contribution under
the [Apache License Version 2.0, January 2004](http://www.apache.org/licenses/LICENSE-2.0)
you have to acknowledge this by using the following check-box.

- [x] I hereby declare this contribution to be licenced under the [Apache License Version 2.0, January 2004](http://www.apache.org/licenses/LICENSE-2.0)
- [ ] In any other case, please file an [Apache Individual Contributor License Agreement](https://www.apache.org/licenses/icla.pdf).

[core-its]: https://maven.apache.org/core-its/core-it-suite/

PR #11686, opened 2026-01-29
docs: clarify deprecation of RELEASE and LATEST version constants

amahi1568:
### What was done
Added JavaDoc guidance for deprecated artifact version constants to clarify recommended alternatives.

### Why
The deprecated RELEASE and LATEST version tokens provided no migration guidance, which could confuse users.

### Scope
Documentation-only change. No functional impact.

### DIFF
--- a/compat/maven-artifact/src/main/java/org/apache/maven/artifact/Artifact.java
+++ b/compat/maven-artifact/src/main/java/org/apache/maven/artifact/Artifact.java
@@ -38,17 +38,17 @@ import org.apache.maven.artifact.versioning.VersionRange;
 public interface Artifact extends Comparable<Artifact> {
 
     /**
-    * @deprecated since 4.0.0
+    * @deprecated since 3.0
     * Use explicit version resolution instead of special version constants.
     */
-    @Deprecated(since = "4.0.0")
+    @Deprecated(since = "3.0")
     String RELEASE_VERSION = "RELEASE";
 
     /**
-    * @deprecated since 4.0.0.
+    * @deprecated since 3.0
     * Use a fixed version or an explicit version range instead of special version tokens.
     */
-    @Deprecated(since = "4.0.0")
+    @Deprecated(since = "3.0")
     String LATEST_VERSION = "LATEST";
 
     String SNAPSHOT_VERSION = "SNAPSHOT";
--- a/compat/maven-artifact/src/main/java/org/apache/maven/artifact/Artifact.java
+++ b/compat/maven-artifact/src/main/java/org/apache/maven/artifact/Artifact.java
@@ -37,9 +37,17 @@ import org.apache.maven.artifact.versioning.VersionRange;
  */
 public interface Artifact extends Comparable<Artifact> {
 
+    /**
+    * @deprecated since 4.0.0
+    * Use explicit version resolution instead of special version constants.
+    */
     @Deprecated(since = "4.0.0")
     String RELEASE_VERSION = "RELEASE";
 
+    /**
+    * @deprecated since 4.0.0.
+    * Use a fixed version or an explicit version range instead of special version tokens.
+    */
     @Deprecated(since = "4.0.0")
     String LATEST_VERSION = "LATEST";
 


amahi1568:
Thanks for the clarification. I misunderstood the deprecation timeline for
LATEST and RELEASE. I’ll close this PR accordingly.

compat/maven-artifact/src/main/java/org/apache/maven/artifact/Artifact.java
```diff
@@ -37,10 +37,18 @@
  */
 public interface Artifact extends Comparable<Artifact> {
 
-    @Deprecated(since = "4.0.0")
+    /**
+    * @deprecated since 3.0
```
elharo, line 41: No need to duplicate the since parameter

elharo (changes requested):

compat/maven-artifact/src/main/java/org/apache/maven/artifact/Artifact.java
```diff
@@ -37,10 +37,18 @@
...
+    * Use explicit version resolution instead of special version constants.
+    */
+    @Deprecated(since = "3.0")
     String RELEASE_VERSION = "RELEASE";
 
-    @Deprecated(since = "4.0.0")
+    /**
+    * @deprecated since 3.0
+    * Use a fixed version or an explicit version range instead of special version tokens.
```
elharo, line 49: never use version ranges

elharo (changes requested):

compat/maven-artifact/src/main/java/org/apache/maven/artifact/Artifact.java
```diff
@@ -37,10 +37,18 @@
  */
 public interface Artifact extends Comparable<Artifact> {
 
-    @Deprecated(since = "4.0.0")
+    /**
+    * @deprecated since 3.0
+    * Use explicit version resolution instead of special version constants.
+    */
+    @Deprecated(since = "3.0")
```
gnodet, line 44: The `since` value should remain `"4.0.0"` — that is when the `@Deprecated` annotation was actually added to these constants. Changing it to `"3.0"` without evidence from the git history is inaccurate.

Also, the Javadoc asterisks are misaligned — they should use ` *` (space-star) to match the project's formatting conventions:

```suggestion
    * @deprecated since 3.0
    * Use explicit version resolution instead of special version constants.
```

should be (if kept at all):

```suggestion
     * @deprecated Since 4.0.0. Use explicit version resolution instead of special version constants.
```

compat/maven-artifact/src/main/java/org/apache/maven/artifact/Artifact.java
```diff
@@ -37,10 +37,18 @@
...
     String RELEASE_VERSION = "RELEASE";
```
gnodet, line 45: Changing `@Deprecated(since = "4.0.0")` to `@Deprecated(since = "3.0")` is incorrect. The deprecation annotation was introduced in Maven 4.0.0. Please revert.

```suggestion
    @Deprecated(since = "4.0.0")
```

gnodet (commented):

Thanks for improving the documentation. A few issues to address:

1. **`since` version change is incorrect**: The `@Deprecated` annotation currently says `since = "4.0.0"` on `master`. Changing it to `"3.0"` is not accurate — the `@Deprecated` annotation itself was added in 4.0.0. While RELEASE/LATEST resolution may have been discouraged earlier, the formal deprecation marker was introduced in 4.x. Please revert this change and keep `since = "4.0.0"`.

2. As @elharo already noted, the Javadoc `@deprecated` tag duplicates what the `@Deprecated(since=...)` annotation already conveys. If you want to add guidance, a plain Javadoc comment without `@deprecated` would be sufficient, since the annotation already handles the deprecation signal.

3. Also as @elharo noted, recommending "version range" as an alternative to LATEST is not good advice.

Please address the existing review feedback before this can move forward.

gnodet (commented):
This PR is fully superseded by the already-merged PR #11584 (commit 4bffb2a6bc, merged 2026-06-17), which added superior Javadoc documentation to the same `RELEASE_VERSION` and `LATEST_VERSION` constants. The PR is now in a CONFLICTING merge state due to the overlapping changes.

**Key issues:**

**1. Incorrect `since` value (high)**
The `@Deprecated(since = "3.0")` change is factually incorrect. The `@Deprecated` annotation was first added to these constants in Maven 4.0.0 (commit 1bac0e6087, 2024-08-16). While RELEASE/LATEST resolution behavior was discouraged earlier, the formal annotation was introduced in 4.x. Multiple reviewers flagged this.

**2. Superseded by merged work (high)**
The current master already contains better documentation from PR #11584, which correctly preserves `since = "4.0.0"`, mentions non-reproducible builds and supply chain attacks, and recommends using explicit versions.

**3. Poor migration advice (medium)**
Recommending _"an explicit version range"_ as an alternative to LATEST is poor advice — version ranges have their own reproducibility and security pitfalls, as noted by @elharo.

This PR should be closed as it would actually degrade the existing documentation if applied on top of the current master.

slachiewicz:
The last CI run on this branch was 2026-02-03, and GitHub has deleted the logs since, so there's no way to tell whether the change still builds against master.

The branch also conflicts with master now, so that needs resolving before CI can say anything. Say the word if you'd like a hand with it.

elharo:
should have been closed after @gnodet's last comment

PR #11690, opened 2026-02-01
Add Javadoc for deprecated Artifact constants

Vaibhavi410:
Adds Javadoc to the deprecated RELEASE_VERSION and LATEST_VERSION
constants in Artifact to indicate preferred alternatives, as requested
in issue #11504.

### DIFF
--- a/compat/maven-artifact/src/main/java/org/apache/maven/artifact/Artifact.java
+++ b/compat/maven-artifact/src/main/java/org/apache/maven/artifact/Artifact.java
@@ -38,17 +38,14 @@ import org.apache.maven.artifact.versioning.VersionRange;
  */
 public interface Artifact extends Comparable<Artifact> {
     /**
-     * @deprecated since 4.0.0.
-     *             Use an explicit version instead.
+     * @deprecated Use {@link #VERSION} instead.
      */
-
     @Deprecated(since = "4.0.0")
     String RELEASE_VERSION = "RELEASE";
-    /**
-     * @deprecated since 4.0.0.
-     *             Avoid using "LATEST"; prefer an explicit version instead.
-     */
 
+    /**
+     * @deprecated Avoid using {@code LATEST}; prefer an explicit version instead.
+     */
     @Deprecated(since = "4.0.0")
     String LATEST_VERSION = "LATEST";
 


compat/maven-artifact/src/main/java/org/apache/maven/artifact/Artifact.java
```diff
@@ -32,13 +32,22 @@
...
 
 /**
- * Maven Artifact interface. Notice that it mixes artifact definition concepts (groupId, artifactId, version)
+ * Maven Artifact interface. Notice that it mixes artifact definition concepts
+ * (groupId, artifactId, version)
  * with dependency information (version range, scope).
  */
 public interface Artifact extends Comparable<Artifact> {
+    /**
+     * @deprecated since 4.0.0.
```
elharo: put this comment on one line.
Do we need the since text here too oir does it duplciate what's in the annotation parameter?

compat/maven-artifact/src/main/java/org/apache/maven/artifact/Artifact.java
```diff
@@ -32,13 +32,22 @@
...
+     *             Use an explicit version instead.
+     */
 
     @Deprecated(since = "4.0.0")
     String RELEASE_VERSION = "RELEASE";
+    /**
+     * @deprecated since 4.0.0.
```
elharo: put this comment on one line.
Do we need the since text here too oir does it duplciate what's in the annotation parameter?

compat/maven-artifact/src/main/java/org/apache/maven/artifact/Artifact.java
desruisseaux: I agree, the "since 4.0.0" text is redundant. Please replace it by a text explaining what is the replacement, or why there is no replacement.

compat/maven-artifact/src/main/java/org/apache/maven/artifact/Artifact.java
desruisseaux: Idem: replace the redundant "since 4.0.0" text by a text explaining what is the replacement, or why there is no replacement.

compat/maven-artifact/src/main/java/org/apache/maven/artifact/Artifact.java
Vaibhavi410: Thanks for the review!
I’ve updated the Javadoc to a single-line comment and removed the redundant since 4.0.0 text, relying on the @Deprecated(since = "4.0.0") annotation instead.

compat/maven-artifact/src/main/java/org/apache/maven/artifact/Artifact.java
Vaibhavi410: Thanks for the review!
I’ve updated the Javadoc to a single-line comment and removed the redundant since 4.0.0 text, relying on the @Deprecated(since = "4.0.0") annotation instead.

compat/maven-artifact/src/main/java/org/apache/maven/artifact/Artifact.java
```diff
@@ -32,14 +32,20 @@
...
 /**
- * Maven Artifact interface. Notice that it mixes artifact definition concepts (groupId, artifactId, version)
+ * Maven Artifact interface. Notice that it mixes artifact definition concepts
+ * (groupId, artifactId, version)
  * with dependency information (version range, scope).
  */
 public interface Artifact extends Comparable<Artifact> {
-
+    /**
+     * @deprecated Use {@link #VERSION} instead.
```
desruisseaux, line 41: There is no `VERSION` field in this class. The intend was probably "use a specific version number instead"?

Vaibhavi410: Fixed 👍
Updated the deprecated Javadoc to remove the invalid VERSION reference and clarify using an explicit version string instead.

PR #11692, opened 2026-02-01
docs: fix broken reference links

amahi1568:
## Description

This pull request addresses a single documentation-related issue by fixing a minor typo / improving clarity in the existing documentation.

There are **no functional or behavioral code changes** included in this PR.  
The change is limited strictly to documentation and does not impact runtime behavior.

## Checklist

- [ ] Write unit tests that match behavioral changes (not applicable for documentation-only changes).
- [ ] Run `mvn verify` (not required for documentation-only changes).
- [ ] You have run the Core IT (not applicable).

## License Declaration

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -139,7 +139,7 @@ under the License.
     <distributionId>apache-maven</distributionId>
     <distributionShortName>Maven</distributionShortName>
     <distributionName>Apache Maven</distributionName>
-    <maven.site.path>ref/4-LATEST</maven.site.path>
+    <maven.site.path>ref/4.0.0</maven.site.path>
     <project.build.outputTimestamp>2025-06-24T20:18:00Z</project.build.outputTimestamp>
     <!-- various versions -->
     <asmVersion>9.9.1</asmVersion>


elharo (approved):

pom.xml
```diff
@@ -139,7 +139,7 @@ under the License.
     <distributionId>apache-maven</distributionId>
     <distributionShortName>Maven</distributionShortName>
     <distributionName>Apache Maven</distributionName>
-    <maven.site.path>ref/4-LATEST</maven.site.path>
+    <maven.site.path>ref/4.0.0</maven.site.path>
```
slawekjaranowski, line 142: on master branch it is ok, we should change it on maven-4.0.x branch to something like `ref/4.0-LATEST`
@gnodet, @cstamas what do you think?

is is used to publish site candidate during release
https://maven.apache.org/components/ref/

maybe we should also add redirects to:
https://github.com/apache/maven-site/blob/master/content/filtered-resources/.htaccess


@gnodet, @cstamas what do you think?

slawekjaranowski (changes requested):

pom.xml
gnodet, line 142: This change should not be made. `4-LATEST` is the correct value — it's an intentional alias that always resolves to the latest 4.x release docs on the Maven site.

You can verify:
- https://maven.apache.org/ref/4-LATEST/ → HTTP 200 ✓
- https://maven.apache.org/ref/4.0.0/ → HTTP 404 ✗

Additionally, `master` is currently on `4.1.0-SNAPSHOT`, so hardcoding `4.0.0` would be doubly wrong.

gnodet (commented):

This change would actually **break** the site path rather than fix it.

- `ref/4-LATEST` resolves correctly (HTTP 200): https://maven.apache.org/ref/4-LATEST/
- `ref/4.0.0` returns HTTP 404: https://maven.apache.org/ref/4.0.0/

The `master` branch is currently developing `4.1.0-SNAPSHOT`. The `4-LATEST` path is intentionally used so the generated site always points to the latest 4.x reference docs, regardless of the specific release version. Hardcoding `4.0.0` here would be incorrect.

gnodet (changes requested):
This change would break the site links — `ref/4-LATEST` resolves correctly (HTTP 200) while `ref/4.0.0` returns 404. The `4-LATEST` path is intentionally used so generated docs always point to the latest 4.x reference, regardless of release version.

gnodet (commented):
This PR would break the Maven site reference links rather than fix them.

The existing `ref/4-LATEST` path is a working rolling pointer that always resolves to the latest 4.x release documentation (currently serving 4.1.0-SNAPSHOT docs, HTTP 200). The proposed `ref/4.0.0` returns HTTP 404.

The `4-LATEST` value was intentionally set by Guillaume Nodet (commit d62113772e6, 2023-11-17) to provide a stable pointer that updates automatically with each release. Since the master branch is currently developing `4.1.0-SNAPSHOT`, hardcoding `4.0.0` would also be immediately stale.

Two project members (slawekjaranowski and gnodet) have already requested changes for the same reason. This PR should be closed.

PR #11700, opened 2026-02-05
Add module-aware resource handling for modular sources

desruisseaux:
This is a combination of pull requests #11505, #11632 and #11702 on master branch.

**Summary:** Maven 4.x introduces a unified `<sources>` element that supports modular project layouts (`src/<module>/<scope>/<lang>`). However, resource handling did not follow the modular layout - resources were always loaded from the legacy `<resources>` element which defaults to `src/main/resources`. This change implements automatic module-aware resource injection.

- For modular projects without resource configuration in `<sources>`, automatically inject resource roots following the modular layout: `src/<module>/main/resources` and `src/<module>/test/resources`.
- Resources configured via `<sources>` take priority over legacy `<resources>`.
- Fail the build (as `ModelProblem`) when explicit legacy resources are ignored.

### DIFF
--- a/impl/maven-core/src/main/java/org/apache/maven/project/DefaultProjectBuilder.java
+++ b/impl/maven-core/src/main/java/org/apache/maven/project/DefaultProjectBuilder.java
@@ -27,6 +27,7 @@ import java.io.File;
 import java.io.IOException;
 import java.io.InputStream;
 import java.nio.charset.StandardCharsets;
+import java.nio.file.Files;
 import java.nio.file.Path;
 import java.util.AbstractMap;
 import java.util.ArrayList;
@@ -63,7 +64,6 @@ import org.apache.maven.api.model.Model;
 import org.apache.maven.api.model.Plugin;
 import org.apache.maven.api.model.Profile;
 import org.apache.maven.api.model.ReportPlugin;
-import org.apache.maven.api.model.Resource;
 import org.apache.maven.api.services.ArtifactResolver;
 import org.apache.maven.api.services.ArtifactResolverException;
 import org.apache.maven.api.services.ArtifactResolverRequest;
@@ -520,7 +520,7 @@ public class DefaultProjectBuilder implements ProjectBuilder {
                 return pomFiles.stream()
                         .map(pomFile -> build(pomFile, recursive))
                         .flatMap(List::stream)
-                        .collect(Collectors.toList());
+                        .toList();
             } finally {
                 Thread.currentThread().setContextClassLoader(oldContextClassLoader);
             }
@@ -566,7 +566,7 @@ public class DefaultProjectBuilder implements ProjectBuilder {
                     project.setCollectedProjects(results(r)
                             .filter(cr -> cr != r && cr.getEffectiveModel() != null)
                             .map(cr -> projectIndex.get(cr.getEffectiveModel().getId()))
-                            .collect(Collectors.toList()));
+                            .toList());
 
                     DependencyResolutionResult resolutionResult = null;
                     if (request.isResolveDependencies()) {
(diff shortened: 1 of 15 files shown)


elharo (commented):
I'm a little uncomfortable that large fixes like this are getting submitted without corresponding issues.

impl/maven-core/src/main/java/org/apache/maven/project/DefaultProjectBuilder.java
```diff
@@ -520,7 +520,7 @@ List<ProjectBuildingResult> doBuild(List<File> pomFiles, boolean recursive) {
                 return pomFiles.stream()
                         .map(pomFile -> build(pomFile, recursive))
                         .flatMap(List::stream)
-                        .collect(Collectors.toList());
+                        .toList();
```
elharo, line 523: This could be a separate small change

impl/maven-core/src/main/java/org/apache/maven/project/DefaultProjectBuilder.java
```diff
@@ -566,7 +566,7 @@ private List<ProjectBuildingResult> build(File pomFile, boolean recursive) {
                     project.setCollectedProjects(results(r)
                             .filter(cr -> cr != r && cr.getEffectiveModel() != null)
                             .map(cr -> projectIndex.get(cr.getEffectiveModel().getId()))
-                            .collect(Collectors.toList()));
+                            .toList());
```
elharo, line 569: ditto; good but not needed in this PR

impl/maven-core/src/main/java/org/apache/maven/project/DefaultProjectBuilder.java
```diff
@@ -660,46 +660,75 @@ private void initProject(MavenProject project, ModelBuilderResult result) {
...
-                    if (Language.JAVA_FAMILY.equals(language)) {
-                        ProjectScope scope = src.scope();
-                        if (ProjectScope.MAIN.equals(scope)) {
-                            hasMain = true;
-                        } else {
-                            hasTest |= ProjectScope.TEST.equals(scope);
-                        }
-                    } else {
-                        hasScript |= Language.SCRIPT.equals(language);
+                    var sourceRoot = DefaultSourceRoot.fromModel(session, baseDir, outputDirectory, source);
```
elharo, line 680: Personally I dislike var. It prioritizes trivial typing speed over code readability

impl/maven-core/src/main/java/org/apache/maven/project/DefaultProjectBuilder.java
```diff
@@ -870,6 +899,49 @@ private void initProject(MavenProject project, ModelBuilderResult result) {
             project.setRemoteArtifactRepositories(remoteRepositories);
         }
 
+        /**
+         * Warns about legacy directory usage in a modular project. Two cases are handled:
+         * <ul>
+         *   <li>Case 1: The default legacy directory exists on the filesystem (e.g., src/main/java exists)</li>
+         *   <li>Case 2: An explicit legacy directory is configured that differs from the default</li>
+         * </ul>
+         * Legacy directories are unconditionally ignored in modular projects because it is not clear
```
elharo: I'm not comfortable with this. If we can't dispatch, I think we need to hard fail the build rather than simply ignoring the content. Let the user figure out how to rearrange the project to make the build pass. A warning is not enough here.

elharo (changes requested):

desruisseaux:
This pull request is a squashed backport of two pull requests from the mater branch, one merged in December 30th and the second one today. The issue describing the second commit on the master branch is #11612, but a more complete description is on the wiki at the link below:

https://cwiki.apache.org/confluence/display/MAVEN/Build+Sources+Validation+-+Discussion+Notes

impl/maven-core/src/main/java/org/apache/maven/project/DefaultProjectBuilder.java
desruisseaux: I agree. If @ascheman also agrees, I guess that we can do this change. But since this pull request is a backport, I would suggest to do the change on master first, then backport here.

elharo: Bleah. 3.x maintenance and 4.x development is hard enough without trying to develop two 4.x versions simultaneously. In any case, sure, let's do the change on master and then backport here, at least until we  come to our senses and finish one 4.x version before starting the next.

elharo:
https://github.com/apache/maven/issues/11701

impl/maven-core/src/main/java/org/apache/maven/project/DefaultProjectBuilder.java
ascheman: Sounds feasible to me. I would think this should be another acceptance criterion (AC8) in the overall concept to document that additional behaviour? It would be consistent with AC6 (where we also fail for mixed modular/classic `<sources>`).

impl/maven-core/src/main/java/org/apache/maven/project/DefaultProjectBuilder.java
ascheman: Personally I agree with you, but perhaps its more a question of programming style. `var` is already used in other places of this codebase, e.g., DefaultModelBuilder.java:
```java
  for (var problem : e.getResult().getProblemCollector()...
  for (var transformer : transformers) {
```

desruisseaux: I don't think that there is a consensus. I tend to use `var` when it avoids a redundancy on the same line, especially when the redundancy is verbose. For example:

```java
MyFooWithLongName foo = new MyFooWithLongName();
```

However, when we don't see the type on the same line (as in `for (var transformer : transformers)`), I agree that it is good to keep the type. I usually try to stick to the rule "use `var` only if we can see the type on the same line", but I may sometime use `var` by mistake.

impl/maven-core/src/main/java/org/apache/maven/project/DefaultProjectBuilder.java
desruisseaux: True, but this pull request is a verbatim backport of commits done in the master branch. By keeping the changes identical, we reduce the risk of conflicts in future merges.

elharo: if you want to do verbatim pulls from master, then it should probably be from one PR in master. Description says "This is a combination of pull requests https://github.com/apache/maven/pull/11505, https://github.com/apache/maven/pull/11632 and https://github.com/apache/maven/pull/11702 on master branch."

desruisseaux: 3 separated pull requests would be clearer if they were 3 independent tasks. But in this case, each new pull request is amending the previous one, for example adding warnings, then changing warnings into build failures. In this case, I think that squashing the 3 pull requests make them easier to analyse, because reviewers can analyse directly the final result instead of spending times on changes that are made obsolete in the next commit.

elharo: Yes, and my point is that when you're merging all these things together you're not doing a straight backport so it's not necessary to mirror exactly what never really existed on master

desruisseaux: The work is ported in one piece instead of 3 pieces, but the result is identical to that we would have if we squashed the 3 commits on master before to port. The important thing (regarding merge conflicts) is that `DefaultProjectBuilder.java` (for example) is almost identical on the `maven-4.0.x` and `master` branches — the differences between the two branches do not come from this pull requests.

If we do not include the change of `.collect(Collectors.toList())` to `.toList()` in this pull request, it will create a difference between the  `maven-4.0.x` and `master` branches. In such case, if a change is made on `master` around those lines, back-porting this change to `maven-4.0.x` will create a merge conflict.

While separating the tasks is good in principle, in this case it increases the effort both in splitting this pull request and in resolving merge conflicts later (unless this pull request is followed by another pull request doing the changes that we skipped).

PR #11702, opened 2026-02-06
Fail on legacy config in modular projects

### LINKED ISSUE #11701: Legacy resources in modular build should fail
### Affected version

HEAD 4.1.x

### Bug description

See  https://github.com/apache/maven/issues/11612 and #11505 and #11632 on master branch.
Maven 4.x introduces a unified `<sources>` element that supports modular project layouts (`src/<module>/<scope>/<lang>`). However, resource handling did not follow the modular layout - resources were always loaded from the legacy `<resources>` element which defaults to `src/main/resources`. This change implements automatic module-aware resource injection.

- For modular projects without resource configuration in `<sources>`, automatically inject resource roots following the modular layout: `src/<module>/main/resources` and `src/<module>/test/resources`.
- Resources configured via `<sources>` take priority over legacy `<resources>`.
- Issue warnings (as `ModelProblem`) when explicit legacy resources are ignored.

The first two are fine. The last isn't. Resources the user clearly meant to include are dropped from the build. This should be a full build failure that requires the user to fix their project one way or another, not a simple warning.

ascheman:
## Summary

* Legacy directories and resources that conflict with modular or new-style source
  configuration now cause the **build to fail** instead of being silently ignored
* Prevents silent loss of user-configured sources/resources
* Covers two acceptance criteria for legacy directory handling with `<sources>`:
  * **AC8** (Modular projects): ALL legacy directories cause the build to fail
  * **AC9** (Non-modular projects): Implicit fallback to legacy directories when
    `<sources>` does not configure that scope/language

Affected configurations:

* Explicit `<sourceDirectory>`/`<testSourceDirectory>` differing from Super POM defaults
* Default `src/main/java` or `src/test/java` existing physically on the filesystem
  (modular projects only)
* Explicit `<resources>`/`<testResources>` differing from Super POM defaults

## Test plan

* `ProjectBuilderTest#testModularWithJavaSourcesRejectsLegacySourceDirectory` passes
* `ProjectBuilderTest#testModularWithPhysicalDefaultLegacyDirectory` passes
* `ProjectBuilderTest#testClassicSourcesWithExplicitLegacyDirectories` passes
* `ProjectBuilderTest#testNonModularResourcesOnlyWithImplicitJavaFallback` passes
* `ProjectBuilderTest#testNonModularResourcesOnlyWithExplicitLegacyDirectoriesRejected` passes

Fixes #11701
See https://github.com/apache/maven/issues/11701#issuecomment-3858462609

### DIFF
--- a/impl/maven-core/src/main/java/org/apache/maven/project/DefaultProjectBuilder.java
+++ b/impl/maven-core/src/main/java/org/apache/maven/project/DefaultProjectBuilder.java
@@ -698,11 +698,14 @@ public class DefaultProjectBuilder implements ProjectBuilder {
                      - All legacy directories are used
 
                   2. MODULAR projects (have <module> in <sources>):
-                     - ALL legacy directories are rejected (cannot dispatch between modules)
-                     - Physical presence of default directories (src/main/java) also triggers ERROR
+                     - ALL legacy directories cause the build to fail (cannot dispatch
+                       between modules)
+                     - The build also fails if default directories (src/main/java)
+                       physically exist on the filesystem
 
                   3. NON-MODULAR projects with <sources>:
-                     - Explicit legacy directories (differ from default) are always rejected
+                     - Explicit legacy directories (differ from default) always cause
+                       the build to fail
                      - Legacy directories for scopes where <sources> defines Java are ignored
                      - Legacy directories for scopes where <sources> has no Java serve as
                        implicit fallback (only if they match the default, e.g., inherited)


ascheman:
Reviews are welcome, in particular from @elharo and @desruisseaux - thanks!

impl/maven-core/src/main/java/org/apache/maven/project/DefaultProjectBuilder.java
```diff
@@ -693,27 +693,27 @@ private void initProject(MavenProject project, ModelBuilderResult result) {
...
                  * `sourceDirectory`, `testSourceDirectory` and `scriptSourceDirectory`
-                 * are ignored if the POM file contains at least one enabled <source> element
+                 * are not used if the POM file contains at least one enabled <source> element
                  * for the corresponding scope and language. This rule exists because
                  * Maven provides default values for those elements which may conflict
                  * with user's configuration.
                  *
                  * Additionally, for modular projects, legacy directories are unconditionally
-                 * ignored because it is not clear how to dispatch their content between
+                 * rejected because it is not clear how to dispatch their content between
```
elharo: I don't think this is what we were looking for. The build fails completely, not simply a warning

impl/maven-core/src/main/java/org/apache/maven/project/DefaultProjectBuilder.java
```diff
@@ -906,15 +906,17 @@ private void initProject(MavenProject project, ModelBuilderResult result) {
...
+         * "Present" means either:
          * <ul>
-         *   <li>Case 1: The default legacy directory exists on the filesystem (e.g., src/main/java exists)</li>
-         *   <li>Case 2: An explicit legacy directory is configured that differs from the default</li>
+         *   <li><b>Configuration presence</b>: an explicit configuration differs from the default</li>
+         *   <li><b>Physical presence</b>: the default directory exists on the filesystem</li>
          * </ul>
-         * Legacy directories are unconditionally ignored in modular projects because it is not clear
-         * how to dispatch their content between different modules.
+         * In both cases, the legacy directory conflicts with modular sources and must not be used.
```
elharo: not "and must not be used" but rather the complete build fails.

impl/maven-core/src/main/java/org/apache/maven/project/SourceHandlingContext.java
```diff
@@ -221,11 +221,19 @@ void handleResourceConfiguration(ProjectScope scope) {
             if (hasResourcesInSources) {
                 // Modular project with resources configured via <sources> - already added above
                 if (hasExplicitLegacyResources(resources, scopeId)) {
-                    LOGGER.warn(
-                            "Legacy {} element is ignored because {} resources are configured via {} in <sources>.",
-                            legacyElement,
-                            scopeId,
-                            sourcesConfig);
+                    String message = String.format(
+                            "Legacy %s element must not be used because %s resources are configured via %s in <sources>.",
```
elharo: must not be used --> cannot be used in modular projects

impl/maven-core/src/main/java/org/apache/maven/project/SourceHandlingContext.java
```diff
@@ -236,13 +244,13 @@ void handleResourceConfiguration(ProjectScope scope) {
                 // Modular project without resources in <sources> - inject module-aware defaults
                 if (hasExplicitLegacyResources(resources, scopeId)) {
                     String message = "Legacy " + legacyElement
-                            + " element is ignored because modular sources are configured. "
+                            + " element must not be used because modular sources are configured. "
```
elharo: must not --> cannot

impl/maven-core/src/main/java/org/apache/maven/project/SourceHandlingContext.java
```diff
@@ -265,11 +273,19 @@ void handleResourceConfiguration(ProjectScope scope) {
             if (hasResourcesInSources) {
                 // Resources configured via <sources> - already added above
                 if (hasExplicitLegacyResources(resources, scopeId)) {
-                    LOGGER.warn(
-                            "Legacy {} element is ignored because {} resources are configured via {} in <sources>.",
-                            legacyElement,
-                            scopeId,
-                            sourcesConfig);
+                    String message = String.format(
+                            "Legacy %s element must not be used because %s resources are configured via %s in <sources>.",
```
elharo: cannot

impl/maven-core/src/test/java/org/apache/maven/project/ProjectBuilderTest.java
```diff
@@ -422,19 +422,21 @@ void testModularSourcesInjectResourceRoots() throws Exception {
...
      * <p>
      * This verifies the behavior described in the design:
-     * - Modular projects with explicit legacy {@code <resources>} configuration should issue a warning
+     * - Modular projects with explicit legacy {@code <resources>} configuration should raise an error
      * - The modular resource roots are injected instead of using the legacy configuration
      * <p>
-     * Acceptance Criterion: AC2 (unified source tracking for all lang/scope combinations)
+     * Acceptance Criteria:
+     * - AC2 (unified source tracking for all lang/scope combinations)
+     * - AC8 (legacy directories error - supersedes AC7 which originally used WARNING)
```
elharo, line 434: not sure how the AC are being used, but shouldn't AC7 be changed rather than superseded?

impl/maven-core/src/test/java/org/apache/maven/project/ProjectBuilderTest.java
```diff
@@ -478,23 +480,23 @@ void testModularSourcesWithExplicitResourcesIssuesWarning() throws Exception {
     }
 
     /**
-     * Tests that legacy sourceDirectory and testSourceDirectory are ignored in modular projects.
+     * Tests that legacy sourceDirectory and testSourceDirectory raise an error in modular projects.
      * <p>
-     * In modular projects, legacy directories are unconditionally ignored because it is not clear
-     * how to dispatch their content between different modules. A warning is emitted if these
-     * properties are explicitly set (differ from Super POM defaults).
+     * In modular projects, legacy directories must not occur because it is not clear
```
elharo: In modular projects, legacy directories must not occur  --> Legacy directories cannot be used in modular projects

impl/maven-core/src/main/java/org/apache/maven/project/DefaultProjectBuilder.java
desruisseaux: Proposal: replace _"A warning is emitted if …"_ by _"The build fails if …"_.

impl/maven-core/src/main/java/org/apache/maven/project/DefaultProjectBuilder.java
desruisseaux: > not "and must not be used" but rather the complete build fails.

Isn't what the next line below is saying?

impl/maven-core/src/main/java/org/apache/maven/project/DefaultProjectBuilder.java
```diff
@@ -693,27 +693,27 @@ private void initProject(MavenProject project, ModelBuilderResult result) {
 
                 /*
                  * `sourceDirectory`, `testSourceDirectory` and `scriptSourceDirectory`
-                 * are ignored if the POM file contains at least one enabled <source> element
+                 * are not used if the POM file contains at least one enabled <source> element
```
elharo: again here don't we need the build to fail in this case?

elharo (changes requested):

impl/maven-core/src/main/java/org/apache/maven/project/DefaultProjectBuilder.java
desruisseaux: Yes, I think that it is a change that was forgot in the documentation. The actual code raises an error as I read it.

ascheman: Looking at this more closely, I think we've uncovered an inconsistency:

**Current behavior:**
- **Resources** (`<resources>`/`<testResources>`): ERROR for **both** modular and classic projects when `<sources>` conflicts with legacy config (handled in `SourceHandlingContext`)
- **Source directories** (`<sourceDirectory>`/`<testSourceDirectory>`): ERROR only for **modular** projects, silently skipped for classic projects

**AC8** specifically scopes the error behavior to modular projects:
> "Legacy directories trigger ERROR in modular projects"

This leaves classic projects with `<sources>` in an inconsistent state - their legacy resources cause an error, but legacy source directories are silently ignored.

**Proposal:** If @desruisseaux and @elharo agree, we could add:

**AC9**: For classic (non-modular) projects, legacy `<sourceDirectory>` and `<testSourceDirectory>` also trigger an ERROR when `<sources>` elements are configured for the same scope/language. This ensures consistency with resource handling and prevents silent configuration loss.

Should I create a follow-up issue for this, or would you prefer to extend the current PR?

desruisseaux: Thanks for spotting that. I suggest to extend the current pull request if this is fine for you.

elharo: I'm not fully up to speed on all of this. I try to avoid modules. However, the general principle I would follow is that if the client project contains code or resources that the developer likely intended to be included and for some reason we are not including it, then Maven should hard fail the build.

desruisseaux: Yes, if I understood correctly @ascheman proposal, it is indeed the plan to make it a hard fail.

ascheman: I think we are totally with you, @elharo: Mixing things up, in particular if they lead to silent ignores (as in the first versions of the implementation) should lead to hard failures. IIRC we wanted to allow for a smooth migration if people start to work with the new module source hiearchy. However, as all of this is completely new (with Maven 4), we should take the opportunity to make such mixed POM specifications hard errors and not let people think they can somehow use both. They have to make a decision to either have a legacy layout or use the new opportunities.

Having said that, we seem to have forgot some combinations in the POM which are handled inconsistent. Happy for your critical thinking and detection of these during the discussion.

ascheman: While working on the change, I would like to re-formulate [AC9](https://github.com/apache/maven/issues/11701#issuecomment-3884782347) to be even more strict: Once you use `<sources>` on your project, you ~~must not~~ _cannot_ use legacy `<sourceDirectory>` / `<testSourceDirectory>` any longer, even if they point to the same implicit (e.g., `src/main/java`) or explicit directory (have the same value for the same directory/scope/language combination). What do you think, @elharo and @desruisseaux ?

desruisseaux: Fine for me.

ascheman: If you agree, we could even simplify AC8 (and won't need a new AC9) as the existence of `<sources>` would then mean that any existence of legacy configurations should fail the build (independent of the existence of modules):

## Revised AC8 Definition

**AC8** (Legacy directory validation for projects using `<sources>`):

For **any project** using `<sources>`, the build fails with an **ERROR** if:

1. An explicit legacy directory configuration (`<sourceDirectory>` or `<testSourceDirectory>`) exists, OR

2. A `<source>` element specifies a custom `<directory>` AND the default legacy directory (e.g., `src/main/java`) exists physically on the filesystem

### Key Point

**The presence of `<sources>` is the trigger, not whether the project is modular or non-modular.**

When a project uses `<sources>`, it opts into the Maven 4.1.0 source model entirely. Legacy directory configurations become invalid regardless of modularity.

desruisseaux: Seems very good to me.

ascheman: > Seems very good to me.

OK, I will implement it this way then.

ascheman: And yet another use case: Assume we have the following structure:

```xml
<build>
    <sources>
      <source>
        <scope>main</scope>
        <lang>resources</lang>
        <directory>src/main/custom-resources</directory>
      </source>
      <!-- No <source lang="java"> - expects implicit src/main/java? -->
    </sources>
  </build>
```

I would think in a legacy directory structure but 4.1.0-model configuration (but no modules), the legacy directories should still be (implicitly) used, unless explicitly overridden. What do you think, @desruisseaux ?

Said that: the classic values were defined as default by the super POM. Do we expect similar defaults by the super POM for `<sources>` as well (currently not configured)?

desruisseaux: Yes, when the `<module>` element is not used I think that we can continue to use the legacy directories.

Regarding the default values specified in the super POM, I suggest to not do any change for now (we may revisit in some future version). The default values of `<source>` elements are implemented in Java code for now, and they are designed in such a way that, when no `<module>` element is used, they match the default values of the super POM.

ascheman: OK, I fully agree that we should not require something on the Super POM by now but just use the classic directories as fallback if nothing is explicitly configured. Nevertheless, I will create a new Acceptance Criterion to express that design choice - we are back at (another) AC9 then 😉:

## AC9 Definition (Non-Modular Projects)

**AC9** (Legacy directory handling for non-modular projects):

For **non-modular projects** using `<sources>`:

1. **When `<sources>` has Java for a scope**: Legacy directories are rejected if they differ from the default (explicit configuration detected)
2. **When `<sources>` has no Java for a scope**: Legacy directories are used as **implicit fallback** only if they match the default (could be inherited)

_I already made some progress implementing and testing this and will deliver test cases for both AC8 and AC9_

impl/maven-core/src/test/java/org/apache/maven/project/ProjectBuilderTest.java
ascheman: As AC7 has already become part of the `master` branch implementation, I do not want to override it now. I consider it better to invalidate it and write a new one. As they all belong to proper legacy source/resource handling wrt to the new `<sources>` configuration element, I proceed with the numbering. I am also preparing some overall documentation, covering all aspects (an update/replacement for [the Confluence page](https://cwiki.apache.org/confluence/x/84TMFw) where it all began.

impl/maven-core/src/main/java/org/apache/maven/project/DefaultProjectBuilder.java
```diff
@@ -1008,7 +1008,7 @@ private void failIfLegacyDirectoryPresent(
                             "Legacy directory '%s' exists but cannot be used in project %s "
                                     + "because sources are configured via <sources>. "
                                     + "Remove or rename the directory.",
-                            defaultPath, projectId);
+                            defaultPath.toString().replace('\\', '/'), projectId);
```
desruisseaux: I suggest to not do this replacement in this code, but in the test instead. It would allow Windows users to see the paths that they are used to.

impl/maven-core/src/main/java/org/apache/maven/project/DefaultProjectBuilder.java
```diff
@@ -692,49 +692,102 @@ private void initProject(MavenProject project, ModelBuilderResult result) {
...
                  *
-                 * Additionally, for modular projects, legacy directories are unconditionally
-                 * ignored because it is not clear how to dispatch their content between
-                 * different modules. A warning is emitted if these properties are explicitly set.
+                 * 1. CLASSIC projects (no <sources>):
+                 *    - All legacy directories are used
+                 *
+                 * 2. MODULAR projects (have <module> in <sources>):
+                 *    - ALL legacy directories are rejected (can't dispatch between modules)
+                 *    - Physical presence of default directories (src/main/java) also triggers ERROR
```
elharo: I could well be missing something since I avoid modules like the plague, but why doesn't this go into the unnamed module?

impl/maven-core/src/main/java/org/apache/maven/project/DefaultProjectBuilder.java
```diff
@@ -692,49 +692,102 @@ private void initProject(MavenProject project, ModelBuilderResult result) {
...
+                 *
+                 * 3. NON-MODULAR projects with <sources>:
+                 *    - Explicit legacy directories (differ from default) are always rejected
```
elharo: This is current behavior, right?

impl/maven-core/src/main/java/org/apache/maven/project/DefaultProjectBuilder.java
```diff
@@ -692,49 +692,102 @@ private void initProject(MavenProject project, ModelBuilderResult result) {
...
+                 *    - If <sources> has Java for a scope: legacy is not used (even if matching default)
```
elharo: This bullet point needs to be rewritten. I cannot parse it.

impl/maven-core/src/main/java/org/apache/maven/project/DefaultProjectBuilder.java
```diff
@@ -692,49 +692,102 @@ private void initProject(MavenProject project, ModelBuilderResult result) {
...
+                 *    - If <sources> has no Java for a scope: legacy is used as implicit fallback
```
elharo: ditto

impl/maven-core/src/main/java/org/apache/maven/project/DefaultProjectBuilder.java
```diff
@@ -692,49 +692,102 @@ private void initProject(MavenProject project, ModelBuilderResult result) {
...
-                 * for the corresponding scope and language. This rule exists because
-                 * Maven provides default values for those elements which may conflict
-                 * with user's configuration.
+                 * Source directory handling depends on project type and <sources> configuration:
                  *
-                 * Additionally, for modular projects, legacy directories are unconditionally
-                 * ignored because it is not clear how to dispatch their content between
-                 * different modules. A warning is emitted if these properties are explicitly set.
+                 * 1. CLASSIC projects (no <sources>):
+                 *    - All legacy directories are used
```
elharo: probably need to use HTML in the javadoc comments for lists

impl/maven-core/src/main/java/org/apache/maven/project/DefaultProjectBuilder.java
```diff
@@ -906,44 +959,60 @@ private void initProject(MavenProject project, ModelBuilderResult result) {
...
 
         /**
-         * Warns about legacy directory usage in a modular project. Two cases are handled:
+         * Validates that legacy directory configuration does not conflict with {@code <sources>}.
+         * <p>
+         * When {@code <sources>} is configured, the build fails if:
          * <ul>
-         *   <li>Case 1: The default legacy directory exists on the filesystem (e.g., src/main/java exists)</li>
-         *   <li>Case 2: An explicit legacy directory is configured that differs from the default</li>
+         *   <li><b>Configuration presence</b>: an explicit legacy configuration differs from the default</li>
```
elharo: probably don't use &lt;b&gt; here

elharo (changes requested):
Is there documentation for this somewhere? I feel like I'm trying to read the tea leaves to understand what's going on. I'm missing the big picture.

In particular I would like to understand how sources are expected to be laid out in the file system and how that maps to different elements in pom.xml.

I'm beginning to worry that the design for this is much more complex and error prone than it needs to be. Why are we failing when there;s a src/main/.java present? Isn't that just the unnamed module?

impl/maven-core/src/main/java/org/apache/maven/project/DefaultProjectBuilder.java
desruisseaux: This is not Javadoc, but an ordinary comment inside the method body. The comment starts with `/*` instead of `/**`, which tells the compiler that this is not javadoc.

elharo: OK, In that case I suggest dropping all the preceding asterisks that make it look like Javadoc. They actually make the text harder to read.

impl/maven-core/src/main/java/org/apache/maven/project/DefaultProjectBuilder.java
desruisseaux: The concept of unnamed modules usually applies to dependencies: a JAR placed on the class-path is an unnamed module. This concept does not really applies to the sources to compile, unless we interpret "unnamed module" as "sources without module-info".

We cannot mix sources for Java modules and sources without module-info in the same compilation unit. Either the Maven sub-project is fully modular, or either it does not use module source hierarchy at all. From  [javac man page](https://docs.oracle.com/en/java/javase/25/docs/specs/man/javac.html#additional-source-files) about the `--source-path` and `--module-source-path` options:

> You can only specify one of these options: if you are not compiling code for a module, or if you are only compiling code for a single module, use the source path; if you are compiling code for multiple modules, use the module source path.

Therefore, a Maven project using `<module>` cannot accept unnamed module. If a developer wants an unnamed module, it must be a separated Maven sub-project.

impl/maven-core/src/main/java/org/apache/maven/project/DefaultProjectBuilder.java
desruisseaux: If by current behaviour we mean Maven 3 behaviour, the current behaviour is to take the source directories from the following elements: `<sourceDirectory>`, `<scriptSourceDirectory>`, `<testSourceDirectory>`, `<resources>`, `<testResources>`. This is what we mean by "legacy directories". The `<source>` element did not existed in Maven 3.

The new `<source>` element provides an unified handling capable to replace all the legacy elements, and more. But we think that if a developer mixes the legacy elements with the new `<source>` element, while it would be technically possible to merge the legacy elements with the `<source>` elements, such mix is more likely to be a configuration error. This is what this item tries to catch.

For projects that do not use the new `<source>` at all, nothing changed.

impl/maven-core/src/main/java/org/apache/maven/project/DefaultProjectBuilder.java
elharo: So is the rule one maven project/subproject per module? My mental model is that there are sources for each module, including the unnamed module. The pom is describing how to produce jars. There is, I think, a 1:1 relationship between jars and modules although there can be multiple jars for the unamed/no-module case. So is the pom describing how to build one exactly one jar for one module?

Again, documentation -- not code -- could help a lot here.

desruisseaux: Using the new proposed terminology where "module" = "Java module" and "sub-project" = "what we were used to call module in Maven 3", then:

* Yes, there is a 1:1 relationship between JARs and modules, i.e.  **Java** modules.
  * This is true in all circumstances, both _package hierarchy_ and _module hierarchy_.
* There is a 1:1 relationship between Maven sub-projects and Java modules when _package hierarchy_ is used.
* Therefore, there is a 1:1 relationship between Maven sub-projects and JARs when _package hierarchy_ is used.
* There is no longer a 1:1 relationship between Maven sub-project and Java modules when _module hierarchy_ is used: a single Maven sub-project can create many Java modules.
* Therefore, there is no longer a 1:1 relationship between Maven sub-project and JARs when _module hierarchy_ is used: a single Maven sub-project can create many JARs.

Package hierarchy is the classical layout of Java source files: _package/class.java_. Module hierarchy is the same as package hierarchy, but prefixed with the module name: _module/package/class.java_. Modules hierarchy are supported by `java`, `javac`, `javadoc` and some other tools (this is not a concept invented by Maven), sometime with distinct options such as `--source-path` versus `--module-source-path` (for package and module hierarchy respectively).

In Maven, the rule is straightforward: if the `<source>` elements contain a `<module>` child element, then the sub-project uses module hierarchy. Otherwise, the sub-project uses package hierarchy. There is no special cases or other complexity, the rule is as straightforward as that.

But another rule is that we cannot mix the two hierarchies: each sub-project must choose one or the other (but different Maven sub-projects can use different hierarchy). However, that latter rule was not enforced by Maven core. If a sub-project breaks that rule, the problem was detected only in the compiler plugin, sometime only when that plugin tries to pass options to `javac`. On the principle that errors are easier to resolve when they are detected early, this pull request tries do detect in advance sub-projects that would fail to build.

This pull request is not technically mandatory: we could perform no verification in Maven core and wait for the Maven Compiler Plugin to fail. This is an attempt to help the users with earlier error report, and also because Maven core has information that the Maven Compiler Plugin does not have: it is possible in Maven core to report the exact line in the `pom.xml` where the problem is detected (although the current pull request does no go that far, but it can be improved in future cycles).

elharo: This is useful information. I will repeat that this should be in the docs. I still don't have a full picture in my head of what's in the pom.xml and how that now relates what files are where in the project. I don't think we need to think about subprojects just yet. If the docs explained how a single project with two named modules and code in the unnamed module needs to be set up, that probably covers 90% of the complexity. Part of this might be the JPMS version of 

https://maven.apache.org/guides/introduction/introduction-to-the-standard-directory-layout.html

Does this exist yet?

The other part would be the JPMS version of the Maven POM reference: https://maven.apache.org/pom.html

Does this exist yet?

desruisseaux: Some documentation of the compiler plugin 4.x cover source directory layout, in particular:

* [Declaration of source directories](https://maven.apache.org/plugins/maven-compiler-plugin-4.x/sources.html)
* [Modular project](https://maven.apache.org/plugins/maven-compiler-plugin-4.x/modules.html)

This documentation has not yet been reflected in the main Maven documentation such as "Introduction to the Standard Directory Layout". A reason is that while the work is completed on the compiler plugin side it is not finished on JAR plugin side (pull request ready, waiting for the next Maven 4 release before to submit for merge) and Surefire plugin side (waiting for the ongoing big simplification).

elharo: OK, I think I'm starting to see the issue. But I'm not yet convinced that having a legacy source base and a modular source base si a deal breaker. Could we run the compiler twice? Once to produce the no-module jar and once to produce the modularized jars passing different sources each time?

It is possible to have Java projects that contain both modular and non-modular jars on the classpath/module-path. So are we requiring that these different jars be in separate Maven projects?

Traditionally a Maven pom.xml produced exactly one library jar (and maybe a test jar, source jars, and so forth but we can ignore that for now). Now with modules we have one pom.xml producing multiple jars. So why can't we take the legacy code in the traditional src/main/java location and put it in the unnamed module? That is, why are ALL legacy directories rejected in a modular project?

ascheman: Well, thats the point: The compiler cannot handle modular and non-modular sources at the same time. We would need two compile passes. 

Therefore, we designed it in a way that you have to choose, either this way or the other (but still with as much convention over configuration as possible - the good old Maven way). If you want both, make two (or more) `<subprojects>` (known as Maven `<modules>` up to Maven 3). This keeps concerns separated and (rather) simple. Imagine the complexity we need if we want to combine both approaches. It is already hard to handle (and describe) as we see here.

desruisseaux: When we compile a multi-module project by a single call to `javac`, we let the Java compiler manages the dependencies between modules. This is one of the reasons for supporting this feature, because Maven dependencies are restricted to acyclic graphs while Java module dependencies accept a limited form of cyclic dependencies (e.g. in qualified exports).

If we run the compiler twice for producing the no-module JAR, we would have to manage its dependency relatively to the modular JARs. Do we compile the legacy source base before the modules, then add the legacy output on the class-path of all modules? Or do we compile the modules first, then add them on the module-path for compiling the legacy source? In the latter case, do we add all modules or a subset of them? There is no `module-info` for telling us which subset of the `pom.xml` dependencies is effective for the legacy JAR.

It was a choice to stick to "one sub-project = one compilation unit" for the main code (ignoring tests and multi-release for this discussion). By requiring legacy source base to be compiled in its own Maven sub-project, we let the developers decide if the legacy source base depends on the modules or the converse, which modules, whether a module is optional, and whether to place them on the class-path or module-path. By contrast, if the Maven Compiler Plugin did two compilations for supporting modular sources and legacy sources in the same sub-project, we would be forced to take many arbitrary decisions.

impl/maven-core/src/main/java/org/apache/maven/project/SourceHandlingContext.java
ascheman: Fixed in an earlier push.

impl/maven-core/src/main/java/org/apache/maven/project/SourceHandlingContext.java
ascheman: Fixed in an earlier push.

impl/maven-core/src/main/java/org/apache/maven/project/SourceHandlingContext.java
ascheman: Fixed in an earlier push.

impl/maven-core/src/test/java/org/apache/maven/project/ProjectBuilderTest.java
ascheman: Fixed in an earlier push.

impl/maven-core/src/main/java/org/apache/maven/project/DefaultProjectBuilder.java
ascheman: Reworded in a5ad905. Changed to "Legacy directories for scopes where `<sources>` defines Java are ignored". Also dropped leading asterisks from the block comment per your suggestion.

impl/maven-core/src/main/java/org/apache/maven/project/DefaultProjectBuilder.java
ascheman: Same commit (a5ad905). Reworded to "Legacy directories for scopes where `<sources>` has no Java serve as implicit fallback (only if they match the default, e.g., inherited)".

impl/maven-core/src/main/java/org/apache/maven/project/DefaultProjectBuilder.java
ascheman: Replaced `<b>` with `<strong>` for better semantics in a5ad905.

impl/maven-core/src/main/java/org/apache/maven/project/DefaultProjectBuilder.java
ascheman: Done in a5ad905. Dropped leading asterisks from the block comment.

impl/maven-core/src/main/java/org/apache/maven/project/DefaultProjectBuilder.java
ascheman: The comment block has been completely rewritten — the old "warning" wording is gone. The code now raises an ERROR and the comment reflects that.

impl/maven-core/src/main/java/org/apache/maven/project/DefaultProjectBuilder.java
ascheman: The "must not be used" wording is gone. The code raises an ERROR and the rewritten comment/Javadoc reflects that.

impl/maven-core/src/main/java/org/apache/maven/project/DefaultProjectBuilder.java
```diff
@@ -692,49 +692,102 @@ private void initProject(MavenProject project, ModelBuilderResult result) {
...
-                 * different modules. A warning is emitted if these properties are explicitly set.
-                 */
-                if (!sourceContext.hasSources(Language.SCRIPT, ProjectScope.MAIN)) {
+                  Source directory handling depends on project type and <sources> configuration:
+
+                  1. CLASSIC projects (no <sources>):
+                     - All legacy directories are used
+
+                  2. MODULAR projects (have <module> in <sources>):
+                     - ALL legacy directories are rejected (cannot dispatch between modules)
```
elharo: what does rejected mean? I think you need to be clear that the build simply fails or that an exception is thrown or whatever.

impl/maven-core/src/main/java/org/apache/maven/project/DefaultProjectBuilder.java
```diff
@@ -692,49 +692,102 @@ private void initProject(MavenProject project, ModelBuilderResult result) {
...
+                     - Physical presence of default directories (src/main/java) also triggers ERROR
```
elharo: triggers ERROR --> ????

what does this actually mean? Are you saying the build fails? an exception is thrown? some error handler is invoked? Be explicit and direct.

impl/maven-core/src/main/java/org/apache/maven/project/DefaultProjectBuilder.java
```diff
@@ -692,49 +692,102 @@ private void initProject(MavenProject project, ModelBuilderResult result) {
...
+
+                  3. NON-MODULAR projects with <sources>:
+                     - Explicit legacy directories (differ from default) are always rejected
```
elharo: ditto

impl/maven-core/src/main/java/org/apache/maven/project/DefaultProjectBuilder.java
ascheman: Good point. Reworded in eb41bc90d3: "rejected" → "cause the build to fail".

impl/maven-core/src/main/java/org/apache/maven/project/DefaultProjectBuilder.java
ascheman: Fixed in eb41bc90d3. The comment now reads "The build also fails if default directories (src/main/java) physically exist on the filesystem".

impl/maven-core/src/main/java/org/apache/maven/project/DefaultProjectBuilder.java
ascheman: Same commit (eb41bc90d3). Changed to "Explicit legacy directories (differ from default) always cause the build to fail".

PR #11703, opened 2026-02-06
Eclipse Sisu 1.0.0-SNAPSHOT

cstamas:
To run ITs w/ latest Eclipse Sisu 1.0.0-SNAPSHOT:
https://github.com/eclipse-sisu/sisu-project/milestone/2

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -516,6 +516,18 @@ under the License.
       <url>https://central.sonatype.com/repository/maven-snapshots</url>
     </repository>
   </repositories>
+  <pluginRepositories>
+    <pluginRepository>
+      <releases>
+        <enabled>false</enabled>
+      </releases>
+      <snapshots>
+        <enabled>true</enabled>
+      </snapshots>
+      <id>central-snapshots</id>
+      <url>https://central.sonatype.com/repository/maven-snapshots</url>
+    </pluginRepository>
+  </pluginRepositories>
 
   <build>
     <pluginManagement>


pom.xml
```diff
@@ -125,7 +125,7 @@ under the License.
 
   <properties>
     <javaVersion>8</javaVersion>
-    <version.sisu-maven-plugin>0.9.0.M4</version.sisu-maven-plugin>
+    <version.sisu-maven-plugin>0.9.0-SNAPSHOT</version.sisu-maven-plugin>
```
Bukama, line 128: Didn't you mention a 1.0.0(-SNAPSHOT) yesterday?

kwin: Sisu is still on version 0.0.9-SNAPSHOT (https://github.com/eclipse-sisu/sisu-project/blob/93c62c42396addd034b15eb35e289ba167c9bec3/pom.xml#L21), but semantically this is 1.0.0-SNAPSHOT (i.e. next release is 1.0.0 soon).

pom.xml
kwin, line 139: @cstamas Is Guice 6.0.0 not yet compatible?

cstamas: Guice 6 is compatible, but users could assume we support jakarta inject as well, while Sisu is not. IMHO better like this: be clear about it, plus there is nothing we miss from guice 6

kwin: There are quite some bugfixes listed in https://github.com/google/guice/wiki/Guice600#guice-core which potentially affect usage in Maven as well.

slachiewicz: see some rationale behind downgrade here https://github.com/apache/maven/pull/2472

pom.xml
slachiewicz: looks like we have final version https://github.com/eclipse-sisu/sisu-project/releases/tag/releases%2F1.0.0

PR #11711, opened 2026-02-11
Bug: SecDispatcher is managed by legacy Plexus DI

cstamas:
And as Plexus DI is capable only of field injection, and 2.0 of SecDispatcher contains `final` field as it was converted to JSR330, it causes WARNINGs on Java 26 due JEP 500.

Fix: drop component from Plexus XML (this was our last managed Plexus component) and provide a drop in replacement, a provider w/ name "maven".

### DIFF
--- a/maven-core/src/main/java/org/apache/maven/internal/secdispatcher/SecDispatcherProvider.java
+++ b/maven-core/src/main/java/org/apache/maven/internal/secdispatcher/SecDispatcherProvider.java
@@ -31,9 +31,9 @@ import org.sonatype.plexus.components.sec.dispatcher.PasswordDecryptor;
 import org.sonatype.plexus.components.sec.dispatcher.SecDispatcher;
 
 /**
- * Provides "maven" named security dispatcher, used by Maven. This component is configured slightly differently than the
- * default one, as default is unaware of Maven specific configuration file locations. The default named component
- * should not be used (injected or referenced in any way) at all in any Maven related codebase.
+ * Provides a security dispatcher named "maven". Maven uses this security dispatcher instead of the default.
+ * The default security dispatcher is unaware of Maven specific configuration file locations and should not be used
+ * (injected or referenced in any way) at all in any Maven related codebase.
  * <p>
  * Note: This whole stuff is really deprecated and replaced with proper security in Maven 4, while this one is
  * just "security through obscurity".


maven-core/src/main/java/org/apache/maven/internal/secdispatcher/SecDispatcherProvider.java
```diff
@@ -0,0 +1,58 @@
...
+@Singleton
+@Named("maven")
+public class SecDispatcherProvider implements Provider<SecDispatcher> {
+    private final SecDispatcher instance;
+
+    @Inject
+    public SecDispatcherProvider(
+            PlexusCipher plexusCipher,
+            Map<String, PasswordDecryptor> decryptors,
+            @Named("${maven.settings.security.configurationFile:-~/.m2/settings-security.xml}")
```
gastaldi, line 52: Is the `-` before `~/.m2/settings-security.xml` intentional? Just out of curiosity

cstamas: Yes, Sisu specialty: `${key:-defValue}`:
https://eclipse.dev/sisu/apidocs/org/eclipse/sisu/wire/package-summary.html

cstamas: AFAIR, origin was bash https://linux.die.net/man/1/bash
Was long time ago :smile: 
```
${parameter:-word}
  Use Default Values. If parameter is unset or null, the expansion of word is substituted. Otherwise, the value of parameter is substituted.
```

maven-core/src/main/java/org/apache/maven/internal/secdispatcher/SecDispatcherProvider.java
```diff
@@ -0,0 +1,58 @@
...
+
+import java.util.Map;
+
+import org.sonatype.plexus.components.cipher.PlexusCipher;
+import org.sonatype.plexus.components.sec.dispatcher.DefaultSecDispatcher;
+import org.sonatype.plexus.components.sec.dispatcher.PasswordDecryptor;
+import org.sonatype.plexus.components.sec.dispatcher.SecDispatcher;
+
+/**
+ * Provides Maven own security dispatcher. It is configured slightly differently than the default one, as
```
elharo: "Maven own security dispatcher" is unclear. Consider rewriting

cstamas: It refers to the fact that managed component is not named "default", as it is almost always the case, but is named "maven". Hence "maven own" component. Am fine to reformulate this in whatever way you want. 

The (somewhat confusing) fact is that `plexus-sec-dispatcher.jar` itself carries a "default" version of SecDispatcher component (and is named "default"), as it contains Sisu index file, so is discovered and on request created and injected, but _it is not configured in the proper way_.

cstamas: Reformulated javadoc, to something hopefully better.

maven-core/src/main/java/org/apache/maven/internal/secdispatcher/SecDispatcherProvider.java
```diff
@@ -0,0 +1,61 @@
...
+
+import java.util.Map;
+
+import org.sonatype.plexus.components.cipher.PlexusCipher;
+import org.sonatype.plexus.components.sec.dispatcher.DefaultSecDispatcher;
+import org.sonatype.plexus.components.sec.dispatcher.PasswordDecryptor;
+import org.sonatype.plexus.components.sec.dispatcher.SecDispatcher;
+
+/**
+ * Provides "maven" named security dispatcher, used by Maven. This component is configured slightly differently than the
```
elharo: Provides a security dispatcher named "maven". Maven uses this security dispatcher instead of the default. The default security dispatcher is unaware of Maven specific configuration file locations and should not be used...

kwin:
But why not cherry-picking what we have in 4.x: compat/maven-settings-builder/src/main/java/org/apache/maven/settings/crypto/MavenSecDispatcher.java

PR #11741, opened 2026-02-24
Fix gh-11740: Add missing null check in validateProfileId()

### LINKED ISSUE #11740: Missing null check in DefaultModelValidator.validateProfileId() can cause NPE
### Affected version

4.1.0-SNAPSHOT (master)

### Bug description

In `impl/maven-impl/src/main/java/org/apache/maven/impl/model/DefaultModelValidator.java`, the `validateProfileId()` method is missing a null check before calling `validProfileIds.contains(id)`.

Since `validProfileIds` is a `ConcurrentHashMap.newKeySet()` (which doesn't allow null keys), passing a null profile ID will throw a `NullPointerException` instead of a proper validation error.

**Inconsistency:**

The same file has two similar validation methods with inconsistent null handling:

```java
// Line ~1743 - HAS null check (correct)
private boolean validateCoordinateId(..., String id, ...) {
    if (id != null && validCoordinatesIds.contains(id)) {
        return true;
    }
    // ...
}

// Line ~1794 - MISSING null check (bug)
private boolean validateProfileId(..., String id, ...) {
    if (validProfileIds.contains(id)) {  // Will throw NPE if id is null
        return true;
    }
    // ...
}
```

### Expected behavior

`validateProfileId()` should handle null IDs gracefully by either:
1. Adding a null check like `validateCoordinateId()` does: `if (id != null && validProfileIds.contains(id))`
2. Or letting the subsequent validation report the proper error message

### Steps to reproduce

1. Create a POM with a profile that has a null/missing ID
2. Run validation with Maven 4
3. Observe NPE instead of proper validation error

### Proposed fix

```java
// Change from:
if (validProfileIds.contains(id)) {

// To:
if (id != null && validProfileIds.contains(id)) {
```

This makes it consistent with `validateCoordinateId()` and the compat module fix in #11739.

---

*Discovered during code review of #11739*

abhu85:
## Summary
Add missing null check before `validProfileIds.contains(id)` in `validateProfileId()` to prevent `NullPointerException` and make it consistent with `validateCoordinateId()`.

## Problem
In `impl/maven-impl/.../DefaultModelValidator.java`, the `validateProfileId()` method calls `validProfileIds.contains(id)` without checking if `id` is null first.

Since `validProfileIds` uses `ConcurrentHashMap.newKeySet()` which doesn't allow null keys, a null profile ID would cause a `NullPointerException` instead of a proper validation error message.

**Inconsistency in the same file:**
```java
// Line 1744 - validateCoordinateId() has null check ✓
if (id != null && validCoordinatesIds.contains(id)) {

// Line 1795 - validateProfileId() was missing null check ✗
if (validProfileIds.contains(id)) {  // NPE if id is null
```

## Solution
Add the same null check pattern used in `validateCoordinateId()`:
```java
if (id != null && validProfileIds.contains(id)) {
```

## Test Plan
- [x] `mvn spotless:apply` - formatting verified
- [x] `mvn test -pl impl/maven-impl -Dtest=DefaultModelValidatorTest` - all 76 tests pass

## Compatibility
This is a backward-compatible fix. Null profile IDs will now be properly handled by `validateStringNotEmpty()` instead of throwing NPE.

Fixes #11740

---

### DIFF
--- a/impl/maven-impl/src/main/java/org/apache/maven/impl/model/DefaultModelValidator.java
+++ b/impl/maven-impl/src/main/java/org/apache/maven/impl/model/DefaultModelValidator.java
@@ -1792,7 +1792,7 @@ public class DefaultModelValidator implements ModelValidator {
             String id,
             @Nullable SourceHint sourceHint,
             InputLocationTracker tracker) {
-        if (validProfileIds.contains(id)) {
+        if (id != null && validProfileIds.contains(id)) {
             return true;
         }
         if (!validateStringNotEmpty(prefix, fieldName, problems, severity, version, id, sourceHint, tracker)) {


gnodet (commented):

LGTM. The fix is correct — `ConcurrentHashMap.newKeySet()` throws `NullPointerException` on `contains(null)`, so this null guard is necessary. It also restores consistency with the identical pattern in `validateCoordinateId()` at line 1744.

gnodet (approved):
LGTM — the null guard is necessary and correct.

PR #11742, opened 2026-02-26
Fix #11715: preserve 4.1.0 namespace/schema in help:effective-pom

### LINKED ISSUE #11715: Maven 4 (rc5) creates effective pom with wrong <project> header
### Affected version

4.0.0-rc5

### Bug description

Hi all.

With a minimal POM
```
<project xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xmlns="http://maven.apache.org/POM/4.1.0"
		 xsi:schemaLocation="http://maven.apache.org/POM/4.1.0 http://maven.apache.org/xsd/maven-4.1.0.xsd">
	<modelVersion>4.1.0</modelVersion>


	<groupId>tmp</groupId>
	<artifactId>tmp</artifactId>
	<version>tmp</version>
	<packaging>pom</packaging>


</project>
```

i noticed, that `mvn help:effective-pom`
generates a pom with **4.0.0** schemaLocation

```
Effective POMs, after inheritance, interpolation, and profiles are applied:

<!-- ====================================================================== -->
<!--                                                                        -->
<!-- Generated by Maven Help Plugin                                         -->
<!-- See: https://maven.apache.org/plugins/maven-help-plugin/               -->
<!--                                                                        -->
<!-- ====================================================================== -->
<!-- ====================================================================== -->
<!--                                                                        -->
<!-- Effective POM for project 'tmp:tmp:pom:tmp'                            -->
<!--                                                                        -->
<!-- ====================================================================== -->
  <modelVersion>4.1.0</modelVersion>
  <groupId>tmp</groupId>
  <artifactId>tmp</artifactId>
  <version>tmp</version>
  <packaging>pom</packaging>

  ...

```

I don't know what are the consequences, but i guess this should be corrected.

greetings
Ralf

arturobernalg:
When generating the effective POM for a modelVersion 4.1.0 project, preserve the root <project> namespace and schemaLocation as 4.1.0 instead of falling back to 4.0.0.

Add/keep coverage with MavenITgh11715EffectivePomNamespaceTest to verify the effective POM header contains:
- xmlns http://maven.apache.org/POM/4.1.0
- schemaLocation .../maven-4.1.0.xsd


gnodet (commented):

The core fix in `MavenXpp3Writer.configureDelegate()` correctly derives the namespace and schema location from the model version, mirroring the existing pattern in `TransformerSupport.write()`. The fallback chain (modelVersion -> namespaceUri -> default 4.0.0) is sensible. Tests and integration test cover the primary scenario well.

Two observations:

1. **Two concerns in one PR**: The `XmlService` classloader fallback change is a separate fix from the namespace preservation. Consider splitting it into its own PR/commit so each can be reviewed and reverted independently.

2. **Minor thread-safety note**: `configureDelegate()` mutates the shared `delegate` field before calling `write()`. This is not thread-safe if a single `MavenXpp3Writer` instance is used concurrently. This is a pre-existing issue (same as `setFileComment`/`setStringFormatter`), but worth noting since `configureDelegate` is now called on every `write`.

compat/maven-model/src/main/java/org/apache/maven/model/io/xpp3/MavenXpp3Writer.java
```diff
@@ -94,9 +103,28 @@ public void write(Writer writer, Model model) throws IOException {
...
+        }
+        if (version == null || version.isBlank()) {
+            String namespaceUri = model.getDelegate().getNamespaceUri();
+            if (namespaceUri != null && namespaceUri.startsWith(NAMESPACE_PREFIX)) {
+                version = namespaceUri.substring(NAMESPACE_PREFIX.length()).trim();
+            }
+        }
+        if (version == null || version.isBlank()) {
+            version = DEFAULT_MODEL_VERSION;
+        }
```
gnodet, line 114: The version extraction from the namespace URI is a nice defensive fallback, but the constants here duplicate values already defined in `TransformerSupport` (`NAMESPACE_FORMAT`, `SCHEMA_LOCATION_FORMAT`). Consider extracting them to a shared location to avoid drift.

api/maven-api-xml/src/main/java/org/apache/maven/api/xml/XmlService.java
```diff
@@ -239,8 +239,15 @@ private static XmlService getService() {
...
     /** Holder class for lazy initialization of the default instance */
     private static final class Holder {
-        static final XmlService INSTANCE = ServiceLoader.load(XmlService.class)
+        static final XmlService INSTANCE = ServiceLoader.load(XmlService.class, XmlService.class.getClassLoader())
                 .findFirst()
+                .or(() -> {
+                    ClassLoader contextClassLoader = Thread.currentThread().getContextClassLoader();
+                    return contextClassLoader != null
+                            ? ServiceLoader.load(XmlService.class, contextClassLoader)
+                                    .findFirst()
```
gnodet: This classloader fallback is a worthwhile robustness improvement, but it addresses a different problem than #11715 (namespace preservation). Consider splitting this into a separate commit/PR so the two fixes can be tracked and reverted independently.

Also, the order here (class classloader first, then TCCL) inverts the typical Java convention where TCCL is checked first. Is this intentional? In most ServiceLoader patterns the TCCL takes precedence because it represents the application's view. If the class's own classloader already sees the provider, the TCCL fallback will never trigger in practice -- which is fine for robustness, but worth documenting the intent.

gnodet (changes requested):

gnodet:
CI is now fully green. Summary of changes on this branch:

1. **`MavenXpp3Writer.configureDelegate`** — before writing the effective POM, sets the namespace and schemaLocation on the `MavenStaxWriter` delegate based on the model's declared `modelVersion`. Falls back to `MavenModelVersion.getModelVersion()` (feature-based minimum) when no version is explicitly declared.

2. **`ModelTest`** — two new unit tests verifying namespace output for both 4.1.0 and 4.0.0 models.

3. **`MavenITgh11715EffectivePomNamespaceTest`** — integration test that runs `help:effective-pom` on a 4.1.0 POM and verifies the output preserves the 4.1.0 namespace and schemaLocation.

Ready for review.

gnodet (approved):

PR #11743, opened 2026-02-26
[MNG-11449] Add Mockito javaagent for maven-cli tests

arturobernalg:
Configure surefire in impl/maven-cli to pass Mockito as a Java agent

- [X ] You have run the [Core IT][core-its] successfully.

### DIFF
--- a/impl/maven-cli/pom.xml
+++ b/impl/maven-cli/pom.xml
@@ -308,6 +308,7 @@ under the License.
         <groupId>org.apache.maven.plugins</groupId>
         <artifactId>maven-surefire-plugin</artifactId>
         <configuration>
+          <argLine>-Xmx256m @{jacocoArgLine} -javaagent:${org.mockito:mockito-core:jar}</argLine>
           <properties>
             <configurationParameters>junit.jupiter.testclass.order.default = org.junit.jupiter.api.ClassOrderer$OrderAnnotation</configurationParameters>
           </properties>
--- a/impl/maven-cli/src/test/java/org/apache/maven/cling/invoker/mvnup/goals/MockitoAgentConfigurationTest.java
+++ b/impl/maven-cli/src/test/java/org/apache/maven/cling/invoker/mvnup/goals/MockitoAgentConfigurationTest.java
@@ -1,0 +1,41 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.maven.cling.invoker.mvnup.goals;
+
+import java.lang.management.ManagementFactory;
+import java.util.List;
+
+import org.junit.jupiter.api.DisplayName;
+import org.junit.jupiter.api.Test;
+
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+@DisplayName("Mockito agent configuration")
+class MockitoAgentConfigurationTest {
+
+    @Test
+    @DisplayName("should run tests with mockito configured as javaagent")
+    void shouldRunTestsWithMockitoAsJavaAgent() {
+        List<String> inputArguments = ManagementFactory.getRuntimeMXBean().getInputArguments();
+
+        assertTrue(
+                inputArguments.stream().anyMatch(a -> a.startsWith("-javaagent:") && a.contains("mockito-core")),
+                () -> "Expected Mockito javaagent in JVM arguments, but got: " + inputArguments);
+    }
+}


gnodet (commented):

The parent pom already has a mechanism for this: `maven-dependency-plugin:properties` sets `org.mockito:mockito-core:jar`, which activates the `mockito` profile that adds `-javaagent:${org.mockito:mockito-core:jar}` to surefire's argLine (lines 942-961 of the root pom).

If that mechanism isn't working for `maven-cli`, it would be better to fix the root cause rather than hardcoding the argLine in this module. Note that the `mockito` profile in the parent pom (line 955) itself has a pre-existing issue: it sets `<argLine>-Xmx256m -javaagent:${org.mockito:mockito-core:jar}</argLine>`, which drops `@{jacocoArgLine}` compared to the default (line 700). That means JaCoCo coverage is silently broken whenever the mockito profile activates.

Could you check why the parent-level profile doesn't activate for `maven-cli`? The module already has `mockito-core` as a test dependency, so the `properties` goal should set the property and the profile should activate.

Also, `MockitoAgentConfigurationTest` is a test that only verifies build infrastructure configuration — it doesn't test any production code. If the javaagent is missing, Mockito-based tests would fail with clear errors anyway, making this test redundant.

impl/maven-cli/pom.xml
```diff
@@ -308,6 +308,7 @@ under the License.
         <groupId>org.apache.maven.plugins</groupId>
         <artifactId>maven-surefire-plugin</artifactId>
         <configuration>
+          <argLine>-Xmx256m @{jacocoArgLine} -javaagent:${org.mockito:mockito-core:jar}</argLine>
```
gnodet, line 311: This hardcodes the argLine at the module level, bypassing the parent pom's `mockito` profile mechanism (root pom lines 942-961). The parent already has `maven-dependency-plugin:properties` + a profile that adds `-javaagent:${org.mockito:mockito-core:jar}` to surefire.

If the parent mechanism doesn't work here, it would be better to fix it at the parent level. Also note that the parent's `mockito` profile (line 955) itself drops `@{jacocoArgLine}` — that's a pre-existing bug worth fixing regardless.

impl/maven-cli/src/test/java/org/apache/maven/cling/invoker/mvnup/goals/MockitoAgentConfigurationTest.java
```diff
@@ -0,0 +1,41 @@
...
+    @Test
+    @DisplayName("should run tests with mockito configured as javaagent")
+    void shouldRunTestsWithMockitoAsJavaAgent() {
+        List<String> inputArguments = ManagementFactory.getRuntimeMXBean().getInputArguments();
+
+        assertTrue(
+                inputArguments.stream().anyMatch(a -> a.startsWith("-javaagent:") && a.contains("mockito-core")),
+                () -> "Expected Mockito javaagent in JVM arguments, but got: " + inputArguments);
+    }
+}
```
gnodet, line 41: This test only verifies build infrastructure (that the javaagent JVM flag is present). It doesn't test any production code. If the mockito agent were missing, any test using `Mockito.mock()` on a non-interface would already fail with a clear error, making this test redundant.

I'd suggest removing this file — the real fix is just the surefire configuration.

gnodet (commented):
## Review Summary

This PR adds a hardcoded Mockito javaagent to the `maven-cli` module's surefire `<argLine>`. However, **this issue has already been fixed on master** via commit `6f4644329f` (merged 2026-07-03, PR #12369), which introduces a superior project-wide mechanism.

### Why this PR is superseded

The master branch now has:
- A `mockito` profile with `maven-dependency-plugin:properties` to resolve the jar path
- Late-binding `@{org.mockito:mockito-core:jar}` interpolation (not `${...}`)
- A composed argLine (`${argLine.xmx} @{jacocoArgLine} ${argLine.mockito}`) in the parent `pluginManagement` that applies to **all modules** and both surefire and failsafe

This PR's approach:
- Uses eager `${...}` interpolation (the exact bug the master fix corrected)
- Only targets surefire in one module (`maven-cli`)
- Overrides the parent's argLine entirely, losing future parent changes

### Recommendation

This PR should be closed as superseded by #12369.

---

impl/maven-cli/pom.xml
gnodet, line 311: Two issues with this line:

1. **Overrides parent argLine**: On master (post-PR #12369), the parent `pluginManagement` sets `argLine` to `${argLine.xmx} @{jacocoArgLine} ${argLine.mockito}`, composing three properties. This per-module override replaces that entire mechanism, losing the ability to set `argLine.xmx` or `argLine.mockito` via profiles.

2. **Eager interpolation**: Uses `${org.mockito:mockito-core:jar}` which resolves at POM parse time — before `dependency:properties` runs. PR #12369 explicitly fixed this project-wide by switching to late-binding `@{org.mockito:mockito-core:jar}`.

```suggestion
          <argLine>${argLine.xmx} @{jacocoArgLine} ${argLine.mockito}</argLine>
```
(But even this is unnecessary — the parent already provides it.)

impl/maven-cli/src/test/java/org/apache/maven/cling/invoker/mvnup/goals/MockitoAgentConfigurationTest.java
```diff
@@ -0,0 +1,41 @@
...
+import org.junit.jupiter.api.Test;
+
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+@DisplayName("Mockito agent configuration")
+class MockitoAgentConfigurationTest {
+
+    @Test
+    @DisplayName("should run tests with mockito configured as javaagent")
+    void shouldRunTestsWithMockitoAsJavaAgent() {
```
gnodet, line 34: This test verifies build infrastructure configuration (javaagent presence) rather than production code. If the Mockito javaagent were missing, every test using `mock()` or `@Mock` on final classes would fail with a clear `IncompatibleClassChangeError` or `MockitoException`, making this infrastructure check redundant.

This test also does not exist on master after #12369.

gnodet (changes requested):
This PR appears to be fully superseded by PR #12369 (commit 6f4644329f, merged 2026-07-03), which addresses the same MNG-11449 issue with a superior project-wide mechanism using composable `argLine` properties and late-binding interpolation.

All three issues below were independently verified:
1. **argLine override** — the per-module `<argLine>` shadows the parent's composable `${argLine.xmx} @{jacocoArgLine} ${argLine.mockito}` mechanism
2. **Eager interpolation** — uses `${org.mockito:mockito-core:jar}` (resolves at POM parse time) instead of `@{org.mockito:mockito-core:jar}` (late-binding, resolves after `dependency:properties`)
3. **Redundant test** — the infrastructure test verifies javaagent presence, but any Mockito test using `mock()` on final classes would fail with a clear error if the agent were missing

Recommendation: close as superseded by #12369.

slachiewicz:
The last CI run on this branch was 2026-02-26, and GitHub has deleted the logs since, so there's no way to tell whether the change still builds against master.

Issue #11767, opened 2026-03-09
Maven 4 (rc5) fails in maven-install-plugin due to consumer POM builder not resolving BOM imports from settings.xml profile repositories


DavidTavoularis:
### Affected version

4.0.0-rc5

### Bug description

Note that I asked Claude Code to help with the analysis and the fix.

```
$ mvn clean install
[...]
[INFO] --- install:3.1.3:install (default-install) @ xxx ---
[INFO] Artifact yyy is present in the local repository, but cached from a remote repository ID that is unavailable in current build context, verifying that is downloadable from [central (https://repo.maven.apache.org/maven2, default, releases)]
[INFO] --------------------------------------------------------------------------------------------------------------------------
[INFO] BUILD FAILURE
[INFO] --------------------------------------------------------------------------------------------------------------------------
[...]
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-install-plugin:3.1.3:install (default-install) on project xxx: Execution default-install of goal org.apache.maven.plugins:maven-install-plugin:3.1.3:install failed: org.apache.maven.api.services.ModelBuilderException: 12 problems were encountered while building the effective model for xxx
[ERROR]     - [ERROR] Non-resolvable import POM: The following artifacts could not be resolved: yyy (present, but unavailable): yyy was not found in https://repo.maven.apache.org/maven2 during a previous attempt. This failure was cached in the local repository and resolution is not reattempted until the update interval of central has elapsed or updates are forced (remote repositories: central (https://repo.maven.apache.org/maven2, default, releases)) @ line 22, column 7
[...]
```
In DefaultProjectBuilder (normal build), the ModelBuilderRequest is constructed with:
  - repositories(...) — passes remote repositories (including those from settings.xml profiles)
  - profiles(...) — passes profiles (from settings.xml)
  - activeProfileIds(...) — passes active profile IDs

```
In DefaultConsumerPomBuilder.buildModel() (used during install/deploy), none of these are set:
  // Line 225-238 - DefaultConsumerPomBuilder.java
  private ModelBuilderResult buildModel(RepositorySystemSession session, ModelSource src) {
      InternalSession iSession = InternalSession.from(session);
      ModelBuilderRequest.ModelBuilderRequestBuilder request = ModelBuilderRequest.builder();
      request.requestType(ModelBuilderRequest.RequestType.BUILD_CONSUMER);
      request.session(iSession);
      request.source(src);
      request.locationTracking(false);
[...]

PR #11768, opened 2026-03-09
[#11767] Fix consumer POM builder failing to resolve BOM imports from…

DavidTavoularis:
… settings.xml profile repositories

DefaultConsumerPomBuilder.buildModel() now passes repositories, profiles, and active profile IDs to the ModelBuilderRequest so that BOM imports from non-central repositories (e.g. defined in settings.xml profiles) can be resolved during consumer POM transformation.

DefaultModelBuilder.derive() now respects the request's repositories instead of always reusing the parent session's, and caches the RepositoryFactory lookup.

Includes unit tests for both fixes and an integration test (gh-11767).

- [] In any other case, please file an [Apache Individual Contributor License Agreement](https://www.apache.org/licenses/icla.pdf).

### DIFF
--- a/impl/maven-core/src/main/java/org/apache/maven/internal/transformation/impl/DefaultConsumerPomBuilder.java
+++ b/impl/maven-core/src/main/java/org/apache/maven/internal/transformation/impl/DefaultConsumerPomBuilder.java
@@ -334,6 +334,12 @@ class DefaultConsumerPomBuilder implements PomBuilder {
             }
             request.activeProfileIds(executionRequest.getActiveProfiles());
             request.inactiveProfileIds(executionRequest.getInactiveProfiles());
+        } else {
+            LOGGER.debug(
+                    "Session is not an InternalMavenSession ({}); settings.xml profiles will not be "
+                            + "passed to the consumer POM model builder. BOM imports from repositories "
+                            + "defined only in settings.xml profiles may fail to resolve.",
+                    iSession.getClass().getName());
         }
         ModelBuilder.ModelBuilderSession mbSession =
                 iSession.getData().get(SessionData.key(ModelBuilder.ModelBuilderSession.class));
--- a/impl/maven-core/src/main/java/org/apache/maven/internal/transformation/impl/DefaultConsumerPomBuilder.java
+++ b/impl/maven-core/src/main/java/org/apache/maven/internal/transformation/impl/DefaultConsumerPomBuilder.java
@@ -48,7 +48,9 @@ import org.apache.maven.api.services.ModelBuilderRequest;
 import org.apache.maven.api.services.ModelBuilderResult;
 import org.apache.maven.api.services.ModelSource;
 import org.apache.maven.api.services.model.LifecycleBindingsInjector;
+import org.apache.maven.execution.MavenExecutionRequest;
 import org.apache.maven.impl.InternalSession;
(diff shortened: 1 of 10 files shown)


gnodet (commented):
## Review

The fix correctly addresses the root cause: `DefaultConsumerPomBuilder.buildModel()` wasn't passing repositories/profiles to the `ModelBuilderRequest`, and `DefaultModelBuilder.derive()` was ignoring request repositories. The integration test is solid. A few observations:

**1. The `derive()` change is broad.** The repository override is applied in the general `derive(ModelBuilderRequest, DefaultModelBuilderResult)` method, which affects all derived sessions — not just BUILD_CONSUMER. This could have unintended side effects on other request types. Consider scoping the override to only apply for BUILD_CONSUMER requests.

**2. Lazy `RepositoryFactory` caching.** The new `getRepositoryFactory()` with lazy init adds a mutable field. This could be simplified by calling `session.getService(RepositoryFactory.class)` directly (likely already cached in the session), avoiding the extra state.

**3. The `instanceof InternalMavenSession` check is fragile.** Profile/activeProfileIds injection only works when the session is an `InternalMavenSession`. In embedded or testing contexts, profiles silently won't be passed. Worth documenting or handling explicitly.

**4. Unit test uses reflection.** `testDeriveSessionUsesRequestRepositories` accesses `mainSession`, `repositories`, and `externalRepositories` via reflection, making it brittle to field renames. Consider adding package-private accessors or testing behavior indirectly (e.g., verify BOM resolution succeeds from a custom repo).

**5. Integration test is well done.** `MavenITConsumerPomBomFromSettingsRepoTest` with the local repo + settings.xml profile setup is the most valuable part — good coverage of the real-world scenario.

gnodet (approved):
## Re-review after feedback commit (4d5cd97)

The second commit cleanly addresses all 5 observations from the previous review:

1. **derive() scoped to BUILD_CONSUMER** -- The repository override in `derive()` is now guarded by `request.getRequestType() == ModelBuilderRequest.RequestType.BUILD_CONSUMER`, so parent POM resolution and other derived sessions are not affected. Good.

2. **RepositoryFactory caching removed** -- The mutable `repositoryFactory` field and `getRepositoryFactory()` method are gone. The service is looked up inline via `session.getService(RepositoryFactory.class)` at both call sites. Simpler and correct.

3. **Non-InternalMavenSession handling** -- The `else` branch now logs a `LOGGER.debug()` message explaining that settings.xml profiles won't be passed in embedded/testing contexts. This makes the limitation visible rather than silent.

4. **Reflection removed from unit test** -- `testBuildConsumerWithExplicitRepositories` uses package-private accessors (`getRepositories()` / `getExternalRepositories()`) on `ModelBuilderSessionState` instead of reflection. The test is in the same package (`org.apache.maven.impl.model`), so this works cleanly and is resilient to refactoring.

5. **Integration test unchanged** -- Still the most valuable part of the PR, verifying end-to-end BOM resolution from a settings.xml profile repository with consumer POM flattening.

The `ConsumerPomBuilderTest` addition using Mockito spy to capture and verify the `ModelBuilderRequest` is also well structured -- it confirms the request carries repositories without depending on internal state.

One minor note: no CI checks have run on this branch yet. Worth confirming CI passes before merge.

LGTM.

hboutemy (approved):
thanks

PR #11770, opened 2026-03-09
feat: Improve the matching of required versions

nielsbasjes:
@cstamas Followup of this discussion on Slack https://the-asf.slack.com/archives/C7Q9JB404/p1773056234905829?thread_ts=1773053661.648449&cid=C7Q9JB404 

When people write in (for example) the maven-invoker-plugin something like `invoker.toolchain.jdk.version = 11` it is implemented as an exact match on the mentioned version specified in the toolchains.xml .
Historically people would write this config manually and then the exact version `11` works as expected.

With the automatic generation of the toolchains.xml more places will contain the `full` version of the JDK like `<version>11.0.29</version>` instead of the more simple `<version>11</version>`.

The effect of people using these automatically generated toolchains.xml files is that now things still work as designed, but not as expected.

This is a quick implementation of my proposal to simply treat the requirement of a single number X as "the major version must be X".
I have also included the same for a "major.minor" requirement to accept all patch versions under that also.

This means a change from a mismatch to a match in cases like this:
```
matcher = RequirementMatcherFactory.createVersionMatcher("1.5.2");
matcher.matches("1.5");
```

**I'm unsure if this change is "too breaking" to be implemented.**

### DIFF
--- a/maven-core/src/main/java/org/apache/maven/toolchain/RequirementMatcherFactory.java
+++ b/maven-core/src/main/java/org/apache/maven/toolchain/RequirementMatcherFactory.java
@@ -18,6 +18,8 @@
  */
 package org.apache.maven.toolchain;
 
+import java.util.regex.Pattern;
+
 import org.apache.maven.artifact.versioning.DefaultArtifactVersion;
 import org.apache.maven.artifact.versioning.InvalidVersionSpecificationException;
 import org.apache.maven.artifact.versioning.VersionRange;
@@ -66,7 +68,7 @@ public final class RequirementMatcherFactory {
         @Override
         public boolean matches(String requirement) {
             try {
-                VersionRange range = VersionRange.createFromVersionSpec(requirement);
+                VersionRange range = convertRequirementToVersionRange(requirement);
                 if (range.hasRestrictions()) {
                     return range.containsVersion(version);
                 } else {
@@ -77,6 +79,29 @@ public final class RequirementMatcherFactory {
                 ex.printStackTrace();
                 return false;
             }
+        }
+
+        private VersionRange convertRequirementToVersionRange(String requirement)
+                throws InvalidVersionSpecificationException {
+            // Specific for Version _requirement_ matching;
+            // If the version is a simple integer (like "25")
+            // then treat this as the requirement "the major version is 25"
+            if (Pattern.matches("^[0-9]+$", requirement)) {
+                int majorVersion = Integer.parseInt(requirement);
+                return VersionRange.createFromVersionSpec("[" + majorVersion + "," + (majorVersion + 1) + ")");
+            }
+
+            // If the version is a major.minor (like "1.5")
+            // then treat this as the requirement "the major version is 1 and the minor is 5"
+            if (Pattern.matches("^[0-9]\\.[0-9]+$", requirement)) {
+                String[] split = requirement.split("\\.", 2);
+                int majorVersion = Integer.parseInt(split[0]);
+                int minorVersion = Integer.parseInt(split[1]);
+                return VersionRange.createFromVersionSpec(
+                        "[" + majorVersion + "." + minorVersion + "," + majorVersion + "." + (minorVersion + 1) + ")");
+            }
+
+            return VersionRange.createFromVersionSpec(requirement);
         }
 
         @Override
--- a/maven-core/src/main/java/org/apache/maven/toolchain/RequirementMatcherFactory.java
+++ b/maven-core/src/main/java/org/apache/maven/toolchain/RequirementMatcherFactory.java
@@ -81,19 +81,21 @@ public final class RequirementMatcherFactory {
             }
         }
 
+        private static final Pattern PATTERN_MAJOR_VERSION = Pattern.compile("^[0-9]+$");
+        private static final Pattern PATTERN_MAJOR_MINOR_VERSION = Pattern.compile("^[0-9]\\.[0-9]+$");
         private VersionRange convertRequirementToVersionRange(String requirement)
                 throws InvalidVersionSpecificationException {
             // Specific for Version _requirement_ matching;
             // If the version is a simple integer (like "25")
             // then treat this as the requirement "the major version is 25"
-            if (Pattern.matches("^[0-9]+$", requirement)) {
+            if (PATTERN_MAJOR_VERSION.matcher(requirement).matches()) {
                 int majorVersion = Integer.parseInt(requirement);
                 return VersionRange.createFromVersionSpec("[" + majorVersion + "," + (majorVersion + 1) + ")");
             }
 
             // If the version is a major.minor (like "1.5")
             // then treat this as the requirement "the major version is 1 and the minor is 5"
-            if (Pattern.matches("^[0-9]\\.[0-9]+$", requirement)) {
+            if (PATTERN_MAJOR_MINOR_VERSION.matcher(requirement).matches()) {
                 String[] split = requirement.split("\\.", 2);
                 int majorVersion = Integer.parseInt(split[0]);
                 int minorVersion = Integer.parseInt(split[1]);
--- a/maven-core/src/main/java/org/apache/maven/toolchain/RequirementMatcherFactory.java
+++ b/maven-core/src/main/java/org/apache/maven/toolchain/RequirementMatcherFactory.java
@@ -83,6 +83,7 @@ public final class RequirementMatcherFactory {
 
         private static final Pattern PATTERN_MAJOR_VERSION = Pattern.compile("^[0-9]+$");
         private static final Pattern PATTERN_MAJOR_MINOR_VERSION = Pattern.compile("^[0-9]+\\.[0-9]+$");
+
         private VersionRange convertRequirementToVersionRange(String requirement)
                 throws InvalidVersionSpecificationException {
             // Specific for Version _requirement_ matching;
--- a/maven-core/src/main/java/org/apache/maven/toolchain/RequirementMatcherFactory.java
+++ b/maven-core/src/main/java/org/apache/maven/toolchain/RequirementMatcherFactory.java
@@ -82,7 +82,7 @@ public final class RequirementMatcherFactory {
         }
 
         private static final Pattern PATTERN_MAJOR_VERSION = Pattern.compile("^[0-9]+$");
-        private static final Pattern PATTERN_MAJOR_MINOR_VERSION = Pattern.compile("^[0-9]\\.[0-9]+$");
+        private static final Pattern PATTERN_MAJOR_MINOR_VERSION = Pattern.compile("^[0-9]+\\.[0-9]+$");
         private VersionRange convertRequirementToVersionRange(String requirement)
                 throws InvalidVersionSpecificationException {
             // Specific for Version _requirement_ matching;
--- a/maven-core/src/test/java/org/apache/maven/toolchain/RequirementMatcherFactoryTest.java
+++ b/maven-core/src/test/java/org/apache/maven/toolchain/RequirementMatcherFactoryTest.java
@@ -50,11 +50,18 @@ public class RequirementMatcherFactoryTest {
     public void testCreateVersionMatcher() {
         RequirementMatcher matcher;
         matcher = RequirementMatcherFactory.createVersionMatcher("1.5.2");
-        assertFalse(matcher.matches("1.5"));
-        assertTrue(matcher.matches("1.5.2"));
+        assertTrue(matcher.matches("1")); // Major matches
+        assertTrue(matcher.matches("1.5")); // Major.Minor matches
+        assertTrue(matcher.matches("1.5.2")); // Full match
+        assertFalse(matcher.matches("1.6")); // Wrong minor
+        assertFalse(matcher.matches("2")); // Wrong major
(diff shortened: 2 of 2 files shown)


maven-core/src/main/java/org/apache/maven/toolchain/RequirementMatcherFactory.java
```diff
@@ -79,6 +81,29 @@ public boolean matches(String requirement) {
             }
         }
 
+        private VersionRange convertRequirementToVersionRange(String requirement)
+                throws InvalidVersionSpecificationException {
+            // Specific for Version _requirement_ matching;
+            // If the version is a simple integer (like "25")
+            // then treat this as the requirement "the major version is 25"
+            if (Pattern.matches("^[0-9]+$", requirement)) {
```
slawekjaranowski, line 89: We can move pattern compile to static filed

gnodet (commented):

Nice idea to make toolchain version matching more lenient for the auto-generated toolchains.xml use case. Two issues I spotted:

1. **Regex bug**: The `major.minor` pattern only allows a single-digit major version, so `"21.3"` would not match. Needs `[0-9]+` for the major part too.
2. **Pattern precompilation**: As @slawekjaranowski already noted, the patterns should be compiled once into static fields.
3. **Missing test coverage**: No tests for multi-digit major versions (e.g., requirement `"21"` against version `"21.0.2"`, or `"21.3"` against `"21.3.1"`).

maven-core/src/main/java/org/apache/maven/toolchain/RequirementMatcherFactory.java
```diff
@@ -79,6 +81,29 @@ public boolean matches(String requirement) {
...
+                int majorVersion = Integer.parseInt(requirement);
+                return VersionRange.createFromVersionSpec("[" + majorVersion + "," + (majorVersion + 1) + ")");
+            }
+
+            // If the version is a major.minor (like "1.5")
+            // then treat this as the requirement "the major version is 1 and the minor is 5"
+            if (Pattern.matches("^[0-9]\\.[0-9]+$", requirement)) {
```
gnodet: Bug: the major version part is `[0-9]` (single digit only), so a requirement like `"21.3"` would silently fall through to the default path and be treated as an exact match. It should be `[0-9]+` to mirror the first pattern.

```suggestion
            if (Pattern.matches("^[0-9]+\\.[0-9]+$", requirement)) {
```

Also, as @slawekjaranowski mentioned, both patterns should be pre-compiled into `private static final Pattern` fields to avoid recompilation on every call.

maven-core/src/test/java/org/apache/maven/toolchain/RequirementMatcherFactoryTest.java
```diff
@@ -50,11 +50,18 @@ public void testCreateExactMatcher() {
     public void testCreateVersionMatcher() {
         RequirementMatcher matcher;
         matcher = RequirementMatcherFactory.createVersionMatcher("1.5.2");
-        assertFalse(matcher.matches("1.5"));
-        assertTrue(matcher.matches("1.5.2"));
+        assertTrue(matcher.matches("1")); // Major matches
+        assertTrue(matcher.matches("1.5")); // Major.Minor matches
+        assertTrue(matcher.matches("1.5.2")); // Full match
+        assertFalse(matcher.matches("1.6")); // Wrong minor
+        assertFalse(matcher.matches("2")); // Wrong major
```
gnodet, line 57: Consider adding tests for multi-digit major versions to catch the regex bug above, e.g.:
```java
matcher = RequirementMatcherFactory.createVersionMatcher("21.0.2");
assertTrue(matcher.matches("21"));     // major-only
assertTrue(matcher.matches("21.0"));   // major.minor
assertFalse(matcher.matches("21.1")); // wrong minor
assertFalse(matcher.matches("2"));     // wrong major
```

gnodet (commented):
## Review Summary

This PR improves toolchain version requirement matching so that a simple requirement like `"11"` matches version `"11.0.29"`, and `"1.5"` matches `"1.5.2"`. This is a reasonable enhancement motivated by auto-generated `toolchains.xml` files containing full JDK versions.

However, there are several issues that need addressing — most of which were already flagged by @slawekjaranowski and @gnodet in their earlier reviews.

### Issues

**1. Regex bug — multi-digit major versions don't match (critical)**
The `major.minor` pattern on line 96 uses `[0-9]` (single digit) instead of `[0-9]+` for the major version:
```java
Pattern.matches("^[0-9]\\.[0-9]+$", requirement)
```
A requirement like `"21.3"` won't match this pattern and will fall through to exact-match behavior — which is the most common real-world scenario (JDK 11, 17, 21, 25). The `major-only` pattern on line 89 correctly uses `[0-9]+`.

*(Already noted by @gnodet on 2026-05-21)*

**2. Pattern precompilation (major)**
Both `Pattern.matches()` calls recompile the regex on every invocation. These should be extracted to `private static final Pattern` fields:
```java
private static final Pattern MAJOR_ONLY = Pattern.compile("[0-9]+");
private static final Pattern MAJOR_MINOR = Pattern.compile("[0-9]+\\.[0-9]+");
```

*(Already noted by @slawekjaranowski on 2026-04-20 and @gnodet on 2026-05-21)*

**3. Missing test coverage for multi-digit versions (major)**
The test only uses version `"1.5.2"` (single-digit major). There are no test cases with multi-digit versions like `"21.0.2"`, which would expose the regex bug above. @gnodet's suggested test additions from 2026-05-21 would cover this gap.

### Note

This review reinforces the feedback already provided by @slawekjaranowski and @gnodet. The existing review comments contain concrete suggestions — addressing them would make this PR ready for another look.

---

maven-core/src/main/java/org/apache/maven/toolchain/RequirementMatcherFactory.java
gnodet: Bug: `[0-9]` matches only a single digit for the major version. Requirements like `"21.3"`, `"17.0"`, or `"11.0"` will not match this pattern and will fall through to exact-match behavior — exactly the most common modern JDK versions.

The major-only pattern on line 89 correctly uses `[0-9]+`.

```suggestion
            if (Pattern.matches("^[0-9]+\\.[0-9]+$", requirement)) {
```

maven-core/src/test/java/org/apache/maven/toolchain/RequirementMatcherFactoryTest.java
```diff
@@ -50,11 +50,18 @@ public void testCreateExactMatcher() {
     public void testCreateVersionMatcher() {
         RequirementMatcher matcher;
         matcher = RequirementMatcherFactory.createVersionMatcher("1.5.2");
-        assertFalse(matcher.matches("1.5"));
-        assertTrue(matcher.matches("1.5.2"));
+        assertTrue(matcher.matches("1")); // Major matches
+        assertTrue(matcher.matches("1.5")); // Major.Minor matches
+        assertTrue(matcher.matches("1.5.2")); // Full match
```
gnodet, line 55: Consider adding test cases with multi-digit major versions to cover the regex bug above. For example:

```java
matcher = RequirementMatcherFactory.createVersionMatcher("21.0.2");
assertTrue(matcher.matches("21"));    // Major-only
assertTrue(matcher.matches("21.0"));  // Major.minor
assertFalse(matcher.matches("17"));   // Wrong major
```

This would immediately expose the single-digit regex limitation.

gnodet (commented):
The version-matching improvement looks useful — prefix matching for major-only and major.minor requirements is a nice addition. However, there is a regex bug that breaks the most common modern JDK versions (11, 17, 21, 25), plus a few minor items.

Verified findings (2 of 6 original findings were false positives removed):

1. **Regex bug (high)** — The major.minor pattern `^[0-9]\.[0-9]+$` only matches single-digit major versions. Requirements like `"21.3"` or `"17.0"` won't get prefix-matching behavior. The major-only pattern correctly uses `[0-9]+`.
2. **Pattern precompilation (low)** — `Pattern.matches()` recompiles the regex on every call. Pre-compiling to `private static final Pattern` fields would be cleaner, though this is not a hot path. (Already noted by @slawekjaranowski.)
3. **Missing test coverage (medium)** — Tests only use version `"1.5.2"` (single-digit major). Adding a test with e.g. `"21.0.2"` and requirements `"21"` / `"21.3"` would immediately expose the regex bug.
4. **Static method (low)** — `convertRequirementToVersionRange` doesn't reference instance state and could be `static`. Very minor.

maven-core/src/main/java/org/apache/maven/toolchain/RequirementMatcherFactory.java
nielsbasjes: Done

maven-core/src/main/java/org/apache/maven/toolchain/RequirementMatcherFactory.java
```diff
@@ -79,6 +81,31 @@ public boolean matches(String requirement) {
             }
         }
 
+        private static final Pattern PATTERN_MAJOR_VERSION = Pattern.compile("^[0-9]+$");
+        private static final Pattern PATTERN_MAJOR_MINOR_VERSION = Pattern.compile("^[0-9]\\.[0-9]+$");
```
gnodet: This still uses `[0-9]` (single digit) for the major version. Requirements like `"21.0"` or `"17.0"` won't match this pattern. Compare with `PATTERN_MAJOR_VERSION` on the line above which correctly uses `[0-9]+`.

```suggestion
    private static final Pattern PATTERN_MAJOR_MINOR_VERSION = Pattern.compile("^[0-9]+\\.[0-9]+$");
```

gnodet (changes requested):
**Re-review after commit f4fdc9be ("fix: Compile regexes only once")**

The pattern precompilation is a good improvement — thanks for addressing that feedback. However, the regex bug flagged in all three prior reviews is still present:

`PATTERN_MAJOR_MINOR_VERSION` uses `[0-9]` (single digit) for the major version instead of `[0-9]+`. This means requirements like `"21.0"`, `"17.0"`, and `"11.0"` silently fall through to exact-match behavior — exactly the modern JDK versions this feature is meant to help with.

The fix is a single character — adding `+` after `[0-9]` on line 85. The major-only pattern on line 84 already uses `[0-9]+` correctly.

Also, the test suite only uses `"1.5.2"` (single-digit major), so it doesn't catch this. Adding a test with e.g. `"21.0.2"` would immediately expose it.

maven-core/src/main/java/org/apache/maven/toolchain/RequirementMatcherFactory.java
nielsbasjes: Yes, good catch. Fixed.

maven-core/src/test/java/org/apache/maven/toolchain/RequirementMatcherFactoryTest.java
nielsbasjes: Yes, did that.

gnodet (approved):
**Re-review after commit 8c439aa7 ("fix: Regex bug not supporting multi digit major versions")**

All previously flagged issues are now resolved:

1. ✅ **Regex bug fixed** — `PATTERN_MAJOR_MINOR_VERSION` now correctly uses `[0-9]+` for the major version, matching the already-correct `PATTERN_MAJOR_VERSION` pattern.
2. ✅ **Multi-digit test coverage added** — new `testCreateVersionMatcherMultiDigit` test with `"11.55.22"` would have caught the original bug.
3. ✅ **Pattern precompilation** — patterns extracted to `private static final` fields.

Looks good to merge. 🎉

maven-core/src/main/java/org/apache/maven/toolchain/RequirementMatcherFactory.java
nielsbasjes: You're quick :)
I was going through the comments step by step.
Just fixed this.

Also note that the AI generated improvement is incorrect (missing '^' and '$'):
```
private static final Pattern MAJOR_ONLY = Pattern.compile("[0-9]+");
private static final Pattern MAJOR_MINOR = Pattern.compile("[0-9]+\\.[0-9]+");
```

nielsbasjes:
@gnodet Is there anything else you want me to improve this with?

gnodet (approved):
### Summary

Well-motivated enhancement that relaxes toolchain version matching for auto-generated `toolchains.xml`. The behavioral change is intentional and additive — it only turns non-matches into matches (e.g., requirement `"11"` now matches version `"11.0.29"`), never the reverse. The code is correct and well-tested.

**Note:** Competing PR #11786 by @cstamas targets the same branch with a broader implementation (prefix matching via `startsWith`, `+`/`-` suffixes, `GenericVersionScheme`). The maintainers should decide which approach best fits the project's direction before merging either PR.

### 📋 PR Metadata

| Milestone | _(none)_ | `3.9.17` |

### 🔀 Backport Status

⚠️ This enhancement targets `maven-3.9.x` but no forward-port was found for:
- `maven-3.10.x` — `RequirementMatcherFactory.java` exists there with identical code

Note: On `master`/`maven-4.0.x` the class lives in `compat/maven-compat/` and is `@Deprecated(since = "4.0.0")`, so a forward-port to 4.x is likely unnecessary.

PR #11775, opened 2026-03-10
[GH-11772] Generate dual consumer POMs for parent POMs with 4.1.0 features

### LINKED ISSUE #11772: Consumer POM of parent POMs currently remain at modelVersion 4.1.0, breaking Maven 3 and Gradle consumers
### Affected version

4.0.0-rc5, probably caused by #11370 in 4.0.0-rc5

## Summary

When a project is built with Maven 4 using model version 4.1.0, the consumer POM transformation successfully downgrades child artifact POMs (jars, etc.) to model version 4.0.0. However, **parent POMs** (POM-packaged projects) by design remain at model version 4.1.0 in their consumer POM, because the transformation only strips some 4.1.0 features (`root`, `subprojects`, `preserveModelVersion`) but not others (`condition`/`packaging` in profile activation, extension `configuration`, build `sources`).

But since child artifact consumer POMs retain a `<parent>` reference (with only `relativePath` removed), Maven 3 and Gradle resolve the parent POM, encounter model version 4.1.0, and fail.

## Reproduction / Example

This is happening in the wild with **JLine 4.0.0**, which was built with Maven 4 using model version 4.1.0:

- The child artifact POMs (e.g., `jline-reader-4.0.0.pom`) are correctly transformed to model version 4.0.0
- But they reference `org.jline:jline-parent:4.0.0` as their parent
- The parent POM [`jline-parent-4.0.0.pom`](https://repo1.maven.org/maven2/org/jline/jline-parent/4.0.0/jline-parent-4.0.0.pom) has `<modelVersion>4.1.0</modelVersion>`
- Maven 3 and Gradle cannot parse model version 4.1.0 and fail

See: https://github.com/jline/jline3/issues/1688

## Impact

This is a **critical compatibility issue**. Any project built with Maven 4 that uses model version 4.1.0 features in its parent POM chain will produce artifacts that cannot be consumed by:

- Maven 3
- Gradle (all versions)
- Any other tool that only supports model version 4.0.0

This effectively breaks the backwards compatibility promise of Maven 4's consumer POM mechanism.

## Analysis

The consumer POM transformation in `DefaultConsumerPomBuilder.transformPom()` does the following for POM-packaged projects:
- Strips `root`, `modules`, `subprojects` ✓
- Strips `preserveModelVersion` ✓
- Strips `relativePath` from parent ✓
- Computes the minimum model version via `MavenModelVersion.getModelVersion()` ✓

But several 4.1.0 features are **not** stripped, causing the model version to remain 4.1.0:
- Profile activation `condition`
- Profile activation `packaging`
- Extension `configuration`
- Build `sources` (only conditionally stripped)

Even if all these were stripped, the problem is **recursive**: the parent POM's own parent may also be 4.1.0, and so on up the chain. The fundamental issue is that the consumer POM of a POM-packaged project retains the `<parent>` reference, and there is no guarantee that the entire parent chain is 4.0.0-compatible.

gnodet:
## Problem

When a POM-packaged project (e.g., a parent POM) uses model version 4.1.0 features such as profile condition activation (`<condition>`), `<subprojects>`, or build `<sources>`, the current consumer POM transformation simply strips these features and emits a 4.0.0 POM. However, this means that **any Maven 4 project consuming such a parent loses all 4.1.0 features** — the full-fidelity model is never published.

This effectively makes it impossible to use 4.1.0 features in any parent POM that will be deployed to a repository, since downstream projects resolving the parent from the repository would get the stripped 4.0.0 version.

Fixes #11772

### Stripping features is a dead end

The current approach of stripping 4.1.0 features to produce a single 4.0.0-compatible consumer POM works as a short-term compatibility measure, but it's fundamentally unsustainable:

1. **Model version 4.1.0** introduces features like profile condition activation, packaging-based activation, build sources, and extension configuration. These can be stripped without losing essential content because they're mostly syntactic sugar or build-time concerns.

2. **Model version 4.2.0 introduces mixins** — structural composition of POM configuration from multiple sources. Mixins bring in actual content (dependencies, plugins, properties) that becomes part of the project's effective model. **You cannot strip a mixin reference without losing the content it provides.** There is no 4.0.0 equivalent — the content simply disappears.

This means the "strip and flatten" approach has a hard ceiling. Without dual consumer POMs, Maven would be stuck on model version 4.0.0 for any published parent POM, making the entire model evolution story dead on arrival.

### The right architecture

By publishing **two** consumer POMs, we get the best of both worlds:

- **Main POM** (no classifier): A 4.0.0-compatible POM with 4.1.0+ features stripped. This ensures backward compatibility with Maven 3, Gradle, and any tool that only understands 4.0.0.
- **Consumer classifier POM** (`consumer` classifier): The full-fidelity 4.1.0+ POM. Maven 4 consumers resolve this via a `classifier` field on the `<parent>` reference, getting the complete model with all features intact.

This architecture scales naturally to future model versions (4.2.0 mixins, etc.) without requiring any changes to the dual POM mechanism itself.

### Relationship to #11346 (managed dependency removal)

During analysis, we examined whether the real issue was about flattening managed dependencies rather than model version features. The key insight from the #11346 IT is:

- When a consumer POM's dependency management entries are removed because they appear "unused" in the current dependency tree, this can break downstream consumers who override transitive dependency versions, bringing in new transitive deps that the removed management would have controlled.
- However, this is a **separate concern** from model version compatibility. The managed dependency removal issue exists regardless of whether we use single or dual consumer POMs.
- The "depth" concern raised in #11346 (about flattening parent hierarchy) is a red herring: dependency resolution depth counts **graph hops between dependencies** (A→B→C), not levels in the POM parent hierarchy. Flattening parent-to-child inheritance doesn't change the dependency graph depth.

The managed dependency removal issue should be addressed in a separate PR by not filtering out "unused" managed dependencies during consumer POM transformation.

### POM generation (`DefaultConsumerPomBuilder`)

When a POM-packaged project has model version > 4.0.0:
1. Build the normal consumer POM (flattened, with inheritance resolved)
2. Create the **4.1.0 consumer POM**: add `classifier="consumer"` to the parent reference so downstream 4.1.0 consumers can resolve the full-fidelity parent
3. Create the **4.0.0 main POM**: strip all 4.1.0+ features (condition activation, packaging activation, build sources, extension config) and set model version to 4.0.0

### Artifact handling (`ConsumerPomArtifactTransformer`)

During the build, both POMs are attached as artifacts:
- The main POM (4.0.0-compatible) replaces the original POM artifact (no classifier)
- The consumer POM (4.1.0 full-fidelity) is deployed with the `consumer` classifier
- The original build POM is deployed with the `build` classifier (as before)

### Parent resolution (`DefaultModelBuilder` / `DefaultModelResolver`)

- When a parent reference includes a `classifier`, the model resolver passes it to the artifact resolver, which fetches the classifier POM from the repository
- Local parent resolution is skipped when a classifier is set (classified POMs are only relevant for repository-resolved parents)

### Model schema (`maven.mdo`)

A `classifier` field is added to the `Parent` type for model version 4.1.0+, allowing consumer POMs to reference the full-fidelity parent POM.

## Changes

- `api/maven-api-model/src/main/mdo/maven.mdo`: Add `classifier` field to `Parent` (4.1.0+)
- `impl/maven-core/.../DefaultConsumerPomBuilder.java`: Add `stripTo400()` and dual POM generation
- `impl/maven-core/.../ConsumerPomArtifactTransformer.java`: Handle dual artifact lifecycle and remap during install/deploy
- `impl/maven-impl/.../DefaultModelResolver.java`: Pass parent classifier during resolution
- `impl/maven-impl/.../DefaultModelBuilder.java`: Skip local parent resolution when classifier is set

## Test plan

- [x] New IT `MavenITgh11772DualConsumerPomTest` verifies:
  - Parent main POM exists and is model version 4.0.0 with condition activation stripped
  - Parent consumer POM exists and is model version 4.1.0 (full fidelity)
  - Parent build POM exists (original)
  - Child POM parent reference has no classifier
- [x] Existing consumer POM ITs pass without regressions (MavenITmng8414, MavenITmng8527, MavenITgh11084, MavenITgh11456, MavenITmng8645)

### DIFF
--- a/api/maven-api-model/src/main/mdo/maven.mdo
+++ b/api/maven-api-model/src/main/mdo/maven.mdo
@@ -1814,6 +1814,16 @@
           </description>
           <type>String</type>
         </field>
+        <field>
+          <name>classifier</name>
+          <version>4.1.0+</version>
+          <description>
+            The classifier of the parent POM to resolve from the repository.
+            This is used by consumer POMs to reference the full-fidelity consumer POM
+            of the parent project, which is deployed with a classifier (e.g., {@code consumer}).
+          </description>
+          <type>String</type>
+        </field>
       </fields>
       <codeSegments>
         <codeSegment>
--- a/impl/maven-core/src/main/java/org/apache/maven/internal/transformation/impl/ConsumerPomArtifactTransformer.java
+++ b/impl/maven-core/src/main/java/org/apache/maven/internal/transformation/impl/ConsumerPomArtifactTransformer.java
@@ -32,11 +32,12 @@ import java.nio.file.Paths;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.List;
+import java.util.Map;
 import java.util.Set;
+import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.CopyOnWriteArraySet;
 
 import org.apache.maven.api.feature.Features;
-import org.apache.maven.api.model.Model;
 import org.apache.maven.api.services.ModelBuilderException;
 import org.apache.maven.api.services.ModelSource;
 import org.apache.maven.api.services.Source;
@@ -58,11 +59,15 @@ import org.eclipse.sisu.PreDestroy;
 @Singleton
 @Named
 class ConsumerPomArtifactTransformer extends TransformerSupport {
-    private static final String CONSUMER_POM_CLASSIFIER = "consumer";
+    static final String CONSUMER_POM_CLASSIFIER = "consumer";
+
+    private static final String CONSUMER_POM_FULL_CLASSIFIER = "consumer-full";
 
     private static final String BUILD_POM_CLASSIFIER = "build";
 
     private final Set<Path> toDelete = new CopyOnWriteArraySet<>();
+    private final Map<Path, byte[]> consumerFullCache = new ConcurrentHashMap<>();
+    private final Set<Path> consumerFullPaths = ConcurrentHashMap.newKeySet();
 
     private final PomBuilder builder;
 
@@ -89,7 +94,15 @@ class ConsumerPomArtifactTransformer extends TransformerSupport {
                     : Files.createTempFile(CONSUMER_POM_CLASSIFIER + "-", ".pom");
             deferDeleteFile(consumer);
 
+            Path consumerFull = buildDir != null
+                    ? Files.createTempFile(buildDir, CONSUMER_POM_FULL_CLASSIFIER + "-", ".pom")
+                    : Files.createTempFile(CONSUMER_POM_FULL_CLASSIFIER + "-", ".pom");
+            deferDeleteFile(consumerFull);
+            consumerFullPaths.add(consumerFull.toAbsolutePath());
+
             project.addAttachedArtifact(createConsumerPomArtifact(project, consumer, session));
+            project.addAttachedArtifact(
+                    createConsumerPomArtifact(project, consumerFull, session, CONSUMER_POM_FULL_CLASSIFIER));
         } else if (project.getModel().getDelegate().isRoot()) {
             throw new IllegalStateException(
                     "The use of the root attribute on the model requires the buildconsumer feature to be active");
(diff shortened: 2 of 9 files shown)


slawekjaranowski:
@gnodet do you need a branch as PR is closed?

PR #11776, opened 2026-03-10
Bug: Settings interpolation and strict parsing

### LINKED ISSUE #11011: Support JVM-Config and environment variables for proxy configuration
### Affected version

* v3 series

### Bug description

Despite dozens of community discussions claiming otherwise on Reddit, Stack Overflow, old Apache bug trackers, and new GitHub Apache bug trackers, neither Maven nor Maven Wrapper support `.mvn/jvm.config` nor environment variables for simple proxy configuration.

None of the industry standard mechanisms available to Gradle, to other programming languages, to *Ant*, work in v3 Maven.

* Industry standard, cURL based environment variables: `http_proxy`, `http_proxy`, `no_proxy`
* UPPERCASE environment variables: `HTTP_PROXY`, `HTTPS_PROXY`, `NO_PROXY`
* Maven wrapper `.mvn/jvm.config` files
* Maven's environment variables: `MAVEN_ARGS`, `MAVEN_OPTS`
* JVM standard `JAVA_OPTS` environment variable

As a workaround, `settings.xml` can be hardcoded to apply a proxy configuration. But this makes it significantly harder to override partial Maven configurations, compared with environment variables.

cstamas:
Problem: today in none of Maven 3 is possible to use placeholders and apply interpolation to non-String fields (ie `proxy.active` or `proxy.port`). Moreover, given Maven is "forgiving" in settings parsing, and only WARNs, it is very easy for user to miss information.

This PR:
* fixes interpolation for whole settings
* allows users to apply "strict" settings parsing that makes Maven FAIL if settings was not properly parsed or interpolated by setting `-Dmaven.settings.strictParsing` in user or system properties.

Fixes #11011

### DIFF
--- a/maven-settings-builder/src/main/java/org/apache/maven/settings/building/DefaultSettingsBuilder.java
+++ b/maven-settings-builder/src/main/java/org/apache/maven/settings/building/DefaultSettingsBuilder.java
@@ -24,8 +24,9 @@ import javax.inject.Singleton;
 
 import java.io.File;
 import java.io.IOException;
+import java.io.InputStream;
 import java.io.StringReader;
-import java.io.StringWriter;
+import java.nio.charset.StandardCharsets;
 import java.util.Collections;
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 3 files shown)


cstamas:
> For reference

I cannot find any related PR.

slawekjaranowski:
@cstamas we the same code in 4.x .... 
https://github.com/apache/maven/blob/2e18d8f42b4ae8c23ff4fd5be08916039f75807b/compat/maven-settings-builder/src/main/java/org/apache/maven/settings/building/DefaultSettingsBuilder.java

should also be updated?

cstamas:
Yes, once the 'sprint' for 3.10.x is done, we will need forward porting stuff. Same for version filters, it is much improved compared to 4 (inclusion added, h/l is G:A addressable, etc).

PR #11778, opened 2026-03-10
Feat: Maven 3.x with Resolver 2.x

cstamas:
This is Maven 3.x line using Resolver 2.x.

Notable:
* the Resolver 1.x "http" HTTP transport got renamed to "apache" transport (IT MavenITmng7470ResolverTransportTest needs change for this)
* IT MavenITmng4368TimestampAwareArtifactInstallerTest is fluke: Resolver 2.x unlike Resolver 1.x _always installs if asked to install_. Resolver 1.x had some rudimentary logic (file timestamp, file size) to perform "change detection" that was removed from Resolver 2.x. If user wants "conditional install" then the install API call needs to be wrapped into condition.

Temporary changes applied to ITs on `maven-3.9.x-resolver-2.x` branch:
https://github.com/apache/maven-integration-testing/pull/420


pom.xml
```diff
@@ -509,6 +514,16 @@ under the License.
...
 
+  <repositories>
+    <repository>
+      <snapshots>
+        <enabled>false</enabled>
+      </snapshots>
+      <id>maven-2404</id>
+      <url>https://repository.apache.org/content/repositories/maven-2404/</url>
+    </repository>
+  </repositories>
```
slawekjaranowski: Is it needed?

cstamas: Dropped

jglick:
Does this expose 3.10 to https://github.com/apache/maven/issues/11856? Could not grasp what this is about from a quick read.

cstamas:
This PR updates Maven dependency on Resolver from 1.9.x to 2.0.x.

Yes, as Resolver 2.x automatically resolves prefixes (configurable, see `aether.remoteRepositoryFilter.prefixes.resolvePrefixFiles` here https://maven.apache.org/resolver/configuration.html).

But, 3.10.0 GA will use Resolver 2.0.21 (currently on vote) that has this:
https://github.com/apache/maven-resolver/pull/1976

PR #11779, opened 2026-03-10
Feat: Promote session.topDirectory and session.rootDirectory

cstamas:
To real properties usable in POM as well, not only in config interpolation.

### DIFF
--- a/maven-core/src/main/java/org/apache/maven/execution/MavenSession.java
+++ b/maven-core/src/main/java/org/apache/maven/execution/MavenSession.java
@@ -165,7 +165,7 @@ public class MavenSession implements Cloneable {
 
     /**
      * Returns top directory if discovered or {@code null}.
-     * Top directory is directory where top level POM resides with respect of any reactor limiting or shaping like use 
+     * Top directory is directory where top level POM resides with respect of any reactor limiting or shaping like use
      * of {@code -f} or alike CLI options.
      *
      * @since 3.10.0


maven-core/src/main/java/org/apache/maven/execution/MavenSession.java
```diff
@@ -163,6 +163,24 @@ public MavenExecutionResult getResult() {
         return result;
     }
 
+    /**
+     * Returns top directory if discovered or {@code null}.
+     *
+     * @since 3.10.0
+     */
```
slawekjaranowski, line 172: more clear documentation will be appreciated what is difference between top and root directory

maven-core/src/main/java/org/apache/maven/execution/MavenSession.java
```diff
@@ -165,6 +165,8 @@ public MavenExecutionResult getResult() {
 
     /**
      * Returns top directory if discovered or {@code null}.
+     * Top directory is directory where top level POM resides with respect of any reactor limiting or shaping like use
```
jglick: How is “top-level POM” defined exactly? It is possible for a `pom`-packaging module to have `<module>`s whose `<parent>` is something else. Or for modules to not be subdirectories.

cstamas: Top level directory is the directory where very first POM being loaded up exists. It is usually `$CWD`, but may be something different, ie. if you use `mvn -f subproject` or `mvn -f module/pom.xml` etc.

PR #11780, opened 2026-03-10
[GH-11772] Fail-fast consumer POM validation for non-4.0.0 model versions

### LINKED ISSUE #11772: Consumer POM of parent POMs currently remain at modelVersion 4.1.0, breaking Maven 3 and Gradle consumers

gnodet:
## Summary

Fixes #11772

When a POM-packaged project (parent POM) uses Maven 4.1.0 features like profile conditions that cannot be stripped during consumer POM transformation, the build now **fails fast** with actionable guidance instead of silently deploying a consumer POM that Maven 3 and Gradle cannot resolve.

### Changes

- **Consumer POM validation** (`DefaultConsumerPomBuilder`): After building the consumer POM for a POM-packaged project on the non-flatten path, validates that the computed model version is 4.0.0. If not, throws a `MavenException` listing the available options:
  1. Enable flattening (`maven.consumer.pom.flatten=true`)
  2. Preserve the model version (`preserve.model.version=true`)
  3. Remove the incompatible features

- **Managed dependency removal flag** (`Constants` / `Features`): Added `maven.consumer.pom.removeUnusedManagedDependencies` property (default: `true`) to control whether unused managed dependencies are removed during consumer POM flattening.

### Scope

The validation only applies to:
- **POM-packaged projects** (parent POMs) — non-POM projects (jar, war, etc.) are consumed as dependencies where Maven 3 ignores unknown elements, so a higher model version is acceptable (only a warning is logged)
- The **non-flatten path** — when flattening is enabled, the transformation handles what it can (e.g. mixins are intentionally kept for POM-packaged projects with model version 4.2.0)
- Projects **without** `preserve.model.version=true`

### Use cases

| Scenario | Outcome |
|---|---|
| 4.1.0 parent with `root`/`subprojects` only (e.g. JLine) | Consumer POM is naturally 4.0.0 after transformation — no intervention needed |
| 4.1.0 parent with profile conditions | Build fails with actionable error message |
| 4.2.0 POM with mixins + flatten | Existing behavior preserved — mixins kept, model version 4.2.0 |
| 4.2.0 POM with mixins, no flatten | Existing mixin-specific error preserved |

## Test plan

- Unit test: `testParentWithConditionsFailsConsumerPom` — validates fail-fast for profile conditions
- Unit test: `testMultiModuleConsumerPreserveModelVersion` — validates `preserveModelVersion` bypass
- IT: `MavenITgh11772ConsumerPom410Test` — verifies consumer POMs are 4.0.0 and build POMs retain 4.1.0 content for a parent+child project using 4.1.0 namespace

### DIFF
--- a/impl/maven-core/src/main/java/org/apache/maven/internal/transformation/impl/DefaultConsumerPomBuilder.java
+++ b/impl/maven-core/src/main/java/org/apache/maven/internal/transformation/impl/DefaultConsumerPomBuilder.java
@@ -144,7 +144,30 @@ class DefaultConsumerPomBuilder implements PomBuilder {
             if (isBom) {
                 return buildBomWithoutFlatten(session, project, src);
             } else {
-                return buildPom(session, project, src);
+                Model result = buildPom(session, project, src);
+                // Validate POM-packaged projects (parent POMs): if the consumer POM cannot be
+                // downgraded to 4.0.0, Maven 3 / Gradle cannot resolve the parent.
+                // Non-POM projects are consumed as dependencies where unknown elements are
+                // ignored, so a higher model version is acceptable (only a warning is logged
+                // by transformNonPom/transformPom).
+                if (POM_PACKAGING.equals(packaging)
+                        && !model.isPreserveModelVersion()
+                        && !ModelBuilder.MODEL_VERSION_4_0_0.equals(result.getModelVersion())) {
+                    throw new MavenException("The consumer POM for " + project.getId()
+                            + " cannot be downgraded to model version 4.0.0 because it contains"
+                            + " features that require a newer model version."
+                            + " Since consumer POM flattening is disabled, the parent reference is"
+                            + " preserved, which requires consumers to resolve the parent POM." + System.lineSeparator()
+                            + "You have the following options to resolve this:" + System.lineSeparator()
+                            + "  1. Enable flattening by setting the property 'maven.consumer.pom.flatten=true'"
+                            + " to inline parent content and produce a self-contained 4.0.0 consumer POM"
+                            + System.lineSeparator()
+                            + "  2. Preserve the model version by setting 'preserve.model.version=true'"
+                            + " on the <project> element (Maven 4 consumers only)"
+                            + System.lineSeparator()
+                            + "  3. Remove the features that require a newer model version");
+                }
+                return result;
             }
         }
         // Default behavior: flatten the consumer POM
@@ -192,6 +215,8 @@ class DefaultConsumerPomBuilder implements PomBuilder {
         InternalSession iSession = InternalSession.from(session);
         ModelBuilderResult result = buildModel(session, src);
         Model model = result.getEffectiveModel();
+        boolean removeUnusedManagedDeps =
+                Features.consumerPomRemoveUnusedManagedDependencies(session.getConfigProperties());
 
         if (model.getDependencyManagement() != null
                 && !model.getDependencyManagement().getDependencies().isEmpty()) {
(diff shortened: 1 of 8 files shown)


impl/maven-core/src/main/java/org/apache/maven/internal/transformation/impl/DefaultConsumerPomBuilder.java
```diff
@@ -408,13 +411,66 @@ static Model transformPom(Model model, MavenProject project) {
...
+            }
+        }
+        return profile;
+    }
+
+    private static Build stripBuildTo400(Build build) {
+        if (!build.getExtensions().isEmpty()) {
+            List<Extension> strippedExtensions = build.getExtensions().stream()
+                    .map(ext -> ext.getConfiguration() != null ? ext.withConfiguration(null) : ext)
+                    .collect(Collectors.toList());
```
desruisseaux: Can be replaced by `toList()`.

impl/maven-core/src/main/java/org/apache/maven/internal/transformation/impl/DefaultConsumerPomBuilder.java
```diff
@@ -408,13 +411,66 @@ static Model transformPom(Model model, MavenProject project) {
...
+     * This removes features that Maven 3 cannot understand:
+     * <ul>
+     *   <li>Profile condition activation</li>
+     *   <li>Profile packaging activation</li>
+     *   <li>Extension configuration</li>
+     * </ul>
+     * Note: mixins cannot be stripped this way (they bring content).
+     * If the model still requires &gt; 4.0.0 after stripping, the caller should handle it.
+     */
+    static Model stripTo400(Model model) {
```
desruisseaux: Maybe this method can be kept `private`?

**EDIT:** above comment is invalid, I just saw that this method is invoked in JUnit tests.

cstamas (approved):

cstamas (approved):

PR #11786, opened 2026-03-11
Feat: Improve matching of Toolchains versions

cstamas:
This feature was totally off from what it was documented. Have to note, that in this case versions (and hence, their matching) _reflect exactly_ the version that Toolchains discovered (ie there is no "automatic" handling of `1.8` vs `8`, JDK reported version is used as is).

Doco: https://maven.apache.org/plugins/maven-invoker-plugin/examples/selector-conditions.html

Expressions and changes:
* "11" (did equality) is now "starts with `11.`" (as per doco, as range `[1.5,)` is equivalent of this)
* "11+" (unsure what this did) is now "contained in `[11,)`" (as per doco)
* "11-" (unsure what this did) is now "contained in `(,11)`" (as per doco)
* otherwise parse it as range and check for contains

### DIFF
--- a/maven-core/src/main/java/org/apache/maven/toolchain/RequirementMatcherFactory.java
+++ b/maven-core/src/main/java/org/apache/maven/toolchain/RequirementMatcherFactory.java
@@ -86,31 +86,29 @@ public final class RequirementMatcherFactory {
         private static final VersionScheme VERSION_SCHEME = new GenericVersionScheme();
 
         private static boolean matchesRequirement(String version, String requirement) {
-            boolean interval = false;
-            boolean included = false;
-            if (requirement.endsWith("+")) {
-                interval = true;
-                included = true;
-                requirement = requirement.substring(0, requirement.length() - 1);
-            } else if (requirement.endsWith("-")) {
-                interval = true;
-                requirement = requirement.substring(0, requirement.length() - 1);
-            }
-            final String req = requirement;
-
             // if requirement is not a version range itself
-            if (!req.contains("[") && !req.contains("(") && !req.contains(",")) {
+            if (!requirement.contains("[") && !requirement.contains("(") && !requirement.contains(",")) {
+                boolean interval = false;
+                boolean included = false;
+                if (requirement.endsWith("+")) {
+                    interval = true;
+                    included = true;
+                    requirement = requirement.substring(0, requirement.length() - 1);
+                } else if (requirement.endsWith("-")) {
+                    interval = true;
+                    requirement = requirement.substring(0, requirement.length() - 1);
+                }
                 if (!interval) {
-                    return version.startsWith(req + "."); // "11" -> "11.xxx"
+                    return version.startsWith(requirement + "."); // "11" -> "11.xxx"
                 } else {
                     try {
                         if (included) {
                             return VERSION_SCHEME
-                                    .parseVersionRange("[" + req + ",)")
+                                    .parseVersionRange("[" + requirement + ",)")
                                     .containsVersion(VERSION_SCHEME.parseVersion(version)); // "11+" -> "[11,)"
                         } else {
                             return VERSION_SCHEME
-                                    .parseVersionRange("(," + req + ")")
+                                    .parseVersionRange("(," + requirement + ")")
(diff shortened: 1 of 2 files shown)


maven-core/src/main/java/org/apache/maven/toolchain/RequirementMatcherFactory.java
```diff
@@ -57,31 +61,78 @@ public String toString() {
...
+            } else if (requirement.endsWith("-")) {
+                interval = true;
+                requirement = requirement.substring(0, requirement.length() - 1);
+            }
+            final String req = requirement;
+
+            // if requirement is not a version range itself
+            if (!req.contains("[") && !req.contains("(") && !req.contains(",")) {
+                if (!interval) {
+                    return version.startsWith(req + "."); // "11" -> "11.xxx"
```
HannesWell: I was about to propose that logic as well for https://github.com/apache/maven/pull/11770, i.e. when one just specifies a version requirement that the actual version should just start with the requirement to handle cases like
- 11
- 11.1
- 11.1.1

Of course one has to handle the Java-8 and older versions. But maybe this can be done by removing a leading `1.` to 'norm' the versions to all start with their, what's now called, `feature` version of the JDK.

In this case I wonder how it works if one requires `1.5.2` and has a version `1.5.2` since a trailing dot is added? But I see this case covered in tests.
Just in case that still has slipped, one could test for equals or starts with dot suffix, i.e. the current code.

maven-core/src/main/java/org/apache/maven/toolchain/RequirementMatcherFactory.java
```diff
@@ -57,31 +61,78 @@ public String toString() {
...
+            boolean interval = false;
+            boolean included = false;
+            if (requirement.endsWith("+")) {
+                interval = true;
+                included = true;
+                requirement = requirement.substring(0, requirement.length() - 1);
+            } else if (requirement.endsWith("-")) {
+                interval = true;
+                requirement = requirement.substring(0, requirement.length() - 1);
+            }
```
HannesWell: Can't this code be moved into the if block that has checked that the requirement isn't a version-range? In case of the latter this seems to be ignored any-ways.

maven-core/src/main/java/org/apache/maven/toolchain/RequirementMatcherFactory.java
cstamas: Before this call the normalized strings are tested for _equality_. This is what I missed in my comment above as well. So `"1.5.2".equalsIgnoreCase("1.5.2")` will yield `true`

cstamas: Re 1.8 support (or older): IMHO, user must write versions as reported by JDK, especially as toolchains XML may be generated today. And if user consistently writes `1.8` there is no problem. The only problem would be `8+` that does not include Java 1.8... unsure about this.

cstamas: For possible 1.8 support, see this similar case: https://github.com/apache/maven/commit/c2e4be2ee0a71c5031a59e371a21c2719643efb5

cstamas: Legacy code for sure wrote "1.4", "1.6" or "1.8"... and given toolchain does not have "feature version" to match against (nor could?) I'd say we are fine IF users use versions "as reported by Java".

HannesWell: > Before this call the normalized strings are tested for _equality_.

Oversaw that, thanks.

> I'd say we are fine IF users use versions "as reported by Java".

Sounds good. I would try to keep it simple and as long as there is a way to handle it, it's fine.
Also given that the number of users at <=1.8 is hopefully declining, this should also become less of a problem.

slawekjaranowski:
I would like not introduce patterns like 11+ only for toolchains version, we have everywhere range for version. It can be inconsistent with other place in Maven so can make user confused.

PR #11805, opened 2026-03-16
fix: preserve processor path types across conflict resolution

mariuszs:
## Summary

Fixes the issue described in [maven-compiler-plugin#1042 (comment)](https://github.com/apache/maven-compiler-plugin/pull/1042#issuecomment-4063925384): when a dependency is both a direct dep (e.g., `modular-jar`) and a transitive dep of a processor, `ConflictResolver` removes the transitive occurrence before `TypeDeriver` can assign processor path types. The artifact ends up only on `--module-path`, missing from `--processor-module-path`.

This builds on PR #11380 which introduced `TypeDeriver`.

### Root cause

`ConflictMarker` keys on `groupId:artifactId:extension:classifier`. Since `modular-jar` and `modular-processor` both have `extension=jar, classifier=""`, they share the same conflict key. `ConflictResolver` (which runs before `TypeDeriver`) picks one winner and **removes** the loser entirely (verbosity=NONE). `TypeDeriver` then never sees the transitive processor occurrence.

### Fix: two-phase approach (no maven-resolver changes needed)

| Phase | Transformer | When | What |
|-------|------------|------|------|
| 1 | `TypeCollector` | **Before** ConflictResolver | Walks pre-conflict tree, records `{GA:ext:cls → derived processor type}` in context |
| 2 | `ProcessorTypeMerger` (in `TypeDeriver`) | **After** ConflictResolver | Reads context, sets `PROCESSOR_TYPE` property on surviving nodes that need both paths |
| 3 | `DefaultDependencyResolverResult` | At path assignment | Reads `PROCESSOR_TYPE`, adds artifact to processor paths too |

`TypeCollector` **short-circuits** when no processor-type direct deps exist, so builds without processors pay zero cost.

### Changes

- New `TypeCollector` transformer (impl + compat)
- New `PROCESSOR_TYPE` artifact property in `MavenArtifactProperties`
- `TypeDeriver` enhanced with `ProcessorTypeMerger` inner visitor
- `DefaultDependencyResolverResult.addDependency()` handles `PROCESSOR_TYPE`
- `MavenSessionBuilderSupplier` wires `TypeCollector` before `ConflictResolver`

## Test plan

- [x] New test `TypeDeriverWithConflictResolutionTest.sharedDependencyLosesProcessorType` — reproduces the exact scenario and verifies the fix
- [x] Control test `typeDeriverAloneWorksCorrectly` — proves TypeDeriver logic is correct in isolation
- [x] Existing `TypeDeriverTest` (2 tests) still pass
- [x] Checkstyle: 0 violations
- [x] End-to-end IT in [maven-compiler-plugin@f8a6fd1](https://github.com/apache/maven-compiler-plugin/pull/1042/commits/f8a6fd14f62dd85280d92a88f3b047ea255bb318) (`modular-processor-shared-dep`) — 3-module project with shared-lib as both `modular-jar` direct dep and transitive dep of `modular-processor`, verifies `javac` compiles successfully

## Related

- [maven-compiler-plugin#1039](https://github.com/apache/maven-compiler-plugin/issues/1039) — the original issue (`FindException` with modular processors)
- [maven-compiler-plugin#1042](https://github.com/apache/maven-compiler-plugin/pull/1042) — compiler-plugin workaround (supplements processor-module-path with module-path entries)
- [maven#11380](https://github.com/apache/maven/pull/11380) — introduced `TypeDeriver` (first step, but doesn't handle the conflict resolution case)

### DIFF
--- a/compat/maven-resolver-provider/src/main/java/org/apache/maven/repository/internal/MavenSessionBuilderSupplier.java
+++ b/compat/maven-resolver-provider/src/main/java/org/apache/maven/repository/internal/MavenSessionBuilderSupplier.java
@@ -25,6 +25,7 @@ import org.apache.maven.api.DependencyScope;
 import org.apache.maven.repository.internal.artifact.FatArtifactTraverser;
 import org.apache.maven.repository.internal.scopes.Maven4ScopeManagerConfiguration;
 import org.apache.maven.repository.internal.type.DefaultTypeProvider;
+import org.apache.maven.repository.internal.type.TypeCollector;
 import org.apache.maven.repository.internal.type.TypeDeriver;
 import org.eclipse.aether.RepositorySystem;
 import org.eclipse.aether.RepositorySystemSession.CloseableSession;
@@ -112,6 +113,7 @@ public class MavenSessionBuilderSupplier implements Supplier<SessionBuilder> {
 
     protected DependencyGraphTransformer getDependencyGraphTransformer() {
         return new ChainedDependencyGraphTransformer(
+                new TypeCollector(),
                 new ConflictResolver(
                         new ConfigurableVersionSelector(), new ManagedScopeSelector(getScopeManager()),
                         new SimpleOptionalitySelector(), new ManagedScopeDeriver(getScopeManager())),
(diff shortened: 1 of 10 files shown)


mariuszs:
Yep, I'm all set on my end.

PR #11809, opened 2026-03-18
document modelVersion only supported value: 4.0.0

hboutemy:
while at it, having clear documentation on required fields will be useful
=> Modello feature requested https://github.com/codehaus-plexus/modello/issues/549

### DIFF
--- a/maven-model/src/main/mdo/maven.mdo
+++ b/maven-model/src/main/mdo/maven.mdo
@@ -86,7 +86,12 @@
           <name>modelVersion</name>
           <version>4.0.0+</version>
           <required>true</required>
-          <description>Declares to which version of project descriptor this POM conforms.</description>
+          <description>
+            <![CDATA[
+            Declares to which version of project descriptor this POM conforms to: <code>4.0.0</code> is the only
+            value supported by Maven 3.
+            ]]>
+          </description>
           <type>String</type>
         </field>
 


maven-model/src/main/mdo/maven.mdo
```diff
@@ -86,7 +86,12 @@
           <name>modelVersion</name>
           <version>4.0.0+</version>
           <required>true</required>
-          <description>Declares to which version of project descriptor this POM conforms.</description>
+          <description>
+            <![CDATA[
+            Declares to which version of project descriptor this POM conforms: <code>4.0.0</code> is the only
```
elharo: Declares the version of the project descriptor this POM conforms to:

hboutemy: please add a commit to the branch: this is what contributing is about when you are a committer
thanks

elharo: Pair programming excepted, in my experience most of the time most PR authors prefer to make changes themselves and object when other devs edit their branches directly.

hboutemy: usual non-helpful comment: you are the english teacher, this is a typo, just do it, stop applying a solution to a non-adapted context

but I'm loosing time trying to discuss, it seems
as usual, loosing less time by just ignoring
it's sad...

PR #11810, opened 2026-03-18
document modelVersion supported values

hboutemy:
equivalent to #11809 for Maven 3.x

IIUC from CLI tests, in Maven 4.0.0-rc5, not defining any value is also supported, which will trigger inference based on namespace and/or xsd: we'll need to write something about it

notice that Modello model saying required=true for this field is not true any more...

based on current discussions on ML and PR #10952, maybe this will change in the future, we still don't have a definitive choice done: documenting the current state remains useful, we'll update later if necessary


api/maven-api-model/src/main/mdo/maven.mdo
```diff
@@ -92,7 +92,8 @@
           <name>modelVersion</name>
           <version>4.0.0+</version>
           <required>true</required>
-          <description>Declares to which version of project descriptor this POM conforms.</description>
+          <description>Declares to which version of project descriptor this POM conforms:
```
elharo: Declares to which version of the project descriptor this POM conforms to:

(Yes, ending a sentence with a preposition is perfectly fine in English, whatever second grade English teachers say.) 

https://www.merriam-webster.com/grammar/prepositions-ending-a-sentence-with

hboutemy: please add a commit to the branch: this is what contributing is about when you are a committer
thanks

gnodet (approved):
Grammar fix looks good — addresses elharo's feedback (added "the") and cleans up the double preposition.

PR #11820, opened 2026-03-21
[MNG-8604] Fix concurrent builder crash with reactor plugins

gnodet:
## Summary

- Fixes a crash (`Step $plan$ not found`) when using `--builder concurrent` on projects where a reactor module is also used as a build plugin (e.g., Apache CXF's `cxf-codegen-plugin`)
- Moves reactor plugin ordering logic from `calculateLifecycleMappings()` to `buildInitialPlan()`, where PLAN/READY steps already exist
- Simplifies plugin resolution to only eagerly resolve non-reactor plugins

## Details

The root cause is that `calculateLifecycleMappings()` calls `plan.requiredStep(project, PLAN)` to set up ordering between reactor plugins and their consuming projects, but the PLAN steps are only created later in `buildInitialPlan()`. Moving this logic to after the PLAN/READY step creation fixes the crash.

Tested successfully on Apache CXF, Maven Resolver, Commons Lang, and Maven Surefire with `-T4 --builder concurrent`.

## Test plan

- [x] Existing `BuildPlanCreatorTest` tests pass
- [x] Verified on Apache CXF (`mvn verify -T4 --builder concurrent`) — previously crashed, now builds successfully
- [x] Verified on Maven Resolver, Commons Lang, Maven Surefire with concurrent builder

### DIFF
--- a/impl/maven-core/src/main/java/org/apache/maven/lifecycle/internal/concurrent/BuildPlanExecutor.java
+++ b/impl/maven-core/src/main/java/org/apache/maven/lifecycle/internal/concurrent/BuildPlanExecutor.java
@@ -268,6 +268,21 @@ public class BuildPlanExecutor {
                 Stream.of(pplan, setup, teardown).forEach(step -> plan.addStep(project, step.name, step));
             }
 
+            // Handle reactor plugins: if a project uses a plugin that is also a reactor module,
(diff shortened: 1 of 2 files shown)


nkonev:
milestone 4.0.0-rc-6 but branch is master, not maven-4.0.x

PR #11823, opened 2026-03-23
Fix consumer POM serialization of prefixed XML attributes (fixes #11760)

### LINKED ISSUE #11760: Maven 4 consumer pom does not carry over mvn xml namespace
### Affected version

4.0.0-rc-5

### Bug description

I am attempting to upgrade a project to Maven 4. The project includes a build parent pom, which is a sibling module to others that use the parent.

```
jdbi/
  pom.xml
  jdbi-build-parent/
    pom.xml
  jdbi-core/
    pom.xml (inherit build-parent)
    src/...
  jdbi-other-integration/...
```

The build parent pom itself inherits from a generic template pom ( [basepom](https://basepom.org/) )
In order to combine e.g. compiler arguments in our build parent and basepom, in our build parent we declare:
```xml
<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xmlns:mvn="http://maven.apache.org/POM/4.0.0" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 https://maven.apache.org/xsd/maven-4.0.0.xsd">
...
<plugins>
    <plugin>
        <artifactId>maven-compiler-plugin</artifactId>
        <executions>
            <execution>
                <id>default-compile</id>
                <configuration>
                    <compilerArgs mvn:combine.children="append">
                        <arg>-Xlint:deprecation</arg>
                        <arg>-Xlint:unchecked</arg>
                    </compilerArgs>
                </configuration>
            </execution>
    </plugin>
...
```

We added the `mvn` namespace since it seemed to be the upgrade path from the old `combine.children` directive.

When building the multi-module project, all is well.

```
% mvn clean install
...
[INFO] BUILD SUCCESS
```

However, if you then try to use an individual project alone, without the full reactor build, you get a surprising error:

```
% mvn test -pl :jdbi-other-integration
...

[WARNING] The POM for org.jdbi:jdbi-core:jar:tests:4.0.0-SNAPSHOT is invalid, transitive dependencies (if any) will not be available: 1 problem was 
    - [FATAL] Non-parseable POM org.jdbi:jdbi-build-parent:4.0.0-SNAPSHOT: Unable to read model: Undeclared namespace prefix "mvn" (for attribute "combine.children")
 at [row,col {unknown-source}]: [786,64]

...

[ERROR] org.jdbi.guava.codec.TypeResolvingCodecFactoryTest.testCollectionCodec -- Time elapsed: 0.101 s <<< ERROR!
java.lang.NoClassDefFoundError: io/leangen/geantyref/GenericTypeReflector
```

Sure enough, the installed consumer pom lost the `xmlns:mvn` on the opening `<project>` tag:
```
% head ~/.m2/repository/org/jdbi/jdbi-build-parent/4.0.0-SNAPSHOT/jdbi-build-parent-4.0.0-SNAPSHOT.pom 
```

So it looks like Maven, while transforming our build parent pom into its consumer version for installation to the local repository, drops the `<project xmlns:mvn` declaration and then writes an invalid POM file to the local repo.

The source is available, although it's not a minimal example: https://github.com/jdbi/jdbi/pull/2935

gnodet:
## Summary

Fix consumer POM serialization so that prefixed attributes (like `mvn:combine.children`) produce valid XML even when the namespace declaration (`xmlns:mvn`) was on an ancestor element not part of the Maven model.

Fixes #11760

## Problem

The [jdbi project](https://github.com/apache/maven/issues/11760) uses `xmlns:mvn="http://maven.apache.org/POM/4.0.0"` on `<project>` with `mvn:combine.children="append"` on plugin configuration elements. During consumer POM transformation:

1. The `xmlns:mvn` declaration on `<project>` is lost (not part of the Maven model)
2. Prefixed attributes on `XmlNode` configuration trees survive
3. Serialization fails with "Undeclared namespace prefix" because the prefix has no declaration

**Note:** `mvn:combine.children` never triggered merge/append behavior — not in Maven 3 or 4. Per the [XML namespace spec](https://www.w3.org/TR/xml-names/#defaulting), unprefixed attributes are in no namespace, so `combine.children` and `mvn:combine.children` are different attributes.

## Changes

- **`XmlNode` API** (`api/maven-api-xml`): add `namespaces()` method returning inherited prefix→URI bindings
- **Parser** (`DefaultXmlService`): accumulate namespace declarations during parsing and propagate to child nodes
- **StAX writer** (`writer-stax.vm`): resolve prefixed attributes against local declarations, then namespace context; auto-declare namespaces from context; strip orphaned prefixes as last resort
- **XPP3 writer** (`writer.vm`): same fix for the deprecated writer template
- **Merge**: preserve dominant node's namespace context

## Test plan

28 namespace-specific tests covering: parsing (prefix inheritance, shadowing, default NS exclusion), writing (round-trip, orphaned prefix stripping, multiple namespaces, `xml:space`), merging (namespace preservation), builder API, and consumer POM simulation.

All existing tests pass: 70 in maven-xml, 172 in maven-model.

### DIFF
--- a/impl/maven-xml/src/main/java/org/apache/maven/internal/xml/DefaultXmlService.java
+++ b/impl/maven-xml/src/main/java/org/apache/maven/internal/xml/DefaultXmlService.java
@@ -162,9 +162,7 @@ public class DefaultXmlService extends XmlService {
     private void writeNode(XMLStreamWriter xmlWriter, XmlNode node) throws XMLStreamException {
         xmlWriter.writeStartElement(node.prefix(), node.name(), node.namespaceUri());
 
-        for (Map.Entry<String, String> attr : node.attributes().entrySet()) {
-            xmlWriter.writeAttribute(attr.getKey(), attr.getValue());
-        }
+        writeAttributes(xmlWriter, node.attributes());
 
         for (XmlNode child : node.children()) {
             writeNode(xmlWriter, child);
@@ -176,6 +174,42 @@ public class DefaultXmlService extends XmlService {
         }
 
         xmlWriter.writeEndElement();
+    }
+
+    static void writeAttributes(XMLStreamWriter xmlWriter, Map<String, String> attributes) throws XMLStreamException {
+        // Write namespace declarations first, then regular attributes
+        for (Map.Entry<String, String> attr : attributes.entrySet()) {
+            String key = attr.getKey();
+            if ("xmlns".equals(key)) {
+                xmlWriter.writeDefaultNamespace(attr.getValue());
+            } else if (key.startsWith("xmlns:")) {
+                xmlWriter.writeNamespace(key.substring(6), attr.getValue());
+            }
+        }
+        for (Map.Entry<String, String> attr : attributes.entrySet()) {
+            String key = attr.getKey();
+            String value = attr.getValue();
+            if ("xmlns".equals(key) || key.startsWith("xmlns:")) {
+                continue; // already written above
+            } else if (key.startsWith("xml:")) {
+                xmlWriter.writeAttribute("http://www.w3.org/XML/1998/namespace", key.substring(4), value);
+            } else if (key.contains(":")) {
+                int colon = key.indexOf(':');
+                String prefix = key.substring(0, colon);
+                String localName = key.substring(colon + 1);
+                String nsUri = attributes.get("xmlns:" + prefix);
+                if (nsUri != null) {
+                    xmlWriter.writeAttribute(prefix, nsUri, localName, value);
+                } else {
+                    // No namespace declaration found for this prefix; write as unprefixed
+                    // to produce valid XML (the namespace declaration may have been lost
+                    // during consumer POM transformation)
+                    xmlWriter.writeAttribute(localName, value);
+                }
+            } else {
+                xmlWriter.writeAttribute(key, value);
+            }
+        }
     }
 
     /**
--- a/impl/maven-xml/src/main/java/org/apache/maven/internal/xml/DefaultXmlService.java
+++ b/impl/maven-xml/src/main/java/org/apache/maven/internal/xml/DefaultXmlService.java
@@ -79,10 +79,10 @@ public class DefaultXmlService extends XmlService {
             Map<String, String> parentNamespaces)
             throws XMLStreamException {
         boolean spacePreserve = false;
-        String lPrefix = null;
-        String lNamespaceUri = null;
-        String lName = null;
-        String lValue = null;
+        String elementPrefix = null;
+        String elementNamespaceUri = null;
+        String elementName = null;
+        String elementValue = null;
         Object location = null;
         Map<String, String> attrs = null;
         Map<String, String> nsContext = null;
@@ -93,62 +93,65 @@ public class DefaultXmlService extends XmlService {
             if (eventType == XMLStreamReader.START_ELEMENT) {
                 lastStartTag = parser.getLocation().getLineNumber() * 1000
                         + parser.getLocation().getColumnNumber();
-                if (lName == null) {
+                // The first START_ELEMENT we encounter is "this" element;
+                // subsequent START_ELEMENTs are children, handled in the else branch.
+                if (elementName == null) {
                     int namespacesSize = parser.getNamespaceCount();
-                    lPrefix = parser.getPrefix();
-                    lNamespaceUri = parser.getNamespaceURI();
-                    lName = parser.getLocalName();
+                    elementPrefix = parser.getPrefix();
+                    elementNamespaceUri = parser.getNamespaceURI();
+                    elementName = parser.getLocalName();
                     location = locationBuilder != null ? locationBuilder.toInputLocation(parser) : null;
-                    // Build the namespace context: start with inherited, add local declarations
+                    // Build the namespace context: start with inherited, add local declarations.
+                    // The default namespace (empty prefix) is excluded because per the XML namespace
+                    // spec (Section 6.2), default namespace declarations do NOT apply to attributes.
                     nsContext = new HashMap<>(parentNamespaces);
-                    for (int i = 0; i < namespacesSize; i++) {
-                        String nsPrefix = parser.getNamespacePrefix(i);
-                        String nsUri = parser.getNamespaceURI(i);
-                        if (nsPrefix != null && !nsPrefix.isEmpty()) {
-                            nsContext.put(nsPrefix, nsUri);
-                        }
-                    }
                     int attributesSize = parser.getAttributeCount();
                     if (attributesSize > 0 || namespacesSize > 0) {
                         attrs = new HashMap<>();
                         for (int i = 0; i < namespacesSize; i++) {
                             String nsPrefix = parser.getNamespacePrefix(i);
                             String nsUri = parser.getNamespaceURI(i);
-                            attrs.put(nsPrefix != null && !nsPrefix.isEmpty() ? "xmlns:" + nsPrefix : "xmlns", nsUri);
+                            if (nsPrefix != null && !nsPrefix.isEmpty()) {
+                                nsContext.put(nsPrefix, nsUri);
+                                attrs.put("xmlns:" + nsPrefix, nsUri);
+                            } else {
+                                attrs.put("xmlns", nsUri);
+                            }
                         }
                         for (int i = 0; i < attributesSize; i++) {
-                            String aName = parser.getAttributeLocalName(i);
-                            String aValue = parser.getAttributeValue(i);
-                            String aPrefix = parser.getAttributePrefix(i);
-                            if (aPrefix != null && !aPrefix.isEmpty()) {
-                                aName = aPrefix + ":" + aName;
+                            String attrName = parser.getAttributeLocalName(i);
+                            String attrValue = parser.getAttributeValue(i);
+                            String attrPrefix = parser.getAttributePrefix(i);
+                            if (attrPrefix != null && !attrPrefix.isEmpty()) {
+                                attrName = attrPrefix + ":" + attrName;
                             }
-                            attrs.put(aName, aValue);
-                            spacePreserve = spacePreserve || ("xml:space".equals(aName) && "preserve".equals(aValue));
+                            attrs.put(attrName, attrValue);
+                            spacePreserve =
+                                    spacePreserve || ("xml:space".equals(attrName) && "preserve".equals(attrValue));
                         }
                     }
                 } else {
                     if (children == null) {
                         children = new ArrayList<>();
                     }
-                    XmlNode child = doBuild(parser, trim, locationBuilder, nsContext != null ? nsContext : Map.of());
+                    XmlNode child = doBuild(parser, trim, locationBuilder, nsContext);
                     children.add(child);
                 }
             } else if (eventType == XMLStreamReader.CHARACTERS || eventType == XMLStreamReader.CDATA) {
                 String text = parser.getText();
-                lValue = lValue != null ? lValue + text : text;
+                elementValue = elementValue != null ? elementValue + text : text;
             } else if (eventType == XMLStreamReader.END_ELEMENT) {
                 boolean emptyTag = lastStartTag
                         == parser.getLocation().getLineNumber() * 1000
                                 + parser.getLocation().getColumnNumber();
-                if (lValue != null && trim && !spacePreserve) {
-                    lValue = lValue.trim();
+                if (elementValue != null && trim && !spacePreserve) {
+                    elementValue = elementValue.trim();
                 }
                 return XmlNode.newBuilder()
-                        .prefix(lPrefix)
-                        .namespaceUri(lNamespaceUri)
-                        .name(lName)
-                        .value(children == null ? (lValue != null ? lValue : emptyTag ? null : "") : null)
+                        .prefix(elementPrefix)
+                        .namespaceUri(elementNamespaceUri)
+                        .name(elementName)
+                        .value(children == null ? (elementValue != null ? elementValue : emptyTag ? null : "") : null)
                         .attributes(attrs)
                         .namespaces(nsContext)
                         .children(children)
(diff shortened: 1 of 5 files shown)


elharo (changes requested):
This seems to be based on a common misunderstanding of how namespaces work in attributes. Unprefixed attributes are never in any namespace. Not the default namespace of a document, not the namespace of the parent element. The only way an attribute has a namespace name is if it has a prefix.

gnodet (commented):

@elharo You're absolutely right — thanks for catching this. The previous version of the PR was based on the incorrect assumption that `mvn:combine.children` and `combine.children` are semantically identical when the prefix maps to the same namespace. Per the XML Namespaces spec, unprefixed attributes are in **no namespace**, so they are indeed different attributes.

We verified this with a test project on Maven 3.9.11:
- `combine.children="append"` (unprefixed) → works correctly, both parent and child compiler args are present
- `mvn:combine.children="append"` (prefixed) → **silently ignored**, only child args are present (default merge replaces by name)

So `mvn:combine.children` **never worked** in Maven 3 either. The Maven 4 regression is narrower: the consumer POM transformation drops the `xmlns:mvn` declaration (which lives on `<project>` and is lost during model serialization) but the prefixed attributes survive in plugin configuration `XmlNode` trees, producing invalid XML with undeclared namespace prefixes.

The PR has been reworked to be a **write-side-only fix**:
- No read-side normalization (the XML is preserved as-is)
- `xmlns:prefix` entries are written as proper namespace declarations via `writeNamespace()`
- Prefixed attributes with no corresponding `xmlns:` declaration have their prefix stripped to produce valid XML
- A test explicitly confirms that `mvn:combine.children` does NOT trigger merge behavior (correct per the XML spec)

impl/maven-xml/src/main/java/org/apache/maven/internal/xml/DefaultXmlService.java
```diff
@@ -178,6 +176,42 @@ private void writeNode(XMLStreamWriter xmlWriter, XmlNode node) throws XMLStream
         xmlWriter.writeEndElement();
     }
 
+    static void writeAttributes(XMLStreamWriter xmlWriter, Map<String, String> attributes) throws XMLStreamException {
```
elharo: This looks like it can be private

impl/maven-xml/src/main/java/org/apache/maven/internal/xml/DefaultXmlService.java
```diff
@@ -178,6 +176,42 @@ private void writeNode(XMLStreamWriter xmlWriter, XmlNode node) throws XMLStream
...
+        // Write namespace declarations first, then regular attributes
+        for (Map.Entry<String, String> attr : attributes.entrySet()) {
```
elharo: Not sure about what XML library you're using here. Some do not include namespace declarations in their attribute map. If you're using DOM I think it depends on how the parser is configured, though I'd really need to review to make remember exactly how that works.

impl/maven-xml/src/main/java/org/apache/maven/internal/xml/DefaultXmlService.java
```diff
@@ -178,6 +176,42 @@ private void writeNode(XMLStreamWriter xmlWriter, XmlNode node) throws XMLStream
...
+            String key = attr.getKey();
+            if ("xmlns".equals(key)) {
+                xmlWriter.writeDefaultNamespace(attr.getValue());
+            } else if (key.startsWith("xmlns:")) {
+                xmlWriter.writeNamespace(key.substring(6), attr.getValue());
+            }
+        }
+        for (Map.Entry<String, String> attr : attributes.entrySet()) {
```
elharo: attr --> attribute

impl/maven-xml/src/main/java/org/apache/maven/internal/xml/DefaultXmlService.java
```diff
@@ -178,6 +176,42 @@ private void writeNode(XMLStreamWriter xmlWriter, XmlNode node) throws XMLStream
...
+            String key = attr.getKey();
+            String value = attr.getValue();
+            if ("xmlns".equals(key) || key.startsWith("xmlns:")) {
+                continue; // already written above
+            } else if (key.startsWith("xml:")) {
```
elharo, line 232: The xml prefix is predefined and understood by all parsers. It does not need to be (must not be?) explicitly declared. You do need the else if case here to keep it from being written, but it should be empty.

impl/maven-xml/src/test/java/org/apache/maven/internal/xml/XmlNodeImplTest.java
```diff
@@ -715,6 +717,198 @@ public Object toInputLocation(XMLStreamReader parser) {
...
+                .attributes(java.util.Map.of(
+                        "xmlns:custom", "http://example.com/custom",
+                        "custom:myattr", "value"))
+                .children(java.util.List.of(XmlNode.newBuilder()
+                        .name("arg")
+                        .value("-Xlint:deprecation")
+                        .build()))
+                .build();
+
+        java.io.StringWriter writer = new java.io.StringWriter();
```
elharo: I'd just import this class instead of using the fully package qualified name

impl/maven-xml/src/test/java/org/apache/maven/internal/xml/XmlNodeImplTest.java
```diff
@@ -715,6 +717,198 @@ public Object toInputLocation(XMLStreamReader parser) {
...
+        String output = writer.toString();
+
+        XmlNode reRead = toXmlNode(output);
+        assertNotNull(reRead);
+        assertEquals("value", reRead.attribute("custom:myattr"));
+        assertEquals("http://example.com/custom", reRead.attribute("xmlns:custom"));
+    }
+
+    /**
+     * Verifies that when a prefixed attribute's xmlns declaration is on a parent
```
elharo: Not sure what this is trying to test but this doesn't sound right. I don't think the prefix should be removed.

elharo (changes requested):

impl/maven-xml/src/main/java/org/apache/maven/internal/xml/DefaultXmlService.java
gnodet: _Claude Code on behalf of Guillaume Nodet_

Good catch — made it `private`. It's not called from the generated template (which has its own copy of the logic).

impl/maven-xml/src/main/java/org/apache/maven/internal/xml/DefaultXmlService.java
gnodet: _Claude Code on behalf of Guillaume Nodet_

Good point. Maven uses StAX (XMLStreamReader) for parsing. The namespace declarations are reported separately via `getNamespaceCount()`/`getNamespacePrefix()`/`getNamespaceURI()` — they are NOT included in the attribute list (`getAttributeCount()`). However, during parsing in `doBuild()`, we explicitly store them as `xmlns:prefix` entries in the attributes map (see line 99-103). So by the time we write, they are reliably present in the map if they were on the original element.

impl/maven-xml/src/main/java/org/apache/maven/internal/xml/DefaultXmlService.java
gnodet: _Claude Code on behalf of Guillaume Nodet_

Fixed — renamed to `attribute`.

impl/maven-xml/src/main/java/org/apache/maven/internal/xml/DefaultXmlService.java
gnodet: _Claude Code on behalf of Guillaume Nodet_

You're right that the `xml:` prefix is predefined and must not be declared. However, we can't make this branch empty — attributes like `xml:space="preserve"` still need to be written to the output. If we skip them, whitespace preservation would be lost on round-trip. The `writeAttribute(namespaceURI, localName, value)` call is the correct StAX way to write them — it writes the attribute with the proper namespace without declaring the prefix (since it's predefined). Added a comment to explain this.

impl/maven-xml/src/test/java/org/apache/maven/internal/xml/XmlNodeImplTest.java
gnodet: _Claude Code on behalf of Guillaume Nodet_

Fixed — added proper imports for `StringWriter`, `Map`, and `List`.

impl/maven-xml/src/test/java/org/apache/maven/internal/xml/XmlNodeImplTest.java
gnodet: _Claude Code on behalf of Guillaume Nodet_

This test reproduces the exact bug from issue #11760. Here's what happens in the consumer POM transformation:

1. The original POM has `xmlns:mvn="http://maven.apache.org/POM/4.0.0"` on the `<project>` element
2. Plugin configuration (e.g., `<compilerArgs mvn:combine.children="append">`) is stored as `XmlNode` trees — the attribute `mvn:combine.children` survives
3. During consumer POM transformation, the model is re-serialized from the `Model` object. The `xmlns:mvn` declaration on `<project>` is **lost** (it's not part of the Maven model)
4. When the `XmlNode` is written, `mvn:combine.children` has no corresponding namespace declaration → "Undeclared namespace prefix" error

The choice is: produce invalid XML (crash) or strip the orphaned prefix. We chose to strip the prefix since it's the only way to produce valid XML when the namespace declaration has been lost. Updated the test javadoc to explain this context.

gnodet: _Claude Code on behalf of Guillaume Nodet_

You were right — the prefix should not be removed when the namespace URI is known. We've now addressed this properly in the latest commit by adding a `namespaces()` method to `XmlNode` that captures the full namespace context (prefix → URI) inherited from ancestor elements during parsing.

When writing, if a prefixed attribute like `mvn:combine.children` appears but `xmlns:mvn` is not in the local attribute map, the writer now looks up the prefix in the namespace context, auto-declares the namespace, and writes the attribute with its proper prefix and namespace URI.

The prefix is only stripped as a last resort when no namespace context is available at all (e.g., a programmatically built `XmlNode` with no namespace bindings). This produces valid XML in all cases.

impl/maven-xml/src/main/java/org/apache/maven/internal/xml/DefaultXmlService.java
```diff
@@ -93,6 +99,15 @@ private XmlNode doBuild(XMLStreamReader parser, boolean trim, InputLocationBuild
                     lNamespaceUri = parser.getNamespaceURI();
                     lName = parser.getLocalName();
                     location = locationBuilder != null ? locationBuilder.toInputLocation(parser) : null;
+                    // Build the namespace context: start with inherited, add local declarations
+                    nsContext = new HashMap<>(parentNamespaces);
+                    for (int i = 0; i < namespacesSize; i++) {
+                        String nsPrefix = parser.getNamespacePrefix(i);
+                        String nsUri = parser.getNamespaceURI(i);
+                        if (nsPrefix != null && !nsPrefix.isEmpty()) {
```
elharo: so you're not putting the default namespace in the map? That's fine as long as the calling code expects that

impl/maven-xml/src/main/java/org/apache/maven/internal/xml/DefaultXmlService.java
elharo: I'm not sure why you check lName == null here. It might be fine but naively I'd expect this to be set unconditionally on a startElement

impl/maven-xml/src/main/java/org/apache/maven/internal/xml/DefaultXmlService.java
```diff
@@ -68,10 +69,14 @@ public XmlNode doRead(Reader reader, @Nullable XmlService.InputLocationBuilder l
...
 
-    private XmlNode doBuild(XMLStreamReader parser, boolean trim, InputLocationBuilder locationBuilder)
+    private XmlNode doBuild(
+            XMLStreamReader parser,
+            boolean trim,
+            InputLocationBuilder locationBuilder,
+            Map<String, String> parentNamespaces)
             throws XMLStreamException {
         boolean spacePreserve = false;
         String lPrefix = null;
```
elharo: I'm  not sure what all these l's stand for. local? location? something else? Unabbreviated names for the win.

elharo (approved):

impl/maven-xml/src/main/java/org/apache/maven/internal/xml/DefaultXmlService.java
gnodet: _Claude Code on behalf of Guillaume Nodet_

Agreed — renamed `lPrefix` → `elementPrefix`, `lName` → `elementName`, `lNamespaceUri` → `elementNamespaceUri`, `lValue` → `elementValue`. Also renamed `aName`/`aValue`/`aPrefix` → `attrName`/`attrValue`/`attrPrefix` in the attribute loop.

impl/maven-xml/src/main/java/org/apache/maven/internal/xml/DefaultXmlService.java
gnodet: _Claude Code on behalf of Guillaume Nodet_

Good question — the `elementName == null` check is how this method distinguishes the first `START_ELEMENT` event (which is "this" element being parsed) from subsequent `START_ELEMENT` events (which are children, handled in the `else` branch). The method is called recursively for each child, and `elementName` starts as `null` — once the first start tag is seen, it gets set, so all further start tags within the same call are children. Added a comment to clarify this.

impl/maven-xml/src/main/java/org/apache/maven/internal/xml/DefaultXmlService.java
gnodet: _Claude Code on behalf of Guillaume Nodet_

Correct — the default namespace (empty prefix) is intentionally excluded from `namespaces()`. Per the XML namespace spec (Section 6.2), default namespace declarations do NOT apply to attributes, so it would never be useful for resolving prefixed attributes. Added a comment explaining this. We also have `testParseDefaultNamespaceNotInNamespacesMap` that verifies this behavior.

PR #11824, opened 2026-03-23
Fix MAVEN_ARGS parsing to handle quoted strings with spaces

### LINKED ISSUE #11487: [regression] Maven4 unable to pass settings path on Windows
### Affected version

4.0.0-rc-5

### Bug description

Updating to Maven4 breaks my usage scenario:
```
$ echo $MAVEN_ARGS
-s D:\projects\mmm\conf\mvn\settings.xml -Dsettings.security=D:\projects\mmm\conf\mvn\settings-security.xml
$ mvn -v
[ERROR] The specified user settings file does not exist: D:\projects\mmm\projectsmmmconfmvnsettings.xml
```

To me it looks as if maven4 does some transformation with backslashes in path's.
Maybe also the wrapper script calling maven via Java is causing this effect.
I diffed the `mvn` wrapper script and found that in 3.9.9 it was containing this relevant block that was removed in maven 4.0:
```
# For MinGW, ensure paths are in Unix format before anything is touched
if $mingw ; then
  [ -n "$MAVEN_HOME" ] &&
    MAVEN_HOME=`(cd "$MAVEN_HOME"; pwd)`
  [ -n "$JAVA_HOME" ] &&
    JAVA_HOME=`(cd "$JAVA_HOME"; pwd)`
  # TODO classpath?
fi
```
UPDATE: I do not believe this is related to the bug. I rather expect the bug to be in the Java code.

When I downgrade to maven 3.x it works again:
```
$ mvn -v
WARNING: A restricted method in java.lang.System has been called
WARNING: java.lang.System::load has been called by org.fusesource.jansi.internal.JansiLoader in an unnamed module (file:/D:/projects/mmm/software/mvn/lib/jansi-2.4.1.jar)
WARNING: Use --enable-native-access=ALL-UNNAMED to avoid a warning for callers in this module
WARNING: Restricted methods will be blocked in a future release unless native access is enabled

Apache Maven 3.9.9 (8e8579a9e76f7d015ee5ec7bfcdc97d260186937)
Maven home: D:\projects\mmm\software\mvn
Java version: 25, vendor: Eclipse Adoptium, runtime: D:\projects\mmm\software\java
Default locale: en_US, platform encoding: UTF-8
OS name: "windows 11", version: "10.0", arch: "amd64", family: "windows"
```

gnodet:
## Summary
- Replace naive `String.split(" ")` with a proper argument parser that handles single and double quoted strings, preserving backslashes for Windows paths
- Applied to both `EmbeddedMavenExecutor` and `ForkedMavenExecutor`
- Added unit tests for both executors covering simple args, quoted paths, empty input, and extra whitespace

Fixes #11487

## Test plan
- [x] Unit tests for `parseArguments()` method in both executor implementations
- [ ] Manual test with `MAVEN_ARGS='-f "C:\Program Files\project\pom.xml"'` on Windows

### DIFF
--- a/impl/maven-executor/src/main/java/org/apache/maven/cling/executor/embedded/EmbeddedMavenExecutor.java
+++ b/impl/maven-executor/src/main/java/org/apache/maven/cling/executor/embedded/EmbeddedMavenExecutor.java
@@ -31,7 +31,6 @@ import java.net.URLClassLoader;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.util.ArrayList;
-import java.util.Arrays;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.HashMap;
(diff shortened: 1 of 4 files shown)


gnodet:
Closing — the `parseArguments()` fix with proper quote handling is already on master, incorporated during the executor refactoring in #11976 / #12004.

Issue #11827, opened 2026-03-23
Maven DI crashes if the file org.apache.maven.api.di.Inject contains empty lines


jafarre-bi:
### Affected version

4.0.0-rc-5

### Bug description

I'm concatenating several org.apache.maven.api.di.Inject files using the AppendingTransformer of the Shade plugin. This transformer adds an empty line after each appended file. Maven DI treats empty lines as if they contained an empty class name, resulting in errors.
Stack trace:
```
mar 23, 2026 3:31:04 P. M. com.google.inject.internal.MessageProcessor visit
INFO: An exception was caught and reported. Message: java.lang.ClassNotFoundException:
org.apache.maven.di.impl.DIException: Error while discovering DI classes from classLoader
        at org.apache.maven.di.impl.InjectorImpl.discover(InjectorImpl.java:111)
        at org.apache.maven.internal.impl.SisuDiBridgeModule.configure(SisuDiBridgeModule.java:87)
        at com.google.inject.AbstractModule.configure(AbstractModule.java:66)
        at com.google.inject.spi.Elements$RecordingBinder.install(Elements.java:409)
        at org.eclipse.sisu.space.QualifiedTypeBinder.installModule(QualifiedTypeBinder.java:153)
        at org.eclipse.sisu.space.QualifiedTypeBinder.hear(QualifiedTypeBinder.java:119)
        at org.eclipse.sisu.plexus.PlexusTypeBinder.hear(PlexusTypeBinder.java:57)
        at org.eclipse.sisu.space.QualifiedTypeVisitor.leaveClass(QualifiedTypeVisitor.java:136)
        at org.eclipse.sisu.plexus.PlexusTypeVisitor.leaveClass(PlexusTypeVisitor.java:119)
        at org.eclipse.sisu.space.SpaceScanner$1.visitEnd(SpaceScanner.java:218)
        at org.objectweb.asm.ClassReader.accept(ClassReader.java:749)
        at org.objectweb.asm.ClassReader.accept(ClassReader.java:425)
        at org.eclipse.sisu.space.SpaceScanner.accept(SpaceScanner.java:143)
        at org.eclipse.sisu.space.SpaceScanner.accept(SpaceScanner.java:109)
        at org.eclipse.sisu.space.SpaceModule.scanForElements(SpaceModule.java:218)
        at org.eclipse.sisu.space.SpaceModule.configure(SpaceModule.java:168)
[...]

PR #11828, opened 2026-03-23
Fixes issue #11827 - Maven DI crashes if the file org.apache.maven.api.di.Inject contains empty lines

jafarre-bi:
Adds a check to ignore blank lines when reading org.apache.maven.api.di.Inject.

### DIFF
--- a/impl/maven-di/src/main/java/org/apache/maven/di/impl/InjectorImpl.java
+++ b/impl/maven-di/src/main/java/org/apache/maven/di/impl/InjectorImpl.java
@@ -100,7 +100,9 @@ public class InjectorImpl implements Injector {
                             BufferedReader reader =
                                     new BufferedReader(new InputStreamReader(Objects.requireNonNull(is)))) {
                         for (String line : reader.lines()
-                                .filter(l -> !l.isBlank() && !l.startsWith("#"))
+                                .map(String::trim)
+                                .filter(l -> !l.isEmpty())
+                                .filter(l -> !l.startsWith("#"))
                                 .toList()) {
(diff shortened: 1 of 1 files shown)


impl/maven-di/src/main/java/org/apache/maven/di/impl/InjectorImpl.java
```diff
@@ -99,8 +99,9 @@ public Injector discover(@Nonnull ClassLoader classLoader) {
                     try (InputStream is = url.openStream();
                             BufferedReader reader =
                                     new BufferedReader(new InputStreamReader(Objects.requireNonNull(is)))) {
-                        for (String line :
-                                reader.lines().filter(l -> !l.startsWith("#")).toList()) {
+                        for (String line : reader.lines()
+                                .filter(l -> !l.isBlank() && !l.startsWith("#"))
```
slawekjaranowski: we can also add `.map(String::trim)` before filters
each condition can be in separate `filter` call

jafarre-bi: Well, the trim isn't necessary with isBlank(). As for one or two filters, I don't mind either way. Should I change it?

jafarre-bi: I understand. If we add the trim, we prevent problems if a class name has leading or trailing whitespace. Let's do it.

jafarre-bi: I just pushed the changes. There's a workflow awaiting approval.

PR #11829, opened 2026-03-23
Refactor JUnit extensions to avoid using static fields.

slawekjaranowski:
Static filed in Junit extension can impact on tests executed in parallel.

For static methods preserver current test context in local thread.

---

### DIFF
--- a/impl/maven-testing/src/main/java/org/apache/maven/api/di/testing/MavenDIExtension.java
+++ b/impl/maven-testing/src/main/java/org/apache/maven/api/di/testing/MavenDIExtension.java
@@ -19,6 +19,7 @@
 package org.apache.maven.api.di.testing;
 
 import java.io.File;
+import java.util.Optional;
 
 import org.apache.maven.di.Injector;
 import org.apache.maven.di.Key;
@@ -52,9 +53,11 @@ import org.junit.jupiter.api.extension.ExtensionContext;
  * </pre>
  */
 public class MavenDIExtension implements BeforeEachCallback, AfterEachCallback {
-    protected static ExtensionContext context;
-    protected Injector injector;
-    protected static String basedir;
+
+    private static final ExtensionContext.Namespace MAVEN_DI_EXTENSION =
+            ExtensionContext.Namespace.create("maven-di-extension");
+
+    private static final ThreadLocal<ExtensionContext> EXTENSION_CONTEXT_THREAD_LOCAL = new ThreadLocal<>();
 
     /**
      * Initializes the test environment before each test method execution.
@@ -65,7 +68,6 @@ public class MavenDIExtension implements BeforeEachCallback, AfterEachCallback {
      */
     @Override
     public void beforeEach(ExtensionContext context) throws Exception {
-        basedir = getBasedir();
         setContext(context);
         getInjector().bindInstance((Class<Object>) context.getRequiredTestClass(), context.getRequiredTestInstance());
         getInjector().injectInstance(context.getRequiredTestInstance());
@@ -76,10 +78,13 @@ public class MavenDIExtension implements BeforeEachCallback, AfterEachCallback {
      *
      * @param context The extension context to store
      */
-    protected void setContext(ExtensionContext context) {
-        MavenDIExtension.context = context;
+    protected static void setContext(ExtensionContext context) {
+        EXTENSION_CONTEXT_THREAD_LOCAL.set(context);
     }
 
+    protected static Optional<ExtensionContext> getContext() {
+        return Optional.ofNullable(EXTENSION_CONTEXT_THREAD_LOCAL.get());
+    }
     /**
      * Creates and configures the DI container for test execution.
      * Performs component discovery and sets up basic bindings.
@@ -87,7 +92,7 @@ public class MavenDIExtension implements BeforeEachCallback, AfterEachCallback {
      * @throws IllegalStateException if the ExtensionContext is null, the required test class is unavailable,
      *         the required test instance is unavailable, or if container setup fails
      */
-    protected void setupContainer() {
+    protected Injector setupContainer(ExtensionContext context) {
         if (context == null) {
             throw new IllegalStateException("ExtensionContext must not be null");
         }
@@ -101,11 +106,12 @@ public class MavenDIExtension implements BeforeEachCallback, AfterEachCallback {
         }
 
         try {
-            injector = Injector.create();
+            Injector injector = Injector.create();
             injector.bindInstance(ExtensionContext.class, context);
             injector.discover(testClass.getClassLoader());
             injector.bindInstance(Injector.class, injector);
             injector.bindInstance(testClass.asSubclass(Object.class), (Object) testInstance); // Safe generics handling
+            return injector;
         } catch (final Exception e) {
             throw new IllegalStateException(
                     String.format(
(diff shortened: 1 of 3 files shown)


impl/maven-testing/src/main/java/org/apache/maven/api/plugin/testing/MojoExtension.java
```diff
@@ -331,23 +333,23 @@ private static String getGoalFromMojoImplementationClass(Class<?> cl) throws IOE
...
-                .orElse(pluginBasedir);
-        if (basedir != null) {
-            if (basedir.isEmpty()) {
-                basedir = pluginBasedir + "/target/tests/"
-                        + context.getRequiredTestClass().getSimpleName() + "/"
-                        + context.getRequiredTestMethod().getName();
-            } else {
-                basedir = basedir.replace("${basedir}", pluginBasedir);
-            }
+                .orElse(null);
```
gnodet: **Bug: behavioral change when `@Basedir` is absent.**

The original code uses `.orElse(pluginBasedir)`, so when `@Basedir` is not present, `basedir = pluginBasedir` (a real, non-empty path). Then the `if (basedir.isEmpty())` check is false, so `basedir` stays as `pluginBasedir`.

With this change (`.orElse(null)`), when `@Basedir` is absent, `basedir == null` enters the `if` branch and `basedir = pluginBasedir + "/target/tests/ClassName/methodName"` — a completely different directory.

This contradicts the `@Basedir` Javadoc: _"If not specified, the plugin's base directory will be used as the default."_ and will break any plugin test that doesn't use `@Basedir` but expects `getBasedir()` to return the plugin root.

Suggested fix — restore original semantics:
```java
.orElse(pluginBasedir);

if (basedir.isEmpty()) {
    basedir = pluginBasedir + "/target/tests/"
            + context.getRequiredTestClass().getSimpleName() + "/"
            + context.getRequiredTestMethod().getName();
} else {
    basedir = basedir.replace("${basedir}", pluginBasedir);
}
```

impl/maven-testing/src/main/java/org/apache/maven/api/plugin/testing/MojoExtension.java
```diff
@@ -437,8 +439,8 @@ public void beforeEach(ExtensionContext context) throws Exception {
             model = new MavenMerger().merge(tmodel, defaultModel, false, null);
         }
         tmodel = new DefaultModelPathTranslator(new DefaultPathTranslator())
-                .alignToBaseDirectory(tmodel, Paths.get(getBasedir()), null);
-        context.getStore(ExtensionContext.Namespace.GLOBAL).put(Model.class, tmodel);
+                .alignToBaseDirectory(model, Paths.get(getBasedir()), null);
```
gnodet: **Undocumented behavior change: `tmodel` to `model`.**

The original passes `tmodel` (the raw parsed POM) to `alignToBaseDirectory`, while this PR passes `model` (the merged model). This is arguably an improvement — it fixes a potential NPE when `tmodel` is null, and aligning the merged model makes more sense. But it's a behavioral change mixed into a refactoring PR without being called out.

Consider either:
- Documenting this in the PR description as an intentional fix, or
- Splitting it into a separate commit

impl/maven-testing/src/main/java/org/apache/maven/api/di/testing/MavenDIExtension.java
```diff
@@ -76,18 +78,21 @@ public void beforeEach(ExtensionContext context) throws Exception {
      *
      * @param context The extension context to store
      */
-    protected void setContext(ExtensionContext context) {
-        MavenDIExtension.context = context;
+    protected static void setContext(ExtensionContext context) {
+        EXTENSION_CONTEXT_THREAD_LOCAL.set(context);
     }
 
+    protected static Optional<ExtensionContext> getContext() {
```
gnodet, line 90: Nit: missing blank line before the Javadoc block (style consistency with the rest of the file).

impl/maven-testing/src/main/java/org/apache/maven/api/di/testing/MavenDIExtension.java
```diff
@@ -76,18 +78,21 @@ public void beforeEach(ExtensionContext context) throws Exception {
      *
```
gnodet, line 78: Note: this changes the method signature from `protected` instance to `protected static`. This is a public API change — any downstream class overriding `setContext` would break at compile time. In this codebase only `MojoExtension` extends `MavenDIExtension` (and doesn't override it), but `maven-testing` is a public artifact used by plugin developers.

gnodet (commented):
## Review Summary

The direction of this PR is sound — replacing static mutable state with `ThreadLocal` and JUnit `ExtensionContext.Store` is the idiomatic JUnit 5 approach for parallel test support. The namespace isolation (`MAVEN_DI_EXTENSION` / `MOJO_EXTENSION`) instead of `Namespace.GLOBAL` is also a good improvement.

However, there are a couple of behavioral changes mixed into this refactoring that need attention:

| # | Severity | Issue |
|---|----------|-------|
| 1 | **Bug** | `@Basedir` absent → basedir now resolves to a test-specific subdirectory instead of `pluginBasedir` |
| 2 | **Medium** | `alignToBaseDirectory(tmodel, ...)` → `alignToBaseDirectory(model, ...)` is a behavior change that should be documented or separated |
| 3 | **Nit** | Missing blank line before Javadoc |
| 4 | **Info** | `setContext` changed from `protected` instance to `protected static` — public API change |

impl/maven-testing/src/main/java/org/apache/maven/api/plugin/testing/MojoExtension.java
slawekjaranowski: ok revert it .... but looks like bug - to fix in separate one

impl/maven-testing/src/main/java/org/apache/maven/api/plugin/testing/MojoExtension.java
slawekjaranowski: looks like `basedir` can be only empty if system property `basedir` will be empty
more tests are needed for basedir annotation

impl/maven-testing/src/main/java/org/apache/maven/api/di/testing/MavenDIExtension.java
```diff
@@ -123,10 +135,11 @@ protected void setupContainer() {
      */
     @Override
     public void afterEach(ExtensionContext context) throws Exception {
+        Injector injector = context.getStore(MAVEN_DI_EXTENSION).get(Injector.class, Injector.class);
         if (injector != null) {
```
gnodet: Suggestion: if `injector.dispose()` throws, the ThreadLocal is never cleaned up. Consider wrapping in try-finally:
```java
try {
    Injector injector = context.getStore(MAVEN_DI_EXTENSION).get(Injector.class, Injector.class);
    if (injector != null) {
        injector.dispose();
    }
} finally {
    EXTENSION_CONTEXT_THREAD_LOCAL.remove();
}
```

gnodet (commented):

Thanks for the quick fixes! The basedir behavior is now correct, the `tmodel`/`model` change has been reverted, the blank line is fixed, and `setContext` stays as a `protected` instance method. One minor remaining suggestion:

impl/maven-testing/src/main/java/org/apache/maven/api/di/testing/MavenDIExtension.java
slawekjaranowski: done

Bukama (approved):
These are the small buter, but important changes for improved performance :)

gnodet (approved):

gnodet:

Note: as discussed during review, there's a pre-existing issue in `MojoExtension.beforeEach()` — `alignToBaseDirectory` is called on `tmodel` (the raw parsed POM) instead of `model` (the merged model). This means:
1. If no POM is found, `tmodel` is `null` and `alignToBaseDirectory` will NPE
2. The stored model misses the defaults from the merge (groupId, artifactId, build paths, etc.)

This should be addressed in a follow-up.

PR #11851, opened 2026-03-26
Trim `threadConfiguration` to accept input surrounded with spaces

### LINKED ISSUE #11850: Allow argument to `--threads` to be surrounded by spaces

1. Given the command:
```
mvn -V -q help:help
```
2. The command, with `.mvn/maven.config`:
```
-ntp
--threads
  1C
```
gives:
```
Apache Maven 3.9.14 (996c630dbc656c76214ce58821dcc58be960875b)
...
Java version: 21.0.10, vendor: Red Hat, Inc., runtime: ...
Default locale: en_GB, platform encoding: UTF-8
OS name: "linux" ...
```
It's fine.
3. The same command with `.mvn/maven.config`:
```
-ntp
--threads
  1
```
gives:
```
Apache Maven 3.9.14 (996c630dbc656c76214ce58821dcc58be960875b)
...
Java version: 21.0.10, vendor: Red Hat, Inc., runtime: ...
Default locale: en_GB, platform encoding: UTF-8
OS name: "linux" ...
[ERROR] Invalid threads value: ' 1'. Supported are integer values.
```
4. Command could be also used w/o `maven.config` like so:
```
mvn -V -q help:help -T ' 1C'             # a
mvn -V -q help:help -T ' 1'              # b
```
which also works for case _a_ and doesn't for _b_.
5. Having leading or trailing space in option gives something like
```
org.apache.commons.cli.ParseException: Unrecognized maven.config file entries: [ -ntp ]
```
or
```
Unable to parse maven.config file options: Unrecognized option: -ntp
```
which might come from `commons-cli`. 

But as parsing threads value is done in Maven - could it be always trimmed first?

This is for `Float.parseFloat` (`valueOf`): https://docs.oracle.com/javase/8/docs/api/java/lang/Float.html#valueOf-java.lang.String-:
> Leading and trailing whitespace characters in `s` are ignored. Whitespace is removed as if by the [String.trim()](https://docs.oracle.com/javase/8/docs/api/java/lang/String.html#trim--) method; that is, both ASCII space and control characters are removed. **The rest** of `s` should constitute a FloatValue ...

But nothing like that happens in `Integer.parseInt`.

pzygielo:
- space-prefixed core multiplier - is already _supported_ - added support for space-suffix
- extended that behaviour for direct number of threads argument
- can be used in bizarre command line like `mvn -T ' 1 '` or - why I'm here - with value indented in `maven.config`:
```
--threads
  1
```

-----
https://github.com/apache/maven/blob/9ac23b040017e7b14877fbc9e74c696d36cba3f9/CONTRIBUTING.md?plain=1#L58

```
[WARNING] The requested profile "run-its" could not be activated because it does not exist.
```

-----

-----

- #11850

### DIFF
--- a/maven-embedder/src/main/java/org/apache/maven/cli/MavenCli.java
+++ b/maven-embedder/src/main/java/org/apache/maven/cli/MavenCli.java
@@ -1477,17 +1477,17 @@ public class MavenCli {
             }
         } else {
             try {
-                int threads = Integer.parseInt(originalThreadConfiguration);
+                int threads = Integer.parseInt(threadConfiguration);
 
                 if (threads <= 0) {
                     throw new IllegalArgumentException(
-                            "Invalid threads value: '" + originalThreadConfiguration + "'. Value must be positive.");
+                            "Invalid threads value: '" + threadConfiguration + "'. Value must be positive.");
                 }
 
                 return threads;
             } catch (NumberFormatException e) {
                 throw new IllegalArgumentException(
-                        "Invalid threads value: '" + originalThreadConfiguration + "'. Supported are integer values.");
+                        "Invalid threads value: '" + threadConfiguration + "'. Supported are integer values.");
             }
         }
     }
--- a/maven-embedder/src/main/java/org/apache/maven/cli/MavenCli.java
+++ b/maven-embedder/src/main/java/org/apache/maven/cli/MavenCli.java
@@ -1453,7 +1453,8 @@ public class MavenCli {
         return request;
     }
 
-    int calculateDegreeOfConcurrency(String threadConfiguration) {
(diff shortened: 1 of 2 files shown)


slachiewicz:
but maybe also good to backport it to 3.9 line?

pzygielo:
> we are starting on next minor version 3.10.x so new PR should be target to this branch, I will try to change a base branch ...

So... this was retargetted to 3.10.x, but...

> but maybe also good to backport it to 3.9 line?

didn't get ported :disappointed: 

@slawekjaranowski What's the plan to have 3.10 out?

slawekjaranowski:
@pzygielo  for 3.9.x

- https://github.com/apache/maven/pull/12042

PR #11857, opened 2026-03-28
Nuance the message about filename-based automodules detected on the module path

desruisseaux:
Instead of _"Please don't publish this project to a public artifact repository"_, said _"This project may not work in applications that do not use these dependencies with exactly the same filenames as listed above"_.

**Rational:** it can be considered okay to publish the project on a public Maven repository, because the artifact filenames are quite deterministic with Maven. For a pure-Maven project and for bundle generated by the Maven assembly plugin, there is no reason why these filenames would not be stable. Therefore, we can warn the users about the fragility of filename-based automodules but let them decide whether they want to publish anyway.

### DIFF
--- a/impl/maven-impl/src/main/java/org/apache/maven/impl/PathModularization.java
+++ b/impl/maven-impl/src/main/java/org/apache/maven/impl/PathModularization.java
@@ -236,7 +236,7 @@ final class PathModularization {
     /**
      * If the module has no name, adds the filename of the JAR file in the given collection.
      * This method should be invoked for dependencies placed on {@link JavaPathType#MODULES}
-     * for preparing a warning asking to not deploy the build artifact on a public repository.
+     * for preparing a warning saying that this project may fail to resolve these modules.
      * If the module has an explicit name either with a {@code module-info.class} file or with
      * an {@code "Automatic-Module-Name"} attribute in the {@code META-INF/MANIFEST.MF} file,
      * then this method does nothing.
--- a/impl/maven-impl/src/main/java/org/apache/maven/impl/PathModularization.java
+++ b/impl/maven-impl/src/main/java/org/apache/maven/impl/PathModularization.java
@@ -234,12 +234,15 @@ final class PathModularization {
     }
 
     /**
-     * If the module has no name, adds the filename of the JAR file in the given collection.
-     * This method should be invoked for dependencies placed on {@link JavaPathType#MODULES}
-     * for preparing a warning saying that this project may fail to resolve these modules.
-     * If the module has an explicit name either with a {@code module-info.class} file or with
-     * an {@code "Automatic-Module-Name"} attribute in the {@code META-INF/MANIFEST.MF} file,
-     * then this method does nothing.
+     * If the JAR has no {@code module-info.class} entry and no {@code Automatic-Module-Name}
+     * attribute in JAR's manifest, adds the filename of the JAR file in the given collection.
+     * This method should be invoked for all dependencies placed on the module-path,
+     * e.g. all dependencies of type {@link JavaPathType#MODULES}.
+     * If the module described by the {@code PathModularization} instance has an explicit name
+     * either with a {@code module-info.class} file or with an {@code "Automatic-Module-Name"}
+     * attribute in the {@code META-INF/MANIFEST.MF} file, then this method does nothing.
+     * Otherwise, this method adds an element in the given {@code automodulesDetected} collection.
+     * That collection will be used later for preparing a warning message.
      */
     public void addIfFilenameBasedAutomodules(Collection<String> automodulesDetected) {
         if (descriptors.isEmpty()) {
--- a/impl/maven-impl/src/main/java/org/apache/maven/impl/PathModularizationCache.java
+++ b/impl/maven-impl/src/main/java/org/apache/maven/impl/PathModularizationCache.java
@@ -189,10 +189,12 @@ final class PathModularizationCache {
             return Optional.empty();
         }
         String lineSeparator = System.lineSeparator();
+        String fileSeparator = lineSeparator + "  - ";
         var joiner = new StringJoiner(
-                lineSeparator + "  - ",
-                "Filename-based automodules detected on the module path: " + lineSeparator + "  - ",
-                lineSeparator + "Please don't publish this project to a public artifact repository.");
+                fileSeparator,
+                "Filename-based automodules detected on the module path: " + fileSeparator,
+                lineSeparator + "This project may not work in applications that do not use "
+                        + "these dependencies with exactly the same filenames as listed above.");
         automodulesDetected.forEach(joiner::add);
         return Optional.of(joiner.toString());
     }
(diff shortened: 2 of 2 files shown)


impl/maven-impl/src/main/java/org/apache/maven/impl/PathModularization.java
```diff
@@ -236,7 +236,7 @@ public JavaPathType getPathType() {
     /**
      * If the module has no name, adds the filename of the JAR file in the given collection.
      * This method should be invoked for dependencies placed on {@link JavaPathType#MODULES}
```
elharo: not new here, but this doc comment really is hard to follow or understand. It should be rewritten.

impl/maven-impl/src/main/java/org/apache/maven/impl/PathModularizationCache.java
```diff
@@ -189,10 +189,12 @@ Optional<String> warningForFilenameBasedAutomodules(Collection<Path> modulePaths
...
         String lineSeparator = System.lineSeparator();
+        String fileSeparator = lineSeparator + "  - ";
         var joiner = new StringJoiner(
-                lineSeparator + "  - ",
-                "Filename-based automodules detected on the module path: " + lineSeparator + "  - ",
-                lineSeparator + "Please don't publish this project to a public artifact repository.");
+                fileSeparator,
+                "Filename-based automodules detected on the module path: " + fileSeparator,
+                lineSeparator + "This project may not work in applications that do not use "
+                        + "these dependencies with exactly the same filenames as listed above.");
```
elharo: Is this only an issue when JPMS is used?

desruisseaux: Yes, this is an issue if non-modular JARs are placed on the `--module-path` option. While not recommended, this is not forbidden neither in order to allow gradual transition, because the effect is not the same as keeping those JARs on `--class-path`. The `javac` tool already emits a warning for such cases, and developers may accept the risk if they understand the condition in which it works, which is to keep the JAR filenames unchanged.

impl/maven-impl/src/main/java/org/apache/maven/impl/PathModularizationCache.java
```diff
@@ -189,10 +189,12 @@ Optional<String> warningForFilenameBasedAutomodules(Collection<Path> modulePaths
...
         automodulesDetected.forEach(joiner::add);
```
gnodet: The new suffix is fairly long and may be hard to parse in a log. Consider shortening it slightly, e.g.:

```suggestion
                lineSeparator + "This project may not work if consumers use "
                        + "these dependencies with different filenames.");
```

The original phrasing ("do not use these dependencies with exactly the same filenames as listed above") requires a double-negative mental parse ("not ... exactly the same"). "Different filenames" says the same thing more directly.

Feel free to ignore if you prefer the current wording.

impl/maven-impl/src/main/java/org/apache/maven/impl/PathModularization.java
```diff
@@ -236,7 +236,7 @@ public JavaPathType getPathType() {
...
-     * for preparing a warning asking to not deploy the build artifact on a public repository.
+     * for preparing a warning saying that this project may fail to resolve these modules.
```
gnodet: Nit: the Javadoc on this method still says "for preparing a warning asking to not deploy the build artifact on a public repository", which no longer matches what the warning actually says. Should be updated to stay consistent with the new message.

```suggestion
     * for preparing a warning saying that this project may fail to resolve these modules.
```

gnodet (commented):

The rationale is sound: the old message was overly prescriptive ("don't publish") when in practice Maven artifact filenames are deterministic. The new wording correctly shifts from a prohibition to an informational warning about the fragility.

Two minor observations below.

gnodet (approved):
LGTM — the new wording is a clear improvement. Minor nits noted inline.

gnodet:

I've pushed a follow-up commit to address the review suggestion about the warning message wording. The commit simplifies the double-negative phrasing:

**Before:**
> This project may not work in applications that do not use these dependencies with exactly the same filenames as listed above.

**After:**
> This project may not work if consumers use these dependencies with different filenames.

The Javadoc update in `PathModularization.java` was already applied in the existing commit.

The commit is available at [`gnodet/maven@nuance-automodule-warning`](https://github.com/gnodet/maven/tree/nuance-automodule-warning) — you can cherry-pick it with:

```
git remote add gnodet https://github.com/gnodet/maven.git
git fetch gnodet nuance-automodule-warning
git cherry-pick gnodet/nuance-automodule-warning
```

Or I can merge this into the main repo directly if you enable "Allow edits from maintainers" on this PR.

elharo (approved):

desruisseaux:
I couldn't find the "Allow edits from maintainers" checkbox. Even the [GitHub documentation](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork) does not match what I see. I will cherry pick. Thanks!

impl/maven-impl/src/main/java/org/apache/maven/impl/PathModularization.java
desruisseaux: Rewriting as below:

```java
    /**
     * If the JAR has no {@code module-info.class} entry and no {@code Automatic-Module-Name}
     * attribute in JAR's manifest, adds the filename of the JAR file in the given collection.
     * This method should be invoked for all dependencies placed on the module-path,
     * e.g. all dependencies of type {@link JavaPathType#MODULES}.
     * If the module described by the {@code PathModularization} instance has an explicit name
     * either with a {@code module-info.class} file or with an {@code "Automatic-Module-Name"}
     * attribute in the {@code META-INF/MANIFEST.MF} file, then this method does nothing.
     * Otherwise, this method adds an element in the given {@code automodulesDetected} collection.
     * That collection will be used later for preparing a warning message.
     */
```

impl/maven-impl/src/main/java/org/apache/maven/impl/PathModularization.java
desruisseaux: In this case, Claude evaluated the line which was suppressed by the commit instead of its replacement. This line is part of the paragraph rewritten in [above comment](https://github.com/apache/maven/pull/11857#discussion_r3292575014) anyway.

impl/maven-impl/src/main/java/org/apache/maven/impl/PathModularizationCache.java
desruisseaux: I clicked on "commit suggestion" too quickly. The replacement proposed by Claude is good, but it is replacing the wrong lines. I will fix and do a forced push.

desruisseaux:
I rebased this pull request on latest commit of Maven master branch and did `push --force`. Because of that, the changes reported above seem huge, but the real changes since the original pull request are [only this commit](https://github.com/apache/maven/pull/11857/changes/6edbc6fb5692e6ef0d8420f5f8f52e6d739ca611).

PR #11868, opened 2026-03-31
fix: propagate addResource() to model Build for Maven 3 compat

gnodet:
## Summary

- When a Maven 3 plugin dynamically adds a resource via `project.addResource()` or `project.getResources().add()`, propagate the change to the model's `Build.resources` so that `project.getBuild().getResources()` stays in sync with `project.getResources()`
- After project building, sync `Build.resources`/`Build.testResources` from the `sources` set so that resources defined via `<sources>` (Maven 4.1.0) are also visible through `project.getBuild().getResources()`
- This fixes resource ordering issues when running Maven 3 plugins (e.g. maven-source-plugin 3.x) on Maven 4

### Root cause

In Maven 3, `project.getResources()` and `project.getBuild().getResources()` return the **same** mutable list. In Maven 4's compatibility layer, they are disconnected: `getResources()` returns a virtual list backed by the internal `sources` set, while `getBuild().getResources()` returns a WrapperList backed by the model.

When `maven-remote-resources-plugin` adds a resource directory, only the `sources` set was updated. The `maven-source-plugin`'s `createArchiver()` method uses `project.getBuild().getResources()` to find the `maven-shared-archive-resources` directory and add META-INF files first in the JAR — but in Maven 4 it couldn't find it, causing the META-INF entries to appear at the end instead of the beginning.

Additionally, when resources are defined via `<sources>` (Maven 4.1.0 model), they were only added to the `sources` set but not reflected in `Build.resources`, making them invisible to Maven 3 plugins accessing resources via `project.getBuild().getResources()`.

See: https://github.com/apache/maven-source-plugin/pull/281

## Test plan

- [x] Existing `ResourceIncludeTest` passes (10 tests)
- [x] Full `maven-core` test suite passes (545 tests)
- [ ] Verify with maven-source-plugin 3.x reproducible IT on Maven 4

### DIFF
--- a/impl/maven-core/src/main/java/org/apache/maven/project/DefaultProjectBuilder.java
+++ b/impl/maven-core/src/main/java/org/apache/maven/project/DefaultProjectBuilder.java
@@ -778,11 +778,13 @@ public class DefaultProjectBuilder implements ProjectBuilder {
                     sourceContext.handleResourceConfiguration(ProjectScope.TEST);
                 }
 
-                // Sync model Build resources from sources set so that
-                // project.getBuild().getResources() is consistent with project.getResources().
-                // This is needed for Maven 3 plugin compatibility, since some plugins
-                // (e.g. maven-source-plugin) access resources via project.getBuild().getResources().
-                project.syncBuildResources();
+                // When resources are defined via <sources> (4.1.0 model), sync them to
+                // the model's Build so project.getBuild().getResources() is consistent.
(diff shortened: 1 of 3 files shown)


hboutemy:
@gnodet I tested the change on maven-source-plugin 3.x ITs, bot on this master branch and on 4.0.x one: I confirm this fixes the issue

anything else to check before merging?

rmannibucau (approved):

Issue #11498, opened 2025-11-27
Another race condition involving MessageUtils


oehme:
### Affected version

4.0.x

### Bug description

MessageUtils uses a `FastTerminal`, which calls `systemInstall` in a background thread. I already found one race condition where uninstall was called before this background thread finishes. https://github.com/apache/maven/issues/11048

~Unfortunately, there is another race in [LookupInvoker](https://github.com/apache/maven/blob/5834e3899eed5c6a088a937a3c8d171d72d869f3/impl/maven-cli/src/main/java/org/apache/maven/cling/invoker/LookupInvoker.java#L399). It also calls `System.setOut`  and depending on the timing, this can happen before the FastTerminal background thread finishes. When that happens, you get a StackOverflowError on the next log message.~

guylabs:
@cstamas Are you maybe able to have a look at this?

guylabs:
Thanks for the new JLine release. Once this pull request is merged, we can see if that solves our issue.

- https://github.com/apache/maven/pull/11757

Will the `maven-4.0.x` branch be merged back to `master` once the above is merged?

PR #11901, opened 2026-04-07
Fix logging setup/teardown order

oehme:
I occasionally saw stackoverflow exceptions which pointed at the logging and terminal setup being in an inconsistent state. I suspect that this was due to the shutdown being in the wrong order. The build log appender should be deregistered before the terminal is shut down. To do this, the closeables need to be registered in the opposite order than they were previously.

Relates to https://github.com/apache/maven/issues/11498

A backport to 4.0 would be appreciated, because that's when we started seeing the stackoverflows.

### DIFF
--- a/impl/maven-cli/src/main/java/org/apache/maven/cling/invoker/LookupInvoker.java
+++ b/impl/maven-cli/src/main/java/org/apache/maven/cling/invoker/LookupInvoker.java
@@ -311,11 +311,6 @@ public abstract class LookupInvoker<C extends LookupContext> implements Invoker 
 
     protected final void createTerminal(C context) {
         if (context.terminal == null) {
-            // Create the build log appender; also sets MavenSimpleLogger sink
-            ProjectBuildLogAppender projectBuildLogAppender =
-                    new ProjectBuildLogAppender(determineBuildEventListener(context));
-            context.closeables.add(projectBuildLogAppender);
-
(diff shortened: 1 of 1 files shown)


oehme:
Note that I can't reliably reproduce the stackoverflow with a simple example, they only happen when running hundreds of embedded Maven tests on a heavily loaded machine. But the above code seemed like it was clearly the wrong order, so I'd like to rule it out as a possible cause.

cstamas:
Looks reasonable. @gnodet  ?

oehme:
Ping :)

cstamas (approved):

gnodet (approved):
The fix is correct and well-targeted. The reorder properly ensures `ProjectBuildLogAppender` (which routes log messages through `context.terminal.writer()`) is deregistered before the terminal is destroyed during shutdown. `LookupContext.close()` reverses the closeables list (LIFO), so moving the appender registration after `MessageUtils::systemUninstall` means it gets closed first.

As a secondary benefit, moving the appender creation after `context.terminal` is set also eliminates a potential NPE: the default writer lambda at `LookupInvoker.java:422` accesses `context.terminal.writer()` lazily, but previously the appender was installed before `context.terminal` was assigned.

No behavioral regression — log messages during terminal setup won't go through the build event listener, but that's expected and harmless.

PR #11916, opened 2026-04-09
Feat: migrate core to JSR330

cstamas:
Changes:
* migrate off plexus DI and Plexus component XML (but NO logger should be `static`!)
* replace Plexus Logger with Slf4j Logger where it is trivial to do so
* plexus DI migration is 100%, only module on Plexus DI is maven-compat now
* Logger migration is not 100%, as there are places where Plexus Logger are still being passed on


cstamas:
```
Error:  Failures: 
Error:    MavenITmng4385LifecycleMappingFromExtensionInReactorTest.testit:55->AbstractMavenIntegrationTestCase.fail:387 Build leaked lifecycle mapping from extension of first project into second project!
[INFO] 
Error:  Tests run: 867, Failures: 1, Errors: 0, Skipped: 42
```

maven-core/src/main/java/org/apache/maven/lifecycle/internal/DefaultLifecyclePluginAnalyzer.java
```diff
@@ -79,7 +88,22 @@ public Set<Plugin> getPluginsBoundByDefaultToAllLifecycles(String packaging) {
...
+        Map<String, LifecycleMapping> filtered = new HashMap<>();
+        // filter by visibility (plexus vs sisu diff; "realms" are plexus thing)
+        // in some tests container is not injected
+        if (this.plexusContainer != null) {
+            for (String name : this.lifecycleMappings.keySet()) {
+                try {
+                    filtered.put(name, plexusContainer.lookup(LifecycleMapping.class, name));
+                } catch (ComponentLookupException e) {
+                    // skip it
+                }
```
slawekjaranowski, line 100: Can we use `hasComponent` instead of catch exception?

maven-core/src/main/java/org/apache/maven/artifact/handler/types/FatJarArtifactHandler.java
```diff
@@ -0,0 +1,48 @@
...
+
+import javax.inject.Named;
+import javax.inject.Provider;
+import javax.inject.Singleton;
+
+import org.apache.maven.artifact.handler.ArtifactHandler;
+import org.apache.maven.artifact.handler.ArtifactHandlerImpl;
+
+/**
+ * Fat-JAR Artifact Handler.
```
slawekjaranowski: It is something new ... If needed I would like to add as separate PR to show change

maven-core/src/main/java/org/apache/maven/artifact/handler/types/JavadocArtifactHandler.java
```diff
@@ -0,0 +1,49 @@
...
+@Singleton
+@Named(JavadocArtifactHandler.NAME)
+public final class JavadocArtifactHandler implements Provider<ArtifactHandler> {
+    public static final String NAME = "javadoc";
+
+    private final ArtifactHandler instance;
+
+    public JavadocArtifactHandler() {
+        this.instance =
+                new ArtifactHandlerImpl(NAME, "jar", "javadoc", null, false, ArtifactHandlerImpl.LANGUAGE_JAVA, true);
```
slawekjaranowski, line 42: Interesting that we add javadocs to classpath .... 
It is ok as we allready such in 3.x

maven-core/src/main/java/org/apache/maven/artifact/handler/types/FatJarArtifactHandler.java
cstamas: Yes, it creates handy new type, where the dependency "has it all" (or does not need transitives pulled in). Will pull this out.

maven-core/src/main/java/org/apache/maven/artifact/handler/types/JavadocArtifactHandler.java
cstamas: Yes, is surprising as well, and as you say, all Maven 3 did this.

maven-core/src/main/java/org/apache/maven/artifact/handler/types/FatJarArtifactHandler.java
cstamas: Dropped

maven-core/src/main/java/org/apache/maven/lifecycle/internal/DefaultLifecyclePluginAnalyzer.java
cstamas: Nope, we cannot; there is a slight semantic difference between (lookup w/ class) vs (lookup w/ roleString and for all hasComponent). Latter returns _plexus registered beans only_. Hence, as now the core packagings are migrated to JSR330, hasComponent would return `false`, and similarly, CLEx would be thrown if we'd do `lookup(LifecycleMapping.class.getName(), name)`.

maven-core/src/main/java/org/apache/maven/eventspy/internal/EventSpyDispatcher.java
```diff
@@ -18,32 +18,33 @@
...
-    private List<EventSpy> eventSpies;
+    private final List<EventSpy> eventSpies;
 
-    public void setEventSpies(List<EventSpy> eventSpies) {
-        // make copy to get rid of needless overhead for dynamic lookups
-        this.eventSpies = new ArrayList<>(eventSpies);
+    @Inject
+    public EventSpyDispatcher(List<EventSpy> eventSpies) {
+        this.eventSpies = eventSpies;
     }
```
slawekjaranowski, line 48: Why here we use constructor injection?

maven-core/src/main/java/org/apache/maven/lifecycle/internal/DefaultLifecyclePluginAnalyzer.java
```diff
@@ -79,7 +88,22 @@ public Set<Plugin> getPluginsBoundByDefaultToAllLifecycles(String packaging) {
                     + Thread.currentThread().getContextClassLoader());
         }
 
-        LifecycleMapping lifecycleMappingForPackaging = lifecycleMappings.get(packaging);
+        Map<String, LifecycleMapping> filtered = new HashMap<>();
+        // filter by visibility (plexus vs sisu diff; "realms" are plexus thing)
+        // in some tests container is not injected
+        if (this.plexusContainer != null) {
```
slawekjaranowski, line 94: Is it possible to have null value for `@Inject`

If it is problem in tests maybe we should fix test?

cstamas: Yes, UT only. At runtime guice will fail if nothing to inject (and injection point is not nullable)

maven-core/src/main/java/org/apache/maven/eventspy/internal/EventSpyDispatcher.java
cstamas: Coz it had setter and ctor and used field injection and it was confusing. This is really internal component, so here i did not care.

We can undo it somehow, if must.

maven-core/src/main/java/org/apache/maven/lifecycle/internal/DefaultLifecyclePluginAnalyzer.java
slawekjaranowski: ok, we can try fix in next round 😄

slawekjaranowski (approved):

PR #11921, opened 2026-04-10
Feat: Add Automatic Module Names to Maven JAR Manifests

cstamas:
Note: by doing this, Maven is not modular, as there are split packages still present.

This is more like a tinkerer PR. Idea is to create "throwaway" module names (`v3`) as with Maven 4 things will totally change.


pom.xml
```diff
@@ -527,6 +532,21 @@ under the License.
...
       <plugins>
+        <plugin>
+          <groupId>org.apache.maven.plugins</groupId>
+          <artifactId>maven-jar-plugin</artifactId>
+          <configuration>
+            <archive>
+              <manifest>
+                <addDefaultSpecificationEntries>true</addDefaultSpecificationEntries>
+                <addDefaultImplementationEntries>true</addDefaultImplementationEntries>
+              </manifest>
```
slawekjaranowski, line 543: We have it in parent
https://github.com/apache/maven-apache-parent/blob/b4d96ca4e6fec318c8781779658b01fd2852c8cc/pom.xml#L229-L232

slawekjaranowski (approved):

PR #11923, opened 2026-04-10
Fix profile property interpolation in repository URLs

utafrali:
Profile-scoped properties weren't getting interpolated in repository URLs. The validator was checking for unresolved expressions before profile properties were injected, which caused it to reject valid property references.

Added a skipExpressionValidation flag so profile repositories can skip the expression check since property injection happens later in the process. Regular repos and distribution management still validate strictly. Includes a test case and verified with mvn verify locally.


utafrali:
closing, lmk if you want me to reopen with changes

PR #11926, opened 2026-04-10
Warn when profile ID matches a lifecycle phase name

### LINKED ISSUE #10310: [MNG-8762] Print warning, when profile has name of a lifecycle phase
**[Matthias Bünger](https://issues.apache.org/jira/secure/ViewProfile.jspa?name=JIRAUSER302100)** opened **[MNG-8762](https://issues.apache.org/jira/browse/MNG-8762?redirect=false)** and commented

As written in Slack I had a situation where I listed multiple profiles not correctly, ending up that another lifecycle phase was triggered was one of the profiles was named "integration-test".

Hervè came up with the idea to print a warning, when a profile has the same name as a lifecycle phase

---
No further details from [MNG-8762](https://issues.apache.org/jira/browse/MNG-8762?redirect=false)

utafrali:
When a profile ID matches a default Maven lifecycle phase name (e.g. "test", "deploy", "integration-test"), users can accidentally trigger that phase. This adds a validation warning to alert about the potential confusion.

Includes a new test case with a POM containing both phase-matching and safe profile IDs.

Closes #10310

### DIFF
--- a/impl/maven-impl/src/main/java/org/apache/maven/impl/model/DefaultModelValidator.java
+++ b/impl/maven-impl/src/main/java/org/apache/maven/impl/model/DefaultModelValidator.java
@@ -99,6 +99,34 @@ public class DefaultModelValidator implements ModelValidator {
 
     private static final String EMPTY = "";
 
+    private static final Set<String> LIFECYCLE_PHASES = Set.of(
+            "clean",
+            "validate",
+            "initialize",
+            "generate-sources",
+            "process-sources",
+            "generate-resources",
+            "process-resources",
+            "compile",
+            "process-classes",
+            "generate-test-sources",
+            "process-test-sources",
+            "generate-test-resources",
+            "process-test-resources",
+            "test-compile",
+            "process-test-classes",
+            "test",
+            "prepare-package",
+            "package",
+            "pre-integration-test",
+            "integration-test",
+            "post-integration-test",
+            "verify",
+            "install",
+            "deploy",
+            "site",
+            "site-deploy");
+
     private record ActivationFrame(String location, Optional<? extends InputLocationTracker> parent) {}
 
     private static class ActivationWalker extends MavenTransformer {
@@ -563,6 +591,20 @@ public class DefaultModelValidator implements ModelValidator {
                 String prefix = "profiles.profile[" + profile.getId() + "].";
 
                 validateProfileId(prefix, "id", problems, Severity.ERROR, Version.V40, profile.getId(), null, model);
+
+                if (profile.getId() != null && LIFECYCLE_PHASES.contains(profile.getId())) {
+                    addViolation(
+                            problems,
+                            Severity.WARNING,
+                            Version.BASE,
+                            "profiles.profile.id",
+                            null,
+                            "Profile has the same id '"
+                                    + profile.getId() + "' as a Maven lifecycle phase. "
+                                    + "This can accidentally trigger the lifecycle phase "
+                                    + "when the profile name is misused on the command line.",
+                            profile);
+                }
 
                 if (!profileIds.add(profile.getId())) {
                     addViolation(
--- a/impl/maven-impl/src/test/java/org/apache/maven/impl/model/DefaultModelValidatorTest.java
+++ b/impl/maven-impl/src/test/java/org/apache/maven/impl/model/DefaultModelValidatorTest.java
@@ -482,6 +482,18 @@ class DefaultModelValidatorTest {
     }
 
     @Test
+    void testLifecyclePhaseProfileId() throws Exception {
+        SimpleProblemCollector result = validateFile("lifecycle-phase-profile-id.xml");
+
+        assertViolations(result, 0, 0, 2);
+
+        assertTrue(result.getWarnings().get(0).contains("integration-test"));
+        assertTrue(result.getWarnings().get(0).contains("lifecycle phase"));
+        assertTrue(result.getWarnings().get(1).contains("deploy"));
+        assertTrue(result.getWarnings().get(1).contains("lifecycle phase"));
+    }
+
+    @Test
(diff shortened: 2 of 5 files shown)


kwin (changes requested):
There was no consensus in #10310 that adding a warning when profile ids clash with phases is a good idea. Also this PR only considers a hard coded set of phases, but not ones potentially contributed by extensions.

impl/maven-impl/src/main/java/org/apache/maven/impl/model/DefaultModelValidator.java
```diff
@@ -99,6 +99,34 @@ public class DefaultModelValidator implements ModelValidator {
...
+            "test",
+            "prepare-package",
+            "package",
+            "pre-integration-test",
+            "integration-test",
+            "post-integration-test",
+            "verify",
+            "install",
+            "deploy",
+            "site",
```
gnodet, line 127: This hardcoded set duplicates phase names that are already defined as constants in `org.apache.maven.api.Lifecycle.Phase` (in `maven-api-core`, which is a dependency of this module). It also mixes Maven 3 legacy phase names (e.g. `generate-sources`, `process-sources`, `process-classes`, `prepare-package`, `pre-integration-test`, `post-integration-test`, `site-deploy`) with the standard ones, while missing Maven 4 phases like `all`, `each`, `build`, `sources`, `resources`, `ready`, and `unit-test`.

Please build this set from the existing `Lifecycle.Phase` constants instead of duplicating strings. For example:

```suggestion
    private static final Set<String> LIFECYCLE_PHASES = Set.of(
            Lifecycle.Phase.ALL,
            Lifecycle.Phase.EACH,
            Lifecycle.Phase.BUILD,
            Lifecycle.Phase.INITIALIZE,
            Lifecycle.Phase.VALIDATE,
            Lifecycle.Phase.SOURCES,
            Lifecycle.Phase.RESOURCES,
            Lifecycle.Phase.COMPILE,
            Lifecycle.Phase.READY,
            Lifecycle.Phase.PACKAGE,
            Lifecycle.Phase.VERIFY,
            Lifecycle.Phase.UNIT_TEST,
            Lifecycle.Phase.TEST_SOURCES,
            Lifecycle.Phase.TEST_RESOURCES,
            Lifecycle.Phase.TEST_COMPILE,
            Lifecycle.Phase.TEST,
            Lifecycle.Phase.INTEGRATION_TEST,
            Lifecycle.Phase.INSTALL,
            Lifecycle.Phase.DEPLOY,
            Lifecycle.Phase.CLEAN,
            Lifecycle.SITE);
```

You may also want to include the legacy Maven 3 aliases (via `Lifecycle.aliases()`) if the goal is to catch those as well, but at minimum the canonical names should come from the constants.

gnodet (commented):

The intent is good -- warning users when a profile ID collides with a lifecycle phase name is a useful safeguard.

However, the hardcoded `LIFECYCLE_PHASES` set has two problems:

1. **It will become stale.** Maven 4 defines different phase names in `org.apache.maven.api.Lifecycle.Phase` (e.g. `all`, `each`, `build`, `sources`, `resources`, `ready`, `unit-test`) which are missing from this list. Conversely, some of the names in the list (`generate-sources`, `process-sources`, `process-resources`, `process-classes`, `generate-test-sources`, `process-test-sources`, `generate-test-resources`, `process-test-resources`, `process-test-classes`, `prepare-package`, `pre-integration-test`, `post-integration-test`, `site-deploy`) are Maven 3 legacy aliases. Maintaining a manually-curated duplicate is error-prone.

2. **Constants already exist.** The `Lifecycle.Phase` interface in `maven-api-core` (which `maven-impl` already depends on) defines all the canonical phase name constants. The set should be derived from those constants rather than duplicated as raw strings.

The `LIFECYCLE_PHASES` set should be built from the `Lifecycle.Phase` constants and the `Lifecycle.CLEAN` / `Lifecycle.SITE` lifecycle names, so it stays in sync automatically. You could also consider querying the lifecycle registry at validation time, but a constant set from the API is a simpler first step.

gnodet (commented):
The intent — warning when a profile ID collides with a lifecycle phase name — addresses a genuine user confusion scenario (issue #10310). However, the implementation has several design concerns that should be addressed:

**1. Hardcoded phase list is incomplete (high)**
The `LIFECYCLE_PHASES` set contains 26 entries but is missing 9 Maven 4 phases (`all`, `each`, `build`, `sources`, `resources`, `ready`, `unit-test`, `test-sources`, `test-resources`) while including 13+ Maven 3 legacy alias names. The `Lifecycle.Phase` constants already exist as a compile-time dependency from `maven-api-core` (which `maven-impl` depends on) and could be used directly instead of duplicating raw strings that will drift out of sync.

**2. Extensible phases not covered (high)**
As @kwin noted in their review and @cstamas raised on issue #10310, lifecycle phases are extensible — extensions can contribute additional phases at runtime. The `Lifecycle` interface extends `ExtensibleEnum`, confirming this by design. A static compile-time set can never cover extension-contributed phases. The validator would need access to the lifecycle registry at validation time to be complete, or the warning's scope and limitations need to be explicitly documented.

**3. Test maintenance side-effect (medium)**
Renaming existing test profile IDs from `"test"` to `"test-profile"` in `duplicate-plugin-execution.xml` and `duplicate-plugin.xml` is a side-effect of the new validation changing expected violation counts. Any future phase name added to the set could silently break other tests whose POMs happen to use that name as a profile ID.

**4. Warning message clarity (low)**
The message says _"when the profile name is misused on the command line"_ which is vague. A clearer message would describe the specific confusion scenario — e.g., _"may be confused with the lifecycle phase if used without -P on the command line."_

Given that @kwin has explicitly requested changes and @cstamas expressed skepticism on the approach, the author may want to seek consensus on the design direction before further implementation effort.

PR #11887, opened 2026-04-03
Fix #11885: Disable ANSI colors when stdout is piped on JDK 22+

gnodet:
## Summary

- Change `ForcedSysOut` to `SysOut` in JLine's `TerminalBuilder` configuration so that JLine properly checks whether stdout is a TTY before creating a system terminal
- `ForcedSysOut` bypasses TTY detection, causing a non-dumb terminal to be created even when stdout is piped (because stdin is still a TTY), resulting in ANSI escape codes in piped output on JDK 22+ where the FFM provider creates a terminal from stdin alone

## Root Cause

JLine's `SystemOutput.ForcedSysOut` always returns `SystemStream.Output` in `select()` without checking if stdout is actually a TTY. On JDK 21, the exec/JNI providers would fail to create a system terminal when stdout was piped, falling back to dumb. On JDK 22+ with the FFM provider, the terminal is created successfully (stdin is a TTY), producing a non-dumb terminal with colors enabled even when piped.

`SystemOutput.SysOut` does the right thing: it checks if stdout is a TTY first. If not, `systemStream` is null, and JLine falls back to a dumb terminal — exactly the correct behavior.

## Test plan

- [ ] `mvn help:help -X | less` should show plain text (no ANSI codes) on JDK 25+
- [ ] `mvn help:help -X > /tmp/out.txt` should produce plain text on JDK 25+
- [ ] `mvn help:help -X` should still show colored output on a real terminal
- [ ] `mvn --color=always help:help -X | less` should still show ANSI codes (forced)
- [ ] `mvn --color=never help:help -X` should show plain text
- [ ] Behavior unchanged on JDK 17/21

Fixes #11885

_Claude Code on behalf of Guillaume Nodet_

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #11927, opened 2026-04-10
Fix ANSI colors not disabled when piped on JDK 25+

### LINKED ISSUE #11885: ANSI colors are not disabled in case of redirect to pipe or file on JDK25+
### Affected version

4.0.0-rc-5

### Bug description

# Test command:

```
mvn help:help -X | less
```

# Oputputs

## JDK 21

```
Apache Maven 4.0.0-rc-5 (fb3ecaef88106acb40467a450248dfdbd75f3b35)
Maven home: /Users/slawomir.jaranowski/.sdkman/candidates/maven/4.0.0-rc-5
Java version: 21.0.10, vendor: Eclipse Adoptium, runtime: /Users/slawomir.jaranowski/.sdkman/candidates/java/21.0.10-tem
Default locale: en_PL, platform encoding: UTF-8
OS name: "mac os x", version: "26.3.1", arch: "aarch64", family: "mac"


...
[DEBUG] Message scheme: plain
```

## JDK 25

```
ESC[1mApache Maven 4.0.0-rc-5 (fb3ecaef88106acb40467a450248dfdbd75f3b35)ESC[0m
Maven home: /Users/slawomir.jaranowski/.sdkman/candidates/maven/4.0.0-rc-5
Java version: 25.0.2, vendor: Eclipse Adoptium, runtime: /Users/slawomir.jaranowski/.sdkman/candidates/java/25.0.2-tem
Default locale: en_PL, platform encoding: UTF-8
OS name: "mac os x", version: "26.3.1", arch: "aarch64", family: "mac"

...

[ESC[36;1mDEBUGESC[0m] Message scheme: color

```

utafrali:
On JDK 22+, `System.console()` returns a non-null Console even when stdout is redirected to a pipe or file. This caused Maven to select the color message scheme when output was piped.

Uses `Console.isTerminal()` (JDK 22+) via reflection to correctly detect piped output. On JDK < 22, falls back to the previous `System.console() != null` check.

Applied to both Maven 4 (LookupInvoker) and Maven 3 compat (MavenCli) code paths.

Closes #11885

### DIFF
--- a/compat/maven-embedder/src/main/java/org/apache/maven/cli/MavenCli.java
+++ b/compat/maven-embedder/src/main/java/org/apache/maven/cli/MavenCli.java
@@ -532,6 +532,10 @@ public class MavenCli {
                             || commandLine.hasOption(CLIManager.NON_INTERACTIVE));
             if (isBatchMode || commandLine.hasOption(CLIManager.LOG_FILE)) {
                 MessageUtils.setColorEnabled(false);
+            } else if (!isConsoleTerminal()) {
+                // On JDK 22+, System.console() returns non-null even when stdout is
+                // redirected to a pipe or file. Use Console.isTerminal() to detect this.
+                MessageUtils.setColorEnabled(false);
             }
         }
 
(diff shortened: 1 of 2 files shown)


kwin:
Isn't #11887 the simpler fix?

gnodet:

@kwin is right — #11887 is the simpler fix and is already merged.

#11887 addresses the root cause: JLine's `TerminalBuilder` was configured with `ForcedSysOut`, which bypasses TTY detection on stdout. Changing it to `SysOut` lets JLine correctly detect piped output and fall back to a dumb terminal, disabling ANSI colors. It's a 3-line change with no reflection or JDK version branching.

This PR works around the symptom instead — using reflection to call `Console.isTerminal()` (JDK 22+) in two places, with duplicated helper methods. Since the root cause is already fixed, this can be closed.

The one additional thing this PR does is patch the Maven 3 compat path (`MavenCli`), but that code path has never done pipe detection (it only disables color for batch mode or log file), so adding a reflection-based workaround there doesn't seem worth the complexity.

PR #11948, opened 2026-04-14
Feat: Resolver 2.x update policy control via CLI

### LINKED ISSUE #11918: Resolver 2 supports split policies for artifacts and metadata

Right now `-U` sets both to same value (always). Maybe provide `-apa` and `-apm`?

As we saw with prefix resolution relying on metadata, it can be handy.

cstamas:
Resolver 2.x supports split policies for artifacts and metadata. Introduce 4 new CLI options, that if used, renders old "all" options ignored (-U and -nsu). New options are:
* `-UA` similar to existing `-U` but forces only artifact updates (resolves cached 404)
* `-UM` similar to existing `-U` but forces only metadata updates (updates ranges and snapshots)
* two long opts (`--artifacts-update-policy=<policy>` and `--metadata-update-policy=<policy>`) where user can explicitly fine tune override policies for artifacts and metadata.

If any of these 4 above are set, the old `-U` is ignored, and if any of `-UA` or `-UM` is set, the long counterpart is ignored.

Fixes #11918

---

Explanation: currently Maven had two update policy altering CLI options, `-U` and `-nsu`, and former set global session policy to "always" and latter to "never". Since Resolver 2.x update policies for artifacts and metadata are split and are handled as such. 

Why it matters:
* plugin prefix resolution happens **via metadata** -- "always" policy will refetch only metadata
* snapshot resolution happens **via metadata** -- "always" makes sure any new ones are discovered
* "remembered" 404 for missing artifact happens **via artifacts**

Before, when user issued `-U` he triggered all these above (forced refresh). Now, while `-U` still works, first two can be triggered with `--metadata-update-policy=always` (or suppress with never), and last one can be triggered with `--artifact-update-policy=always` without affecting the others (ie. discovery of new snapshots).


maven-embedder/src/main/java/org/apache/maven/cli/CLIManager.java
```diff
@@ -109,7 +109,11 @@ public class CLIManager {
 
     public static final String IGNORE_TRANSITIVE_REPOSITORIES = "itr";
 
-    private static final String RAW_STREAMS = "raw-streams";
+    public static final String RAW_STREAMS = "raw-streams";
+
+    public static final String ARTIFACTS_UPDATE_POLICY = "artifacts-update-policy";
+
+    public static final String METADATA_UPDATE_POLICY = "metadata-update-policy";
```
slawekjaranowski, line 116: What about short version of options?

cstamas: I wanted an option like `-U` is but more versatile:
* `-mup` would work like `-U` (set always for metadata)
* but optionally, user could set actual policy `-mup never` or `-mup always`

But the problem with optional support in Commons CLI, is that it does not work like? Or at least I could not make it work, ie CLI like this `mvn -mup clean` would assume "clean" is a policy and it would explode (unknown policy "clean").

So, I ended up with long options with mandatory param that is nice and explicit `--metadata-update-policy=never`, and on CLI this is also clear `mvn --metadata-update-policy=never clean` and it is perfectly clear what is written here.

cstamas: Solved; introduced `-UA` and `-UM` that is like "half of" `-U` :smile: And the other two long opts allow user to fine tune policies, ie. use `"never"` policy on thing he wants.

slawekjaranowski (approved):

PR #11967, opened 2026-04-20
Make def session builder have interface

cstamas:
To be a real component, "hidden" behind interface.


maven-core/src/main/java/org/apache/maven/internal/aether/DefaultRepositorySystemSessionFactory.java
```diff
@@ -214,8 +218,18 @@ public class DefaultRepositorySystemSessionFactory {
...
+    @Deprecated
+    public DefaultRepositorySystemSession newRepositorySession(MavenExecutionRequest request) {
+        return new DefaultRepositorySystemSession(
+                newRepositorySessionBuilder(request).build());
+    }
+
     @SuppressWarnings("checkstyle:methodlength")
-    public RepositorySystemSession.SessionBuilder newRepositorySession(MavenExecutionRequest request) {
+    @Override
+    public RepositorySystemSession.SessionBuilder newRepositorySessionBuilder(MavenExecutionRequest request) {
```
cstamas, line 232: Make new method `private`. nobody should within Maven create new session (I think)

slawekjaranowski (approved):

PR #11968, opened 2026-04-20
Promote java version in JavaToolchain

slawekjaranowski:
Many plugins using toolchains has requirements to detect java version selected by toolchains in runtime.

Expose java version in JavaToolchain can simplify plugins code.

similar to:
 - https://github.com/apache/maven/pull/11954

### DIFF
--- a/maven-core/src/main/java/org/apache/maven/toolchain/java/JavaToolchain.java
+++ b/maven-core/src/main/java/org/apache/maven/toolchain/java/JavaToolchain.java
@@ -18,6 +18,7 @@
  */
 package org.apache.maven.toolchain.java;
 
+import org.apache.maven.artifact.versioning.ArtifactVersion;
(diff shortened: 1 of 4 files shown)


maven-core/src/main/java/org/apache/maven/toolchain/java/JavaToolchain.java
```diff
@@ -18,6 +18,7 @@
  */
 package org.apache.maven.toolchain.java;
 
+import org.apache.maven.artifact.versioning.ArtifactVersion;
```
cstamas, line 21: Are you sure this class should be used? I'd not use it for anything new.

cstamas: Oh, I missed this is already merged on master, nvm

cstamas (approved):

PR #11969, opened 2026-04-20
Optimize log level rendering in MavenSimpleLogger

slawekjaranowski:
What changed:
- Cached rendered log level strings to improve performance.
- Removed redundant calls to the builder for each log level.
- Initialize rendered levels only once during the first call.

Why:
- Enhances efficiency by reducing object creation during logging.
- Improves overall logging performance in Maven.

### DIFF
--- a/maven-slf4j-provider/src/main/java/org/slf4j/simple/MavenSimpleLogger.java
+++ b/maven-slf4j-provider/src/main/java/org/slf4j/simple/MavenSimpleLogger.java
@@ -29,24 +29,38 @@ import static org.apache.maven.jline.MessageUtils.builder;
  */
 public class MavenSimpleLogger extends SimpleLogger {
 
+    private String traceRenderedLevel;
+    private String debugRenderedLevel;
+    private String infoRenderedLevel;
+    private String warnRenderedLevel;
+    private String errorRenderedLevel;
+
     MavenSimpleLogger(String name) {
         super(name);
     }
 
     @Override
     protected String renderLevel(int level) {
+        if (traceRenderedLevel == null) {
+            traceRenderedLevel = builder().trace("TRACE").build();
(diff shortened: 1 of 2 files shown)


cstamas (approved):

maven-slf4j-provider/src/main/java/org/slf4j/simple/MavenSimpleLogger.java
```diff
@@ -29,24 +29,38 @@
...
+    private String warnRenderedLevel;
+    private String errorRenderedLevel;
+
     MavenSimpleLogger(String name) {
         super(name);
     }
 
     @Override
     protected String renderLevel(int level) {
+        if (traceRenderedLevel == null) {
```
cstamas: What happens if multiple threads simultaneously log?

slawekjaranowski: we will create it twice .... next time last object will be used, the same is in master
I think synchronization is not needed here

cstamas (approved):

cstamas (approved):

maven-slf4j-provider/src/main/java/org/slf4j/simple/MavenSimpleLogger.java
olamy: maybe could be simply done in the constructor?

olamy (approved):

maven-slf4j-provider/src/main/java/org/slf4j/simple/MavenSimpleLogger.java
slawekjaranowski: can not be done in constructor ...

PR #11970, opened 2026-04-20
Feat: Ability to disable site lifecycle

### LINKED ISSUE #11759: Add ability to disable maven-site-plugin

Right now even if plugin is not bound to any phases, it is resolved which one can check by running mvn dependency:resolve-plugins. This plugin is very heavy and has a lot of dependencies, it would be nice to have ability to remove it from plugin resolution so it is not installed each time on clean build.

cstamas:
Ability to suppress site lifecycle from Maven "default lifecycles". This PR **does not** remove "site" lifecycle component, but Maven will not consider site lifecycle among "defaults", if disabled.

Note: hence, with this option enabled, the IT suite will never (and cannot) pass, as IT suite contains ITs running/expecting site lifecycle.

Fixes #11759

### DIFF
--- a/maven-core/src/main/java/org/apache/maven/lifecycle/DefaultLifecycles.java
+++ b/maven-core/src/main/java/org/apache/maven/lifecycle/DefaultLifecycles.java
@@ -52,8 +52,7 @@ import org.codehaus.plexus.util.StringUtils;
 public class DefaultLifecycles {
     private static final List<String> BUILT_IN_LIFECYCLES =
             Arrays.asList(CleanLifecycleProvider.NAME, DefaultLifecycleProvider.NAME, SiteLifecycleProvider.NAME);
-    private static final boolean SITE_ENABLED =
-            Boolean.parseBoolean(System.getProperty("maven.site.lifecycle.enabled", "true"));
+    private static final boolean SITE_DISABLED = Boolean.getBoolean("maven.site.lifecycle.disabled");
 
     /**
      * @deprecated Use {@link #getStandardLifecycles()} instead.
@@ -81,7 +80,7 @@ public class DefaultLifecycles {
 
     private boolean isEnabled(String lifecycleId) {
         if (SiteLifecycleProvider.NAME.equals(lifecycleId)) {
-            return SITE_ENABLED;
+            return !SITE_DISABLED;
         } else {
             return true;
         }
--- a/maven-core/src/main/java/org/apache/maven/lifecycle/DefaultLifecycles.java
+++ b/maven-core/src/main/java/org/apache/maven/lifecycle/DefaultLifecycles.java
@@ -30,7 +30,11 @@ import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
+import java.util.stream.Collectors;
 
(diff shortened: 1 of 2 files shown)


maven-core/src/main/java/org/apache/maven/lifecycle/DefaultLifecycles.java
```diff
@@ -46,7 +50,18 @@
 @Singleton
 @Named
 public class DefaultLifecycles {
-    public static final String[] STANDARD_LIFECYCLES = {"clean", "default", "site"};
+    private static final List<String> BUILT_IN_LIFECYCLES =
+            Arrays.asList(CleanLifecycleProvider.NAME, DefaultLifecycleProvider.NAME, SiteLifecycleProvider.NAME);
+    private static final boolean SITE_ENABLED =
+            Boolean.parseBoolean(System.getProperty("maven.site.lifecycle.enabled", "true"));
```
slawekjaranowski: can we also use a user properties for it?
or repos session config props?

cstamas: That is too late. This component is constructed as part of object graph before any session is being created... Will look more around tomorrow

cstamas: Looked around, and in many places session is not present/passed over. While we could alter these classes, that would lead to many potential breakages.

I consider this use case "special" or "very edge", so IMO Java System property is good to cover it.

michael-o: Why not `Boolean.getBoolean()`?

cstamas: How to default it to `true`?

michael-o: Indeed, you can't, but maybe we should invert the logic anyway? Site lifecycle is always enabled unless you explicitly disable this, e.g., `-DskipTests=true` and other skippers. `-Dmaven.site.lifecycle.skip`?

cstamas: Well, I could agree. But this is not skipping it, it is like "pretending is not there" (filter it out, remove it from default ones).

cstamas: Maybe just `maven.site.lifecycle.disabled` ?

michael-o: > Well, I could agree. But this is not skipping it, it is like "pretending is not there" (filter it out, remove it from default ones).

Yes, skipping is wrong. I noticed that after reflection.

michael-o: > Maybe just `maven.site.lifecycle.disabled` ?

maven.lifecycle.site.disabled if we talk about a properties tree.

cstamas: Unsure what tree you mean, but I am more to group properties by feature.

PR #11983, opened 2026-04-24
Fix mvn script expanding ${...} in CLI arguments

### LINKED ISSUE #11978: mvn script eval expands ${...} in CLI arguments
## Summary

The Maven 4 `mvn` shell script uses `eval exec "$cmd"` to launch the JVM.
This causes the shell to expand `${...}` patterns in user-provided CLI arguments,
breaking any argument that contains Maven property placeholders like
`${surefire.threadNumber}` or `${project.basedir}`.

Maven 3's `mvn` script uses `exec ... "$@"` which passes arguments verbatim.

## Reproducer

```sh
mvn validate -DtestProp='value_${some.placeholder}'
# Maven 3: works fine, passes the literal string to Java
# Maven 4: /path/to/mvn: line 302: bad substitution
```

## Root cause

In `apache-maven/src/assembly/maven/bin/mvn`
([lines 278-302](https://github.com/apache/maven/blob/a5558f892d/apache-maven/src/assembly/maven/bin/mvn#L278-L302)):

```sh
for arg in "$@"; do
    cmd="$cmd \"$arg\""
done
eval exec "$cmd"
```

User arguments are concatenated into the `cmd` string.
When `eval` executes it, the shell re-parses the entire string and expands
`${...}` as shell variables.

Maven 3's approach (`exec ... "$@"`) passes arguments directly without
re-interpretation.

## Impact

This affects any Maven plugin that passes `${...}` placeholders via `-D`
system properties on the command line. Known affected tests in
maven-surefire:

- [`ForkCountIT.java` line 161](https://github.com/apache/maven-surefire/blob/98cfb3d78b4485639f12af89b88aa5f9d8aef615/surefire-its/src/test/java/org/apache/maven/surefire/its/ForkCountIT.java#L161) — passes `-DtestProperty=testValue_${surefire.threadNumber}_${surefire.forkNumber}`
- [`ForkCountMultiModuleIT.java` line 126](https://github.com/apache/maven-surefire/blob/98cfb3d78b4485639f12af89b88aa5f9d8aef615/surefire-its/src/test/java/org/apache/maven/surefire/its/ForkCountMultiModuleIT.java#L126) — same pattern
- `ForkCountTestNGIT` — extends `ForkCountIT`

20 maven-surefire integration tests fail with `bad substitution` when run with
Maven 4. See also apache/maven-surefire#3345.

## Proposed fix

Keep `eval` for the base command (needed for `$MAVEN_OPTS` word splitting)
but pass user arguments directly via `"$@"`:

```sh
# before:
for arg in "$@"; do
    cmd="$cmd \"$arg\""
done
eval exec "$cmd"

# after:
eval exec "$cmd" '"$@"'
```

Verified locally: all 20 `bad substitution` errors eliminated.
`MAVEN_OPTS` word splitting and arguments with spaces continue to work correctly.

ascheman:
## Summary

The `eval` in the `mvn` script causes shell expansion of `${...}` patterns
in user-provided CLI arguments, breaking any argument that contains Maven
property placeholders like `${surefire.threadNumber}` or `${project.basedir}`.

## Problem

The current script concatenates user arguments into a command string and
then uses `eval exec` which re-parses the string and triggers shell variable
expansion:

```sh
for arg in "$@"; do
    cmd="$cmd \"$arg\""
done
eval exec "$cmd"
```

Maven 3's `mvn` script uses `exec ... "$@"` which passes arguments verbatim.

## Fix

Pass user arguments directly via `"$@"` instead of concatenating them into
the eval string. Only the base command (containing `$MAVEN_OPTS` etc.) uses
`eval` for word splitting:

```sh
eval exec "$cmd" '"$@"'
```

## Verification

Tested locally with Maven 4.0.0-rc-5:
- `${...}` in `-D` arguments: no longer causes `bad substitution`
- `MAVEN_OPTS` with spaces: still works (word splitting via `eval`)
- Arguments with spaces: still works (`"$@"` preserves quoting)
- 20 maven-surefire integration tests that previously failed with
  `bad substitution` now run successfully

Fixes #11978
Related: apache/maven-surefire#3345

Note: The same fix applies to the `maven-4.0.x` branch where the `mvn`
script is identical.

### DIFF
--- a/apache-maven/src/assembly/maven/bin/mvn
+++ b/apache-maven/src/assembly/maven/bin/mvn
@@ -291,7 +291,7 @@ cmd="\"$JAVACMD\" \
 
 if [ -n "$MAVEN_DEBUG_SCRIPT" ]; then
   echo "[DEBUG] Launching JVM with command:" >&2
-  echo "[DEBUG]   $cmd" "$@" >&2
+  printf '[DEBUG]   %s' "$cmd" >&2; printf ' "%s"' "$@" >&2; echo >&2
 fi
 
 # User arguments ("$@") are passed directly to preserve literal values
--- a/apache-maven/src/assembly/maven/bin/mvn
+++ b/apache-maven/src/assembly/maven/bin/mvn
@@ -275,7 +275,7 @@ handle_args() {
 handle_args "$@"
 MAVEN_MAIN_CLASS=${MAVEN_MAIN_CLASS:=org.apache.maven.cling.MavenCling}
 
-# Build command string for eval
+# Build base command string for eval (only contains Maven-controlled values)
 cmd="\"$JAVACMD\" \
   $MAVEN_OPTS \
   $MAVEN_DEBUG_OPTS \
@@ -289,14 +289,12 @@ cmd="\"$JAVACMD\" \
   $LAUNCHER_CLASS \
   $MAVEN_ARGS"
 
-# Add remaining arguments with proper quoting
-for arg in "$@"; do
-    cmd="$cmd \"$arg\""
-done
-
 if [ -n "$MAVEN_DEBUG_SCRIPT" ]; then
   echo "[DEBUG] Launching JVM with command:" >&2
-  echo "[DEBUG]   $cmd" >&2
+  echo "[DEBUG]   $cmd" "$@" >&2
 fi
 
-eval exec "$cmd"
+# User arguments ("$@") are passed directly to preserve literal values
+# like ${...} Maven property placeholders without shell expansion.
+# Only the base command uses eval for MAVEN_OPTS word splitting.
+eval exec "$cmd" '"$@"'
--- a/its/core-it-suite/src/test/java/org/apache/maven/it/MavenITgh11978PlaceholderInCliArgTest.java
+++ b/its/core-it-suite/src/test/java/org/apache/maven/it/MavenITgh11978PlaceholderInCliArgTest.java
@@ -1,0 +1,107 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.maven.it;
+
+import java.nio.file.Path;
+import java.util.Properties;
+
+import org.junit.jupiter.api.Test;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+/**
+ * This is a test set for <a href="https://github.com/apache/maven/issues/11978">gh-11978</a>.
+ *
+ * Verifies that the launcher script does not expand <code>${...}</code> patterns
+ * in CLI arguments. Regression test for the {@code eval exec} shell expansion
+ * that broke any argument containing Maven property placeholders.
+ *
+ * <h2>Real-world scenario</h2>
+ *
+ * <p>This bug was originally surfaced by maven-surefire-plugin integration tests.
+ * Surefire allows users to declare system properties for the forked test JVM with
+ * placeholders that surefire substitutes <em>at fork time</em>, e.g.
+ * {@code -DtestProperty=testValue_${surefire.threadNumber}_${surefire.forkNumber}}.
+ * The intended flow is:
+ * <ol>
+ *   <li>The user invokes {@code mvn} with the literal placeholder on the CLI.</li>
+ *   <li>The launcher script must pass the literal {@code ${...}} verbatim to
+ *       the JVM as a system property value.</li>
+ *   <li>The surefire plugin reads the system property, performs <em>its own</em>
+ *       interpolation when forking each test JVM, replacing
+ *       {@code ${surefire.threadNumber}} and {@code ${surefire.forkNumber}}
+ *       with the actual fork/thread numbers.</li>
+ *   <li>Each forked test JVM thus sees a unique substituted value.</li>
+ * </ol>
+ *
+ * <p>The bug broke step 2: the launcher script's {@code eval exec} re-parsed
+ * the command string and invoked shell variable expansion on {@code ${...}}.
+ * Names containing dots (such as {@code surefire.threadNumber}) are invalid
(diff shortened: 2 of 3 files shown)


hboutemy:
I see that the code was introduced in #2213
are you clear that the new approach does not break the intent in #2213 (I took time to search when the code was done, I have absolute no clue: that's why I'm asking)

ascheman:
> I see that the code was introduced in #2213 are you clear that the new approach does not break the intent in #2213 (I took time to search when the code was done, I have absolute no clue: that's why I'm asking)

I was sure that my change didn't break the fix from #2213 wrt. the still running integration tests for [MNG-4559](https://issues.apache.org/jira/browse/MNG-4559). 

But I took the chance to introduce an IT for my own change. At least it proves that there is no shell replacement of injected '${...}' any longer. Testing correct replacement in a nested `mvn` call as surefire does it is way harder to prove with the given IT test capabilities of Maven core. However, I tested that with surefire on apache/maven-surefire#3345.

gnodet (approved):

Solid fix for a real regression. The core change — `eval exec "$cmd" '"$@"'` — is the correct shell idiom: `eval` still word-splits `$MAVEN_OPTS` as needed, while `"$@"` passes user arguments verbatim without shell re-interpretation. The Windows `mvn.cmd` already uses `%*` directly so is unaffected.

The IT correctly uses `setForkJvm(true)` to exercise the actual launcher script, and the `verifyErrorFreeLog()` assertion is the key signal that catches `bad substitution` failures.

Minor suggestions (non-blocking):

1. **Debug output**: `echo "[DEBUG]   $cmd" "$@"` will print user args but without the quoted form that the old `echo "[DEBUG]   $cmd"` showed (since the args were embedded in `$cmd`). Consider using `printf` or printing `"$@"` in a loop for better debug visibility, though this is cosmetic.

2. **Javadoc verbosity**: The 60+ line Javadoc on the IT is quite heavy compared to the codebase style (see `MavenITgh10937QuotedPipesInMavenOptsTest` for comparison). The issue link + a few sentences would suffice — the detailed surefire context is better in the issue itself.

3. **Test assertion**: The `assertEquals("-value__end-", ...)` depends on Maven resolving undefined `${some.maven.placeholder}` to empty. The real proof is `verifyErrorFreeLog()` which catches the shell crash. Worth a brief inline comment noting this priority.

LGTM overall — the fix is correct, minimal, and well-motivated.

apache-maven/src/assembly/maven/bin/mvn
```diff
@@ -289,14 +289,12 @@ cmd="\"$JAVACMD\" \
...
 
-# Add remaining arguments with proper quoting
-for arg in "$@"; do
-    cmd="$cmd \"$arg\""
-done
-
 if [ -n "$MAVEN_DEBUG_SCRIPT" ]; then
   echo "[DEBUG] Launching JVM with command:" >&2
-  echo "[DEBUG]   $cmd" >&2
+  echo "[DEBUG]   $cmd" "$@" >&2
```
gnodet: Since user args are now separate from `$cmd`, this `echo` concatenates them losing boundary information — args containing spaces become ambiguous in debug output. `printf` can show each arg quoted:

```suggestion
  printf '[DEBUG]   %s' "$cmd" >&2; printf ' "%s"' "$@" >&2; echo >&2
```

its/core-it-suite/src/test/java/org/apache/maven/it/MavenITgh11978PlaceholderInCliArgTest.java
```diff
@@ -0,0 +1,107 @@
...
+ *   <li>The resulting value is {@code -value__end-}.</li>
+ * </ul>
+ *
+ * <p>Note: this test cannot easily replicate surefire's late-binding interpolation
+ * (step 3 above) because Maven's standard property interpolation does not
+ * recursively resolve POM properties for placeholders embedded inside a system
+ * property value. Surefire performs that substitution in its own plugin code at
+ * fork time. The signal that proves the fix works is the absence of a
+ * {@code bad substitution} shell error, not the resolved value itself.
+ */
```
gnodet, line 35: The detailed surefire scenario and test-design sections bring this Javadoc to ~56 lines, which is quite heavy compared to other ITs (e.g., `MavenITgh10937QuotedPipesInMavenOptsTest`). The issue link + a short summary should suffice — the surefire context lives better in the issue itself.

```suggestion
/**
 * This is a test set for <a href="https://github.com/apache/maven/issues/11978">gh-11978</a>.
 *
 * <p>Verifies that the launcher script does not expand {@code ${...}} patterns
 * in CLI arguments. The placeholder name intentionally contains dots (invalid as a
 * shell variable name) so without the fix, {@code eval exec} aborts with
 * {@code bad substitution} before Maven even starts.
 */
```

gnodet (commented):

Inline suggestions for the minor (non-blocking) items mentioned in the review.

its/core-it-suite/src/test/java/org/apache/maven/it/MavenITgh11978PlaceholderInCliArgTest.java
```diff
@@ -0,0 +1,107 @@
...
+        // With the fix, the literal ${...} arrives at Maven.
+        verifier.addCliArgument("-Dtest.placeholder=value_${some.maven.placeholder}_end");
+        verifier.addCliArgument("validate");
+        verifier.execute();
+        verifier.verifyErrorFreeLog(); // key check: fails immediately on shell crash
+
+        // Sanity check: the value flowed through Maven, which resolves the unknown
+        // ${some.maven.placeholder} to empty during recursive interpolation.
+        Properties props = verifier.loadProperties("target/pom.properties");
+        assertEquals("-value__end-", props.getProperty("project.properties.pom.placeholder"));
```
gnodet: Minor: the existing comments already capture the priority. A small wording tweak makes "primary vs. secondary" parallel:

```suggestion
        verifier.verifyErrorFreeLog(); // primary check: shell crash produces non-zero exit

        // Secondary: verify the literal placeholder flowed through Maven.
        // Maven resolves the unknown ${some.maven.placeholder} to empty.
        Properties props = verifier.loadProperties("target/pom.properties");
        assertEquals("-value__end-", props.getProperty("project.properties.pom.placeholder"));
```

apache-maven/src/assembly/maven/bin/mvn
ascheman: Resolved in 0c33a93d7c — thanks for the review!

its/core-it-suite/src/test/java/org/apache/maven/it/MavenITgh11978PlaceholderInCliArgTest.java
ascheman: Resolved in 0c33a93d7c — thanks for the review!

its/core-it-suite/src/test/java/org/apache/maven/it/MavenITgh11978PlaceholderInCliArgTest.java
ascheman: Resolved in 0c33a93d7c — thanks for the review!

ascheman:
> _Claude Code on behalf of Guillaume Nodet_
> 
> Solid fix for a real regression. 
...
> LGTM overall — the fix is correct, minimal, and well-motivated.

Claude instances talking to each other 😀

PR #11984, opened 2026-04-24
Fix #11899: Default addLocationInformation to false in Settings and Toolchains XML writers

### LINKED ISSUE #11899: `DefaultSettingsXmlFactory.write()` Always Emits Location Comments

abhu85:
## Summary

- Default `addLocationInformation` to `false` in `DefaultSettingsXmlFactory.write()` and `DefaultToolchainsXmlFactory.write()`
- Respect `inputLocationFormatter` from `XmlWriterRequest` — only emit location comments when explicitly requested
- Align behavior with `DefaultModelXmlFactory.write()`, which already handles this correctly

## Problem

`DefaultSettingsXmlFactory.write()` always emits `InputLocation` comments (e.g. `<!-- /path/to/settings.xml:163 -->`) because it creates a bare `new SettingsStaxWriter()` whose `addLocationInformation` field defaults to `true`. It also ignores the `inputLocationFormatter` from the `XmlWriterRequest`.

This is inconsistent with `DefaultModelXmlFactory.write()`, which correctly defaults location tracking to `false` and only enables it when an explicit `inputLocationFormatter` is provided.

The same issue exists in `DefaultToolchainsXmlFactory.write()`.

## Solution

Applied the same pattern from `DefaultModelXmlFactory.write()` (lines 165-174) to both:
1. Create the StaxWriter instance
2. Set `addLocationInformation(false)` by default
3. Check `request.getInputLocationFormatter()` — if provided, enable location information and set the string formatter

## Test Plan

- [x] `mvn -pl impl/maven-impl -am test` — 461 tests pass, 0 failures
- [x] `mvn spotless:apply` — formatting applied
- [x] `git diff --check` — no whitespace issues

## Compatibility

- No breaking changes — callers that don't provide `inputLocationFormatter` now get cleaner XML
- Callers that explicitly provide `inputLocationFormatter` will continue to get location comments

Fixes #11899

### DIFF
--- a/impl/maven-impl/src/main/java/org/apache/maven/impl/DefaultSettingsXmlFactory.java
+++ b/impl/maven-impl/src/main/java/org/apache/maven/impl/DefaultSettingsXmlFactory.java
@@ -22,6 +22,7 @@ import java.io.InputStream;
 import java.io.OutputStream;
 import java.io.Reader;
 import java.io.Writer;
+import java.util.function.Function;
 
 import org.apache.maven.api.annotations.Nonnull;
 import org.apache.maven.api.di.Named;
@@ -31,6 +32,7 @@ import org.apache.maven.api.services.xml.XmlReaderException;
 import org.apache.maven.api.services.xml.XmlReaderRequest;
 import org.apache.maven.api.services.xml.XmlWriterException;
 import org.apache.maven.api.services.xml.XmlWriterRequest;
+import org.apache.maven.api.settings.InputLocation;
 import org.apache.maven.api.settings.InputSource;
 import org.apache.maven.api.settings.Settings;
 import org.apache.maven.settings.v4.SettingsStaxReader;
@@ -79,10 +81,20 @@ public class DefaultSettingsXmlFactory implements SettingsXmlFactory {
             throw new IllegalArgumentException("writer or outputStream must be non null");
         }
         try {
+            SettingsStaxWriter xmlWriter = new SettingsStaxWriter();
+            xmlWriter.setAddLocationInformation(false);
+
+            Function<Object, String> formatter = request.getInputLocationFormatter();
+            if (formatter != null) {
+                xmlWriter.setAddLocationInformation(true);
+                Function<InputLocation, String> adapter = formatter::apply;
+                xmlWriter.setStringFormatter(adapter);
+            }
+
             if (writer != null) {
-                new SettingsStaxWriter().write(writer, content);
+                xmlWriter.write(writer, content);
             } else {
-                new SettingsStaxWriter().write(outputStream, content);
+                xmlWriter.write(outputStream, content);
             }
         } catch (Exception e) {
             throw new XmlWriterException("Unable to write settings: " + getMessage(e), getLocation(e), e);
--- a/impl/maven-impl/src/main/java/org/apache/maven/impl/DefaultToolchainsXmlFactory.java
+++ b/impl/maven-impl/src/main/java/org/apache/maven/impl/DefaultToolchainsXmlFactory.java
@@ -23,6 +23,7 @@ import java.io.OutputStream;
 import java.io.Reader;
 import java.io.Writer;
 import java.util.Objects;
+import java.util.function.Function;
 
 import org.apache.maven.api.annotations.Nonnull;
 import org.apache.maven.api.di.Named;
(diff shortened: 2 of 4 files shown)


gnodet (approved):
LGTM — the fix correctly aligns `DefaultSettingsXmlFactory` and `DefaultToolchainsXmlFactory` with the existing pattern in `DefaultModelXmlFactory` (lines 166-174): default `addLocationInformation` to `false`, only enable when an explicit `inputLocationFormatter` is provided.

I pushed one commit adding:
- `DefaultToolchainsXmlFactoryTest` — mirrors the Settings test for the Toolchains writer (both paths: no formatter → no comments, with formatter → comments present)
- Positive content assertions (`assertTrue(result.contains("<localRepository>"))` / `<type>jdk</type>`) in both test classes to guard against false positives from empty output

All 4 tests pass.

PR #12004, opened 2026-04-30
Feat: Pull out maven-executor into its own project

cstamas:
The main goal of `maven-executor` is ability to execute Maven 3/4 in "embedded" and "forked" mode, and has not related with Maven release lifecycle itself (and is dependency-less as well). So, let's move it out into own project.

https://github.com/apache/maven-executor

The new library executes Maven in:
* forked or embedded mode, but introduces other providers as well
* supports Maven 3 and 4
* the main `maven-executor` module is Java 8, dependency-less library (is MR-JAR, and on 9+ is modular and provides ToolProvider SPI integration as well)


.mvn/maven.config
```diff
@@ -1 +1,4 @@
--DsessionRootDirectory=${session.rootDirectory}
+-D
+sessionRootDirectory=${session.rootDirectory}
+-D
+apache.snapshots
```
olamy, line 4: related to this change?

cstamas: Yes, as maven-executor 1.0.0-SNAPSHOT is not a release

olamy: Ah, ok, sorry for the noise. I was more used to seeing a profile for this.

olamy (approved):

gnodet (approved):

rmannibucau (approved):

slachiewicz (approved):

hboutemy:
question: what about `maven-verifier` and `maven-invoker` shared components vs `maven-executor` https://maven.apache.org/shared/index.html ?
if we have a third shared component, we should a t least make visible about the 3 choices, isn't it?

cstamas:
Re site, I did not do anything, and I agree, we need visibility.

Re projects, there were some issues:
https://github.com/apache/maven-verifier/issues/186
https://github.com/apache/maven-invoker/issues/164

hboutemy:
oh, if maven-executor is expected to be the replacement for maven-invoker and maven-verifier, let's make it a shared component and document it from the start, please

and I suppose there are also related issues or PRs in related plugins https://maven.apache.org/plugins/ like maven-invoker-plugin and/or maven-verifier-plugin?

cstamas:
Yes, Maven project was inconsistent (or maybe "organic") in this respect, had wildly different projects meant to perform pretty much same thing, but there were huge differences in capabilities (fork-only vs fork-embedded) and APIs (CLI-like vs fully mimic Maven CLI).

The idea of `maven-executor` is to remain simple, does not stand in a way (ie. no need to release if new Maven CLI option is added) and offer fork-embed (and even more, see docker based ones) running options, without any need for "client code" to change, is just matter of configuration.

slawekjaranowski:
😄 my old mail - https://lists.apache.org/thread/20m5llon639s2vvps4o8qm9crpb2hg8n

cstamas:
maven-executor is now Java 8 baseline with MR-JAR version (9) for module-info and ToolProvider SPI support

PR #12009, opened 2026-05-01
Deps: Upgrade to Resolver 2.0.17

cstamas:
It requires some new internals in object graph, so the suppliers needs change. Also, legacy goes to legacy.

### DIFF
--- a/.mvn/maven.config
+++ b/.mvn/maven.config
@@ -1,1 +1,2 @@
--DsessionRootDirectory=${session.rootDirectory}
+-D
+sessionRootDirectory=${session.rootDirectory}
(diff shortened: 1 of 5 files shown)


cstamas:
This is again JBang stuff: https://github.com/jbangdev/jbang/issues/2325

All failed jobs failed due this issue above.

cstamas:
As chatting with JBang author, this seems plugin problem, as it seems it downloaded a (corrupted) ZIP file and now sticks to it (does not redownload, but re-unpack the same corrupted ZIP).

cstamas:
Superseded by #12112

PR #12024, opened 2026-05-10
[3.9.x] Bump to parent POM 48

cstamas:


slachiewicz (approved):

slawekjaranowski:
We should also remove: `bannerLeft` and `bannerRight` from 
https://github.com/apache/maven/blob/maven-3.9.x/src/site/site.xml

To check if we need  `<edit>`

It will be inherited from parent

cstamas:
So, is `<edit>` needed or not?

slawekjaranowski:
> So, is `<edit>` needed or not?

ok, edit is also not needed is the same in parent:
https://github.com/apache/maven-parent/blob/master/src/site/site.xml#L39

PR #12025, opened 2026-05-10
[3.10.x] Bump to parent POM 48

cstamas:


slawekjaranowski:
to remove `bannerLeft` from 
https://github.com/apache/maven/blob/maven-3.10.x/src/site/site.xml

Issue #11882, opened 2026-04-02
Maven 4 profile activation by <condition> issues


RalfNaujokat:
### Affected version

rc5

### Bug description

I found some issues in the `<condition>` syntax.
Maybe there is room for improvements.

Please have a look at the comments in the example pom.xml

```
<project xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xmlns="http://maven.apache.org/POM/4.1.0"
         xsi:schemaLocation="http://maven.apache.org/POM/4.1.0 http://maven.apache.org/xsd/maven-4.1.0.xsd">
	<modelVersion>4.1.0</modelVersion>


	<groupId>de.ralfn.tmp</groupId>
	<artifactId>maven4-profile-activation-by-condition-tester</artifactId>
	<version>1.0-SNAPSHOT</version>
	<packaging>pom</packaging>


	<properties>
		<root>true</root>
	</properties>


	<profiles>

		<!-- working -->

		<profile>
			<id>profile-1</id>
			<activation>
				<condition>exists( '.profile-1' )</condition>
			</activation>
		</profile>

		<profile>
			<id>profile-1b</id>
			<activation>
				<condition>
					exists( '.profile-1' )
				</condition>
			</activation>
		</profile>

		<!-- working, but kind of ugly -->

		<profile>
			<id>profile-2</id>
			<activation>
				<condition>exists( '.profile-2' ) &amp;&amp; missing( '.profile-1' )</condition>
				<!--
				Why not using something like this? (may be just as an optional version instead of the '&&' operator)
				<condition>exists( '.profile-2' ) AND missing( '.profile-1' )</condition>

				This could be nice, too., but maybe is somewhat harder to implement. We have a not(..) function anyway.
				<condition> and( exists( '.profile-2' ), missing( '.profile-1' ) ) </condition>
				-->
			</activation>
		</profile>

		<!-- working, but still ugly -->

		<profile>
			<id>profile-2b</id>
			<activation>
				<condition><![CDATA[exists( '.profile-2' ) && missing( '.profile-1' )]]></condition>
			</activation>
		</profile>

		<!-- working workaround to get it a bit cleaner -->

		<profile>
			<id>profile-2c</id>
			<activation>
				<condition><![CDATA[
					exists( '.profile-2' ) && missing( '.profile-1' ) && missing( '.profile-0' )
				]]></condition>
			</activation>
		</profile>

		<!-- not working, just because of the line break -->

		<profile>
[...]

PR #12038, opened 2026-05-12
Fix ConditionParser to handle newlines before && operator

blaspat:
Fix for https://github.com/apache/maven/issues/11882

The tokenize() method only treated space as whitespace, so newlines became standalone tokens. When a line break appeared before &&, the operator was parsed as two separate & tokens instead of one &&.

Added \n, \r, \t to whitespace handling and properly tokenize && as a single operator token regardless of surrounding whitespace.

### DIFF
--- a/impl/maven-impl/src/main/java/org/apache/maven/impl/model/profile/ConditionParser.java
+++ b/impl/maven-impl/src/main/java/org/apache/maven/impl/model/profile/ConditionParser.java
@@ -129,8 +129,8 @@ public class ConditionParser {
                 }
                 quoteType = c;
                 sb.append(c);
-            } else if (c == ' ' || c == '\n' || c == '\r' || c == '\t' || c == '(' || c == ')' || c == ',' || c == '+'
-                    || c == '>' || c == '<' || c == '=' || c == '!' || c == '&') {
+            } else if (Character.isWhitespace(c) || c == '(' || c == ')' || c == ',' || c == '+'
+                    || c == '>' || c == '<' || c == '=' || c == '!' || c == '&' || c == '|') {
                 if (!sb.isEmpty()) {
                     tokens.add(sb.toString());
                     sb.setLength(0);
--- a/impl/maven-impl/src/main/java/org/apache/maven/impl/model/profile/ConditionParser.java
+++ b/impl/maven-impl/src/main/java/org/apache/maven/impl/model/profile/ConditionParser.java
@@ -129,17 +129,20 @@ public class ConditionParser {
                 }
                 quoteType = c;
                 sb.append(c);
-            } else if (c == ' ' || c == '(' || c == ')' || c == ',' || c == '+' || c == '>' || c == '<' || c == '='
-                    || c == '!') {
+            } else if (c == ' ' || c == '\n' || c == '\r' || c == '\t' || c == '(' || c == ')' || c == ',' || c == '+'
+                    || c == '>' || c == '<' || c == '=' || c == '!' || c == '&') {
                 if (!sb.isEmpty()) {
                     tokens.add(sb.toString());
                     sb.setLength(0);
                 }
-                if (c != ' ') {
-                    if ((c == '>' || c == '<' || c == '=' || c == '!')
+                if (c != ' ' && c != '\n' && c != '\r' && c != '\t') {
+                    if ((c == '>' || c == '<' || c == '=' || c == '!' || c == '&')
                             && i + 1 < expression.length()
                             && expression.charAt(i + 1) == '=') {
                         tokens.add(c + "=");
+                        i++; // Skip the next character
+                    } else if (c == '&' && i + 1 < expression.length() && expression.charAt(i + 1) == '&') {
+                        tokens.add("&&");
                         i++; // Skip the next character
                     } else {
                         tokens.add(String.valueOf(c));
--- a/impl/maven-impl/src/main/java/org/apache/maven/impl/model/profile/ConditionParser.java
+++ b/impl/maven-impl/src/main/java/org/apache/maven/impl/model/profile/ConditionParser.java
@@ -135,7 +135,7 @@ public class ConditionParser {
                     tokens.add(sb.toString());
                     sb.setLength(0);
                 }
-                if (c != ' ' && c != '\n' && c != '\r' && c != '\t') {
+                if (!Character.isWhitespace(c)) {
                     if ((c == '>' || c == '<' || c == '=' || c == '!' || c == '&')
                             && i + 1 < expression.length()
                             && expression.charAt(i + 1) == '=') {
--- a/impl/maven-impl/src/main/java/org/apache/maven/impl/model/profile/ConditionParser.java
+++ b/impl/maven-impl/src/main/java/org/apache/maven/impl/model/profile/ConditionParser.java
@@ -129,20 +129,32 @@ public class ConditionParser {
                 }
                 quoteType = c;
                 sb.append(c);
-            } else if (Character.isWhitespace(c) || c == '(' || c == ')' || c == ',' || c == '+'
-                    || c == '>' || c == '<' || c == '=' || c == '!' || c == '&' || c == '|') {
+            } else if (Character.isWhitespace(c)
+                    || c == '('
+                    || c == ')'
+                    || c == ','
+                    || c == '+'
+                    || c == '>'
+                    || c == '<'
+                    || c == '='
+                    || c == '!'
+                    || c == '&'
+                    || c == '|') {
                 if (!sb.isEmpty()) {
                     tokens.add(sb.toString());
                     sb.setLength(0);
                 }
                 if (!Character.isWhitespace(c)) {
-                    if ((c == '>' || c == '<' || c == '=' || c == '!' || c == '&')
+                    if ((c == '>' || c == '<' || c == '=' || c == '!')
                             && i + 1 < expression.length()
                             && expression.charAt(i + 1) == '=') {
                         tokens.add(c + "=");
                         i++; // Skip the next character
                     } else if (c == '&' && i + 1 < expression.length() && expression.charAt(i + 1) == '&') {
                         tokens.add("&&");
+                        i++; // Skip the next character
+                    } else if (c == '|' && i + 1 < expression.length() && expression.charAt(i + 1) == '|') {
+                        tokens.add("||");
                         i++; // Skip the next character
                     } else {
                         tokens.add(String.valueOf(c));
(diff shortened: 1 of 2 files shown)


impl/maven-impl/src/main/java/org/apache/maven/impl/model/profile/ConditionParser.java
```diff
@@ -129,18 +129,21 @@ private List<String> tokenize(String expression) {
                 }
                 quoteType = c;
                 sb.append(c);
-            } else if (c == ' ' || c == '(' || c == ')' || c == ',' || c == '+' || c == '>' || c == '<' || c == '='
-                    || c == '!') {
+            } else if (c == ' ' || c == '\n' || c == '\r' || c == '\t' || c == '(' || c == ')' || c == ',' || c == '+'
+                    || c == '>' || c == '<' || c == '=' || c == '!' || c == '&') {
```
gnodet: Use `Character.isWhitespace(c)` instead of enumerating whitespace characters — it is cleaner, handles edge cases like form feed, and is the standard Java idiom. Also add `'|'` to the operator list so `||` can be properly tokenized (see comment below).

```suggestion
            } else if (Character.isWhitespace(c) || c == '(' || c == ')' || c == ',' || c == '+'
                    || c == '>' || c == '<' || c == '=' || c == '!' || c == '&' || c == '|') {
```

impl/maven-impl/src/main/java/org/apache/maven/impl/model/profile/ConditionParser.java
```diff
@@ -129,18 +129,21 @@ private List<String> tokenize(String expression) {
...
                 if (!sb.isEmpty()) {
                     tokens.add(sb.toString());
                     sb.setLength(0);
                 }
-                if (c != ' ') {
-                    if ((c == '>' || c == '<' || c == '=' || c == '!')
+                if (c != ' ' && c != '\n' && c != '\r' && c != '\t') {
```
gnodet: Same here — use `Character.isWhitespace(c)` for consistency.

```suggestion
                if (!Character.isWhitespace(c)) {
```

impl/maven-impl/src/main/java/org/apache/maven/impl/model/profile/ConditionParser.java
```diff
@@ -129,18 +129,21 @@ private List<String> tokenize(String expression) {
...
+                    if ((c == '>' || c == '<' || c == '=' || c == '!' || c == '&')
                             && i + 1 < expression.length()
                             && expression.charAt(i + 1) == '=') {
                         tokens.add(c + "=");
                         i++; // Skip the next character
+                    } else if (c == '&' && i + 1 < expression.length() && expression.charAt(i + 1) == '&') {
+                        tokens.add("&&");
+                        i++; // Skip the next character
```
gnodet: Two issues:

1. **`&` in the `X=` group**: `&` was added to the `>=`/`<=`/`==`/`!=` lookahead, which means `&=` would be tokenized as a valid two-char operator. But `&=` is not valid in this language — remove `|| c == '&'` from this condition.

2. **`||` not handled**: The `||` operator has the exact same tokenization gap that `&&` had. The `|` character is not in the operator list, so `||` only works when surrounded by spaces. Handle it symmetrically with `&&`.

```suggestion
                    if ((c == '>' || c == '<' || c == '=' || c == '!')
                            && i + 1 < expression.length()
                            && expression.charAt(i + 1) == '=') {
                        tokens.add(c + "=");
                        i++; // Skip the next character
                    } else if (c == '&' && i + 1 < expression.length() && expression.charAt(i + 1) == '&') {
                        tokens.add("&&");
                        i++; // Skip the next character
                    } else if (c == '|' && i + 1 < expression.length() && expression.charAt(i + 1) == '|') {
                        tokens.add("||");
                        i++; // Skip the next character
```

Please also add corresponding test cases for `||` with newlines, similar to the `&&` tests.

impl/maven-impl/src/test/java/org/apache/maven/impl/model/profile/ConditionParserTest.java
```diff
@@ -262,6 +262,32 @@ void testPropertyAlias() {
         assertThrows(RuntimeException.class, () -> parser.parse("${unclosed"));
     }
 
+    @Test
+    void testAmpersandAmpersandTokenizerMultiline() {
+        // Regression test for https://github.com/apache/maven/issues/11882
+        // The && operator was not being tokenized correctly when a line break appeared before it.
+        // Uses ${os.name} and ${os.arch} which work reliably in the test environment.
+        // ${os.name} == 'windows' && ${os.arch} == 'amd64' evaluates to true.
```
gnodet: These values come from the mock context in `createMockContext()` (lines 64-68), not from the actual OS. The comment is misleading.

```suggestion
        // Uses ${os.name} and ${os.arch} which are set to 'windows' and 'amd64' in the mock context.
```

gnodet (changes requested):
Thanks for the fix, @blaspat! The root cause analysis is spot-on — whitespace and `&` delimiter handling in the tokenizer were the issue. Here are a few improvements to make the fix more robust before merging:

1. **Use `Character.isWhitespace(c)`** instead of enumerating `' '`, `'\n'`, `'\r'`, `'\t'` — cleaner and handles edge cases
2. **Handle `||` symmetrically with `&&`** — `||` has the same tokenization gap (not in the operator char list)
3. **Remove `&` from the `X=` lookahead** — `&=` is not a valid operator in this language

See inline comments for concrete suggestions.

For the remaining arithmetic operators (`-`, `*`, `/`) that also are not handled as delimiter characters, I have filed https://github.com/apache/maven/issues/12052 as a separate follow-up since those need more careful design (e.g., `-` in property names).

impl/maven-impl/src/test/java/org/apache/maven/impl/model/profile/ConditionParserTest.java
```diff
@@ -262,6 +262,32 @@ void testPropertyAlias() {
...
+        assertTrue((Boolean) parser.parse("${os.arch} == 'amd64'\n&&\n${os.name} == 'windows'"));
+
+        // Case 5: Multiple && with line break before first && (like bad-profile-2d in issue)
+        assertTrue(
+                (Boolean) parser.parse("${os.arch} == 'amd64'\n&& ${os.name} == 'windows' && ${os.name} == 'windows'"));
+    }
+
     @Test
     void testNestedPropertyAlias() {
         functions.put("property", args -> {
```
gnodet, line 315: Please also add a test for `||` with newlines, similar to `testAmpersandAmpersandTokenizerMultiline`. For example:

```java
@Test
void testPipePipeTokenizerMultiline() {
    // || with newlines should work the same as &&
    assertTrue((Boolean) parser.parse("${os.arch} == 'amd64' || ${os.name} == 'linux'"));
    assertTrue((Boolean) parser.parse("${os.arch} == 'amd64'\n|| ${os.name} == 'linux'"));
    assertTrue((Boolean) parser.parse("${os.arch} == 'amd64' ||\n${os.name} == 'linux'"));
    assertTrue((Boolean) parser.parse("${os.arch} == 'amd64'\n||\n${os.name} == 'linux'"));
}
```

gnodet:
@blaspat Thanks for applying the suggestions! Two items still need to be addressed before this can be merged:

**1. `&` is still in the `X=` lookahead (line 139)**

`|| c == '&'` needs to be removed from the `>=`/`<=`/`==`/`!=` condition. Currently `&=` would be tokenized as a valid two-char operator, but it isn't one. The `&&` handling in the `else if` below is correct — it's only the `X=` group that needs the fix.

Current:
```java
if ((c == '>' || c == '<' || c == '=' || c == '!' || c == '&')
        && i + 1 < expression.length()
        && expression.charAt(i + 1) == '=') {
```

Should be:
```java
if ((c == '>' || c == '<' || c == '=' || c == '!')
        && i + 1 < expression.length()
        && expression.charAt(i + 1) == '=') {
```

**2. `||` tokenization is missing — this is now a regression**

`|` was added to the delimiter list (good), but there's no branch to combine two consecutive `|` chars into a single `||` token. This means `a || b` now tokenizes as `["a", "|", "|", "b"]` instead of `["a", "||", "b"]`, which **breaks `||` entirely** — even with spaces, which worked before this PR.

Please add a `||` branch after the `&&` one:
```java
} else if (c == '&' && i + 1 < expression.length() && expression.charAt(i + 1) == '&') {
    tokens.add("&&");
    i++; // Skip the next character
} else if (c == '|' && i + 1 < expression.length() && expression.charAt(i + 1) == '|') {
    tokens.add("||");
    i++; // Skip the next character
} else {
```

**3. Add `||` multiline tests** (as mentioned in the earlier review comment)

gnodet (commented):
Thanks for updating the PR, @blaspat. I appreciate you applying the suggested changes.

Looking at the current diff, two of the three requested changes have been addressed:

1. **`Character.isWhitespace(c)`** — applied, good.
2. **Test comment about mock context** — fixed, good.

However, there are still two outstanding items from the previous review that were not addressed:

- **`&` in the `X=` lookahead**: `c == '&'` is still included in the `>=`/`<=`/`==`/`!=` condition, which means `&=` would be tokenized as a two-char operator. Since `&=` is not a valid operator in this language, `c == '&'` should be removed from that condition. (The `&&` case is handled separately in the `else if` below, so removing it from the `X=` group is safe.)

- **`||` not handled**: The `|` character was correctly added to the operator list (the first `if`), but there is no corresponding `else if` branch to combine `|` + `|` into a `||` token — so `||` will still break when split across a line boundary. This should be handled symmetrically with `&&`, and corresponding tests should be added.

That said, I want to flag that PR #12136 by @Will-thom addresses the same issue (#11882) with a more comprehensive fix that covers all of the above points, includes additional test coverage at the activator and model-building levels, and has already been reviewed. Both PRs are still open, but since #12136 is more thorough, it will likely land first — at which point this PR would be superseded and could be closed.

@blaspat — if #12136 gets merged before this one, you may want to close this PR. Either way, thank you for the contribution and the solid root-cause analysis.

impl/maven-impl/src/test/java/org/apache/maven/impl/model/profile/ConditionParserTest.java
blaspat: test already exists in line 291

gnodet (approved):
All three items from the previous review are now addressed:

1. **`&` removed from `X=` lookahead** — `&=` is no longer tokenized as a valid two-char operator
2. **`||` tokenization added** symmetrically with `&&` — both operators now combine correctly regardless of surrounding whitespace
3. **`||` multiline tests added** — covers line breaks before, after, and around the operator

The tokenizer logic is clean and correct. Note: PR #12136 addresses the same issue (#11882) with a broader approach — whichever lands first should close the other to avoid conflicts.

PR #12069, opened 2026-05-18
Fix site lifecycle reactor dependency resolution

### LINKED ISSUE #12064: Site and dependencies in multi module project
### Affected version

4.x, 3.x

### Bug description

When we have a multi module project and we have dependencies between modules

```
mvn site 
```

will fail as dependencies are not resolved

adding compile phase resolve a problem, like

```
mvn compile site 
```

It can be something special with `default` and `site` lifecycles

Will-thom:
Fixes #12064

This updates Maven's reactor workspace resolution for site lifecycle builds so reports can resolve reactor JAR dependencies even when the producer module has not been compiled or packaged first.

The change keeps compile/default lifecycle behavior unchanged and adds an integration test covering a multi-module site build where a consumer module depends on a producer module. The test verifies that mvn site succeeds without running compile explicitly and without producing producer/target/classes.

This should be a candidate for backport to maven-3.10.x.

### DIFF
--- a/impl/maven-core/src/main/java/org/apache/maven/ReactorReader.java
+++ b/impl/maven-core/src/main/java/org/apache/maven/ReactorReader.java
@@ -40,6 +40,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.Objects;
 import java.util.concurrent.ConcurrentHashMap;
+import java.util.jar.JarOutputStream;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
@@ -217,11 +218,41 @@ class ReactorReader implements MavenWorkspaceReader {
             if (projectHasOutputFromPreviousSession || projectCompiledDuringThisSession) {
                 return outputDirectory;
             }
+
+            if (hasSiteLifecyclePhase(project) && "jar".equals(artifact.getExtension()) && "jar".equals(type)) {
+                return ensureEmptyArtifactFile(artifact);
+            }
         }
 
         // The fall-through indicates that the artifact cannot be found;
         // for instance if package produced nothing or classifier problems.
         return null;
+    }
+
+    private File ensureEmptyArtifactFile(final Artifact artifact) {
+        Path target = getArtifactPath(artifact);
+        synchronized (this) {
+            if (Files.isRegularFile(target)) {
+                return target.toFile();
+            }
+            try {
+                Files.createDirectories(target.getParent());
+                try (JarOutputStream ignored = new JarOutputStream(Files.newOutputStream(target))) {
+                    // create an empty jar for site reports that inspect reactor artifacts before package
+                }
+                return target.toFile();
+            } catch (IOException e) {
+                LOGGER.warn("Unable to create empty reactor artifact for '{}'.", artifact, e);
+                return null;
+            }
+        }
+    }
+
+    private boolean hasSiteLifecyclePhase(MavenProject project) {
+        return project.hasLifecyclePhase("pre-site")
+                || project.hasLifecyclePhase("site")
+                || project.hasLifecyclePhase("post-site")
+                || project.hasLifecyclePhase("site-deploy");
     }
 
     private boolean isPackagedArtifactUpToDate(MavenProject project, File packagedArtifactFile) {
--- a/impl/maven-core/src/main/java/org/apache/maven/ReactorReader.java
+++ b/impl/maven-core/src/main/java/org/apache/maven/ReactorReader.java
@@ -81,6 +81,8 @@ class ReactorReader implements MavenWorkspaceReader {
     private Map<String, Map<String, Map<String, MavenProject>>> projects;
     private Map<String, Map<String, Map<String, MavenProject>>> allProjects;
     private Path projectLocalRepository;
+    private Path emptyArtifactRepository;
+    private final ConcurrentHashMap<Path, File> emptyArtifactCache = new ConcurrentHashMap<>();
     // projectId -> Deque<lifecycle>
     private final Map<String, Deque<String>> lifecycles = new ConcurrentHashMap<>();
 
@@ -219,7 +221,7 @@ class ReactorReader implements MavenWorkspaceReader {
                 return outputDirectory;
             }
 
-            if (hasSiteLifecyclePhase(project) && "jar".equals(artifact.getExtension()) && "jar".equals(type)) {
+            if (isSiteGoalRequested() && "jar".equals(artifact.getExtension()) && "jar".equals(type)) {
                 return ensureEmptyArtifactFile(artifact);
             }
         }
@@ -230,29 +232,35 @@ class ReactorReader implements MavenWorkspaceReader {
     }
 
     private File ensureEmptyArtifactFile(final Artifact artifact) {
-        Path target = getArtifactPath(artifact);
-        synchronized (this) {
-            if (Files.isRegularFile(target)) {
-                return target.toFile();
+        Path target = getEmptyArtifactPath(artifact);
+        return emptyArtifactCache.computeIfAbsent(target, path -> {
+            if (Files.isRegularFile(path)) {
+                return path.toFile();
             }
             try {
-                Files.createDirectories(target.getParent());
-                try (JarOutputStream ignored = new JarOutputStream(Files.newOutputStream(target))) {
+                Files.createDirectories(path.getParent());
+                try (JarOutputStream ignored = new JarOutputStream(Files.newOutputStream(path))) {
                     // create an empty jar for site reports that inspect reactor artifacts before package
                 }
-                return target.toFile();
+                return path.toFile();
             } catch (IOException e) {
                 LOGGER.warn("Unable to create empty reactor artifact for '{}'.", artifact, e);
                 return null;
             }
-        }
+        });
     }
 
-    private boolean hasSiteLifecyclePhase(MavenProject project) {
-        return project.hasLifecyclePhase("pre-site")
-                || project.hasLifecyclePhase("site")
-                || project.hasLifecyclePhase("post-site")
-                || project.hasLifecyclePhase("site-deploy");
+    private boolean isSiteGoalRequested() {
+        return session.getGoals().stream().anyMatch(ReactorReader::isSiteGoal);
+    }
+
+    private static boolean isSiteGoal(String goal) {
+        return "pre-site".equals(goal)
+                || "site".equals(goal)
+                || "post-site".equals(goal)
+                || "site-deploy".equals(goal)
+                || goal.endsWith(":site")
+                || goal.endsWith(":site-deploy");
     }
 
     private boolean isPackagedArtifactUpToDate(MavenProject project, File packagedArtifactFile) {
@@ -516,12 +524,26 @@ class ReactorReader implements MavenWorkspaceReader {
         String version = artifact.getBaseVersion();
         String classifier = artifact.getClassifier();
         String extension = artifact.getExtension();
-        return getArtifactPath(groupId, artifactId, version, classifier, extension);
+        return getArtifactPath(getProjectLocalRepo(), groupId, artifactId, version, classifier, extension);
     }
 
     private Path getArtifactPath(
             String groupId, String artifactId, String version, String classifier, String extension) {
-        Path repo = getProjectLocalRepo();
+        return getArtifactPath(getProjectLocalRepo(), groupId, artifactId, version, classifier, extension);
+    }
+
+    private Path getEmptyArtifactPath(Artifact artifact) {
+        return getArtifactPath(
+                getEmptyArtifactRepo(),
+                artifact.getGroupId(),
+                artifact.getArtifactId(),
+                artifact.getBaseVersion(),
+                artifact.getClassifier(),
+                artifact.getExtension());
+    }
+
+    private Path getArtifactPath(
+            Path repo, String groupId, String artifactId, String version, String classifier, String extension) {
         return repo.resolve(groupId)
                 .resolve(artifactId)
                 .resolve(version)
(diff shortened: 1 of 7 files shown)


slawekjaranowski:
@Will-thom please preserve in PR description last paragraph  from template:

> To make clear that you license your contribution under the Apache License Version 2.0, January 2004 you have to acknowledge this by using the following check-box.
> ...

impl/maven-core/src/main/java/org/apache/maven/ReactorReader.java
```diff
@@ -217,13 +218,43 @@ private File determineBuildOutputDirectoryForArtifact(final MavenProject project
             if (projectHasOutputFromPreviousSession || projectCompiledDuringThisSession) {
                 return outputDirectory;
             }
+
+            if (hasSiteLifecyclePhase(project) && "jar".equals(artifact.getExtension()) && "jar".equals(type)) {
```
gnodet: This checks whether the *producer* project has entered a site lifecycle phase. But dependency resolution for a consumer module can happen before the producer's mojos have started, meaning `project.hasLifecyclePhase("pre-site")` etc. would all return `false` at the time we need it most.

Wouldn't it be more reliable to check the session's top-level goals instead? Something like:
```suggestion
            if (isSiteGoalRequested() && "jar".equals(artifact.getExtension()) && "jar".equals(type)) {
```
with a helper that inspects `session.getGoals()` for site-related goals. That way the check doesn't depend on mojo execution order.

impl/maven-core/src/main/java/org/apache/maven/ReactorReader.java
```diff
@@ -217,13 +218,43 @@ private File determineBuildOutputDirectoryForArtifact(final MavenProject project
...
 
         // The fall-through indicates that the artifact cannot be found;
         // for instance if package produced nothing or classifier problems.
         return null;
     }
 
+    private File ensureEmptyArtifactFile(final Artifact artifact) {
+        Path target = getArtifactPath(artifact);
+        synchronized (this) {
+            if (Files.isRegularFile(target)) {
```
gnodet: Using `synchronized(this)` here locks the entire `ReactorReader` for every artifact resolution, which could become a bottleneck in large reactor builds. Since you only need mutual exclusion per target path, consider using a `ConcurrentHashMap<Path, Path>` with `computeIfAbsent` to avoid global contention:
```java
private final ConcurrentHashMap<Path, File> emptyArtifactCache = new ConcurrentHashMap<>();

private File ensureEmptyArtifactFile(final Artifact artifact) {
    Path target = getArtifactPath(artifact);
    return emptyArtifactCache.computeIfAbsent(target, p -> {
        // create empty jar ...
    });
}
```

impl/maven-core/src/main/java/org/apache/maven/ReactorReader.java
```diff
@@ -217,13 +218,43 @@ private File determineBuildOutputDirectoryForArtifact(final MavenProject project
...
+                return target.toFile();
+            }
+            try {
+                Files.createDirectories(target.getParent());
+                try (JarOutputStream ignored = new JarOutputStream(Files.newOutputStream(target))) {
+                    // create an empty jar for site reports that inspect reactor artifacts before package
+                }
+                return target.toFile();
```
gnodet: This empty JAR is written to the project-local-repo. If the user subsequently runs `mvn package` (without `clean`), `findInProjectLocalRepository` will find this empty JAR and `isPackagedArtifactUpToDate` may consider it valid (if no output directory exists yet). This could silently produce a broken build.

Consider either:
- Writing the empty JAR to a temporary/distinct location that won't be picked up by normal artifact resolution, or
- Adding metadata (e.g., a marker file or attribute) so the reactor reader knows to ignore it in non-site builds.

gnodet (commented):

Thanks for the fix. The approach of synthesizing an empty JAR so site reports can resolve reactor dependencies is creative, but I have a few concerns about correctness and potential side effects.

**Main concern:** The `hasSiteLifecyclePhase(project)` check queries whether the *producer* project has already entered a site lifecycle phase. But `determineBuildOutputDirectoryForArtifact` is typically called when resolving dependencies of a *consumer* project. At that point, the producer module may not have started its own site phase yet (reactor order means the producer's mojos may not have fired). The `lifecyclePhases` set is populated by `MojoStarted` events, which only fire when a mojo actually runs on that project. So this guard could fail to trigger exactly when it's needed.

Could you verify that the producer project has indeed entered its site phase before the consumer tries to resolve it? If not, you may need a different detection mechanism (e.g., checking the session's goals list rather than per-project phase tracking).

**Secondary concerns:**
- The empty JAR is written to `project-local-repo` and could persist across builds. A subsequent `mvn compile package` would find this empty JAR in the project-local-repo and might treat it as up-to-date, leading to a broken build.
- `synchronized(this)` on the entire `ReactorReader` instance is coarse-grained; a per-path lock (or `ConcurrentHashMap.computeIfAbsent`) would be more appropriate for a workspace reader that can be called concurrently.

gnodet (approved):
All three concerns from the previous review are addressed in the follow-up commit:

1. **Detection mechanism**: Replaced per-project lifecycle phase tracking (`hasSiteLifecyclePhase`) with `isSiteGoalRequested()` checking `session.getGoals()`. This is deterministic and available from session start — no longer depends on producer modules having started their own site phase.

2. **Artifact isolation**: Empty JARs now go to a separate `reactor-empty-artifacts` directory (sibling of `project-local-repo`), so a subsequent `mvn package` won't find stale empty JARs. The integration test explicitly asserts this.

3. **Concurrency**: `synchronized(this)` replaced with `ConcurrentHashMap.computeIfAbsent` for per-path locking — correct and efficient.

The `isSiteGoal()` helper properly covers both lifecycle phases (`site`, `pre-site`, `post-site`, `site-deploy`) and direct plugin goals (`*:site`, `*:site-deploy`). Integration test is well-structured. LGTM.

Will-thom:
Addressed the review concerns in a follow-up commit:

- site detection now uses the requested session goals instead of the producer project's lifecycle state;
- synthetic empty JARs are written under a separate `reactor-empty-artifacts` directory instead of `project-local-repo`;
- empty artifact creation is cached per target path with `ConcurrentHashMap.computeIfAbsent`, avoiding the previous coarse `synchronized(this)` lock;
- the IT now asserts that `site` does not leave a synthetic producer JAR in `target/project-local-repo`.

Validation passed locally:

```sh
mvn -Dmaven.repo.local=..\maven-12069-m2 -Drat.skip=true -Dsurefire.failIfNoSpecifiedTests=false -Dtest=ReactorReader -pl impl/maven-core -am test
mvn -Dmaven.repo.local=..\maven-12069-m2 -Drat.skip=true -Dsurefire.failIfNoSpecifiedTests=false -Dits.test=MavenITgh12064SiteReactorDependenciesTest -Prun-its -pl its/core-it-suite -am verify
```

The focused IT report shows 1 test, 0 failures, 0 errors.

gnodet:
Thanks for working on this! A few thoughts after reviewing:

**The underlying issue is real** — `mvn site` failing on multi-module projects with inter-module dependencies is a pain point, and requiring `mvn compile site` is not great UX.

However, I have some concerns about the approach:

**Silent failures for common configurations:** The PR injects empty JARs into the reactor when a site goal is detected. This works fine for basic site generation (project-info-reports, etc.), but many projects have Javadoc, JXR, SpotBugs, or other reports enabled in their POM that inspect actual bytecode or source paths resolved from the classpath. With this change, those reports would silently produce empty/incomplete output instead of failing with a clear error. Today the failure is explicit — you know you need `mvn compile site`. With the empty JAR workaround, it passes but the output is wrong.

**Scope of the change:** Patching `ReactorReader` (a low-level artifact resolution component) with site-specific logic feels like the wrong layer. The site goal detection via hardcoded strings (`pre-site`, `post-site`, `:site`, etc.) is fragile and couples artifact resolution to lifecycle knowledge it shouldn't have.

**Possible alternative:** Would it be better to address this at the lifecycle/phase level — e.g., having the site lifecycle declare a dependency on the compile phase, or handling this in the site plugin itself?

This feels more like a feature/enhancement than a bug fix, and probably not a candidate for rc-6. What do you think?

gnodet (changes requested):
Withdrawing approval — see comment above for concerns about silent failures and the approach.

gnodet (commented):
The underlying issue — `mvn site` failing on multi-module projects when dependencies haven't been compiled — is real. However, injecting empty JAR artifacts into `ReactorReader` raises significant architectural concerns:

**1. Silent incorrect output (high)**
The core behavior change returns an empty JAR when no compiled output exists and a site goal is detected. This masks the real problem instead of surfacing it. Report plugins that inspect bytecode or source paths (maven-javadoc-plugin, maven-jxr-plugin, spotbugs-maven-plugin) would silently produce empty or incomplete output rather than failing with a clear error. The maintainer independently raised this exact concern.

**2. Lifecycle coupling in artifact resolution (confirmed with nuance)**
The `isSiteGoal` method hardcodes lifecycle phase names and uses `endsWith(":site")` pattern matching, coupling a low-level artifact resolution component to lifecycle-specific knowledge it shouldn't have. While the specific claim about Maven 4 phase name incompatibility is overstated (Maven 4 uses the same canonical names with aliases), the design coupling concern is valid.

**3. Cleanup inconsistency (medium)**
The `cleanProjectLocalRepository` method does not clean the `reactor-empty-artifacts` directory, while both directories live under `target/`. `mvn clean` handles both, but the within-session cleanup is inconsistent.

**4. Integration test gap (low)**
The test only exercises basic `project-info-reports`. It does not test what happens when a report plugin that needs compiled classes (javadoc, spotbugs) is configured — which is the failure mode that matters most.

As the maintainer suggested, this would be better addressed at the lifecycle or plugin level — the site lifecycle could declare a dependency on the compile phase, or `maven-site-plugin` could handle missing reactor artifacts gracefully. Patching a core artifact resolution component with lifecycle-specific workarounds sets a problematic precedent.

impl/maven-core/src/main/java/org/apache/maven/ReactorReader.java
```diff
@@ -217,13 +220,49 @@ private File determineBuildOutputDirectoryForArtifact(final MavenProject project
             if (projectHasOutputFromPreviousSession || projectCompiledDuringThisSession) {
                 return outputDirectory;
             }
+
```
gnodet, line 223: **[high]** The empty JAR injection silently masks missing compiled output. Report plugins that inspect bytecode or source paths resolved from the classpath (`maven-javadoc-plugin`, `maven-jxr-plugin`, `spotbugs-maven-plugin`) would silently produce empty or incomplete output instead of failing with a clear error. Today the failure is explicit — users know they need `mvn compile site`. With this workaround, the build passes but the output is wrong.

impl/maven-core/src/main/java/org/apache/maven/ReactorReader.java
```diff
@@ -217,13 +220,49 @@ private File determineBuildOutputDirectoryForArtifact(final MavenProject project
...
+                }
+                return path.toFile();
+            } catch (IOException e) {
+                LOGGER.warn("Unable to create empty reactor artifact for '{}'.", artifact, e);
+                return null;
+            }
+        });
+    }
+
+    private boolean isSiteGoalRequested() {
```
gnodet, line 253: **[medium]** The `isSiteGoal` method hardcodes lifecycle phase names (`pre-site`, `site`, `post-site`, `site-deploy`) and uses `endsWith(":site")` pattern matching, coupling a low-level artifact resolution component to lifecycle-specific knowledge. `ReactorReader`'s responsibility is workspace artifact resolution — it should not contain lifecycle-specific workarounds. This sets a precedent for adding other lifecycle-specific branches in the future.

its/core-it-suite/src/test/java/org/apache/maven/it/MavenITgh12064SiteReactorDependenciesTest.java
```diff
@@ -0,0 +1,55 @@
...
+        verifier.deleteDirectory("consumer/target");
+        verifier.deleteArtifacts("org.apache.maven.its.gh12064");
+        verifier.addCliArgument("site");
+        verifier.execute();
+        verifier.verifyErrorFreeLog();
+
+        verifier.verifyFilePresent("consumer/target/site/dependencies.html");
+        assertFalse(new File(testDir, "producer/target/classes").exists(), "site must not require compile first");
+        assertFalse(
+                new File(
```
gnodet, line 48: **[low]** The integration test only exercises basic `project-info-reports` (`dependencies.html`). It does not test what happens when a report plugin that needs compiled classes (`maven-javadoc-plugin`, `spotbugs-maven-plugin`) is configured, which is the failure mode that matters most. A test that configures `maven-javadoc-plugin` and verifies the output is complete (or at least non-empty) would strengthen confidence in the approach.

gnodet (commented):
### Summary

The PR addresses a real problem ([#12064](https://github.com/apache/maven/issues/12064) — `mvn site` failing on multi-module projects with inter-module dependencies without prior compile), but injecting empty JARs in `ReactorReader` has confirmed drawbacks: (1) it silently masks missing compiled output for bytecode-inspecting report plugins, producing incorrect results instead of clear errors; (2) it couples a low-level artifact resolution component to lifecycle-specific knowledge via hardcoded phase names; (3) the within-session cleanup path is inconsistent with the new artifact directory; and (4) the integration test does not cover the failure mode that matters most.

This review aligns with the existing feedback from the maintainer — the fundamental architectural concern remains that ReactorReader is not the right place to solve a lifecycle ordering problem. The fix would be better addressed at the lifecycle or plugin level.

**Additional note:** `cleanProjectLocalRepository` (existing code) only cleans `getProjectLocalRepo()` paths but does not clean the new `reactor-empty-artifacts` directory from `getEmptyArtifactRepo()`. While `mvn clean` removes everything in `target/`, the within-session cleanup path (triggered by lifecycle phase transitions) could leave stale empty artifacts.

### 📋 PR Metadata

| Milestone | _(none)_ | `4.x` |

PR #12071, opened 2026-05-18
[maven-4.0.x] Upgrade plugin dependencies (extra-enforcer-rules) in mvnup

gnodet:
Backport of #12051 to `maven-4.0.x`.

## Summary

Adds support for upgrading plugin dependencies in `mvnup`, specifically `org.codehaus.mojo:extra-enforcer-rules` to minimum version 1.4 for Maven 4 compatibility.

## Problem

`extra-enforcer-rules` versions before 1.4 use `DependencyGraphBuilder.buildDependencyGraph(MavenProject, ArtifactFilter)` — a method removed in Maven 4. This causes `NoSuchMethodError` at runtime when rules like `BanCircularDependencies` are executed.

Found during maven4-testing across ~960 Apache repositories (e.g., rocketmq uses `extra-enforcer-rules:1.0-beta-4`).

## Changes

- `PluginUpgradeStrategy`: Added `PLUGIN_DEPENDENCY_UPGRADES` list and `upgradePluginDependencies()` method
- Added 2 tests for plugin dependency upgrades

_Claude Code on behalf of Guillaume Nodet_

PR #12092, opened 2026-05-19
Downgrade plexus-classworlds from 2.11.0 to 2.9.0

gnodet:
## Summary

- Downgrade plexus-classworlds from 2.11.0 to 2.9.0 to fix all integration test failures on ubuntu-latest

## Root Cause

Classworlds 2.11.0 introduced a bug in `ConfigurationParser.loadGlob()` ([commit 4f4cfe64](https://github.com/codehaus-plexus/plexus-classworlds/commit/4f4cfe64d1)) where an automated code cleanup incorrectly changed `&&` to `||` in the glob file filter. This causes `load maven-*.jar` in `m2.conf` to match **all** `.jar` files instead of only those starting with `maven-`.

On Linux ext4, `File.listFiles()` returns non-deterministic ordering (inode hash order). After a recent GitHub Actions runner image update (`ubuntu24/20260513.135`), `org.eclipse.sisu.plexus-1.0.0.jar` loads before `maven-embedder-4.0.0-SNAPSHOT.jar`, causing Sisu's `PlexusXmlBeanConverter` (without XmlNode handling) to win the class collision. This breaks lifecycle configuration injection for any component using `XmlNode` fields (e.g., `LifecycleMojo.configuration`).

This is the root cause of all 5 backport PR failures (#12071, #12070, #12073, #12077, #12079) — they are all innocent bystanders.

Upstream fix: https://github.com/codehaus-plexus/plexus-classworlds/pull/147

Once classworlds ships a fix (2.11.1+), we can upgrade again.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -145,7 +145,7 @@ under the License.
     <assertjVersion>3.27.7</assertjVersion>
     <asmVersion>9.9.1</asmVersion>
     <byteBuddyVersion>1.18.8</byteBuddyVersion>
-    <classWorldsVersion>2.11.0</classWorldsVersion>
+    <classWorldsVersion>2.9.0</classWorldsVersion>
     <commonsCliVersion>1.11.0</commonsCliVersion>
     <guiceVersion>5.1.0</guiceVersion>
     <guavaVersion>33.6.0-jre</guavaVersion>


cstamas (approved):

slawekjaranowski (approved):

slawekjaranowski (approved):

slawekjaranowski:
Does 3.x also affected?

gnodet:
Maven 3.9.x also uses classworlds 2.11.0, so the `loadGlob` bug is technically present. However, 3.9.x is **not practically affected** because:

1. Its `m2.conf` uses `load ${maven.home}/lib/*.jar` (empty prefix), so the `||` vs `&&` distinction doesn't matter — `"".startsWith("")` is always true regardless.
2. There is no `PlexusXmlBeanConverter` class collision in 3.9.x (the custom version only exists in 4.0.x's `maven-embedder`).

The 4.0.x `m2.conf` uses `load ${maven.home}/lib/maven-*.jar` followed by `load ${maven.home}/lib/*.jar` — the first glob's `maven-` prefix is what triggers the bug (matching all jars instead of only `maven-*` jars), defeating the intended load ordering.

That said, it would still be good to align 3.9.x with the fix once classworlds ships 2.11.1, to avoid any future surprises.

PR #12112, opened 2026-05-19
Resolver 2018

cstamas:
Bumps `resolverVersion` from 2.0.16 to 2.0.18. Also some related changes.

Changes:
* updates Resolver to 2.0.18
* dropped any existing filter building, is happening now in new component in Resolver 2.0.18 (same as in Maven 3.10.x)
* user configured filter is NOT applied to plugin resolution
* Legacy code uses `LegacyTrackingFileManager` the rest `TrackingFileManagerSupplier`
* Expose `RepositorySystemSession` as session scoped, same as Maven 3.10.x (in fact, Maven 3.9.x did seed it, but Module by mistake never exposed it -- this is one of changes in Maven 3.10.x)

### DIFF
--- a/api/maven-api-core/src/main/java/org/apache/maven/api/Constants.java
+++ b/api/maven-api-core/src/main/java/org/apache/maven/api/Constants.java
@@ -344,23 +344,37 @@ public final class Constants {
     public static final String MAVEN_RELOCATIONS_ENTRIES = "maven.relocations.entries";
 
     /**
-     * User property for version filter expression used in session, applied to resolving ranges: a semicolon separated
-     * list of filters to apply. By default, no version filter is applied (like in Maven 3).
+     * Builds <code>org.eclipse.aether.collection.VersionFilter</code> instances out of input expression string.
+     * <br/>
+     * Expression is a semicolon separated list of filters to apply. By default, no version filter is applied (like in Maven 3).
      * <br/>
      * Supported filters:
      * <ul>
-     *     <li>"h" or "h(num)" - highest version or top list of highest ones filter</li>
-     *     <li>"l" or "l(num)" - lowest version or bottom list of lowest ones filter</li>
-     *     <li>"s" - contextual snapshot filter</li>
-     *     <li>"ns" - unconditional snapshot filter (no snapshots selected from ranges)</li>
-     *     <li>"e(G:A:V)" - predicate filter (leaves out G:A:V from range, if hit, V can be range)</li>
+     *     <li><code>"s"</code> - contextual snapshot filter (project version decides are snapshots allowed or not)</li>
+     *     <li><code>"nosnapshot"</code> - unconditional snapshot filter (no snapshot versions selected from ranges)</li>
+     *     <li><code>"norelease"</code> - unconditional release filter (no release versions selected from ranges)</li>
+     *     <li><code>"nopreview"</code> - unconditional preview filter (no preview versions selected from ranges)</li>
+     *     <li><code>"noprerelease"</code> - unconditional pre-release filter (no preview and rc/cr versions selected from ranges)</li>
+     *     <li><code>"noqualifier"</code> - unconditional any-qualifier filter (no version with any qualifier selected from ranges)</li>
+     *     <li><code>"h"</code> (shorthand of <code>h(1)</code>) or <code>"h(num)"</code> - highest N version (based on version ordering)</li>
+     *     <li><code>"l"</code> (shorthand of <code>l(1)</code>) or <code>"l(num)"</code> - lowest N version (based on version ordering)</li>
+     *     <li><code>"e(V)"</code> - exclusion filter (excludes versions matching V version constraint)</li>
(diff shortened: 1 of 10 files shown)


gnodet (changes requested):

## Review: Resolver 2.0.18 Upgrade

### Overview

This PR upgrades Maven Resolver from 2.0.16 to 2.0.18 and adapts Maven to the resulting API changes: delegating version filter expression parsing to Resolver's new `VersionFilterBuilder`, switching to `LegacyTrackingFileManager`/`TrackingFileManagerSupplier`, introducing `NamedLockFactorySelector`, suppressing version filters during plugin resolution, and exposing `RepositorySystemSession` as session-scoped to align with Maven 3.10.x. It also downgrades `classWorldsVersion` from 2.11.0 to 2.9.0.

### Issues

#### Blocking

**B1. `classWorldsVersion` downgrade to 2.9.0 — should use 2.12.0 instead.**
The downgrade from 2.11.0 is justified (critical glob matching bug in `ConfigurationParser.loadGlob()`), but 2.12.0 was released with the fix (plexus-classworlds#147). Reverting to 2.9.0 loses improvements from 2.10.0+. If 2.12.0 is intentionally avoided, the rationale should be documented.

#### Major

**M1. Test coverage for version filter expressions is deleted with no replacement.**
The `versionFilteringTest()` method (58 lines) in `DefaultRepositorySystemSessionFactoryTest` is removed because parsing moved to Resolver. However, there's no integration-level test verifying that Maven correctly wires the `VersionFilterBuilder` and that `maven.session.versionFilter` still produces a working filter. Even a single smoke test confirming the wiring would catch regressions.

**M2. `@since` on `MAVEN_VERSION_FILTER` changed from `4.0.0` to `3.10.0` — needs verification.**
If this feature was backported to 3.10.x, the change is correct. If it only exists in 4.x, this would be misleading.

#### Minor

**M3. Commit history contains a "WIP" commit and a merge commit.**
The 9 commits include `WIP` and a merge commit — these should be squashed or given meaningful messages before merge.

**M4. Javadoc formatting: unmatched quote.**
In the `MAVEN_VERSION_FILTER` Javadoc example: `"h(5);s;e(1)@org.foo:bar` — the opening `"` has no closing quote.

**M5. `Optional.map()` used for side effect — should be `ifPresent()`.**
```java
versionFilterBuilder
        .buildVersionFilter(mergedProps.get(Constants.MAVEN_VERSION_FILTER), this::parseVersionConstraint)
        .map(sessionBuilder::setVersionFilter);
```
Using `.map()` purely for the side effect of calling `setVersionFilter` is semantically an `.ifPresent()` call. `.ifPresent(sessionBuilder::setVersionFilter)` would express the intent more clearly.

### Suggestions (non-blocking)

- The expanded Javadoc documenting new filter types (`norelease`, `nopreview`, `noprerelease`, `noqualifier`, `i(V)`, scoping with `@G[:A]`) is a welcome improvement.
- The `RepositorySystemSession` session-scope exposure aligns well with 3.10.x. A brief comment in `DefaultMaven.java` noting this mirrors 3.10.x behavior would preserve the "Module by mistake never exposed it" context from the PR description.
- The `VERSION_FILTER_SUPPRESSED` flag in `DefaultPluginDependenciesResolver` is the right approach. Consider whether the same suppression should apply to build/core extension resolution paths if not already covered.

cstamas:
B1 - classWorlds handled on master (merged)
M1 - in Maven there is no filter building happening anymore (see https://github.com/apache/maven-resolver/blob/master/maven-resolver-impl/src/test/java/org/eclipse/aether/internal/impl/collect/DefaultVersionFilterBuilderTest.java)
M2 - that is correct, same Resolver 2.0.18 (with same builder) is being used in Maven 3.10.x, syntax is 100% same (given component comes from resolver)
M3 - rebased
M4 - fixed
M5 - fixed

Suggestions:
S1: unsure what more to tell here?
S2: Fixed
S3: build/core extensions end up at this same execution path (at least in Maven 3.10.x? Guillaume?)

gnodet:

Thanks for addressing all the points!

B1: ack, good.
M1-M5: all clear, thanks for the fixes.

Regarding S3: confirmed on both 3.10.x and 4.x — all three resolution paths converge on `resolveInternal()` where `VERSION_FILTER_SUPPRESSED` is set:

- **Plugins**: `resolve()` / `resolvePlugin()` → `resolveInternal()`
- **Build extensions**: `DefaultMavenPluginManager.resolveExtensionArtifacts()` → `pluginDependenciesResolver.resolve()` → `resolveInternal()`
- **Core extensions**: `resolveCoreExtension()` → `resolveInternal()`

Checked on `origin/maven-3.10.x` at `DefaultPluginDependenciesResolver.java:199` — same suppression point.

LGTM, approving.

gnodet (approved):

All concerns addressed. Build extensions, core extensions, and plugins all flow through `resolveInternal()` where `VERSION_FILTER_SUPPRESSED` is set — S3 confirmed covered.

PR #12117, opened 2026-05-19
Support sealed parameter implementation hints

### LINKED ISSUE #11378: Maven 4: first class support for sealed types in parameter configuration.

Hello.

Proposing that given Maven 4 will baseline to Java 17, sealed classes get a simplified interface for implementation discovery during the parameter convertion process.

## Aims

- Reduce boilerplate for users.
- Allow polymorphic conversion of sealed types using intrinsics.
- Allow further documentation generation automatically from sealed types.

## Sealed type handling

Effectively, suppose I have the following Mojo:

```java
@Mojo(...)
public class MyMojo extends ... {

  @Parameter
  ThirdPartyArtifact artifact;

  ...
}
```

Now, let's assume `ThirdPartyArtifact` is sealed:

```java
public sealed interface ThirdPartyArtifact {
  public final class MavenArtifact implements ThirdPartyArtifact { ... }

  public final class SystemPathArtifact implements ThirdPartyArtifact { ... }

  public final class LocalPathArtifact implements ThirdPartyArtifact { ... }

  public final class RemoteUrlArtifact implements ThirdPartyArtifact { ... }
}
```

Currently, my understanding is that the user would have to provide the canonical class name and package to the `implementation` attribute in their XML configuration, to avoid ambiguity.

Given the types are sealed, we should be able to query `Class#getPermittedSubclasses()` to get all implementations. This would in theory allow us to perform this class lookup based on just the class simple name, allowing a user to simplify configuration from...

```xml
<configuration>
  <artifact implementation="io.github.ascopes.myplugin.models.MavenArtifact.SystemPathArtifact">
    <name>gcc</name>
  </artifact>
</configuration>
```

...to just...

```xml
<configuration>
  <artifact implementation="SystemPathArtifact">
    <name>gcc</name>
  </artifact>
</configuration>
```

...which is far less verbose and far more user-friendly. 

## Possible implementation

A new converter could be added to the Plexus context by default with the following implementation:

```java
public class SealedTypeConverter extends AbstractBasicConverter {

  @Override
  public boolean canConvert(Class<?> cls) {
    return cls.isSealed();
  }

  @Override
  public Object fromConfiguration(
      ConverterLookup lookup,
      PlexusConfiguration configuration,
      Class<?> type,
      Class<?> enclosingType,
      ClassLoader classLoader,
      ExpressionEvaluator evaluator
  ) throws ComponentConfigurationException {
    if (configuration.getChildCount() == 0) {
      return null;
    }

    var hint = configuration.getAttribute("implementation");

    var implementationClass = Stream
        .of(type.getPermittedSubclasses())
        .sorted(Class::getName)
        .filter(cls -> getPermittedClassNames(cls)
            .anyMatch(name -> name.equals(hint))
        .findFirst();

    if (implementationClass.isNotPresent()) {
      throw new ComponentConfigurationException(...);
    }

    return converterLookup.lookupConverterFromType(implementationClass.get())
        .fromConfiguration(...);
  }

  private Stream<String> getPermittedClassNames(Class<?> type) {
    var canonicalNames = Stream.of(
        type.getName(),
        type.getSimpleName()
    );

    var customHints = ofNullable(type.getAnnotation(SealedTypeHint.class)
        .map(SealedTypeHint::value)
        .stream()
        .flatMap(Stream::of);

    return Stream.concat(canonicalNames, customHints);
  }
}
```

## Addition of metadata to `@Parameter`

It would be sensible to add an attribute to `@Parameter` to allow the specification of the default implementation type to consider if information is not provided. I'd envison this being allowed for both sealed and non-sealed types, for example:

```java
@interface Parameter {
  ...

  Class<?> defaultImplementation() default void.class;
}
```

```java
class MyMojo extends ... {
  @Parameter(defaultImplementation = MavenArtifact.class)
  ThirdPartyArtifact artifact;

  // Also useful for cases where code is generated via an annotation processor.
  // Consider the immutables library:
  @Parameter(defaultImplementation = MutableSourceListing.class)
  SourceListing source;
}
```

## Caveats and possible improvements

A caveat to this would be if there is more than one implementor with the same name. Aside from being a very poor design decision, this could be worked around by either raising a warning/exception, or providing the developer with a way of "tagging" implementation names to the sealed class implementations via an annotation.

The annotation for doing this could
be the following:

```java
@Retention(RUNTIME)
@Target({TYPE, RECORD})
public @interface SealedTypeHint {
  String[] value();
}
```

...this would allow defining my example from earlier like so:

```java
public sealed interface ThirdPartyArtifact {
  @SealedTypeHint("maven")
  public final class MavenArtifact implements ThirdPartyArtifact { ... }

  @SealedTypeHint("system-path")
  public final class SystemPathArtifact implements ThirdPartyArtifact { ... }

  @SealedTypeHint("local-path")
  public final class LocalPathArtifact implements ThirdPartyArtifact { ... }

  @SealedTypeHint("remote")
  public final class RemoteUrlArtifact implements ThirdPartyArtifact { ... }
}
```

...enabling the XML example to be just...

```xml
<configuration>
  <artifact implementation="system-path">
    <name>gcc</name>
  </artifact>
</configuration>
```

One other caveat may be the use of `non-sealed` implementation types. However, in this case, the default behaviour of `implementation` should still be allowed for backwards compatibility.

### Further possible improvements

- Records could be supported for deserialization via Plexus to allow developers to create concise parameter definitions.
- Documentation could be generated from the knowledge of the sealed types in `maven-plugin-plugin` to avoid developers having to manually document supported types.
- The `maven-plugin-plugin` could configure the generated XSD schema to have knowledge of the known permitted types for the parameter, allowing IDEs to provide clear linting and autocomplete for polymorphic types without
having to perform expensive classpath analysis of the plugin and dependency tree.
    - This implementation would be compatible with other schema formats like JsonSchema, should Maven decide to allow non-XML configuration in the future.
- The `@SealedTypeHint` _in theory_ could be changed to be just a regular `@TypeHint` or leverage similar Java EE/Jakarta EE metadata annotations for the same purpose. This would potentially require costly classpath scanning though which may hinder startup execution times for plugins. In theory though this information may be possible to collect when `maven-plugin-plugin` is run.

## Workarounds

Right now in Maven 3, parameters have to have a clear type or rely on the user giving a fully qualified canonical class name. This contributes to the outdated archetype that Maven is overly verbose for new users, which can hinder adoption.

Polymorphic types have to be documented explicitly by the developer rather than being generated from JavaDocs when a type is sealed, which leads to documentation mistakes and inconsistencies. Many plugins just do not document this sort of thing properly at all, requiring users of plugins to have to read the source code to have any idea how to configure this sort of thing properly.

Will-thom:
Fixes #11378

## Summary

When a plugin parameter type is a sealed interface/class, users currently must provide the fully qualified class name in the `implementation` attribute. This PR allows using the simple class name instead, by scanning `Class.getPermittedSubclasses()` when standard Plexus class resolution fails.

For example, users can now write:
```xml
<artifact implementation="LocalArtifact">
  <name>local</name>
</artifact>
```
instead of:
```xml
<artifact implementation="com.example.myplugin.LocalArtifact">
  <name>local</name>
</artifact>
```

Fully qualified names continue to work as before.

## Changes

- Add `resolveClassForImplementationHint()` wrapper in `EnhancedConfigurationConverter` that falls back to sealed type resolution when standard Plexus resolution fails
- Handle ambiguous simple names (multiple permitted subclasses sharing the same simple name) with a clear error message
- Unit tests in `DefaultBeanConfiguratorTest` covering simple name, fully qualified name, and missing subclass scenarios
- Integration test with a Maven 4 plugin exercising sealed parameter configuration end-to-end


gnodet (commented):

## Review: GH-11378 Support sealed parameter implementation hints

### Overview

This PR adds sealed-class awareness to Maven 4's plugin parameter configuration. When a plugin parameter type is a sealed interface/class, users can now use simple class names (e.g., `implementation="LocalArtifact"`) instead of fully qualified names in their POM configuration. The implementation wraps the existing Plexus `getClassForImplementationHint` call — on failure, it falls back to scanning `Class.getPermittedSubclasses()` for a match. Includes unit tests and a Maven 4 plugin integration test.

### Verdict

**Looks good overall — minor issues to address.** The approach is sound: backward-compatible, minimally invasive, and properly scoped.

### Issues

**Minor — Missing `cause` in ambiguity exception**

In `getPermittedSubclass()`, the "not found" error path includes the original `cause` exception, but the "ambiguous" path does not. This loses diagnostic context:

```java
// "not found" — includes cause
throw new ComponentConfigurationException(configuration,
    "Cannot find permitted subclass '...'", cause);

// "ambiguous" — missing cause
throw new ComponentConfigurationException(configuration,
    "Implementation hint '...' is ambiguous...");
```

Should add `cause` to the ambiguity exception for consistency.

**Minor — Odd control flow: sort before isEmpty check**

```java
if (matches.size() == 1) {
    return matches.get(0);
}
matches.sort(Comparator.comparing(Class::getName));
if (matches.isEmpty()) {
    throw ...;
}
```

Sorting an empty list is harmless, but checking `isEmpty()` before sorting reads more naturally and makes the intent clearer.

**Minor — No test for the ambiguity case**

The unit tests cover simple name resolution, fully qualified name, and missing class — but there's no test for the ambiguity path (two permitted subclasses with the same simple name). This code path exists and has specific error formatting with sorted class names, so it should be verified.

**Nit — Brittle error message assertion**

`testSealedTypeImplementationHintMustMatchPermittedSubclass` asserts `e.getMessage()` against an exact string. If `BeanConfigurationException` wraps the cause message differently across versions, this could break. Consider using a `contains` check instead.

### Suggestions (non-blocking)

- **Recursive sealed hierarchies**: `getPermittedSubclasses()` only returns direct permits. If a hierarchy has intermediate sealed types (`A permits B` where `B` is sealed and `B permits C`), `C` won't be found from `A`. Fine for initial implementation but worth noting for future work.

- **IT test plugin pom.xml** uses `<maven.version>4.1.0-SNAPSHOT</maven.version>` — correct for development against master but will need updating if the SNAPSHOT version changes before merge.

- The issue mentions several follow-up features (`@SealedTypeHint` annotation, `defaultImplementation` on `@Parameter`, XSD generation). These are clearly out of scope for this PR, which is the right call.

gnodet (commented):
_Claude Code on behalf of Guillaume Nodet_ — inline suggestions below.

impl/maven-core/src/main/java/org/apache/maven/configuration/internal/EnhancedConfigurationConverter.java
```diff
@@ -108,6 +111,56 @@ public Object fromConfiguration(
...
+            return matches.get(0);
+        }
+        matches.sort(Comparator.comparing(Class::getName));
+
+        if (matches.isEmpty()) {
+            throw new ComponentConfigurationException(
+                    configuration,
+                    "Cannot find permitted subclass '" + implementation + "' for sealed type " + type.getName(),
+                    cause);
+        }
```
gnodet, line 154: Minor: sorting before the `isEmpty()` check is harmless but reads oddly. Moving `isEmpty()` first makes the intent clearer:

```suggestion
        if (matches.size() == 1) {
            return matches.get(0);
        }
        if (matches.isEmpty()) {
            throw new ComponentConfigurationException(
                    configuration,
                    "Cannot find permitted subclass '" + implementation + "' for sealed type " + type.getName(),
                    cause);
        }
        matches.sort(Comparator.comparing(Class::getName));
```

impl/maven-core/src/main/java/org/apache/maven/configuration/internal/EnhancedConfigurationConverter.java
```diff
@@ -108,6 +111,56 @@ public Object fromConfiguration(
...
+
+        throw new ComponentConfigurationException(
+                configuration,
+                "Implementation hint '" + implementation + "' is ambiguous for sealed type " + type.getName() + ": "
+                        + matches.stream().map(Class::getName).toList());
```
gnodet: Minor: the "not found" path includes `cause` but the ambiguity path does not. Should be consistent so diagnostic context isn't lost:

```suggestion
        throw new ComponentConfigurationException(
                configuration,
                "Implementation hint '" + implementation + "' is ambiguous for sealed type " + type.getName() + ": "
                        + matches.stream().map(Class::getName).toList(),
                cause);
```

gnodet (commented):
_Claude Code on behalf of Guillaume Nodet_ — inline suggestions.

impl/maven-core/src/test/java/org/apache/maven/configuration/DefaultBeanConfiguratorTest.java
```diff
@@ -31,6 +31,8 @@
 import org.junit.jupiter.api.Test;
 
 import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertInstanceOf;
+import static org.junit.jupiter.api.Assertions.assertThrows;
```
gnodet, line 35: Add `assertTrue` import for the assertion change below:

```suggestion
import static org.junit.jupiter.api.Assertions.assertInstanceOf;
import static org.junit.jupiter.api.Assertions.assertThrows;
import static org.junit.jupiter.api.Assertions.assertTrue;
```

impl/maven-core/src/test/java/org/apache/maven/configuration/DefaultBeanConfiguratorTest.java
```diff
@@ -108,8 +110,72 @@ void testChildConfigurationElement() throws BeanConfigurationException {
...
+        Xpp3Dom config = toConfig("<artifact implementation=\"MissingArtifact\"><name>missing</name></artifact>");
+
+        DefaultBeanConfigurationRequest request = new DefaultBeanConfigurationRequest();
+        request.setBean(bean).setConfiguration(config);
+
+        BeanConfigurationException e =
+                assertThrows(BeanConfigurationException.class, () -> configurator.configureBean(request));
+        assertEquals(
+                "Cannot find permitted subclass 'MissingArtifact' for sealed type " + SealedArtifact.class.getName(),
+                e.getMessage());
```
gnodet: Using `assertEquals` on the full message can be brittle if `BeanConfigurationException` changes how it wraps the cause. A `contains` check is more resilient:

```suggestion
        assertTrue(
                e.getMessage().contains("Cannot find permitted subclass 'MissingArtifact' for sealed type "
                        + SealedArtifact.class.getName()));
```

impl/maven-core/src/test/java/org/apache/maven/configuration/DefaultBeanConfiguratorTest.java
```diff
@@ -108,8 +110,72 @@ void testChildConfigurationElement() throws BeanConfigurationException {
...
+    public static final class LocalArtifact implements SealedArtifact {
+
+        String name;
+    }
+
+    public static final class RemoteArtifact implements SealedArtifact {
+
+        String url;
+    }
 }
```
gnodet, line 213: Consider adding a test for the **ambiguity** code path (two permitted subclasses sharing the same simple name). This also requires adding the helper types:

```suggestion
    }

    @Test
    void testSealedTypeAmbiguousSimpleNameThrowsError() {
        AmbiguousBean bean = new AmbiguousBean();

        Xpp3Dom config = toConfig("<value implementation=\"Ambiguous\"/>");

        DefaultBeanConfigurationRequest request = new DefaultBeanConfigurationRequest();
        request.setBean(bean).setConfiguration(config);

        BeanConfigurationException e =
                assertThrows(BeanConfigurationException.class, () -> configurator.configureBean(request));
        assertTrue(e.getMessage().contains("is ambiguous for sealed type"));
    }

    static class SomeBean {

        File file;
    }

    static class SealedBean {

        SealedArtifact artifact;
    }

    static class AmbiguousBean {

        AmbiguousSealedType value;
    }

    public sealed interface SealedArtifact permits LocalArtifact, RemoteArtifact {}

    public sealed interface AmbiguousSealedType permits Holder1.Ambiguous, Holder2.Ambiguous {}

    public static final class LocalArtifact implements SealedArtifact {

        String name;
    }

    public static final class RemoteArtifact implements SealedArtifact {

        String url;
    }

    public static final class Holder1 {

        public static final class Ambiguous implements AmbiguousSealedType {}
    }

    public static final class Holder2 {

        public static final class Ambiguous implements AmbiguousSealedType {}
    }
}
```

gnodet (commented):
_Claude Code on behalf of Guillaume Nodet_ — additional inline suggestions with code blocks.

gnodet (approved):
All feedback from the previous review has been addressed in the follow-up commit:

- Ambiguity exception now includes the `cause` for consistent diagnostic context
- Control flow reordered: `isEmpty()` check before `sort()`
- Error message assertions use `contains()` instead of exact match
- Ambiguity test added with `Holder1.Ambiguous` / `Holder2.Ambiguous` types
- `assertTrue` import added

The sealed-type resolution logic is clean, backward-compatible, and well-tested (4 unit tests + integration test). LGTM.

Will-thom:
Addressed the review feedback in a follow-up commit:

- moved the empty-match check before sorting;
- preserved the original cause for the ambiguous implementation-hint error;
- relaxed the exact message assertion to a contains check;
- added coverage for ambiguous permitted subclasses with the same simple name.

Focused validation passed locally:

```sh
mvn -Dmaven.repo.local=..\maven-12117-m2 -Drat.skip=true -Dsurefire.failIfNoSpecifiedTests=false -pl impl/maven-core -am -Dtest=DefaultBeanConfiguratorTest test
```

Result: BUILD SUCCESS, `DefaultBeanConfiguratorTest` ran 7 tests with 0 failures/errors.

PR #12136, opened 2026-05-20
Fix multiline profile activation conditions

### LINKED ISSUE #11882: Maven 4 profile activation by <condition> issues

Will-thom:
Fixes apache/maven#11882.

This updates profile activation condition tokenization so logical operators remain valid when separated by line breaks or other whitespace, such as CDATA-formatted conditions.

Changes:
- Treat all Java whitespace as token separators in activation conditions.
- Tokenize logical && and || operators explicitly.
- Add parser/profile activator coverage for multiline and mixed-whitespace logical expressions.
- Add a model-building fixture covering multiline CDATA activation conditions.

Verification:
- `mvn -pl impl/maven-impl -am '-Dtest=ConditionParserTest,ConditionProfileActivatorTest,ComplexActivationTest' '-Dsurefire.failIfNoSpecifiedTests=false' test`
  - Tests run: 62, Failures: 0, Errors: 0, Skipped: 3
- `git diff --check`

Additional local validation notes:
- Full IT suite was attempted locally. On Windows, remaining failure was environment-specific symlink privilege in `MavenITmng8400CanonicalMavenHomeTest`.
- Docker/Linux validation was also attempted; unrelated IT failures were triaged/rerun where practical, but a final clean full run was killed due local disk/memory limits. The tests directly covering this change pass consistently.


impl/maven-impl/src/main/java/org/apache/maven/impl/model/profile/ConditionParser.java
```diff
@@ -129,18 +129,30 @@ private List<String> tokenize(String expression) {
...
-                if (c != ' ') {
+                if (!Character.isWhitespace(c)) {
                     if ((c == '>' || c == '<' || c == '=' || c == '!')
                             && i + 1 < expression.length()
                             && expression.charAt(i + 1) == '=') {
                         tokens.add(c + "=");
                         i++; // Skip the next character
+                    } else if ((c == '&' || c == '|') && i + 1 < expression.length() && expression.charAt(i + 1) == c) {
+                        tokens.add(String.valueOf(c) + c);
+                        i++; // Skip the next character
```
gnodet, line 158: Nit: the existing code on line 151 builds the two-char operator with `c + "="` (char + String). Here the order is reversed to `String.valueOf(c) + c`, which works but is inconsistent. Consider aligning the style:

```suggestion
                    } else if ((c == '&' || c == '|') && i + 1 < expression.length() && expression.charAt(i + 1) == c) {
                        tokens.add("" + c + c);
                        i++; // Skip the next character
```

Or even simpler, use the same `c + ""` pattern that the `>=`/`<=` branch above uses (`c + "="`). That said, this is purely cosmetic -- feel free to ignore.

gnodet, line 158: Minor: if someone writes a lone `&` or `|` (without doubling), it falls through to the `else` branch and gets added as a single-character token, which will then cause `parseExpression()` to throw `"Unexpected tokens after end of expression"`. That error message isn't very helpful for diagnosing a typo. Not a blocker -- the old code didn't support these characters at all -- but worth noting for a future improvement.

gnodet (commented):

Clean fix that addresses the root cause: the tokenizer was only treating `' '` as whitespace (missing `\r`, `\n`, `\t`) and wasn't recognizing `&` / `|` as token-separator characters, so multiline conditions with `&&` / `||` split across lines would fail to parse.

The fix is correct and the test coverage is good (unit-level tokenizer/activator tests plus a model-building integration test with a CDATA multiline condition).

One minor nit below, otherwise LGTM.

Will-thom:
Addressed the operator construction style nit in the latest commit (`9ac5389c7c`). I left the lone `&` / `|` diagnostic behavior unchanged for now since it is non-blocking and can be handled separately if desired.

Validation:
- `git diff --check`
- `mvn -pl impl/maven-impl -am '-Dtest=ConditionParserTest,ConditionProfileActivatorTest,ComplexActivationTest' '-Dsurefire.failIfNoSpecifiedTests=false' test` — Tests run: 62, Failures: 0, Errors: 0, Skipped: 3

gnodet (approved):
Thanks for addressing the style nit. The `"" + c + c` form is now consistent with the `c + "="` pattern above.

The fix is solid: `Character.isWhitespace(c)` replaces the hand-enumerated whitespace check, `&` and `|` are properly recognized as delimiters, and `&&`/`||` are tokenized symmetrically. Test coverage is thorough (unit-level tokenizer/activator tests plus the CDATA multiline integration test).

LGTM.

PR #12145, opened 2026-05-23
Avoid final logger injection in EventSpyDispatcher

### LINKED ISSUE #12115: WARNING: Final field logger in EventSpyDispatcher has been mutated - JDK 26
### Affected version

3.10.0

### Bug description

Reproduce:

```
mvn -V help:help
Apache Maven 3.10.0-SNAPSHOT (bf1ea07b40daa020eaca038148a9a39c72c2c4e7)
Maven home: /Users/slawomir.jaranowski/.sdkman/candidates/maven/3.10.x
Java version: 26.0.1, vendor: Eclipse Adoptium, runtime: /Users/slawomir.jaranowski/.sdkman/candidates/java/26.0.1-tem
Default locale: en_PL, platform encoding: UTF-8, time zone: Europe/Warsaw
OS name: "mac os x", version: "26.3.1", arch: "aarch64", family: "mac"
WARNING: Final field logger in class org.apache.maven.eventspy.internal.EventSpyDispatcher has been mutated reflectively by class org.eclipse.sisu.bean.BeanPropertyField in unnamed module @3c72f59f (file:/Users/slawomir.jaranowski/.sdkman/candidates/maven/3.10.x/lib/org.eclipse.sisu.inject-1.0.0.jar)
WARNING: Use --enable-final-field-mutation=ALL-UNNAMED to avoid a warning

```

Will-thom:
Fixes #12115

This changes EventSpyDispatcher to use a static logger so Sisu does not mutate a final instance logger field reflectively on newer JDKs. It also adds a small regression test that guards against reintroducing a final instance Logger field.

Validation:
- mvn -B -pl maven-core -Dtest=EventSpyDispatcherTest test
- mvn -B -pl maven-core test

### DIFF
--- a/maven-core/src/test/java/org/apache/maven/eventspy/internal/EventSpyDispatcherTest.java
+++ b/maven-core/src/test/java/org/apache/maven/eventspy/internal/EventSpyDispatcherTest.java
@@ -1,0 +1,39 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.maven.eventspy.internal;
+
+import java.lang.reflect.Modifier;
+import java.util.Arrays;
+
+import org.junit.jupiter.api.Test;
+import org.slf4j.Logger;
+
+import static org.junit.jupiter.api.Assertions.assertFalse;
+
+class EventSpyDispatcherTest {
+
+    @Test
+    void doesNotExposeFinalInstanceLoggerFieldForInjection() {
+        boolean hasFinalInstanceLogger = Arrays.stream(EventSpyDispatcher.class.getDeclaredFields())
+                .filter(field -> Logger.class.equals(field.getType()))
+                .anyMatch(field -> !Modifier.isStatic(field.getModifiers()) && Modifier.isFinal(field.getModifiers()));
+
+        assertFalse(hasFinalInstanceLogger);
+    }
+}
--- a/maven-core/src/main/java/org/apache/maven/eventspy/internal/EventSpyDispatcher.java
+++ b/maven-core/src/main/java/org/apache/maven/eventspy/internal/EventSpyDispatcher.java
@@ -38,7 +38,7 @@ import org.slf4j.LoggerFactory;
 @Named
 public class EventSpyDispatcher {
 
-    private final Logger logger = LoggerFactory.getLogger(getClass());
+    private static final Logger LOGGER = LoggerFactory.getLogger(EventSpyDispatcher.class);
 
     private final List<EventSpy> eventSpies;
 
@@ -107,10 +107,10 @@ public class EventSpyDispatcher {
     private void logError(String action, Throwable e, EventSpy spy) {
         String msg = "Failed to " + action + " spy " + spy.getClass().getName() + ": " + e.getMessage();
 
-        if (logger.isDebugEnabled()) {
-            logger.warn(msg, e);
+        if (LOGGER.isDebugEnabled()) {
+            LOGGER.warn(msg, e);
         } else {
-            logger.warn(msg);
+            LOGGER.warn(msg);
         }
     }
 }


gnodet (commented):
The fix itself is correct and follows the established convention in the codebase (`private static final Logger LOGGER`). This avoids Sisu mutating the `final` instance field on JDK 26+.

One thought on the test -- see inline comment.

maven-core/src/test/java/org/apache/maven/eventspy/internal/EventSpyDispatcherTest.java
```diff
@@ -0,0 +1,39 @@
...
+class EventSpyDispatcherTest {
+
+    @Test
+    void doesNotExposeFinalInstanceLoggerFieldForInjection() {
+        boolean hasFinalInstanceLogger = Arrays.stream(EventSpyDispatcher.class.getDeclaredFields())
+                .filter(field -> Logger.class.equals(field.getType()))
+                .anyMatch(field -> !Modifier.isStatic(field.getModifiers()) && Modifier.isFinal(field.getModifiers()));
+
+        assertFalse(hasFinalInstanceLogger);
+    }
```
gnodet, line 38: This test guards against reintroducing a `final` instance `Logger` field, but it would not catch someone reintroducing a *non-final* instance `Logger` field (which would still be subject to Sisu injection and equally problematic). A simpler and broader guard would be to assert that **no** instance `Logger` field exists at all:

```suggestion
        boolean hasInstanceLogger = Arrays.stream(EventSpyDispatcher.class.getDeclaredFields())
                .filter(field -> Logger.class.equals(field.getType()))
                .anyMatch(field -> !Modifier.isStatic(field.getModifiers()));

        assertFalse(hasInstanceLogger, "Logger field should be static to avoid Sisu reflective mutation");
    }
```

slawekjaranowski:
I'm not sure about we finally resolve a problem when  Sisu mutating the final instance field ...

we have many other places in code with final without static.

slawekjaranowski:
@Will-thom please always preserve last paragraph from PR template

> To make clear that you license your contribution under the Apache License Version 2.0, January 2004 you have to acknowledge this by using the following check-box.
...

Please also consider to sign - https://www.apache.org/licenses/icla.pdf

maven-core/src/test/java/org/apache/maven/eventspy/internal/EventSpyDispatcherTest.java
slawekjaranowski: IMHO testing final presence is a little redundant ...

cstamas:
I would rather stop sisu/plexus (and i think this is rather plexus) doing this, AFAIR there was plan to inject Logger instances that was never finalized as Plexus Loggers are still in play (goal was to fully switch to slf4j)

slawekjaranowski:
In this case logger is slf4j ... but in some case sisu try to update such filed

cstamas:
See
https://github.com/eclipse-sisu/sisu-project/blob/a2d1d99282f37d366e7fe417c8c0fea4897c5704/org.eclipse.sisu.plexus/src/main/java/org/eclipse/sisu/plexus/PlexusLifecycleManager.java#L105

But something is off then, as in Maven Core of 3.10.x we should not have Plexus Components anymore, need to look more...

cstamas:
@Will-thom can you provide a reproducer for refd issue?

cstamas:
Also, this issue reminds me of https://github.com/apache/maven/commit/dea5f14a85e26291f654a4c4a29a58210293628b

There, we _thought_ we migrated the component to Sisu (from Plexus) but in fact, we did not. So, if something still defines `EventSpyDispatcher` as Plexus component (instead of Sisu JSR330 component), then yes, this could happen. Otherwise, I don't know what exactly happens here...

gnodet (approved):
Clean, minimal fix. Changing the instance logger to a static logger is the standard SLF4J idiom and avoids Sisu's reflective mutation of a final field on newer JDKs. The regression test using reflection to assert no final instance Logger field exists is a nice defensive guard. LGTM.

maven-core/src/test/java/org/apache/maven/eventspy/internal/EventSpyDispatcherTest.java
elharo: I don't feel strongly, but it does feel like this is testing too much of the internal implementation details.

gnodet (commented):
The code change itself — converting the instance logger to a static logger — follows the standard SLF4J idiom (`private static final Logger LOGGER`) and is a correct best-practice improvement. However, several concerns should be addressed before this PR can move forward:

**1. Underlying issue closed as not reproducible (high)**
The referenced issue #12115 was closed as NOT_PLANNED — neither the reporter nor maintainers could reproduce the JDK 26 warning after rebuilding from source. Framing this change as a fix for #12115 is misleading.

**2. Incomplete scope (medium)**
This fix addresses only `EventSpyDispatcher`, but there are approximately 40 other files using the identical `private final Logger logger = LoggerFactory.getLogger(getClass())` pattern. If Sisu reflective mutation were the real problem, all of those files would be equally affected.

**3. Test concerns (high)**
The test at `EventSpyDispatcherTest` checks only for a `final` instance Logger field but would not catch a non-final instance Logger, which would be equally subject to Sisu reflective mutation. Three reviewers (gnodet, slawekjaranowski, elharo) independently flagged this test as testing implementation details rather than behavior. Since `EventSpyDispatcher` is `@Singleton`, the difference between a static and instance logger is a convention concern, not a runtime correctness issue.

**4. Missing license acknowledgment (low)**
The PR description is missing the Apache License acknowledgment paragraph from the PR template, which slawekjaranowski requested be preserved.

As maintainer cstamas noted, the root cause may be that EventSpyDispatcher is still being treated as a Plexus component by Sisu's `PlexusLifecycleManager` — if so, the proper fix would be in Sisu/Plexus, not in application code.

gnodet (commented):
The static logger change itself follows standard SLF4J conventions and is technically safe — `EventSpyDispatcher` is `@Singleton`, so there is no behavioral difference between a static and instance logger.

However, there are some concerns worth considering:

### Observations

**Underlying issue closed (high):** The referenced issue #12115 was closed as NOT_PLANNED — the JDK 26 warning could not be reproduced by any maintainer. Maintainer @cstamas identified the root cause as likely being in Sisu/Plexus (`PlexusLifecycleManager` processing `EventSpyDispatcher` as a Plexus component), suggesting the fix belongs there rather than in application code.

**Incomplete scope (medium):** There are ~54 files in the codebase using the identical `private final Logger logger = LoggerFactory.getLogger(getClass())` pattern. Fixing only `EventSpyDispatcher` is inconsistent — if Sisu reflective mutation of final fields is the problem, all those files are equally affected.

**Test approach (medium):** The reflection-based test checks field modifiers (final + non-static Logger) rather than testing behavior. Three reviewers (@gnodet, @slawekjaranowski, @elharo) independently flagged this concern.

### 📋 PR Metadata

| Milestone | _(none)_ | `3.10.0` |

Note: On `maven-3.9.x`, `EventSpyDispatcher` uses Plexus `@Requirement private Logger logger` (non-final, Plexus-injected), so this specific issue does not apply there. A forward-port to `master` would be needed if accepted.

cstamas:
I agree with @gnodet this PR misses the point, as Plexus DI is not used anymore in Maven 3.10 (it uses Sisu DI/JSR330).

PR #12146, opened 2026-05-24
Handle missing package metadata in model ids

### LINKED ISSUE #8403: [MNG-6932] NullPointerException if no Package is defined for the current class
**[Bence Sipka](https://issues.apache.org/jira/secure/ViewProfile.jspa?name=sipka)** opened **[MNG-6932](https://issues.apache.org/jira/browse/MNG-6932?redirect=false)** and commented

When the maven-resolver and related classes are used with a `ClassLoader` that doesn't define a java.lang.Package object for the loaded classes, the library throws a `NullPointerException`.

Scenario: I'm using the maven-resolver classes to resolver artifacts from various repositories. I package the app, and deploy it to a custom environment where the `ClassLoader` for the app doesn't define a `Package` for the loaded classes. In these cases, when the maven-resolver library calls `this.getClass().getPackage().getImplementationVersion()` then a `NullPointerException` will occurr.

 The following classes are affected:
`DefaultSuperPomProvider.getSuperModel():81`
`DefaultReportingConverter.<init>:56`

Code around the locations:

```java
String modelId = "org.apache.maven:maven-model-builder:"
    + this.getClass().getPackage().getImplementationVersion() + ":super-pom";
InputSource inputSource = new InputSource();
inputSource.setModelId( modelId );
```

This exception makes the library unportable to different environments.

Workaround:

Implement a custom variant of the above classes that circumvent these parts of the code.

Possible solution:

Null checks, and hard coding the version number.

---

**Affects:** 3.6.3

**Issue Links:**
- [MNG-6593](https://issues.apache.org/jira/browse/MNG-6593) track input location for super-pom
 (_**"is caused by"**_)
- [MNG-6600](https://issues.apache.org/jira/browse/MNG-6600) add input location tracking for default lifecycle bindings executions
 (_**"is caused by"**_)
- [MNG-6960](https://issues.apache.org/jira/browse/MNG-6960) Use RuntimeInformation instead of reading properties
 (_**"depends upon"**_)

0 votes, 5 watchers

Will-thom:
Fixes #8403.

This avoids NullPointerException when Maven code builds internal model ids in environments where Class#getPackage() returns null, while preserving the previous model id contents when package metadata exists but the implementation version is absent. The fallback is applied to the remaining implementation-version based model ids in the current codebase.

Validation:
- mvn -B -am -pl compat/maven-model-builder,impl/maven-impl,impl/maven-core -Dtest=DefaultReportingConverterTest,DefaultSuperPomProviderTest,DefaultLifecycleRegistryTest -Dsurefire.failIfNoSpecifiedTests=false test

### DIFF
--- a/compat/maven-model-builder/src/test/java/org/apache/maven/model/plugin/DefaultReportingConverterTest.java
+++ b/compat/maven-model-builder/src/test/java/org/apache/maven/model/plugin/DefaultReportingConverterTest.java
@@ -1,0 +1,31 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.maven.model.plugin;
+
+import org.junit.jupiter.api.Test;
+
+import static org.junit.jupiter.api.Assertions.assertNull;
+
+class DefaultReportingConverterTest {
+
+    @Test
+    void returnsNullImplementationVersionWhenClassHasNoPackage() {
+        assertNull(DefaultReportingConverter.getImplementationVersion(int.class));
+    }
+}
--- a/impl/maven-core/src/main/java/org/apache/maven/internal/impl/DefaultLifecycleRegistry.java
+++ b/impl/maven-core/src/main/java/org/apache/maven/internal/impl/DefaultLifecycleRegistry.java
@@ -88,7 +88,7 @@ public class DefaultLifecycleRegistry implements LifecycleRegistry {
     private static final String MAVEN_PLUGINS = "org.apache.maven.plugins:";
 
     public static final String DEFAULT_LIFECYCLE_MODELID = "org.apache.maven:maven-core:"
-            + DefaultLifecycleRegistry.class.getPackage().getImplementationVersion()
+            + getImplementationVersion(DefaultLifecycleRegistry.class)
             + ":default-lifecycle-bindings";
 
     public static final InputLocation DEFAULT_LIFECYCLE_INPUT_LOCATION =
@@ -120,6 +120,11 @@ public class DefaultLifecycleRegistry implements LifecycleRegistry {
                 }
             });
         }
+    }
+
+    static String getImplementationVersion(Class<?> clazz) {
+        Package packageInfo = clazz.getPackage();
+        return packageInfo != null ? packageInfo.getImplementationVersion() : null;
     }
 
     @Override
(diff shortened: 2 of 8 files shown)


gnodet (commented):
Thanks for addressing this long-standing issue (MNG-6932). The fix is correct in principle -- guarding against `Class#getPackage()` returning `null` prevents the `NullPointerException`.

A few observations below.

impl/maven-core/src/main/java/org/apache/maven/internal/impl/DefaultLifecycleRegistry.java
```diff
@@ -88,7 +88,7 @@ public class DefaultLifecycleRegistry implements LifecycleRegistry {
     private static final String MAVEN_PLUGINS = "org.apache.maven.plugins:";
 
     public static final String DEFAULT_LIFECYCLE_MODELID = "org.apache.maven:maven-core:"
-            + DefaultLifecycleRegistry.class.getPackage().getImplementationVersion()
+            + getImplementationVersion(DefaultLifecycleRegistry.class)
```
gnodet, line 91: `DEFAULT_LIFECYCLE_MODELID` is a `static final` constant whose value feeds directly into `DEFAULT_LIFECYCLE_INPUT_LOCATION` on the next line. Before this PR, when `getPackage()` returned `null`, the class failed at load time with an `ExceptionInInitializerError` (wrapping an NPE).

With this PR, the class loads successfully, but the model id will contain the literal string `"null"` via string concatenation (e.g. `"org.apache.maven:maven-core:null:default-lifecycle-bindings"`), which may silently produce confusing diagnostics later. The other three call sites have the same characteristic.

It might be worth returning a fallback like `"unknown"` instead of `null`, so the concatenated model id remains parseable/debuggable:

```suggestion
            + getImplementationVersion(DefaultLifecycleRegistry.class)
```

(the suggestion is a no-op to keep this visible -- the actual change would be in `getImplementationVersion` to return `"unknown"` instead of `null`)

impl/maven-core/src/main/java/org/apache/maven/internal/impl/DefaultLifecycleRegistry.java
```diff
@@ -122,6 +122,11 @@ public DefaultLifecycleRegistry(List<LifecycleProvider> providers) {
         }
     }
 
+    static String getImplementationVersion(Class<?> clazz) {
+        Package packageInfo = clazz.getPackage();
+        return packageInfo != null ? packageInfo.getImplementationVersion() : null;
+    }
```
gnodet, line 128: The identical `getImplementationVersion` helper is copy-pasted verbatim into four separate classes (`DefaultReportingConverter`, both `DefaultSuperPomProvider` variants, and here). Consider extracting it into a shared utility to avoid the duplication. If the compat module intentionally avoids depending on impl internals, the duplication there may be acceptable, but the two `impl/` modules could still share one copy.

compat/maven-model-builder/src/test/java/org/apache/maven/model/plugin/DefaultReportingConverterTest.java
```diff
@@ -0,0 +1,31 @@
...
+
+import org.junit.jupiter.api.Test;
+
+import static org.junit.jupiter.api.Assertions.assertNull;
+
+class DefaultReportingConverterTest {
+
+    @Test
+    void returnsNullImplementationVersionWhenClassHasNoPackage() {
+        assertNull(DefaultReportingConverter.getImplementationVersion(int.class));
```
gnodet, line 29: Using `int.class` (a primitive type whose `getPackage()` returns `null`) is a clever way to trigger the null-package path. However, it would strengthen the test to also verify the case where `getPackage()` is non-null but `getImplementationVersion()` itself returns `null` -- which is the far more common scenario in IDE and test-classpath environments (classes loaded without a JAR manifest). A regular class from the test classpath would exercise that path.

Also, all four test classes are identical except for the class under test -- a parameterized test or a shared test helper could reduce the boilerplate.

gnodet (approved):
Simple and correct null-safety fix. Wrapping `clazz.getPackage().getImplementationVersion()` in a null-safe helper prevents the NPE in environments where `Class.getPackage()` returns null (e.g., certain classloaders). Applied consistently to all 4 affected locations. The helper methods are package-private, which is right for testability without polluting the API. LGTM.

PR #12157, opened 2026-05-26
Fix mvnup plugin upgrade for plugins inherited in build/plugins

gnodet:
## Summary

- When a remote parent POM declares a plugin with an explicit version in `<build><plugins>` (not `<pluginManagement>`), `mvnup`'s plugin upgrade strategy now adds the plugin to **both** `<pluginManagement>` and `<build><plugins>` in the child project
- Previously, only a `<pluginManagement>` entry was added, which is insufficient because the parent's `<plugins>` version takes precedence over the child's `<pluginManagement>`
- This fixes the scenario described in the ratis-hadoop-projects issue where `maven-enforcer-plugin:1.4.1` inherited from a parent's `<plugins>` section caused `NoSuchMethodError` in Maven 4

## Test plan

- [x] New test `shouldAddDirectBuildPluginsOverrideForInheritedPlugins` verifies both `pluginManagement` and `build/plugins` entries are added for inherited plugins
- [x] New test `shouldNotDuplicatePluginInBuildPluginsWhenAlreadyDeclared` verifies no duplicates when plugin is already locally declared
- [x] All 29 existing + new PluginUpgradeStrategy tests pass

### DIFF
--- a/impl/maven-cli/src/main/java/org/apache/maven/cling/invoker/mvnup/goals/PluginUpgradeStrategy.java
+++ b/impl/maven-cli/src/main/java/org/apache/maven/cling/invoker/mvnup/goals/PluginUpgradeStrategy.java
@@ -131,10 +131,9 @@ public class PluginUpgradeStrategy extends AbstractUpgradeStrategy {
             Path tempDir = createTempProjectStructure(context, pomMap);
 
             // Phase 2: For each POM, build effective model using the session and analyze plugins
-            Map<Path, Set<String>> pluginsNeedingManagement =
-                    analyzePluginsUsingEffectiveModels(context, pomMap, tempDir);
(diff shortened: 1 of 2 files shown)


gnodet:

Closing in favor of a more precise implementation that only adds direct `<plugins>` entries when the parent's version is truly locked (explicit version in `<plugins>`, not from `<pluginManagement>`). The previous approach added unnecessary `<plugins>` entries even when a `<pluginManagement>` override alone would suffice.

gnodet:

Superseded by #12165, which uses a more precise approach — comparing effective versions in `build/plugins` vs `build/pluginManagement/plugins` upfront to determine the correct override location, avoiding unnecessary `<plugins>` entries.

PR #12160, opened 2026-05-26
Remove invalid combine.self and combine.children attributes in mvnup

gnodet:
## Summary

- **Remove** invalid `combine.self` and `combine.children` attribute values entirely instead of converting them to `merge`, since Maven 3 silently ignored invalid values — removing preserves actual Maven 3 behavior
- Detect **any** invalid value against the full set of valid values (`combine.self`: `override`/`merge`/`remove`; `combine.children`: `append`/`merge`), not just specific ones like `append` or `override`
- Collect matching elements to a list before mutating, replacing the `.peek().count()` pattern which could miss occurrences

## Test plan

- [x] Added tests for removing `combine.self="append"` and arbitrary invalid values
- [x] Added tests for removing `combine.children="override"` and arbitrary invalid values
- [x] Added tests verifying valid values (`override`, `merge`, `remove` for self; `append`, `merge` for children) are preserved
- [x] Added tests for invalid attributes in profile plugin configurations
- [x] Added tests for all occurrences across pluginManagement, plugins, and profiles
- [x] All 31 CompatibilityFixStrategyTest tests pass

### DIFF
--- a/impl/maven-cli/src/main/java/org/apache/maven/cling/invoker/mvnup/goals/CompatibilityFixStrategy.java
+++ b/impl/maven-cli/src/main/java/org/apache/maven/cling/invoker/mvnup/goals/CompatibilityFixStrategy.java
@@ -77,6 +77,10 @@ public class CompatibilityFixStrategy extends AbstractUpgradeStrategy {
 
     private static final Pattern EXPRESSION_PATTERN = Pattern.compile("\\$\\{([^}]+)}");
 
+    private static final Set<String> VALID_COMBINE_SELF_VALUES = Set.of(COMBINE_OVERRIDE, COMBINE_MERGE, "remove");
+
+    private static final Set<String> VALID_COMBINE_CHILDREN_VALUES = Set.of(COMBINE_APPEND, COMBINE_MERGE);
+
     @Override
     public boolean isApplicable(UpgradeContext context) {
         UpgradeOptions options = getOptions(context);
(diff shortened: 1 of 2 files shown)


gnodet:
TODO: Replace hardcoded `"remove"` with `COMBINE_REMOVE` constant once [maveniverse/domtrip#240](https://github.com/maveniverse/domtrip/pull/240) is released.

impl/maven-cli/src/main/java/org/apache/maven/cling/invoker/mvnup/goals/CompatibilityFixStrategy.java
```diff
@@ -164,44 +168,44 @@ public UpgradeResult doApply(UpgradeContext context, Map<Path, Document> pomMap)
...
-                            + "' in " + element.name());
-                })
-                .count();
-        fixed |= fixedCombineChildrenCount > 0;
+        List<Element> invalidElements = findElementsWithInvalidAttribute(
+                        root, COMBINE_CHILDREN, VALID_COMBINE_CHILDREN_VALUES)
+                .toList();
 
-        return fixed;
+        for (Element element : invalidElements) {
```
rmannibucau, line 181: do we want to fail and only run on a valid pom? (to encourage people to review instead of doing a blind silent migration which can just not be desired)?

Bukama: Imo the tool is to also fix the poms for users and show them by printing it out

Bukama (approved):

impl/maven-cli/src/main/java/org/apache/maven/cling/invoker/mvnup/goals/CompatibilityFixStrategy.java
gnodet: Maven 4 already validates these at runtime: `DefaultModelValidator.validateXmlNode()` (lines 815-852) raises a `Severity.ERROR` for any unsupported `combine.self` or `combine.children` value. So invalid POMs will fail the build.

The whole point of mvnup is to fix these proactively — if someone has `combine.self="append"` (which Maven 3 silently ignored), mvnup removes it *before* Maven 4 rejects it. Running mvnup on an invalid POM and having it fix the problem is exactly the expected workflow. Failing and telling the user "your POM is invalid, go fix it yourself" would defeat the purpose.

Issue #12086, opened 2026-05-18
Transitive dependency POMs with uninterpolated repository IDs cause build failures


gnodet:
## Description

Maven 4's stricter validation rejects repositories with uninterpolated property expressions in their IDs (e.g., `${eclipseP2RepoId}`). When such repositories come from transitive dependency POMs — where the defining property is not available — builds fail.

PR #12050 addressed a similar issue for **dependencies** with uninterpolated expressions by filtering them in `DefaultArtifactDescriptorReader`. However, the same problem exists for **repositories** from transitive dependency POMs, which are not currently filtered.

## Steps to Reproduce

1. Build `apache/opennlp-sandbox` with Maven 4
2. The `caseditor-corpus-server-plugin` module fails during `maven-remote-resources-plugin:3.3.0:process`

## Error

```
Caused by: java.lang.IllegalArgumentException: Invalid RemoteRepositories: [central (https://repo.maven.apache.org/maven2, default, releases), ${eclipseP2RepoId} (https://download.eclipse.org/releases/2022-09/, default, releases+snapshots), apache.snapshots (https://repository.apache.org/snapshots, default, ...)]
    Suppressed: java.lang.IllegalArgumentException: Not fully interpolated remote repository ${eclipseP2RepoId} (https://download.eclipse.org/releases/2022-09/, default, releases+snapshots)
```

## Analysis

- The `${eclipseP2RepoId}` repository is NOT defined in the opennlp-sandbox project — it comes from a transitive dependency POM
- The property `eclipseP2RepoId` is defined in the original project context but not available when the POM is consumed as a transitive dependency
- Maven 3 silently accepted uninterpolated repository IDs; Maven 4 rejects them
- PR #12050 filters dependencies with uninterpolated expressions in `DefaultArtifactDescriptorReader`, but does NOT filter repositories

## Proposed Fix

Extend the filtering from PR #12050 to also cover repositories with uninterpolated IDs in the `ArtifactDescriptorResult`. In `DefaultArtifactDescriptorReader`, after `delegate.populateResult()`, filter out repositories where the ID contains `${`:

```java
result.getRepositories().removeIf(repo -> {
    if (repo.getId() != null && repo.getId().contains("${")) {
        logger.debug("Filtered repository with uninterpolated ID: {}", repo);
        return true;
    }
    return false;
});
```

## Related

- PR #12050 — Filters transitive dependencies with uninterpolated expressions
[...]

PR #12175, opened 2026-05-28
Move repository expression validation from raw to effective model

gnodet:
## Summary

- Skip uninterpolated expression checks for root-level repositories during raw model validation, because parent POM properties are not yet available at that stage
- Add the same expression checks to effective model validation (`validate20EffectiveRepository`), where parent properties have been merged and full interpolation is complete
- Truly unresolvable expressions are still caught as errors, but after all property sources have been resolved

## Context

Raw model validation runs before the parent POM is read (`doReadRawModel` → `validateRawModel` happens before `readEffectiveModel` → `readParent`). This means properties defined in parent POMs (e.g., `${eclipseP2RepoId}`) are not available during raw validation, causing false positives.

Profile and distributionManagement repositories already skipped expression checks for the same reason. This change extends that to root-level `repositories` and `pluginRepositories`.

Fixes: MNG-15012, MNG-14390, MNG-10104, MNG-13299, MNG-14733

## Test plan

- [x] Updated raw validation tests to reflect skipped expression checks
- [x] Added effective model validation tests to verify unresolvable expressions are still caught
- [x] All 470 maven-impl tests pass

_Generated with Claude Code_

### DIFF
--- a/impl/maven-impl/src/main/java/org/apache/maven/impl/model/DefaultModelValidator.java
+++ b/impl/maven-impl/src/main/java/org/apache/maven/impl/model/DefaultModelValidator.java
@@ -668,7 +668,7 @@ public class DefaultModelValidator implements ModelValidator {
 
         if (validationLevel > VALIDATION_LEVEL_MINIMAL) {
             validateRawRepositories(
-                    problems, model.getRepositories(), "repositories.repository.", EMPTY, validationLevel, false);
+                    problems, model.getRepositories(), "repositories.repository.", EMPTY, validationLevel, true);
 
             validateRawRepositories(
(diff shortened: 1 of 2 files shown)


gnodet:
Closing in favor of #12086 (filter transitive repos with uninterpolated expressions), which already fixes the 6 affected projects.

Repositories are intentionally available from the raw model — they're used to download parents and extensions before the effective model exists. Moving validation to the effective model stage would mean silently using potentially broken repos, then reporting the error too late. Filtering invalid transitive repos (#12086) is the better approach: those repos genuinely don't belong in the consumer's context.

PR #12180, opened 2026-05-28
Bump default maven-dependency-plugin to 3.11.0

brunoborges:
### Summary

Bumps the Maven 3.9.x Super POM default for `maven-dependency-plugin` from `3.7.0` to `3.11.0`.

### Rationale

The current default resolves short-prefix invocations such as `mvn dependency:add` to `maven-dependency-plugin:3.7.0`. Newer dependency plugin goals, including `dependency:add`, require a newer plugin version, so updating the Super POM default allows the short-form invocation to work without requiring users to spell out the fully qualified plugin coordinate.

### Testing

- `git diff --check`
- `mvn -q -pl maven-model-builder -am -DskipTests package`

### Checklist

- [x] My pull request addresses just one issue, without pulling in other changes.
- [x] I wrote a pull request description that is detailed enough to understand what the pull request does, how, and why.
- [x] Each commit in the pull request has a meaningful subject line and body.
- [x] I ran `mvn verify` or an appropriate targeted build.
- [x] I hereby declare this contribution to be licenced under the Apache License Version 2.0, January 2004.

### DIFF
--- a/maven-model-builder/src/main/resources/org/apache/maven/model/pom-4.0.0.xml
+++ b/maven-model-builder/src/main/resources/org/apache/maven/model/pom-4.0.0.xml
@@ -79,7 +79,7 @@ under the License.
         </plugin>
         <plugin>
           <artifactId>maven-dependency-plugin</artifactId>
-          <version>3.7.0</version>
+          <version>3.11.0</version>
         </plugin>
         <plugin>
           <artifactId>maven-release-plugin</artifactId>


cstamas (approved):

cstamas:
But please do this for `maven-3.10.x` as well

slawekjaranowski:
@cstamas , @brunoborges 

> But please do this for `maven-3.10.x` as well

We have drop it from super pom 😄 in 3.10.x
- https://github.com/apache/maven/pull/11986

By the way each project should manage version of used plugins

brunoborges:
Clarification after checking the newer maintenance branches: this change is specific to Maven 3.9.x. The maven-3.10.x branch no longer carries these default plugin versions in the Super POM, so there is no corresponding Super POM entry for maven-dependency-plugin to update there.

With Maven 3.10.x, the short form `mvn dependency:add` resolves the `dependency` prefix from plugin metadata to `org.apache.maven.plugins:maven-dependency-plugin`, then resolves the plugin version from repository metadata. Since Maven Central currently advertises `maven-dependency-plugin` `<release>3.11.0</release>`, Maven 3.10.x can resolve the canonical short form to 3.11.0 without a Super POM change, subject to normal metadata freshness (for example, `-U` if a local cache is stale).

slawekjaranowski:
in 3.9 you can also use short version of calling plugin

```
mvn dependency:3.11.0:add
```

cstamas:
Yup, I totally forgot the attic cleanup. All good!

brunoborges:
> in 3.9 you can also use short version of calling plugin
> 
> ```
> mvn dependency:3.11.0:add
> ```

@slawekjaranowski is this specific to 3.9.x or is it supported in older versions as well?

brunoborges:
Ok, I tested and `dependency:3.11.0:add` works on Maven 3.8.8 and later.

cstamas:
Yup, https://issues.apache.org/jira/browse/MNG-7353

brunoborges:
Found a bug in 3.11.0 and we will need a bug fix release there. Moving this PR to draft until 3.11.1 is out

slachiewicz:
Are you still interested in working on this pull request? It has not seen any activity for a while.

If you are still on it, please rebase it on the current `master` and leave a comment. Otherwise, it will be marked stale and automatically closed in about 90 days. You can always reopen or push an update later.

brunoborges:
@slachiewicz I need `maven-dependency-plugin` to be released (v 3.11.1) before this PR goes in.

PR #12236, opened 2026-06-04
[4.0.x] [GH-11772] Fail-fast consumer POM validation for non-4.0.0 model versions

gnodet:
Cherry-pick of #11780

Adaptation for 4.0.x:
- Removed `deployBuildPom()` method (constant not yet on 4.0.x)
- Added missing `MavenException` import
- Added `super("[4.0.0-rc-1,)")` constructor to IT test

### DIFF
--- a/api/maven-api-core/src/main/java/org/apache/maven/api/Constants.java
+++ b/api/maven-api-core/src/main/java/org/apache/maven/api/Constants.java
@@ -490,6 +490,21 @@ public final class Constants {
     public static final String MAVEN_CONSUMER_POM_FLATTEN = "maven.consumer.pom.flatten";
 
     /**
+     * User property for controlling removal of unused managed dependencies during consumer POM flattening.
+     * When set to {@code true} (default), managed dependencies that do not appear in the resolved
+     * dependency tree are removed from the consumer POM to keep it lean. This is important when using
+     * BOMs like Spring Boot or Quarkus that contain hundreds of managed dependency entries.
+     * When set to {@code false}, all managed dependencies are preserved in the consumer POM,
+     * which may be needed in rare cases where downstream consumers override transitive dependency
+     * versions and rely on the original managed dependencies for alignment.
+     *
+     * @since 4.1.0
+     */
+    @Config(type = "java.lang.Boolean", defaultValue = "true")
+    public static final String MAVEN_CONSUMER_POM_REMOVE_UNUSED_MANAGED_DEPENDENCIES =
+            "maven.consumer.pom.removeUnusedManagedDependencies";
+
+    /**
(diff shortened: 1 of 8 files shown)


ascheman:
**Real-world adopter feedback — tested apache/plc4x against a current 4.0.x SNAPSHOT (pre-RC-6).** Posting here since this backport first ships with RC-6.

The fail-fast works as designed: PLC4X's parent POM (model 4.1.0, using `<subprojects>` and build `<sources>` — the latter on this PR's non-strippable list) now stops the very first reactor module on `install`, where rc-5 "succeeded" but silently installed a model-4.1.0 consumer POM — exactly the #11772 poison. Root-module matrix:

| Maven | `mvn -N install` |
|---|---|
| 4.0.0-rc-5 | SUCCESS (but consumer POM stays 4.1.0) |
| 4.0.x-SNAPSHOT | fail-fast ✔ |
| 4.0.x-SNAPSHOT + `maven.consumer.pom.flatten=true` | SUCCESS |

With the flatten property in `.mvn/maven.config` we then ran a full A/B (identical sources, separate local repos, tests on): **rc-5 and the SNAPSHOT are behaviorally identical** — 113 modules, 4058/4058 tests green on both, identical installed file lists (156), identical consumer-POM model distribution per artifact (28× self-contained 4.0.0, 1× 4.0.0 with parent, 2× 4.1.0 with parent for the pom-packaged aggregators). So the fail-fast is the only rc-5 → RC-6 behavior change here, and the suggested remedy is safe on both versions.

Three observations that may be worth acting on before RC-6:

1. **Release notes**: our Maven 4 adoption registry currently tracks ~70–85 GitHub projects declaring the 4.1.0 model ([report](https://aschemaven.github.io/maven-simple-reports/maven4-adoption.html)). Every one of them that installs/deploys a parent POM with non-strippable 4.1.0 features will hard-fail on RC-6 upgrade. A release-notes callout with the three remedies from the error message would soften the landing considerably.

2. **The error text slightly over-promises for pom packaging**: option 1 says enabling flattening will "produce a self-contained **4.0.0** consumer POM", but with flattening enabled the pom-packaged project's *own* consumer POM is still published as model 4.1.0 with `<parent>` and `<subprojects>` (identically on rc-5, so not a regression). Children are correctly flattened and no longer reference it — but *direct* consumers inheriting from such a parent would still hit #11772-style breakage. If "handles what it can" is the intended semantics, maybe adjust the error text and document the residual limitation.

3. As a data point for how adopters need to react: PLC4X is now prepared via apache/plc4x#2647 (one line in `.mvn/maven.config`, verified against rc-5 and the SNAPSHOT).

Happy to re-run this against RC-6 staging when it's cut.

PR #12242, opened 2026-06-06
Fix MojoExtension.beforeEach to use merged model instead of raw parsed model

### LINKED ISSUE #12201: Fix MojoExtension.beforeEach to use proper model
Note: as discussed during review, there's a pre-existing issue in `MojoExtension.beforeEach()` — `alignToBaseDirectory` is called on `tmodel` (the raw parsed POM) instead of `model` (the merged model). This means:
1. If no POM is found, `tmodel` is `null` and `alignToBaseDirectory` will NPE
2. The stored model misses the defaults from the merge (groupId, artifactId, build paths, etc.)

This should be addressed in a follow-up.

_Originally posted by @gnodet in https://github.com/apache/maven/issues/11829#issuecomment-4489231354_

UjjwalChitransh:
## Problem

In MojoExtension.beforeEach(), alignToBaseDirectory was called on tmodel (the raw parsed POM) and the result was stored in the extension context. This caused two bugs:

1. NPE when no POM exists: tmodel is null when no POM file is found, causing a NullPointerException.
2. Missing defaults: The stored model was the raw parsed POM, missing defaults merged in from defaultModel (groupId, artifactId, build paths, etc.).

Fixes #12201

## Fix

Call alignToBaseDirectory on model (the merged result of tmodel + defaultModel) and store that aligned model. The createProject provider inside Foo is also updated to use alignedModel consistently.

## Testing

- All 22 existing tests in maven-testing pass.
- The NPE scenario (no POM file) is now safe: model is never null since defaultModel is always the fallback.

### DIFF
--- a/impl/maven-testing/src/main/java/org/apache/maven/api/plugin/testing/MojoExtension.java
+++ b/impl/maven-testing/src/main/java/org/apache/maven/api/plugin/testing/MojoExtension.java
@@ -437,9 +437,9 @@ public class MojoExtension extends MavenDIExtension implements ParameterResolver
         } else {
             model = new MavenMerger().merge(tmodel, defaultModel, false, null);
         }
-        tmodel = new DefaultModelPathTranslator(new DefaultPathTranslator())
-                .alignToBaseDirectory(tmodel, Paths.get(getBasedir()), null);
-        context.getStore(MOJO_EXTENSION).put(Model.class, tmodel);
+        final Model alignedModel = new DefaultModelPathTranslator(new DefaultPathTranslator())
+                .alignToBaseDirectory(model, Paths.get(getBasedir()), null);
+        context.getStore(MOJO_EXTENSION).put(Model.class, alignedModel);
 
         // mojo execution
         // Map<Object, Object> map = getInjector().getContext().getContextData();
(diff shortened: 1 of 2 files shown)


impl/maven-testing/src/main/java/org/apache/maven/api/plugin/testing/MojoExtension.java
```diff
@@ -437,9 +437,9 @@ public void beforeEach(ExtensionContext context) throws Exception {
         } else {
             model = new MavenMerger().merge(tmodel, defaultModel, false, null);
         }
-        tmodel = new DefaultModelPathTranslator(new DefaultPathTranslator())
-                .alignToBaseDirectory(tmodel, Paths.get(getBasedir()), null);
-        context.getStore(MOJO_EXTENSION).put(Model.class, tmodel);
+        final Model alignedModel = new DefaultModelPathTranslator(new DefaultPathTranslator())
+                .alignToBaseDirectory(model, Paths.get(getBasedir()), null);
```
gnodet, line 441: This is the heart of the fix: `model` is guaranteed non-null here (either `defaultModel` or the merge result), so the NPE is eliminated. And since `model` carries the merged defaults, the aligned result now includes build paths and coordinates from `defaultModel` — which was the other half of the bug.

Nit: the local `model` variable (line 434) is now only used as an intermediate to compute `alignedModel`. That's fine — keeping it separate is more readable than inlining the ternary.

gnodet (commented):
Thanks for the fix, @UjjwalChitransh. The root-cause analysis in the PR description is accurate.

**Correctness** — The old code called `alignToBaseDirectory(tmodel, ...)` which was wrong in two ways: `tmodel` is `null` when no POM is provided (NPE), and even when non-null it lacks the defaults merged from `defaultModel` (missing build paths, coordinates, etc.). Passing the merged `model` instead fixes both issues.

**Consistency** — Before this fix there was also a subtle inconsistency: the extension context stored the aligned `tmodel` while `createProject()` used the non-aligned `model`. The two could diverge. Now both consistently use `alignedModel`, which is cleaner.

**Scoping** — The `final` qualifier on `alignedModel` is needed for the inner-class capture in `Foo.createProject()` (line 488+). Looks correct.

**Tests** — Good coverage: the no-POM default path, the merged-model-with-custom-coordinates path, and the absolute-path invariant after alignment. The three tests map directly to the three things the fix is supposed to guarantee.

LGTM from a code-review standpoint. Leaving as COMMENT so Guillaume can give the final approval.

gnodet (approved):

PR #12251, opened 2026-06-10
disable dtd and external entities in stax xml parsers

jmestwa-coder:
**XXE via external entity resolution in the StAX XML parsers**

Maven's hand-written StAX readers build their parser with `XMLInputFactory.newFactory()`, and the Woodstox defaults resolve external entities. A `pom.xml`, `META-INF/maven/extension.xml` or `plugin.xml` read from a downloaded artifact can pull in `file://` or `http://` SYSTEM entities, which gives local file disclosure and SSRF while a build runs. `SUPPORT_DTD` and `IS_SUPPORTING_EXTERNAL_ENTITIES` are now turned off on every input factory used to read XML. None of Maven's own formats carry a DOCTYPE, so this only rejects hostile input.

The added test in `ExtensionDescriptorBuilderTest` leaks a temp file through an external entity in an extension descriptor; it fails on the current tree and passes with the factories hardened.

### DIFF
--- a/compat/maven-model-builder/src/main/java/org/apache/maven/model/root/DefaultRootLocator.java
+++ b/compat/maven-model-builder/src/main/java/org/apache/maven/model/root/DefaultRootLocator.java
@@ -43,7 +43,7 @@ public class DefaultRootLocator implements RootLocator {
         // we're too early to use the modelProcessor ...
         Path pom = dir.resolve("pom.xml");
         try (InputStream is = Files.newInputStream(pom)) {
-            XMLStreamReader parser = XMLInputFactory.newFactory().createXMLStreamReader(is);
+            XMLStreamReader parser = newInputFactory().createXMLStreamReader(is);
             if (parser.nextTag() == XMLStreamReader.START_ELEMENT
                     && parser.getLocalName().equals("project")) {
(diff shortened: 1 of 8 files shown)


gnodet (commented):
Thanks for raising this — the XXE hardening is a good idea.

A few observations:

1. **Duplication**: the PR copies an identical `newInputFactory()` helper into 5 separate classes. A single shared method would be easier to maintain.
2. **Missing coverage**: the Velocity templates (`reader-stax.vm`, `reader.vm`) that generate StAX readers for settings.xml, toolchains.xml, plugin.xml, metadata.xml, etc. also create their own `XMLInputFactory` instances and are not hardened by this PR.
3. **Context**: external entity / XInclude support was an intentional feature (MNG-5862) that was later extracted into [maven-xinclude-extension](https://github.com/apache/maven-xinclude-extension). That extension creates its own factory with entities enabled under a restricted resolver, so it is unaffected by this hardening.

I've opened #12256 as an alternative that centralizes the factory creation into `XmlService.newXMLInputFactory()` and also covers the Velocity templates.

gnodet:
Superseded by #12256 (centralized approach) which has been merged and backported to 4.0.x via #12269.

Issue #11966, opened 2026-04-20
Update documentation at - Introduction to the Build Lifecycle


slawekjaranowski:
### New feature, improvement proposal

At site: https://maven.apache.org/guides/introduction/introduction-to-the-lifecycle.html
We have links: https://maven.apache.org/guides/introduction/introduction-to-the-lifecycle.html#references

In https://github.com/apache/maven/pull/11916 we have droped some of xml files so documentation can be outdated.

Hemanth7723:
Assign me this issue plz

jiteshkhatri11:
Hi, I'd like to work on this issue. Can you guide me on what exactly needs to be updated?

Hiteshsai007:
Hi, I'd like to work on updating the documentation described in this issue. I'll review the affected page and submit a PR if I identify the outdated references.

cstamas:
My proposal https://github.com/apache/maven-help-plugin/pull/374

We do not have plexus XML to include in doco, but this PR makes plugin work for all

PR #12260, opened 2026-06-13
[MNG-11966] Update lifecycles reference to point to LifecycleRegistry

Hiteshsai007:
# MNG-11966: Update lifecycles reference to point to LifecycleRegistry

### Proposed Changes
This PR addresses issue **#11966** ("Update documentation at - Introduction to the Build Lifecycle").

In PR #11916, legacy Plexus XML descriptors (`components.xml` and `default-bindings.xml`) were removed from the `maven-core` module as part of the migration of the core to JSR-330 / `javax.inject`.

Consequently, the source file for the generated references page (`lifecycles.apt.vm`) was still pointing to the old package `org.apache.maven.lifecycle.providers` for the 3 lifecycles, which is no longer the source of truth. 

This change updates `lifecycles.apt.vm` to point to the new public `org.apache.maven.api.services.LifecycleRegistry` where the lifecycles are registered and queried in Maven 4.

### Checklist

- [ ] Write unit tests that match behavioral changes, where the tests fail if the changes to the runtime are not applied. (N/A: Documentation-only change)
- [ ] Run `mvn verify` to make sure basic checks pass. (N/A: Documentation-only change)
- [ ] You have run the [Core IT][core-its] successfully. (N/A: Documentation-only change)

### DIFF
--- a/.github/workflows/maven.yml
+++ b/.github/workflows/maven.yml
@@ -33,7 +33,7 @@ permissions: {}
 
 env:
   # Keep in sync with mimir testing version in pom.xml
-  MIMIR_VERSION: 0.11.4
+  MIMIR_VERSION: 0.12.0
   MIMIR_BASEDIR: ~/.mimir
   MIMIR_LOCAL: ~/.mimir/local
   MAVEN_OPTS: -XX:+HeapDumpOnOutOfMemoryError -XX:HeapDumpPath=./target/java_heapdump.hprof
--- a/apache-maven/src/assembly/maven/bin/JvmConfigParser.java
+++ b/apache-maven/src/assembly/maven/bin/JvmConfigParser.java
@@ -94,14 +94,7 @@ public class JvmConfigParser {
     static String parseJvmConfig(Path jvmConfigPath, String mavenProjectBasedir) throws IOException {
         StringBuilder result = new StringBuilder();
 
-        boolean firstLine = true;
         for (String line : Files.readAllLines(jvmConfigPath, StandardCharsets.UTF_8)) {
-            if (firstLine) {
-                if (line.startsWith("\uFEFF")) {
-                    line = line.substring(1);
-                }
-                firstLine = false;
-            }
             line = processLine(line, mavenProjectBasedir);
             if (line.isEmpty()) {
                 continue;
--- a/apache-maven/src/assembly/maven/bin/JvmConfigParser.java
+++ b/apache-maven/src/assembly/maven/bin/JvmConfigParser.java
@@ -94,7 +94,14 @@ public class JvmConfigParser {
     static String parseJvmConfig(Path jvmConfigPath, String mavenProjectBasedir) throws IOException {
         StringBuilder result = new StringBuilder();
 
+        boolean firstLine = true;
         for (String line : Files.readAllLines(jvmConfigPath, StandardCharsets.UTF_8)) {
+            if (firstLine) {
+                if (line.startsWith("\uFEFF")) {
+                    line = line.substring(1);
+                }
+                firstLine = false;
+            }
             line = processLine(line, mavenProjectBasedir);
             if (line.isEmpty()) {
                 continue;
--- a/impl/maven-core/src/site/apt/lifecycles.apt.vm
+++ b/impl/maven-core/src/site/apt/lifecycles.apt.vm
@@ -25,7 +25,7 @@
 
 Lifecycles Reference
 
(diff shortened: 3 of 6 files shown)


Hiteshsai007:
The `windows-latest` test failed on Java 25 because of Maven 4's Remote Repository Filtering (RRF). 

The test environment runs with the `mimir` extension. In the integration tests, we already disable RRF inside `Verifier#execute()`. However, `Verifier` also makes quick helper calls to Maven under the hood (like when `deleteArtifacts()` looks up the metadata path at the start of `MavenIT0010`). Those helper calls are created through `Verifier#executorRequest()`, which was missing the `-Daether.remoteRepositoryFilter` flags. 

On clean/uncached runners, this caused those helper invocations to run with RRF enabled, query Maven Central, and get blocked because `eu.maveniverse` isn't in Central's prefixes list.

I've added the disable flags directly in `Verifier#executorRequest()` so it gets bypassed for helper Maven processes as well. Tested the change locally, and it compiles fine.

Hiteshsai007:
### Why the old one failed:
On Windows, the integration tests generated the `.mvn/jvm.config` file with a UTF-8 BOM (`\uFEFF`) prepended to it. The old `JvmConfigParser` read this file but did not strip the BOM, passing `\uFEFF-Dprop=val` directly to the Java 21 launcher. The JVM didn't recognize the corrupted `-D` option, failing to set the property and causing test assertions to fail.

### How the new one works:
The new parser checks for and strips the leading BOM (`\uFEFF`) character. The clean `-Dprop=val` is passed to the Java 21 launcher, which successfully sets the system property, allowing all integration tests to pass.

cstamas:
Assessment is wrong, while Central prefixes really does not have `/eu/maveniverse`, it does have `/eu` which still allows `eu.maveniverse...` (and any other `eu.` top level) group IDs.

And due assessment being wrong, the change in this PR is wrong as well.

Hiteshsai007:
You are completely correct. The prefix matching in Maven Resolver allows `/eu` to cover the `eu.maveniverse` namespace, so remote repository filtering was not the blocker.
 
 I have reverted the changes to `Verifier.java` from this PR. 
 
 The actual root cause of the Windows-specific CI failures was the UTF-8 BOM (`\uFEFF`) in the generated `.mvn/jvm.config` file, which was causing the JVM launcher to reject option arguments. The PR now only contains the fix in `JvmConfigParser.java` to strip the leading BOM.

cstamas:
Does UTF-8 care for endianness? How and why did BOM appear in your UTF-8 text file?

Hiteshsai007:
> Does UTF-8 care for endianness? How and why did BOM appear in your UTF-8 text file?

No Sir, UTF-8 doesn't care about endianness at all. 

The BOM showed up because a lot of standard Windows tools (like Notepad or default PowerShell redirections) silently prepend the UTF-8 BOM to distinguish the file from legacy ANSI encoding. If a Windows user creates or edits their `.mvn/jvm.config` with one of these tools, the BOM gets inserted automatically. 

Since our parser wasn't stripping it, that invisible `\uFEFF` character was being passed straight to the JVM launcher (e.g., `\uFEFF-Dprop=val`), causing the JVM to reject the argument. Stripping it just makes Maven resilient against files saved by default Windows text editors.

cstamas:
Given this is not the only file that Maven reads, let's remove this change from this PR (is out of scope for it anyway) and report a Maven issue where we can discuss this.

Hiteshsai007:
> Given this is not the only file that Maven reads, let's remove this change from this PR (is out of scope for it anyway) and report a Maven issue where we can discuss this.

Makes total sense! I've reverted the BOM-stripping changes in `JvmConfigParser.java` so this PR remains strictly focused on the original `lifecycles.apt.vm` documentation update.

elharo (approved):

Hiteshsai007:
Thank you for the review and approval. The BOM-related changes have been reverted, and this PR is now focused solely on the `lifecycles.apt.vm` documentation update. All CI checks are passing.

gnodet (approved):
LGTM — the link correctly points to `LifecycleRegistry` in the aggregated Javadoc.

Hiteshsai007:
Thank you to everyone who reviewed and supported this contribution. I really appreciate the guidance and feedback throughout the process. (@gnodet @elharo @cstamas @slachiewicz )

 I'm also hoping to become a **GSoC 2027** contributor with Apache Maven, so being able to contribute here is an exciting step in that journey. Looking forward to contributing more in the future! 🙏

PR #12269, opened 2026-06-16
Backport #12256: Centralize XXE hardening for StAX XML parsers

gnodet:
Backport of #12256 to 4.0.x — centralizes XXE hardening into `XmlService.newXMLInputFactory()`.

All `XMLInputFactory.newFactory()` call sites now go through this hardened factory method, which disables DTD support and external entity resolution to prevent XXE attacks.

### DIFF
--- a/api/maven-api-xml/src/main/java/org/apache/maven/api/xml/XmlService.java
+++ b/api/maven-api-xml/src/main/java/org/apache/maven/api/xml/XmlService.java
@@ -18,6 +18,7 @@
  */
 package org.apache.maven.api.xml;
 
+import javax.xml.stream.XMLInputFactory;
 import javax.xml.stream.XMLStreamException;
 import javax.xml.stream.XMLStreamReader;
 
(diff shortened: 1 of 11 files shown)


api/maven-api-xml/src/main/java/org/apache/maven/api/xml/XmlService.java
```diff
@@ -103,6 +104,20 @@ public abstract class XmlService {
      */
     public static final String KEYS_COMBINATION_MODE_ATTRIBUTE = "combine.keys";
 
+    /**
+     * Creates a new {@link XMLInputFactory} hardened against XXE attacks.
+     * The returned factory has DTD support and external entity resolution disabled.
+     *
+     * @return a hardened XMLInputFactory
+     * @since 4.1.0
```
cstamas: since 4.0.0-rc-6?

gnodet: Fixed — changed to `@since 4.0.0-rc-6`. Also created #12270 to fix it on master.

PR #12291, opened 2026-06-17
[#12288] Pass settings.xml profile properties to LRM

### LINKED ISSUE #12288: Profile-level properties from a settings.xml-activated profile are dropped before LRM init
### Background

Apache Maven Resolver lets users configure the local repository manager
through `aether.*` properties. The Maven user manual describes
`<properties>` inside a `<profile>` as a regular way to set them, and
the documented activation channels for such a profile are:

- CLI: `mvn -P <id>`
- `settings.xml`: `<activeProfiles><activeProfile>...</activeProfile></activeProfiles>`
- `settings.xml`: `<profile><activation><activeByDefault>true</activeByDefault></activation>...`

The first two channels propagate the profile's `<properties>` to the
resolver session as expected. The third does **not**.

### Bug

When a profile is activated through
`<activation><activeByDefault>true</activeByDefault></activation>` in
`settings.xml`, the properties declared in that profile's
`<properties>` block never reach the
`RepositorySystemSession.configProperties` that the
`EnhancedLocalRepositoryManager` reads at initialization. Consequently,
`aether.*` settings declared in an active-by-default settings.xml
profile have no effect on the LRM — but the *same* `<properties>` block
on the *same* `<profile>` works correctly when the user activates it via
`-P <id>` on the CLI, or via `<activeProfiles><activeProfile>...` in the
same `settings.xml`.

This is observable to the user because
`mvn help:evaluate -Dexpression=aether.enhancedLocalRepository.localPrefix`
*does* report the configured value (the property is in the Maven session
elsewhere), yet the LRM behaves as if no `aether.*` configuration had
been set.

### Reproducer

`pom.xml`:

```xml
<project xmlns="http://maven.apache.org/POM/4.0.0">
  <groupId>org.example.aether</groupId>
  <artifactId>test-artifact</artifactId>
  <version>1.0</version>
  <packaging>pom</packaging>
</project>
```

`settings.xml` (active-by-default — broken case):

```xml
<settings xmlns="http://maven.apache.org/SETTINGS/1.0.0">
  <profiles>
    <profile>
      <id>aether-split-via-settings</id>
      <activation><activeByDefault>true</activeByDefault></activation>
      <properties>
        <aether.enhancedLocalRepository.split>true</aether.enhancedLocalRepository.split>
        <aether.enhancedLocalRepository.localPrefix>mvb-foo</aether.enhancedLocalRepository.localPrefix>
      </properties>
    </profile>
  </profiles>
</settings>
```

For comparison, the same profile activated via `<activeProfiles>` (works correctly):

```xml
<settings xmlns="http://maven.apache.org/SETTINGS/1.0.0">
  <activeProfiles>
    <activeProfile>aether-split-via-settings</activeProfile>
  </activeProfiles>
  <profiles>
    <profile>
      <id>aether-split-via-settings</id>
      <properties>
        <aether.enhancedLocalRepository.split>true</aether.enhancedLocalRepository.split>
        <aether.enhancedLocalRepository.localPrefix>mvb-foo</aether.enhancedLocalRepository.localPrefix>
      </properties>
    </profile>
  </profiles>
</settings>
```

Run with each settings file:

```bash
mvn -s settings.xml -Dmaven.repo.local=$PWD/local-repo install
find local-repo -name 'test-artifact-1.0.pom'
```

| settings.xml activation | Found at |
|---|---|
| `<activeProfiles><activeProfile>...` (works) | `local-repo/mvb-foo/org/example/aether/test-artifact/1.0/test-artifact-1.0.pom` |
| `<activeByDefault>true</activeByDefault>` (broken — actual) | `local-repo/org/example/aether/test-artifact/1.0/test-artifact-1.0.pom` |
| `<activeByDefault>true</activeByDefault>` (expected) | `local-repo/mvb-foo/org/example/aether/test-artifact/1.0/test-artifact-1.0.pom` |

### Versions

- Maven 4.x master (sha `09fdd6a6`, `4.1.0-SNAPSHOT`)
- Maven 3.9.16

Likely affects all versions on both lines; backport candidate.

ascheman:
Fixes #12288.

Profiles activated through `settings.xml` (`<activeProfiles>` or
`<activation><activeByDefault>true</activeByDefault>`) had their
`<properties>` block dropped before the resolver session was built, so
`aether.*` configuration declared in such a profile silently failed.
The same `<properties>` on the same `<profile>` worked when activated
via CLI `-P`.

## Changes

1. Two ITs in `core-it-suite` covering both settings.xml activation
   channels — `testActiveByDefaultProfile` (red on master) and
   `testActiveProfilesList` (green on master, regression guard).
2. `DefaultRepositorySystemSessionFactory#getPropertiesFromRequestedProfiles`
   extended to also consider `request.getActiveProfiles()` and
   `<activeByDefault>true</activeByDefault>` profiles.

Locally verified against the `core-it-suite`:
- master + new ITs: 1 of 2 fail (`testActiveByDefaultProfile`)
- master + new ITs + fix: 2 of 2 pass

## Scope

Only the profile's `<properties>` block is affected — other profile
content (`<plugins>`, `<dependencies>`, `<repositories>`, …) flows
through its existing channels unchanged.

Limitation: this does not re-implement the full `DefaultProfileSelector`
semantics (e.g. `activeByDefault` deactivation when other profiles of
the same source are active). For the property-propagation use case in
scope of #12288 this is sufficient; injecting `ProfileSelector` here
would be a larger refactor in a session-build hot path and is left to a
follow-up if needed.

## Affected versions

Reproduces on Maven 4.x master and on Maven 3.9.16; backport candidate.

## Test plan

- [x] `mvn install` on `apache-maven` from the branch builds cleanly
- [x] `mvn verify -Prun-its -DmavenDistro=… -Dit.test=MavenITSettingsProfileAetherPropertiesTest` on the patched dist: both ITs green
- [x] Same command on vanilla master dist: `testActiveByDefaultProfile` red, `testActiveProfilesList` green
- [ ] `core-it-suite` full run (not done locally — CI takes it from here)

### DIFF
--- a/impl/maven-core/src/main/java/org/apache/maven/internal/aether/DefaultRepositorySystemSessionFactory.java
+++ b/impl/maven-core/src/main/java/org/apache/maven/internal/aether/DefaultRepositorySystemSessionFactory.java
@@ -433,9 +433,15 @@ public class DefaultRepositorySystemSessionFactory implements RepositorySystemSe
         HashSet<String> activeProfileId =
                 new HashSet<>(request.getProfileActivation().getRequiredActiveProfileIds());
         activeProfileId.addAll(request.getProfileActivation().getOptionalActiveProfileIds());
+        // Also include profiles activated via settings.xml <activeProfiles> so that their
+        // properties (notably aether.* configuration) reach the resolver session config in
+        // time for LocalRepositoryManager initialization.
+        activeProfileId.addAll(request.getActiveProfiles());
 
         return request.getProfiles().stream()
-                .filter(profile -> activeProfileId.contains(profile.getId()))
+                .filter(profile -> activeProfileId.contains(profile.getId())
+                        || (profile.getActivation() != null
+                                && profile.getActivation().isActiveByDefault()))
                 .map(ModelBase::getProperties)
                 .flatMap(properties -> properties.entrySet().stream())
                 .filter(e -> e.getValue() != null)
--- a/its/core-it-suite/src/test/java/org/apache/maven/it/MavenITSettingsProfileAetherPropertiesTest.java
+++ b/its/core-it-suite/src/test/java/org/apache/maven/it/MavenITSettingsProfileAetherPropertiesTest.java
@@ -1,0 +1,109 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.maven.it;
+
+import java.io.File;
+
+import org.junit.jupiter.api.Test;
+
+import static org.junit.jupiter.api.Assertions.assertFalse;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+/**
+ * Integration tests proving that {@code aether.*} properties declared in the
+ * {@code <properties>} block of a {@code settings.xml} profile are honored
+ * by the resolver at local repository manager initialization, regardless of
+ * which settings.xml-only activation channel was used.
+ *
+ * <p>Two activation channels are covered:
+ * <ul>
+ *   <li>{@code <activation><activeByDefault>true</activeByDefault></activation>}
+ *       on the profile itself;</li>
+ *   <li>{@code <activeProfiles><activeProfile>...</activeProfile></activeProfiles>}
+ *       at the top of {@code settings.xml}.</li>
+ * </ul>
+ *
+ * <p>In both cases the same profile sets:
+ * <pre>
+ *   aether.enhancedLocalRepository.split       = true
+ *   aether.enhancedLocalRepository.localPrefix = it-custom-prefix
+ * </pre>
+ * and the test asserts that {@code mvn install} writes the installed pom
+ * under {@code <localRepo>/it-custom-prefix/&lt;groupId-path&gt;/...} rather
+ * than the flat or default-split layout.
+ *
+ * <p>The same properties on the same profile work correctly when the
+ * profile is activated via {@code -P <id>} on the CLI; only the
+ * settings.xml activation channels fail, which is what these tests guard
+ * against.
+ */
+public class MavenITSettingsProfileAetherPropertiesTest extends AbstractMavenIntegrationTestCase {
+
+    @Test
+    public void testActiveByDefaultProfile() throws Exception {
+        runAndAssertCustomPrefix("settings-active-by-default.xml");
+    }
+
+    @Test
+    public void testActiveProfilesList() throws Exception {
+        runAndAssertCustomPrefix("settings-active-profiles-list.xml");
+    }
+
+    private void runAndAssertCustomPrefix(String settingsFile) throws Exception {
+        File testDir = extractResources("/settings-profile-aether-properties");
+
+        Verifier verifier = newVerifier(testDir.getAbsolutePath());
+        verifier.setAutoclean(false);
+        verifier.deleteDirectory("target");
+        verifier.deleteArtifacts("org.apache.maven.its.settings.profile.aether");
+
+        verifier.addCliArgument("--settings");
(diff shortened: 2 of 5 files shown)


impl/maven-core/src/main/java/org/apache/maven/internal/aether/DefaultRepositorySystemSessionFactory.java
```diff
@@ -433,9 +433,15 @@ private Map<String, String> getPropertiesFromRequestedProfiles(MavenExecutionReq
         HashSet<String> activeProfileId =
                 new HashSet<>(request.getProfileActivation().getRequiredActiveProfileIds());
         activeProfileId.addAll(request.getProfileActivation().getOptionalActiveProfileIds());
+        // Also include profiles activated via settings.xml <activeProfiles> so that their
+        // properties (notably aether.* configuration) reach the resolver session config in
+        // time for LocalRepositoryManager initialization.
+        activeProfileId.addAll(request.getActiveProfiles());
```
gnodet, line 439: `request.getActiveProfiles()` delegates to `profileActivation.getActiveProfiles()`, which returns all IDs where `active() == true` — the exact union of `getRequiredActiveProfileIds()` + `getOptionalActiveProfileIds()` already collected above. The settings.xml `<activeProfiles>` list is wired via `overwriteActiveProfiles()` → `activateOptionalProfile()`, so those IDs are already in `getOptionalActiveProfileIds()`.

This block should instead collect **inactive** profile IDs, so the `activeByDefault` filter below can respect explicit deactivation (`-P !profileId`):

```suggestion
        // Profiles explicitly deactivated via -P !id must be excluded even if they
        // declare activeByDefault=true.
        HashSet<String> inactiveProfileId =
                new HashSet<>(request.getProfileActivation().getRequiredInactiveProfileIds());
        inactiveProfileId.addAll(request.getProfileActivation().getOptionalInactiveProfileIds());
```

impl/maven-core/src/main/java/org/apache/maven/internal/aether/DefaultRepositorySystemSessionFactory.java
```diff
@@ -433,9 +433,15 @@ private Map<String, String> getPropertiesFromRequestedProfiles(MavenExecutionReq
...
 
         return request.getProfiles().stream()
-                .filter(profile -> activeProfileId.contains(profile.getId()))
+                .filter(profile -> activeProfileId.contains(profile.getId())
+                        || (profile.getActivation() != null
+                                && profile.getActivation().isActiveByDefault()))
```
gnodet, line 444: The `activeByDefault` branch should exclude profiles that have been explicitly deactivated:

```suggestion
                .filter(profile -> activeProfileId.contains(profile.getId())
                        || (!inactiveProfileId.contains(profile.getId())
                                && profile.getActivation() != null
                                && profile.getActivation().isActiveByDefault()))
```

its/core-it-suite/src/test/java/org/apache/maven/it/MavenITSettingsProfileAetherPropertiesTest.java
```diff
@@ -0,0 +1,109 @@
...
+ * and the test asserts that {@code mvn install} writes the installed pom
+ * under {@code <localRepo>/it-custom-prefix/&lt;groupId-path&gt;/...} rather
+ * than the flat or default-split layout.
+ *
+ * <p>The same properties on the same profile work correctly when the
+ * profile is activated via {@code -P <id>} on the CLI; only the
+ * settings.xml activation channels fail, which is what these tests guard
+ * against.
+ */
+public class MavenITSettingsProfileAetherPropertiesTest extends AbstractMavenIntegrationTestCase {
```
gnodet, line 56: Nit: other ITs fixing tracked issues typically follow the `MavenITgh<number>...Test` pattern (e.g. `MavenITgh10210SettingsXmlDecryptTest`). Consider renaming to `MavenITgh12288SettingsProfileAetherPropertiesTest` for consistency and traceability.

gnodet (changes requested):
Thank you for this well-documented PR, @ascheman — the bug report and reproducer are thorough, and the fix targets the right method.

Two issues to address, one correctness gap and one redundancy:

1. **Missing deactivation check:** The `activeByDefault` branch doesn't exclude profiles explicitly deactivated via `-P !profileId`. A user running `mvn -P !aether-split-via-settings` would expect the profile to be deactivated, but the current code still includes its properties.

2. **Redundant `request.getActiveProfiles()` call:** `request.getActiveProfiles()` returns the exact union of `getRequiredActiveProfileIds()` + `getOptionalActiveProfileIds()` (already collected above). The settings.xml `<activeProfiles>` list flows via `overwriteActiveProfiles()` → `activateOptionalProfile()`, so they're already in `getOptionalActiveProfileIds()`. The real fix is the `activeByDefault` check — the `getActiveProfiles()` line adds nothing.

See inline suggestions for concrete fixes.

_Note: this review does not replace specialized AI review tools (CodeRabbit, Sourcery) or static analysis (SonarCloud)._

ascheman:
Superseded by #12297 — closing in favor of @gnodet's reroll. Thanks for picking it up and preserving attribution.

PR #12298, opened 2026-06-17
[#12288] Add -P !profile deactivation regression guard

ascheman:
Hypersedes #12297 (which superseded #12291).

Starts from the head of #12297 and adds one additional integration
test, `testActiveByDefaultDeactivatedViaCli`, that guards the
`-P !profileId` deactivation path introduced in @gnodet's commit
"Address review feedback: fix activeByDefault deactivation and rename
IT". The fix logic itself is unchanged from #12297; this PR adds the
regression test to lock it in.

## What's new vs #12297

* `testActiveByDefaultDeactivatedViaCli` — runs `mvn install` with
  `-P!aether-split-via-settings` and asserts the artifact lands at the
  flat layout (not under the custom prefix), confirming the
  `inactiveProfileId` guard works end-to-end.
* `deleteRecursivelyIfExists` private helper — clears any stale
  `it-custom-prefix/` subtree before the deactivation run, so the
  assertion is robust against test-execution order with sibling tests
  installing under that same prefix.

## Verification

* Patched build: 3/3 ITs green.
* Without the deactivation guard in the fix (i.e. vanilla master):
  `testActiveByDefaultProfile` is RED — the bug — and the deactivation
  test trivially passes because the profile is silently dropped anyway.
  So the new test specifically guards the fix-introduced deactivation
  logic, not a pre-existing path.

Happy to fold the extra test into #12297 instead if you'd prefer to
keep it on a single PR; just say the word.

### DIFF
--- a/its/core-it-suite/src/test/java/org/apache/maven/it/MavenITgh12288SettingsProfileAetherPropertiesTest.java
+++ b/its/core-it-suite/src/test/java/org/apache/maven/it/MavenITgh12288SettingsProfileAetherPropertiesTest.java
@@ -107,10 +107,11 @@ public class MavenITgh12288SettingsProfileAetherPropertiesTest extends AbstractM
         if (!path.exists()) {
             return;
         }
-        Files.walk(path.toPath())
-                .sorted(Comparator.reverseOrder())
-                .map(java.nio.file.Path::toFile)
-                .forEach(File::delete);
+        try (var paths = Files.walk(path.toPath())) {
+            paths.sorted(Comparator.reverseOrder())
+                    .map(java.nio.file.Path::toFile)
+                    .forEach(File::delete);
+        }
     }
 
     private void runAndAssertCustomPrefix(String settingsFile) throws Exception {
--- a/its/core-it-suite/src/test/java/org/apache/maven/it/MavenITgh12288SettingsProfileAetherPropertiesTest.java
+++ b/its/core-it-suite/src/test/java/org/apache/maven/it/MavenITgh12288SettingsProfileAetherPropertiesTest.java
@@ -101,7 +101,6 @@ public class MavenITgh12288SettingsProfileAetherPropertiesTest extends AbstractM
                 customPrefix.exists(),
                 "Found artifact at custom prefix " + customPrefix + " — deactivation via -P ! was ignored.");
     }
-    }
 
     private void runAndAssertCustomPrefix(String settingsFile) throws Exception {
         File testDir = extractResources("/settings-profile-aether-properties");
--- a/its/core-it-suite/src/test/java/org/apache/maven/it/MavenITgh12288SettingsProfileAetherPropertiesTest.java
+++ b/its/core-it-suite/src/test/java/org/apache/maven/it/MavenITgh12288SettingsProfileAetherPropertiesTest.java
@@ -19,6 +19,8 @@
 package org.apache.maven.it;
 
 import java.io.File;
+import java.nio.file.Files;
+import java.util.Comparator;
 
 import org.junit.jupiter.api.Test;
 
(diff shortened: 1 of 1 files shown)


gnodet (approved):
Clean addition — the new IT guards the deactivation logic from #12297 end-to-end, and CI is green across all platforms.

One minor suggestion inline about resource handling in the helper method, but not blocking.

_Note: this review does not replace specialized AI review tools (CodeRabbit, Sourcery) or static analysis (SonarCloud)._

gnodet (commented):
Thanks @ascheman! I pushed a small cleanup commit to replace the custom `deleteRecursivelyIfExists` helper with `FileUtils.deleteDirectory()` from commons-io (already on the classpath). This handles non-existent paths gracefully and is the pattern used elsewhere in the IT suite.

Note: I could not push directly to your fork, so I will apply this change at squash-merge time.

its/core-it-suite/src/test/java/org/apache/maven/it/MavenITgh12288SettingsProfileAetherPropertiesTest.java
```diff
@@ -19,6 +19,8 @@
 package org.apache.maven.it;
 
 import java.io.File;
+import java.nio.file.Files;
+import java.util.Comparator;
 
 import org.junit.jupiter.api.Test;
```
gnodet, line 24: Replace `Files`/`Comparator` imports with `FileUtils` (same one the `Verifier` uses internally):

```suggestion
import java.io.File;

import org.codehaus.plexus.util.FileUtils;
import org.junit.jupiter.api.Test;
```

its/core-it-suite/src/test/java/org/apache/maven/it/MavenITgh12288SettingsProfileAetherPropertiesTest.java
```diff
@@ -65,6 +67,53 @@ public void testActiveProfilesList() throws Exception {
...
+        Verifier verifier = newVerifier(testDir.getAbsolutePath());
+        verifier.setAutoclean(false);
+        verifier.setLogFileName("log-deactivation.txt");
+        verifier.deleteDirectory("target");
+        verifier.deleteArtifacts("org.apache.maven.its.settings.profile.aether");
+
+        // Sibling tests in this class install under the same custom prefix; clear it to
+        // avoid false positives if those tests ran first.
+        File customPrefixSubtree = new File(verifier.getLocalRepository(), "it-custom-prefix");
+        deleteRecursivelyIfExists(customPrefixSubtree);
```
gnodet: Use `FileUtils.deleteDirectory()` which handles non-existent paths gracefully:

```suggestion
        FileUtils.deleteDirectory(customPrefixSubtree);
```

gnodet (commented):
The custom `deleteRecursivelyIfExists` helper can be replaced by `FileUtils.deleteDirectory()`, already used by other ITs in this suite (e.g. `MavenITmng7819FileLockingWithSnapshotsTest`, `MavenIT0108SnapshotUpdateTest`). Three suggestions below.

its/core-it-suite/src/test/java/org/apache/maven/it/MavenITgh12288SettingsProfileAetherPropertiesTest.java
```diff
@@ -65,6 +67,52 @@ public void testActiveProfilesList() throws Exception {
...
+
+    private static void deleteRecursivelyIfExists(File path) throws Exception {
+        if (!path.exists()) {
+            return;
+        }
+        Files.walk(path.toPath())
+                .sorted(Comparator.reverseOrder())
+                .map(java.nio.file.Path::toFile)
+                .forEach(File::delete);
+    }
```
gnodet: Remove the custom helper — `FileUtils.deleteDirectory()` replaces it:

```suggestion
```

its/core-it-suite/src/test/java/org/apache/maven/it/MavenITgh12288SettingsProfileAetherPropertiesTest.java
```diff
@@ -65,6 +66,43 @@ public void testActiveProfilesList() throws Exception {
...
+
+        assertTrue(
+                flatLayout.exists(),
+                "Expected artifact at flat layout (profile deactivated via -P !), but not found at " + flatLayout);
+
+        assertFalse(
+                customPrefix.exists(),
+                "Found artifact at custom prefix " + customPrefix + " — deactivation via -P ! was ignored.");
+    }
+    }
```
gnodet: Stray closing brace left over from applying the previous suggestion — sorry about that!

```suggestion
    }
```

PR #12316, opened 2026-06-18
[#12302] Fix TransitiveDependencyManager version downgrade from parent-inherited depMgmt

### LINKED ISSUE #12302: TransitiveDependencyManager changes dependency mediation, breaking enforcer rules and scope derivation
## Description

Maven 4's default `TransitiveDependencyManager` (enabled when `maven3Personality=false`) changes dependency mediation behavior compared to Maven 3's `ClassicDependencyManager`. This causes enforcer rule violations due to managed version application to transitive dependencies.

~~### 1. Scope derivation excludes transitive compile deps through provided parents~~

**Update:** The projects originally reported under this category (`hadoop-api-shim`, `flink-connector-hive`) are **not** affected by scope derivation changes. Both projects fail because their `hadoop-project` parent POM (versions 3.1.0 and 3.2.0) contains invalid XML (`<Xlint:-unchecked/>` — the `-` after the colon is not a valid XML name start character). Maven 3 silently parses this in lenient mode, while Maven 4 correctly rejects it, causing all transitive dependencies from those POMs to be unavailable. This is a project-side issue, not a Maven 4 regression.

### ~~2.~~ Managed version application to transitive dependencies

`TransitiveDependencyManager` applies `<dependencyManagement>` versions at ALL depths during collection, not just to direct dependencies. This can **downgrade** transitive dependency versions, triggering `RequireUpperBoundDeps` and `BannedDependencies` enforcer rule violations that pass with Maven 3.

**Affected projects:** `guacamole-client`, `logging-log4j-samples`, `netbeans-html4j`

Example (guacamole-client):
```
Require upper bound dependencies error for com.google.code.findbugs:jsr305:3.0.1
+-guacamole-auth-header:1.6.1
  +-guice:5.1.0
    +-guava:32.1.3-jre (managed) <-- guava:30.1-jre
      +-jsr305:3.0.1 (managed) <-- jsr305:3.0.2
```

Note the `(managed)` annotations — Maven 4 is applying the parent's `dependencyManagement` to transitive deps, downgrading jsr305 from 3.0.2 to 3.0.1.

## Root cause

`TransitiveDependencyManager.manageDependency()` applies managed versions/scopes at depth ≥ 2 during collection. Maven 3's `ClassicDependencyManager` did not do this — version management and scope derivation were handled post-collection in the graph transformer.

Key code difference:

| | `ClassicDependencyManager` | `TransitiveDependencyManager` |
|---|---|---|
| `deriveUntil` | `2` (root depMgmt only) | `Integer.MAX_VALUE` (all POMs) |
| Version management sources | Root `<dependencyManagement>` only | Every POM in the graph |

## Workaround

`-Dmaven.maven3Personality=true` reverts to `ClassicDependencyManager` behavior.

## Suggested fix

Apply `TransitiveDependencyManager`'s full-depth collection only for Maven 4 POMs, or only apply collected managed versions when no version has already been established by a higher-priority (nearer-to-root) source.

## Reproducers

```bash
# Enforcer mediation issue
git clone https://github.com/apache/guacamole-client.git && cd guacamole-client
mvn -B -e clean package -DskipTests
```

## Context

Found during [Maven 4 compatibility testing](https://github.com/gnodet/maven4-testing/issues/13307) of Apache projects.

gnodet:
## Summary

- Strip parent-inherited `<dependencyManagement>` entries from transitive dependency descriptors to prevent `TransitiveDependencyManager` from propagating them through the dependency graph
- Add integration test reproducing the version downgrade scenario (RED test turned GREEN)

## Problem

When a transitive dependency's parent POM declares `<dependencyManagement>`, those entries are inherited into the child's effective model. `TransitiveDependencyManager` collects depMgmt from every POM in the graph and applies it at depth 2+, causing parent-inherited management rules to override version declarations in deeper transitive dependencies.

Example: `root → module-a (parent=parent-a, manages lib-c:1.0) → module-b → lib-c:2.0`. Parent-a's `lib-c:1.0` management leaks through module-a and downgrades `lib-c` from 2.0 to 1.0.

## Fix

In `DefaultArtifactDescriptorReader.loadPom()`, compare the effective model's depMgmt with the parent model's depMgmt. Entries that appear in both with the same key and version are parent-inherited and stripped. This is safe because `ArtifactDescriptorReader` is only called for transitive dependencies — the root project's depMgmt goes via `CollectRequest.setManagedDependencies()`.

Applied to both `impl/maven-impl` and `compat/maven-resolver-provider`.

## Test plan

- [x] `gh-12302` IT: `lib-c:2.0` resolved correctly (not downgraded to 1.0)
- [x] `MNG-8347` transitive mode: level4 managed by level2's own depMgmt, level3 unmanaged, root wins for level5/6
- [x] `MNG-7982` non-transitive mode: only root depMgmt applies (Maven 3 behavior preserved)

Fixes #12302


gnodet:
Closing this PR — after discussion with Tamas, the current behavior is correct.

`TransitiveDependencyManager` applies depMgmt "down the tree" within the same branch. When module-a inherits `lib-c:1.0` management from parent-a, that management correctly applies to `lib-c` at depth 3 in module-a's subtree. This is by design, not a bug.

The fix for consumers who need a different version is to explicitly declare the dependency in their own `<dependencyManagement>` — root's depMgmt is collected at depth 0 and wins via `putIfAbsent` semantics.

Issue #12301, opened 2026-06-18
StackOverflowError in DefaultModelBuilder.readFileModel with recursive parent resolution


gnodet:
## Description

Building `jackrabbit-filevault` with Maven from the `maven-4.0.x` branch causes a `StackOverflowError` in `DefaultModelBuilder$ModelBuilderSessionState.readFileModel`.

The stack shows infinite recursion through this cycle:
```
readFileModel → getEnhancedProperties → doReadFileModel → cache → readFileModel
```

## Stack trace (tail)

```
at org.apache.maven.impl.model.DefaultModelBuilder$ModelBuilderSessionState.cache(DefaultModelBuilder.java:2140)
at org.apache.maven.impl.model.DefaultModelBuilder$ModelBuilderSessionState.readFileModel(DefaultModelBuilder.java:1484)
at org.apache.maven.impl.model.DefaultModelBuilder$ModelBuilderSessionState.getEnhancedProperties(DefaultModelBuilder.java:705)
at org.apache.maven.impl.model.DefaultModelBuilder$ModelBuilderSessionState.doReadFileModel(DefaultModelBuilder.java:1653)
at org.apache.maven.impl.model.DefaultModelBuilder$ModelBuilderSessionState.lambda\$cache\$14(DefaultModelBuilder.java:2140)
at org.apache.maven.impl.cache.AbstractRequestCache.request(AbstractRequestCache.java:69)
... (repeating)
```

## Reproducer

```bash
git clone https://github.com/apache/jackrabbit-filevault.git
cd jackrabbit-filevault
mvn -B -e clean package -DskipTests
```

Build fails in ~5s with StackOverflowError during POM resolution, never reaching compilation.

## Environment

- Maven: built from `maven-4.0.x` branch (bootstrapped with 4.0.0-rc-5)
- JDK: 17 (Temurin)
- OS: Ubuntu (GitHub Actions runner)

## Context

Found during [Maven 4 compatibility testing](https://github.com/gnodet/maven4-testing/issues/20577) of Apache projects.

_Claude Code on behalf of Guillaume Nodet_

gnodet:
## Root Cause Analysis

### The Project Pattern That Triggers This

Jackrabbit FileVault uses an **internal parent** anti-pattern in combination with CI-friendly versions:

- The **root POM** (`pom.xml`) declares `<parent>` pointing to `parent/pom.xml` (a reactor module in the same repo)
[...]

PR #12317, opened 2026-06-18
[#12301] Fix StackOverflowError with internal parent and CI-friendly revision

gnodet:
Forward-port of #12314 from `maven-4.0.x` to `master`.

Fixes StackOverflowError when a project has an internal parent in a subdirectory using CI-friendly `${revision}` and a `.mvn/` root marker (GH-12301).

The fix adds cycle detection in `doReadFileModel()` via a shared `activeModelReads` set that tracks paths currently being read, preventing recursive loops between parent resolution and enhanced property resolution.

IT adapted for master (no version-range constructor).


slawekjaranowski:
@gnodet please check it, look like master branch stop build after this merge

PR #12322, opened 2026-06-19
[#12303] Fix CI-friendly ${revision} not interpolated for non-build POM reads

### LINKED ISSUE #12303: CI-friendly ${revision} not interpolated when used by maven-remote-resources-plugin
## Description

When building `bigtop-manager` with Maven from the `maven-4.0.x` branch, the `maven-remote-resources-plugin:3.0.0` attempts to resolve the project's own artifact using the literal string `${revision}` instead of the interpolated version value.

## Error

```
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-remote-resources-plugin:3.0.0:process
  (process-resource-bundles) on project bigtop-manager-stack-bigtop:
  Execution process-resource-bundles of goal
  org.apache.maven.plugins:maven-remote-resources-plugin:3.0.0:process failed:
  Invalid Version Range Request: org.apache.bigtop:bigtop-manager:pom:${revision}
```

The `${revision}` placeholder should have been interpolated to the actual version before the plugin tried to resolve it.

## Context

The project uses CI-friendly versions (`${revision}`) as documented in Maven's CI-friendly documentation. This works correctly with Maven 3.x but fails with Maven 4.x.

## Reproducer

```bash
git clone https://github.com/apache/bigtop-manager.git
cd bigtop-manager
mvn -B -e clean package -DskipTests
```

## Environment

- Maven: built from `maven-4.0.x` branch (bootstrapped with 4.0.0-rc-5)
- JDK: 17 (Temurin)

## Context

Found during [Maven 4 compatibility testing](https://github.com/gnodet/maven4-testing/issues/20595).

gnodet:
## Summary

Fixes the `${revision}` CI-friendly version placeholder not being interpolated when Maven reads dependency or parent POMs outside of a build request (e.g. when the `maven-remote-resources-plugin` resolves the project's own POM).

- In `DefaultModelBuilder.doReadFileModel()`, the CI-friendly version interpolation for non-build requests was guarded by `else if (modelSource.getPath() != null)`. Since `ResolvedPathSource.getPath()` always returns `null`, the `replaceCiFriendlyVersion()` call was dead code.
- Changed to an unconditional `else` block so `getEnhancedProperties()` and `replaceCiFriendlyVersion()` always run for non-build requests.
- Added IT that exercises the fix with a dependency whose parent POM uses `${revision}` in its version.

Verified against the original use case (bigtop-manager project from #12303): without the fix, `process-resources` fails with `Invalid Version Range Request: ...pom:${revision}`; with the fix, the version resolves correctly to `1.2.0-SNAPSHOT`.

Fixes #12303

## Test plan

- [x] IT `MavenITgh12303CIFriendlyRevisionRemoteResourcesTest` passes
- [x] RED/GREEN verified: IT fails without fix, passes with fix
- [x] Verified against original bigtop-manager use case from #12303
- [ ] Existing CI-friendly version tests still pass
- [ ] No regressions in model builder behavior for build requests


Bukama:
> Run 'mvn spotless:apply' to fix these violations.

:)

Bukama (approved):

PR #12327, opened 2026-06-19
Feat: Align CP ordering with Maven 4

cstamas:
This PR aligns classpath ordering with Maven 4 (level order). All Maven 2/3 versions released so far has "pre-order" flattening (historical reasons, Maven 2 did it like that). Maven 3.10.0 switches to Resolver 2, where resolver exposes flattening configuration too, and defaults to safer "level order", preventing issues like described here https://arxiv.org/abs/2407.18760v3

For resolver configuration, see configuration key `aether.system.dependencyVisitor` in 
https://maven.apache.org/resolver/configuration.html If _full Maven 3 behaviour needed_, user can simply set `-Daether.system.dependencyVisitor=preOrder` and it restores CP ordering to that used in Maven 3.9 and older.

This PR changes:
* remove scattered "flattening" happening in multiple spots in Maven, it is (and in fact was, but result was thrown away) fully delegating to Resolver.
* simplify spots that was doing repeated deprecated pre-order flattening (which already happened in plugin dependencies resolver)

### DIFF
--- a/maven-core/src/main/java/org/apache/maven/plugin/internal/DefaultPluginDependenciesResolver.java
+++ b/maven-core/src/main/java/org/apache/maven/plugin/internal/DefaultPluginDependenciesResolver.java
@@ -57,6 +57,7 @@ import org.eclipse.aether.resolution.ArtifactRequest;
 import org.eclipse.aether.resolution.ArtifactResolutionException;
 import org.eclipse.aether.resolution.DependencyRequest;
 import org.eclipse.aether.resolution.DependencyResolutionException;
+import org.eclipse.aether.resolution.DependencyResult;
 import org.eclipse.aether.util.artifact.JavaScopes;
 import org.eclipse.aether.util.filter.AndDependencyFilter;
 import org.eclipse.aether.util.filter.ScopeDependencyFilter;
@@ -98,6 +99,7 @@ public class DefaultPluginDependenciesResolver implements PluginDependenciesReso
                 session.getArtifactTypeRegistry().get("maven-plugin"));
     }
 
+    @Override
     public Artifact resolve(Plugin plugin, List<RemoteRepository> repositories, RepositorySystemSession session)
             throws PluginResolutionException {
         RequestTrace trace = RequestTrace.newChild(null, plugin);
@@ -161,10 +163,36 @@ public class DefaultPluginDependenciesResolver implements PluginDependenciesReso
             List<RemoteRepository> repositories,
             RepositorySystemSession session)
             throws PluginResolutionException {
+        return resolveInternal(plugin, null /* pluginArtifact */, dependencyFilter, repositories, session)
+                .getRoot();
+    }
+
+    /**
+     * @since 3.10.0
+     */
+    public DependencyResult resolveCoreExtensionAndFlatten(
+            Plugin plugin,
+            DependencyFilter dependencyFilter,
+            List<RemoteRepository> repositories,
+            RepositorySystemSession session)
+            throws PluginResolutionException {
         return resolveInternal(plugin, null /* pluginArtifact */, dependencyFilter, repositories, session);
     }
 
+    @Override
     public DependencyNode resolve(
+            Plugin plugin,
+            Artifact pluginArtifact,
+            DependencyFilter dependencyFilter,
+            List<RemoteRepository> repositories,
+            RepositorySystemSession session)
+            throws PluginResolutionException {
+        return resolveInternal(plugin, pluginArtifact, dependencyFilter, repositories, session)
+                .getRoot();
+    }
+
+    @Override
+    public DependencyResult resolveAndFlatten(
             Plugin plugin,
             Artifact pluginArtifact,
             DependencyFilter dependencyFilter,
@@ -174,7 +202,7 @@ public class DefaultPluginDependenciesResolver implements PluginDependenciesReso
         return resolveInternal(plugin, pluginArtifact, dependencyFilter, repositories, session);
     }
 
-    private DependencyNode resolveInternal(
+    private DependencyResult resolveInternal(
             Plugin plugin,
             Artifact pluginArtifact,
             DependencyFilter dependencyFilter,
@@ -189,8 +217,6 @@ public class DefaultPluginDependenciesResolver implements PluginDependenciesReso
 
         DependencyFilter collectionFilter = new ScopeDependencyFilter("provided", "test");
         DependencyFilter resolutionFilter = AndDependencyFilter.newInstance(collectionFilter, dependencyFilter);
-
-        DependencyNode node;
 
         try {
             DependencySelector selector =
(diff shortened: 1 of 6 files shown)


maven-core/src/main/java/org/apache/maven/plugin/internal/PluginDependenciesResolver.java
```diff
@@ -61,12 +62,34 @@ Artifact resolve(Plugin plugin, List<RemoteRepository> repositories, RepositoryS
      * @param session The repository session to use for resolving the plugin artifacts, must not be {@code null}.
      * @return The dependency tree denoting the resolved plugin class path, never {@code null}.
      * @throws PluginResolutionException If any dependency could not be resolved.
+     * @deprecated This method should be avoided, as it needs manual flattening, instead to let Resolver do it. Use {@link #resolveAndFlatten(Plugin, Artifact, DependencyFilter, List, RepositorySystemSession)} instead.
```
gnodet: Nit: `"instead to let Resolver do it"` reads awkwardly. Suggestion:

```suggestion
     * @deprecated This method should be avoided, as it requires manual flattening; use {@link #resolveAndFlatten(Plugin, Artifact, DependencyFilter, List, RepositorySystemSession)} instead to let Resolver handle it.
```

maven-core/src/main/java/org/apache/maven/plugin/internal/PluginDependenciesResolver.java
```diff
@@ -61,12 +62,34 @@ Artifact resolve(Plugin plugin, List<RemoteRepository> repositories, RepositoryS
...
+     * @param plugin The plugin for which to resolve the dependencies, must not be {@code null}.
+     * @param pluginArtifact The plugin's main artifact, may be {@code null}.
+     * @param dependencyFilter A filter to exclude artifacts from resolution (but not collection), may be {@code null}.
+     * @param repositories The plugin repositories to use for resolving the plugin artifacts, must not be {@code null}.
+     * @param session The repository session to use for resolving the plugin artifacts, must not be {@code null}.
+     * @return The dependency resolution result having the resolved plugin class path but also the tree, never {@code null}.
+     * @throws PluginResolutionException If any dependency could not be resolved.
+     * @since 3.10.0
+     */
+    DependencyResult resolveAndFlatten(
```
gnodet: Question: on master (Maven 4), the equivalent method is named `resolvePlugin()` and `resolveCoreExtension()` already returns `DependencyResult` directly. The `AndFlatten` suffix is arguably more descriptive, but it introduces a naming divergence from Maven 4 — which is the opposite of the PR's stated alignment goal. Was this intentional, or would `resolvePlugin` / keeping `resolveCoreExtension` (with a changed return type) be preferred for consistency?

maven-core/src/main/java/org/apache/maven/plugin/internal/DefaultPluginDependenciesResolver.java
```diff
@@ -161,20 +163,46 @@ public DependencyNode resolveCoreExtension(
...
             RepositorySystemSession session)
             throws PluginResolutionException {
+        return resolveInternal(plugin, null /* pluginArtifact */, dependencyFilter, repositories, session)
+                .getRoot();
+    }
+
+    /**
+     * @since 3.10.0
+     */
+    public DependencyResult resolveCoreExtensionAndFlatten(
```
gnodet, line 176: This method is `public` but not declared on the `PluginDependenciesResolver` interface. `BootstrapCoreExtensionManager` calls it, so it depends on the concrete type. The old `resolveCoreExtension` had the same pattern, but since the interface is already being touched to add `resolveAndFlatten`, would it make sense to also promote this to the interface?

gnodet (commented):
Nice work aligning Maven 3.10.x classpath ordering with Maven 4 by delegating flattening to Resolver 2. The core refactoring is clean and consistent across call sites.

A few observations:

**Scope drift:** The `getTransferStartTime()` → `getStartTime().toEpochMilli()` changes in the two transfer listener files are a Resolver 2 API migration unrelated to classpath ordering. Consider splitting them into a separate commit (or PR) to keep scope clean and ease bisection.

**Master porting:** I compared against master (Maven 4) and the plugin dependency resolver refactoring is already done there — `resolvePlugin()` returns `DependencyResult`, callers use it directly, no `PreorderNodeListGenerator` anywhere. **No porting needed** for the core change.

However, there is one transfer listener inconsistency on master worth fixing separately: `impl/maven-cli/.../Slf4jMavenTransferListener.java:91` uses `Instant.ofEpochMilli(resource.getTransferStartTime())` while `impl/maven-cli/.../AbstractMavenTransferListener.java:86` already uses the newer `resource.getStartTime()` directly. The `Slf4jMavenTransferListener` should be updated to match.

_Note: this review does not replace specialized AI review tools (CodeRabbit, Sourcery) or static analyzers (SonarCloud)._

maven-core/src/main/java/org/apache/maven/plugin/internal/PluginDependenciesResolver.java
cstamas: I dont see `resolveCoreExtension()` on https://github.com/apache/maven/blob/bfc30f1fa89741fdafdfa26725b5e0795a619708/impl/maven-core/src/main/java/org/apache/maven/plugin/internal/PluginDependenciesResolver.java

But yes, those should be aligned, and new methods marked as `@since 3.10.0`

maven-core/src/main/java/org/apache/maven/plugin/internal/DefaultPluginDependenciesResolver.java
cstamas: Agreed, will add new (and old) methods to interface. I just did not want to "stir too much".

cstamas:
**Scope drift**: fixed by https://github.com/apache/maven/pull/12328
**Master porting**: aligned master with https://github.com/apache/maven/pull/12329

PR comments applied as well

gnodet (approved):
All feedback from the first review has been addressed:

- **Scope drift fixed**: transfer listener changes extracted to #12328
- **Master alignment**: addressed via #12329
- **Interface promotion**: both `resolveCoreExtensionAndFlatten` and `resolvePluginAndFlatten` are now on the `PluginDependenciesResolver` interface
- **Javadoc typo**: fixed
- **Bonus**: `BootstrapCoreExtensionManager` now injects the `PluginDependenciesResolver` interface instead of the `DefaultPluginDependenciesResolver` concrete type — proper DI

The PR is now cleanly focused on its stated goal: delegating classpath flattening to Resolver 2 and aligning with Maven 4's level-order default. The refactoring is consistent across all call sites, `PreorderNodeListGenerator` is fully removed, and the escape hatch via `-Daether.system.dependencyVisitor=preOrder` is documented.

_Note: this review does not replace specialized AI review tools (CodeRabbit, Sourcery) or static analyzers (SonarCloud)._

PR #12329, opened 2026-06-19
Forward port Maven 3.10.0 PluginDependenciesResolver changes

cstamas:
Also, restore binary compatibility that was broken in 4-beta-2.

This PR aligns PluginDependenciesResolver with changes happened in Maven 3.10.x, and also restores binary incompatibility happened in DefaultPluginDependenciesResolver that used to be injected directly.

### DIFF
--- a/impl/maven-core/src/main/java/org/apache/maven/plugin/internal/PluginDependenciesResolver.java
+++ b/impl/maven-core/src/main/java/org/apache/maven/plugin/internal/PluginDependenciesResolver.java
@@ -35,6 +35,7 @@ import org.eclipse.aether.resolution.DependencyResult;
  * changed or deleted without prior notice.
  *
  * @since 3.0
+ * @author Benjamin Bentmann
  */
 public interface PluginDependenciesResolver {
 
@@ -61,7 +62,9 @@ public interface PluginDependenciesResolver {
      * @param session The repository session to use for resolving the plugin artifacts, must not be {@code null}.
      * @return The dependency tree denoting the resolved plugin class path, never {@code null}.
      * @throws PluginResolutionException If any dependency could not be resolved.
+     * @deprecated This method should be avoided, as it requires manual flattening; use {@link #resolvePluginAndFlatten(Plugin, Artifact, DependencyFilter, List, RepositorySystemSession)} instead to let Resolver handle it.
      */
+    @Deprecated
     DependencyNode resolve(
             Plugin plugin,
             Artifact pluginArtifact,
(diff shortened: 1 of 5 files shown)


cstamas:
@gnodet master is broken with this same error

impl/maven-core/src/main/java/org/apache/maven/plugin/internal/PluginDependenciesResolver.java
```diff
@@ -35,6 +35,7 @@
  * changed or deleted without prior notice.
  *
  * @since 3.0
+ * @author Benjamin Bentmann
```
gnodet: Minor: The Maven 4.x codebase consistently avoids `@author` tags (zero files in `impl/maven-core/src/main/java/` use them). This appears carried over from the 3.10.x change but is inconsistent with the 4.x conventions. Consider removing it.

gnodet (commented):
Nice work aligning the Maven 4.x `PluginDependenciesResolver` with the 3.10.x changes and restoring the binary compatibility broken in 4-beta-2. The deprecation+delegation pattern is correctly applied, and switching to interface-based injection in `BootstrapCoreExtensionManager` is the right move.

A couple more observations not tied to specific lines:

- **TODO comment**: Both `BootstrapCoreExtensionManager` files (compat and impl) have a TODO comment above the call site that still references `resolveCoreExtension`. Since the method is now `resolveCoreExtensionAndFlatten` and the method is now on the interface, the TODO may be partially addressed and could be refined.
- **`@since 3.10.0` on new interface methods**: The new methods use `@since 3.10.0`, which correctly references where they originated, but since Maven 4.x and 3.10.x are separate release lines, consumers reading the Maven 4.x javadoc might find this confusing. Flagging for awareness only.
- **No tests added**: No new or modified tests, which is reasonable for an interface evolution / binary compat PR where the behavior is unchanged (new methods delegate to the same `resolveInternal` logic).

_This review does not replace specialized AI review tools (CodeRabbit, Sourcery) or static analysis (SonarCloud)._

impl/maven-core/src/main/java/org/apache/maven/plugin/internal/PluginDependenciesResolver.java
```diff
@@ -70,11 +72,63 @@ DependencyNode resolve(
...
+    /**
+     * Resolves the runtime dependencies of the specified core extension (as {@link Plugin} as GAV carrier).
+     *
+     * @param plugin The plugin for which to resolve the dependencies, must not be {@code null}.
+     * @param dependencyFilter A filter to exclude artifacts from resolution (but not collection), may be {@code null}.
+     * @param repositories The plugin repositories to use for resolving the plugin artifacts, must not be {@code null}.
+     * @param session The repository session to use for resolving the plugin artifacts, must not be {@code null}.
+     * @return The dependency resolution result having the resolved extension class path but also the tree, never {@code null}.
+     * @throws PluginResolutionException If any dependency could not be resolved.
+     * @since 3.10.0
```
slawekjaranowski, line 84: maybe we can also add a 4.x version here, or use only 4.x as @gnodet mention

slawekjaranowski (approved):

PR #12334, opened 2026-06-20
[#12331] Fix clean/site lifecycle bindings executing after pom.xml goals

### LINKED ISSUE #12331: goal bindings in pom.xml executed BEFORE lifecycle ones with Maven 4
### Affected version

4.0.0-RC5

### Bug description

found while working on https://github.com/apache/maven-scm-publish-plugin/pull/152#issuecomment-4717823539

for default lifecycle, Maven 4 works as Maven 3: goal binding from `pom.xml` runs after lifecycle one

but for `site` or `clean`, Maven 4 runs binding from `pom.xml` first

reproducer: [pom.xml](https://github.com/user-attachments/files/29162847/pom.xml)

with Maven 3:
```
❯ mvn package && mvn -V clean
Apache Maven 3.9.12 (848fbb4bf2d427b72bdb2471c22fced7ebd9a7a1)
Maven home: /home/herve/.sdkman/candidates/maven/3.9.12
Java version: 25, vendor: Azul Systems, Inc., runtime: /home/herve/.sdkman/candidates/java/25-zulu
Default locale: fr_FR, platform encoding: UTF-8
OS name: "linux", version: "6.8.0-124-generic", arch: "amd64", family: "unix"
[INFO] 
[INFO] ----------------< org.apache.maven.its:test-lifecycle >-----------------
[INFO] Building Test lifecycle bindings vs added ones 1.0-SNAPSHOT
[INFO]   from pom.xml
[INFO] --------------------------------[ jar ]---------------------------------
[INFO] 
[INFO] --- clean:3.2.0:clean (default-clean) @ test-lifecycle ---
[INFO] Deleting /home/herve/dev/maven/sources/plugins/tools/target
[INFO] 
[INFO] --- antrun:3.2.0:run (target) @ test-lifecycle ---
[INFO] Executing tasks
```
(failure expected)

with Maven 4:
```
❯ mvn package && mvn -V clean
Apache Maven 4.0.0-SNAPSHOT (61b7c60e4211f024564ff89964dac825afb3146e)
Maven home: /home/herve/.sdkman/candidates/maven/4.0.0-SNAPSHOT
Java version: 25, vendor: Azul Systems, Inc., runtime: /home/herve/.sdkman/candidates/java/25-zulu
Default locale: fr_FR, platform encoding: UTF-8, time zone: Europe/Paris
OS name: "linux", version: "6.8.0-124-generic", arch: "amd64", family: "unix"
[INFO] 
[INFO] -----------------------------------------< org.apache.maven.its:test-lifecycle >------------------------------------------
[INFO] Building Test lifecycle bindings vs added ones 1.0-SNAPSHOT
[INFO]   from pom.xml
[INFO] ---------------------------------------------------------[ jar ]----------------------------------------------------------
[INFO] 
[INFO] --- antrun:3.2.0:run (target) @ test-lifecycle ---
[INFO] Executing tasks
[INFO] Executed tasks
[INFO] 
[INFO] --- clean:3.4.0:clean (default-clean) @ test-lifecycle ---
[INFO] Deleting /home/herve/dev/maven/sources/plugins/tools/target
```

same results with `mvn site`

hboutemy:
## Summary

Fix goal bindings in `pom.xml` being executed **before** lifecycle bindings for
the `clean` and `site` lifecycles (regression vs Maven 3, fixes #12331).

## Root cause

`Lifecycles.plugin()` creates executions for built-in `clean` and `site`
lifecycle bindings (e.g. `default-clean`) without setting a priority, so they
defaulted to `0` — the same as user-defined executions from `pom.xml`.

`DefaultLifecycleMappingDelegate` sorts executions within a phase by priority.
When both a lifecycle binding and a pom.xml execution share priority `0`,
ordering falls back to plugin iteration order. Because `LifecycleBindingsMerger`
inserts pom.xml plugins before lifecycle-only plugins, pom.xml executions ended
up running first.

For the `default` lifecycle this was not a problem: packaging bindings (e.g. for
`jar`) are built by `DefaultPackagingRegistry` with `priority(i - mojos.size())`
— always negative — so they already sort before user executions.

## Fix

Add `.priority(-1)` to the `PluginExecution` built in `Lifecycles.plugin()`,
making `clean` and `site` lifecycle bindings consistent with how `default`
lifecycle packaging bindings work.

## Test plan

- [x] `LifecycleExecutorTest`, `DefaultLifecyclesTest`, `BuildPlanCreatorTest`
      all pass (561 tests, 0 failures in `maven-core`)
- [x] Manual reproduction from #12331: `mvn clean` with an antrun execution
      bound to the `clean` phase now runs `default-clean` (maven-clean-plugin)
      before the pom.xml antrun execution, matching Maven 3 behavior

### DIFF
--- a/impl/maven-core/src/main/java/org/apache/maven/internal/impl/Lifecycles.java
+++ b/impl/maven-core/src/main/java/org/apache/maven/internal/impl/Lifecycles.java
@@ -74,6 +74,7 @@ public class Lifecycles {
                 .version(c[2])
                 .executions(Collections.singletonList(PluginExecution.newBuilder()
                         .id("default-" + c[3])
+                        .priority(-1)
                         .phase(phase)
                         .goals(Collections.singletonList(c[3]))
                         .location("", DefaultLifecycleRegistry.DEFAULT_LIFECYCLE_INPUT_LOCATION)
--- a/impl/maven-core/src/test/java/org/apache/maven/lifecycle/LifecycleExecutorTest.java
+++ b/impl/maven-core/src/test/java/org/apache/maven/lifecycle/LifecycleExecutorTest.java
@@ -314,6 +314,24 @@ class LifecycleExecutorTest extends AbstractCoreMavenComponentTestCase {
     }
 
     @Test
+    void testCleanLifecycleBindingRunsBeforePomExecution() throws Exception {
+        File pom = getProject("project-with-clean-phase-execution");
+        MavenSession session = createMavenSession(pom);
+        assertEquals(
+                "project-with-clean-phase-execution",
+                session.getCurrentProject().getArtifactId());
+        List<MojoExecution> executionPlan = getExecutions(calculateExecutionPlan(session, "clean"));
+        assertEquals(2, executionPlan.size());
+        assertListEquals(
+                List.of("clean:clean", "it:it"),
+                executionPlan.stream()
+                        .map(e -> e.getMojoDescriptor().getFullGoalName())
+                        .toList());
+        assertEquals("default-clean", executionPlan.get(0).getExecutionId());
+        assertEquals("user-clean-execution", executionPlan.get(1).getExecutionId());
+    }
+
+    @Test
     void testInvalidGoalName() throws Exception {
         File pom = getProject("project-basic");
         MavenSession session = createMavenSession(pom);
--- a/impl/maven-core/src/test/projects/lifecycle-executor/project-with-clean-phase-execution/pom.xml
+++ b/impl/maven-core/src/test/projects/lifecycle-executor/project-with-clean-phase-execution/pom.xml
@@ -1,0 +1,92 @@
+<?xml version="1.0" encoding="UTF-8"?>
+
+<!--
+Licensed to the Apache Software Foundation (ASF) under one
+or more contributor license agreements.  See the NOTICE file
+distributed with this work for additional information
+regarding copyright ownership.  The ASF licenses this file
+to you under the Apache License, Version 2.0 (the
+"License"); you may not use this file except in compliance
+with the License.  You may obtain a copy of the License at
+
+  http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing,
+software distributed under the License is distributed on an
+"AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+KIND, either express or implied.  See the License for the
+specific language governing permissions and limitations
+under the License.
+-->
+
+<project>
+  <modelVersion>4.0.0</modelVersion>
+
+  <groupId>org.apache.maven.lifecycle.test</groupId>
+  <artifactId>project-with-clean-phase-execution</artifactId>
+  <version>1.0</version>
+  <packaging>jar</packaging>
+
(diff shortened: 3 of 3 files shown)


gnodet (commented):

## Review of PR #12334

### Root Cause Analysis — Verified

I traced the full execution path. The PR description is accurate.

`DefaultLifecycleRegistry.CleanLifecycle` (line 429) and `SiteLifecycle` (line 547) call `Lifecycles.plugin(...)` to create lifecycle bindings. Before this PR, the `PluginExecution` was built without setting `priority`, defaulting to `0` — the same as user-defined executions from pom.xml.

`DefaultLifecycleMappingDelegate` (line 99) sorts executions within a phase using a `TreeMap<PhaseId, ...>` with `PhaseComparator`, which sorts by `Integer.compare(p1.priority(), p2.priority())` (line 65) — lower priority number first. When priorities are tied (both `0`), ordering falls to plugin iteration order.

`LifecycleBindingsMerger.mergePluginContainer_Plugins` inserts pom.xml plugins into the `LinkedHashMap` first (as target), lifecycle plugins second (as source). So at equal priority, pom.xml executions come first — causing the regression.

For comparison, `DefaultPackagingRegistry` (line 155) sets `priority(i - mojos.size())` for `default` lifecycle packaging bindings, always producing negative values. This is why the `default` lifecycle was unaffected.

### The Fix — Correct

Adding `.priority(-1)` makes clean/site bindings sort before user executions (priority `0`) in both execution paths:

1. **Sequential path**: `DefaultLifecycleMappingDelegate` → `PhaseComparator` → `Integer.compare(-1, 0)` → lifecycle binding first
2. **Concurrent path**: `BuildPlanExecutor` (line 600) → `BuildStep.addMojo(mojo, priority)` → `TreeMap<Integer, ...>` key ordering → lifecycle binding first

The `LifecycleBindingsMerger.mergePluginExecution_Priority` also handles this correctly — it keeps `min(target, source)`, so the `-1` from lifecycle bindings is preserved during model merging. The compat layer (`DefaultLifecycleBindingsInjector`, line 154) uses `Math.min()` — same correct behavior.

### Observations

**The value `-1` is consistent.** `DefaultPackagingRegistry` uses `i - mojos.size()` which for a single mojo (i=0, size=1) produces exactly `-1`. Since `Lifecycles.plugin()` always creates a single execution, hardcoding `-1` is equivalent.

**No test for the fixed behavior.** Existing tests pass and manual reproduction confirms the fix, but there's no new test that asserts lifecycle bindings execute before pom.xml executions in the `clean`/`site` lifecycle. This is the kind of regression that could easily reoccur if someone refactors lifecycle binding creation. A unit test or integration test verifying the execution order (like the manual reproducer from #12331) would guard against that.

**The PR description is excellent** — clear root cause, comparison with the working `default` lifecycle path, and a concise fix section.

gnodet (approved):

Fix is correct, minimal, and now has test coverage. Approving.

The `.priority(-1)` value is consistent with `DefaultPackagingRegistry` (which produces `-1` for single-mojo bindings via `i - mojos.size()`), both execution paths (sequential via `PhaseComparator` and concurrent via `BuildStep.mojos` TreeMap) sort correctly, and the merger preserves the lower priority during model merging.

PR #12337, opened 2026-06-22
Retry MNG-4559 jvm.config ITs on sporadic Windows failure

gnodet:
## Summary

- On Windows CI, `mvn.cmd` parses `.mvn/jvm.config` via a temp file that `for /f` occasionally fails to read (likely antivirus or filesystem flush delay), leaving system properties as unresolved `${...}` placeholders
- Add a targeted retry in `MavenITmng4559MultipleJvmArgsTest` and `MavenITmng4559SpacesInJvmOptsTest`: if the first Maven execution produces unresolved properties, re-execute once before asserting
- No-op on Linux/macOS since the retry path only triggers when properties contain `${...}` placeholders

## Test plan

- [ ] Verify Windows CI passes consistently for both MNG-4559 tests
- [ ] Verify Linux/macOS CI is unaffected (retry path never triggered)


cstamas (approved):

gnodet:
Superseded by #12378 which fixes the root cause in mvn.cmd (replacing `for /f` with `set /p` to avoid Windows file lock issues) rather than adding test-level retries.

PR #12338, opened 2026-06-22
Switch default resolver transport from JDK/methanol to Apache HttpClient

gnodet:
## Summary

- The JDK HTTP transport (backed by methanol) has proven flaky on Windows CI with race conditions in the gzip/inflater code path (`NullPointerException: Inflater has been closed`)
- Remove `maven-resolver-transport-jdk` from the distribution entirely, making Apache HttpClient the only bundled HTTP transport
- Users who need JDK transport can still add the dependency manually to their extension classpath

## Changes

- `apache-maven/pom.xml`: removed `maven-resolver-transport-jdk` dependency from the distribution; updated comment on `transport-apache` to reflect it is now the default
- `impl/maven-cli/pom.xml`: switched `mvnup` dependency from `transport-jdk` to `transport-apache`
- `AbstractUpgradeStrategy` (mvnup): switched DI provider from `JdkTransporterFactory` to `ApacheTransporterFactory`
- `MavenITmng7470ResolverTransportTest`: default transport test now expects `ApacheTransporter` log snippet; removed `testResolverTransportJdk` test (JDK transport no longer shipped)

## Notes

- `DefaultRepositorySystemSessionFactory` is **not** changed — the `"default"` transport mode still falls through to `auto`, which selects the highest-priority available transport. Since `transport-jdk` is no longer on the classpath, auto now resolves to Apache HttpClient.
- `-Dmaven.resolver.transport=jdk` will fail at runtime since the transport is no longer bundled.

## Test plan

- [x] `MavenITmng7470ResolverTransportTest` — default transport now logs `ApacheTransporter`
- [ ] Full CI run on all platforms

### DIFF
--- a/impl/maven-core/src/main/java/org/apache/maven/internal/aether/DefaultRepositorySystemSessionFactory.java
+++ b/impl/maven-core/src/main/java/org/apache/maven/internal/aether/DefaultRepositorySystemSessionFactory.java
@@ -312,7 +312,9 @@ public class DefaultRepositorySystemSessionFactory implements RepositorySystemSe
         Object transport =
                 mergedProps.getOrDefault(Constants.MAVEN_RESOLVER_TRANSPORT, MAVEN_RESOLVER_TRANSPORT_DEFAULT);
         if (MAVEN_RESOLVER_TRANSPORT_DEFAULT.equals(transport)) {
-            // The "default" mode (user did not set anything) from now on defaults to AUTO
+            // Default transport is Apache HTTP client; JDK transport available via -Dmaven.resolver.transport=jdk
+            configProps.put(FILE_TRANSPORTER_PRIORITY_KEY, RESOLVER_MAX_PRIORITY);
+            configProps.put(APACHE_HTTP_TRANSPORTER_PRIORITY_KEY, RESOLVER_MAX_PRIORITY);
         } else if (MAVEN_RESOLVER_TRANSPORT_JDK.equals(transport)) {
             // Make sure (whatever extra priority is set) that resolver file/jdk is selected
             configProps.put(FILE_TRANSPORTER_PRIORITY_KEY, RESOLVER_MAX_PRIORITY);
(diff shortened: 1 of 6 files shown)


cstamas:
@gnodet bundle re-declares it, so it needs to be removed as well
https://github.com/apache/maven/blob/master/apache-maven/pom.xml#L99-L103

impl/maven-core/src/main/java/org/apache/maven/internal/aether/DefaultRepositorySystemSessionFactory.java
```diff
@@ -312,7 +312,9 @@ public SessionBuilder newRepositorySessionBuilder(MavenExecutionRequest request)
         Object transport =
                 mergedProps.getOrDefault(Constants.MAVEN_RESOLVER_TRANSPORT, MAVEN_RESOLVER_TRANSPORT_DEFAULT);
         if (MAVEN_RESOLVER_TRANSPORT_DEFAULT.equals(transport)) {
-            // The "default" mode (user did not set anything) from now on defaults to AUTO
+            // Default transport is Apache HTTP client; JDK transport available via -Dmaven.resolver.transport=jdk
```
cstamas: This is wrong. You just removed jdk transport, so it is not available via....

impl/maven-core/src/test/java/org/apache/maven/internal/aether/DefaultRepositorySystemSessionFactoryTest.java
```diff
@@ -364,8 +364,18 @@ void transportConfigurationTest() throws InvalidRepositoryException {
         MavenExecutionRequest request = new DefaultMavenExecutionRequest();
         request.setLocalRepository(getLocalRepository());
 
-        // native
+        // default (no property set) should select Apache
```
cstamas: No need to do anything, JDK is not on classpath, hence apache is default.

cstamas (approved):

kwin:
@gnodet Do you have some stack trace for `NullPointerException: Inflater has been closed`? Please report upstream at https://github.com/apache/maven-resolver in order to get this fixed in the JDK Client Transporter. If only GZIP is flaky we should try to debug in the according Methanol code. Also we could temporarily just disable that compression (https://mizosoft.github.io/methanol/decompression/).

PR #12369, opened 2026-06-25
[#11449] Fix Mockito agent: use late binding @{} interpolation and extend to failsafe plugin

### LINKED ISSUE #11449: Add Mockito as an agent

When running `mvnup/ApplicabilityTests` tests, there is a warning of Mockito that it should be added as an agent.

> Mockito is currently self-attaching to enable the inline-mock-maker. This will no longer work in future releases of the JDK. Please add Mockito as an agent to your build as described in Mockito's documentation: https://javadoc.io/doc/org.mockito/mockito-core/latest/org.mockito/org/mockito/Mockito.html#0.3
WARNING: A Java agent has been loaded dynamically (C:\Users\matth\.m2\repository\net\bytebuddy\byte-buddy-agent\1.17.7\byte-buddy-agent-1.17.7.jar)
WARNING: If a serviceability tool is in use, please run with -XX:+EnableDynamicAgentLoading to hide this warning
WARNING: If a serviceability tool is not in use, please run with -Djdk.instrument.traceUsage for more information
WARNING: Dynamic loading of agents will be disallowed by default in a future release
OpenJDK 64-Bit Server VM warning: Sharing is only supported for boot loader classes because bootstrap classpath has been appended

Hiteshsai007:
The mockito profile activated by dependency:properties used Maven property interpolation (\) which resolves at POM parse time - before the dependency:properties goal runs and sets the org.mockito:mockito-core:jar property. This caused the javaagent path to remain unresolved.

Fix: use Surefire/Failsafe late-binding property interpolation (@{...}) which resolves the property at test execution time, after dependency:properties has already set it.

Also:
- Restore @{jacocoArgLine} placeholder (was dropped in previous attempt)
- Extend the profile to cover maven-failsafe-plugin as well

Fixes: https://github.com/apache/maven/issues/11449

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -172,6 +172,8 @@ under the License.
     <woodstoxVersion>7.2.1</woodstoxVersion>
     <xmlunitVersion>2.12.0</xmlunitVersion>
     <jacocoArgLine />
+    <argLine.xmx>-Xmx256m</argLine.xmx>
+    <argLine.mockito />
 
     <!-- Set as system property in surefire/failsafe to propagate to tests and IT Verifier -->
     <toolboxVersion>0.14.1</toolboxVersion>
@@ -697,7 +699,7 @@ under the License.
           <artifactId>maven-surefire-plugin</artifactId>
           <configuration>
             <forkNode implementation="org.apache.maven.plugin.surefire.extensions.SurefireForkNodeFactory" />
-            <argLine>-Xmx256m @{jacocoArgLine}</argLine>
+            <argLine>${argLine.xmx} @{jacocoArgLine} ${argLine.mockito}</argLine>
             <systemPropertyVariables combine.children="append">
               <version.toolbox>${toolboxVersion}</version.toolbox>
             </systemPropertyVariables>
@@ -708,7 +710,7 @@ under the License.
           <artifactId>maven-failsafe-plugin</artifactId>
           <configuration>
             <forkNode implementation="org.apache.maven.plugin.surefire.extensions.SurefireForkNodeFactory" />
-            <argLine>-Xmx256m @{jacocoArgLine}</argLine>
+            <argLine>${argLine.xmx} @{jacocoArgLine} ${argLine.mockito}</argLine>
             <systemPropertyVariables combine.children="append">
               <version.toolbox>${toolboxVersion}</version.toolbox>
             </systemPropertyVariables>
@@ -945,26 +947,9 @@ under the License.
           <name>org.mockito:mockito-core:jar</name>
         </property>
       </activation>
-      <build>
-        <pluginManagement>
-          <plugins>
-            <plugin>
-              <groupId>org.apache.maven.plugins</groupId>
-              <artifactId>maven-surefire-plugin</artifactId>
-              <configuration>
-                <argLine>-Xmx256m @{jacocoArgLine} -javaagent:@{org.mockito:mockito-core:jar}</argLine>
-              </configuration>
-            </plugin>
-            <plugin>
-              <groupId>org.apache.maven.plugins</groupId>
-              <artifactId>maven-failsafe-plugin</artifactId>
-              <configuration>
-                <argLine>-Xmx256m @{jacocoArgLine} -javaagent:@{org.mockito:mockito-core:jar}</argLine>
-              </configuration>
-            </plugin>
-          </plugins>
-        </pluginManagement>
-      </build>
+      <properties>
+        <argLine.mockito>-javaagent:@{org.mockito:mockito-core:jar}</argLine.mockito>
+      </properties>
     </profile>
     <profile>
       <id>graph</id>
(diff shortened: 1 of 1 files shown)


gnodet (commented):
## Review — Fix Mockito agent: use late-binding `@{}` interpolation and extend to failsafe plugin

**What the PR does:** This 1-file, 8-line change fixes the `mockito` Maven profile so the Mockito Java agent path resolves correctly at test execution time rather than at POM parse time. It switches the surefire `<argLine>` from `${org.mockito:mockito-core:jar}` (eager, parse-time interpolation) to `@{org.mockito:mockito-core:jar}` (Surefire/Failsafe late-binding, resolves after `dependency:properties` has run). It also restores the `@{jacocoArgLine}` placeholder that the old mockito-profile argLine was missing, and extends the configuration to `maven-failsafe-plugin` for integration tests.

**Verdict: The change looks correct.** The fix aligns with how `@{jacocoArgLine}` is already used throughout this POM (lines 700, 711, 1220) and properly handles the case where `jacocoArgLine` is empty (it's defaulted to empty at line 174). The `pluginManagement` placement is consistent with the existing surefire entry in the profile, and Maven's configuration merge semantics mean only `<argLine>` is overridden — the `<forkNode>` and `<systemPropertyVariables>` from the default pluginManagement are inherited correctly.

### Findings

**No confirmed or plausible bugs were found in this PR's changes.**

### Pre-existing note (not introduced by this PR)

`pom.xml:1220` — When both the `jacoco` and `mockito` profiles are active simultaneously, the `jacoco` profile's **direct** `<build><plugins>` surefire config (`<argLine>-Xmx1G @{jacocoArgLine}</argLine>`) takes precedence over the `mockito` profile's `<build><pluginManagement>` surefire config, causing the `-javaagent:@{org.mockito:mockito-core:jar}` to be silently dropped for unit tests. This was already the case before this PR (the old `${...}` syntax had the same override behavior), so it's not a regression — but it may be worth a follow-up if both profiles are expected to be used together. Failsafe is unaffected since the jacoco profile doesn't override failsafe's argLine.

---

Hiteshsai007:
Thanks for the review! I've updated the PR to resolve the profile collision highlighted in the pre-existing note:

* **The Issue:** When both `jacoco` and `mockito` profiles were active, the `jacoco` profile's direct `surefire` plugin configuration overrode the `mockito` profile's `pluginManagement` configuration, dropping the Mockito `-javaagent`.
* **The Fix:** Refactored the `surefire` and `failsafe` `argLine` in `pom.xml` to be dynamically composed using properties:
  - Added default `<argLine.xmx>-Xmx256m</argLine.xmx>` and `<argLine.mockito />` properties.
  - Set the default plugin configuration to: `${argLine.xmx} @{jacocoArgLine} ${argLine.mockito}`.
  - The `mockito` profile now only defines the `<argLine.mockito>` property.
  - The `jacoco` profile now only overrides `<argLine.xmx>` to `-Xmx1G` (and no longer overrides the plugin's `argLine` configuration directly).

This ensures both profiles can be active at the same time without conflicting or dropping each other's parameters.

gnodet (commented):
## Follow-up Review — Commit 2: property-based composition

The second commit (`2db4fbbe`) replaces the competing `<argLine>` overrides with composable Maven properties — a much cleaner design that eliminates the profile collision entirely.

### How it works

1. Two new properties with safe defaults: `<argLine.xmx>-Xmx256m</argLine.xmx>` and `<argLine.mockito />` (empty).
2. A single shared argLine template in pluginManagement for **both** surefire and failsafe:
   ```xml
   <argLine>${argLine.xmx} @{jacocoArgLine} ${argLine.mockito}</argLine>
   ```
3. Profiles now compose by overriding properties instead of fighting over `<argLine>`:
   - **mockito** profile sets `<argLine.mockito>-javaagent:@{org.mockito:mockito-core:jar}</argLine.mockito>`
   - **jacoco** profile sets `<argLine.xmx>-Xmx1G</argLine.xmx>` and removes its own `<argLine>` from surefire config

When **both** profiles are active, the effective argLine becomes `-Xmx1G @{jacocoArgLine} -javaagent:@{org.mockito:mockito-core:jar}` — everything composes correctly. ✅

### Interpolation layering

The two-step interpolation is correct:
- **Parse time** (`${...}`): `${argLine.mockito}` expands to the literal string `-javaagent:@{org.mockito:mockito-core:jar}` — Maven's `${}` interpolation does not touch the `@{}` tokens.
- **Fork time** (`@{...}`): Surefire's late-binding resolves `@{org.mockito:mockito-core:jar}` after `dependency:properties` has set it. ✅

### One minor behavior change to be aware of

Previously, the `jacoco` profile only overrode surefire's `-Xmx` to `1G` — failsafe kept the default `256m`. With the property-based approach, `<argLine.xmx>-Xmx1G</argLine.xmx>` now applies to **both** surefire and failsafe when jacoco is active. This is probably desirable (JaCoCo instrumentation uses more memory regardless of test type), but it is a change from the old behavior.

### Verdict

No bugs found. This is a clean, well-designed approach that solves the original issue AND the profile collision. The property-based composition is the right pattern for Maven profile interop. 👍

---

PR #12402, opened 2026-06-29
Try to use RAT 0.18 - help needed with Maven project structure

ottlinger:
Trying to add configuration for RAT 0.18, but the basic scan does not pass and reports many thousand new unlicensed files .... as I'm not sure about the Maven project layout I don't know where RAT is configured for the whole project.

Can you help out? @slachiewicz @slawekjaranowski

### DIFF
--- a/NOTICE
+++ b/NOTICE
@@ -2,4 +2,4 @@ Apache Maven
 Copyright 2001-2026 The Apache Software Foundation
 
 This product includes software developed at
-The Apache Software Foundation (http://www.apache.org/).
+The Apache Software Foundation (https://www.apache.org/).
--- a/its/pom.xml
+++ b/its/pom.xml
@@ -372,7 +372,7 @@ under the License.
                   <pluginExecutionFilter>
                     <groupId>org.apache.rat</groupId>
                     <artifactId>apache-rat-plugin</artifactId>
-                    <versionRange>[0.11,)</versionRange>
+                    <versionRange>[0.18,)</versionRange>
                     <goals>
                       <goal>check</goal>
                     </goals>
--- a/pom.xml
+++ b/pom.xml
@@ -810,25 +810,29 @@ under the License.
         <plugin>
           <groupId>org.apache.rat</groupId>
           <artifactId>apache-rat-plugin</artifactId>
+          <version>0.18</version><!-- Remove if correct version comes from parent -->
           <configuration>
-            <excludes>
+            <inputExcludeStds>
+              <exclude>ALL</exclude>
+            </inputExcludeStds>
+            <inputExcludeParsedScm>GIT</inputExcludeParsedScm>
+            <inputExcludes>
               <exclude>Jenkinsfile</exclude>
               <exclude>**/.gitattributes</exclude>
-              <exclude>src/main/resources/META-INF/services/**</exclude>
+              <exclude>src/main/resources/META-INF/services</exclude>
               <exclude>src/test/resources*/**</exclude>
-              <exclude>src/test/projects/**</exclude>
-              <exclude>src/test/remote-repo/**</exclude>
-              <exclude>its/**</exclude>
+              <exclude>src/test/projects</exclude>
+              <exclude>src/test/remote-repo</exclude>
+              <exclude>its</exclude>
               <exclude>**/*.odg</exclude>
               <exclude>**/*.svg</exclude>
               <exclude>.asf.yaml</exclude>
-              <exclude>.mvn/**</exclude>
-              <exclude>.jbang/**</exclude>
+              <exclude>.jbang</exclude>
               <!--
                 ! Excluded the license files itself cause they do not have a license of themselves.
               -->
               <exclude>src/main/appended-resources/licenses/**</exclude>
-            </excludes>
+            </inputExcludes>
           </configuration>
         </plugin>
       </plugins>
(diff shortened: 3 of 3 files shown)


cstamas:
The other day was tinkering with same problem, and we have issues as well:
https://issues.apache.org/jira/browse/RAT-567
https://issues.apache.org/jira/browse/RAT-554

Idea is that Maven POM should **really not contain RAT config**, merely a plugin definition with one liner config, that points to some sort of RAT config file, presumably in repository root. This way, that file would end up in source bundle as well, and could be picked up by tools like ATR is, as currently we have divergence between config used in build by Maven, ATR reports unrelated errors, etc.

cstamas:
RAT plugin as I found, top to bottom:
ASF parent POM: https://github.com/apache/maven-apache-parent/blob/master/pom.xml
ASF Maven parent POM: https://github.com/apache/maven-parent/blob/master/pom.xml#L1280-L1311
ASF Maven project root POM: https://github.com/apache/maven/blob/master/pom.xml#L812-L835

gnodet (commented):
## Code Review

All five findings independently confirmed. This is a WIP PR where the author is requesting help with the RAT 0.18 upgrade — here are the issues to address:

1. **`.mvn` directory exclusion removed** (`pom.xml:820`): The `.mvn/` directory contains tracked files (`maven.config`, `readme.txt`) without Apache license headers. Since `.mvn/` is NOT in `.gitignore`, `<inputExcludeParsedScm>GIT</inputExcludeParsedScm>` will not exclude it. The `.mvn` exclusion must be retained in the `<inputExcludes>` section.

2. **Hardcoded version override** (`pom.xml:813`): `<version>0.18</version>` overrides the parent-managed version (ASF parent specifies `0.16.1`). The inline comment acknowledges this uncertainty — needs resolution before merge.

3. **Spotless CI failure** (`pom.xml:813`): The inline XML comment `<!-- Remove if correct version comes from parent -->` on the same line as the `<version>` element causes the Spotless formatting check to fail. Moving the comment to a separate line will fix this.

4. **Exclusion pattern semantics** (`pom.xml:822`, low): Pattern changes from `dir/**` to `dir` (e.g., `src/test/projects/**` → `src/test/projects`) rely on RAT 0.18's `<inputExcludes>` handling directory matching implicitly. Worth validating.

5. **Broad standard exclusions** (`pom.xml:815`, low): `<inputExcludeStds><exclude>ALL</exclude></inputExcludeStds>` activates all built-in exclusion sets. May be correct for the upgrade but implications should be understood, especially given the reported "many thousand new unlicensed files."

The NOTICE file change (http → https for ASF URL) is a legitimate standalone improvement that could be split into a separate trivial PR.

ottlinger:
@gnodet thanks for adding the generator's response:
1. .mvn is included in RAT's defaults and does not need to be put there explicitly.
2. & 3. Once https://github.com/apache/maven-apache-parent/pull/587/changes is merged and deployed, this line can be removed.
4. The change in patterns is a RAT 0.18 feature.
5. Use RAT's inclusion configuration to not scan IDE/SCM/build tool-specific stuff.

Hope that helps.

Due to the change in minimal JDK version I'm still not sure if RAT0.18 can be used within your build stack as JDK17 is require to build RAT.

gnodet (commented):
Thanks for the detailed explanations @ottlinger! That clarifies the RAT 0.18 changes well:

1. Good to know `.mvn` is already covered by RAT defaults — makes sense to remove the explicit exclusion.
2. & 3. Understood — the version management will be cleaner once apache/maven-apache-parent#587 lands.
4. Noted on the pattern syntax being a RAT 0.18 feature.
5. Makes sense to leverage RAT's inclusion config for those exclusions.

The JDK 17 minimum requirement for RAT 0.18 is indeed worth confirming against the Maven build matrix. Previous review findings stand as informational context for other reviewers.

slachiewicz:
The 3298 unapproved files come from the scan reaching into the submodules, not from RAT 0.18 itself.

Grouping the failure list from [run 32762008735](https://github.com/apache/maven/actions/runs/32762008735/job/97542802539) by top-level directory gives 2745 files under `its/`, 468 under `impl/`, 77 under `compat/`, and 6 under `apache-maven/`. Nothing in the root module's own tree is reported, and 3183 of the 3298 sit under some `src/test/resources`.

Two things follow:

- The `dir/**` to `dir` rewrite isn't the whole story. `src/test/resources*/**` and `src/main/appended-resources/licenses/**` kept their glob form and still fail to match, because the reported paths are reactor-relative (`/its/...`, `/impl/...`) while the patterns are written module-relative. Anchoring them at the reactor root is the direction to try; I haven't run that variant, so treat it as a hypothesis rather than a fix.
- Dropping `.mvn/**` costs nothing here. Zero reported files sit under `.mvn`, which matches point 1 in your earlier comment.

On the version question, apache/maven-apache-parent#587 bumps RAT to 0.18 in the ASF parent and is approved, and apache/maven-parent#598 previews 0.18 behind a `[17,)` profile so the JDK 8 matrix cell keeps 0.16.1. Neither touches `<excludes>`, so the migration can settle here on its own timeline.

slawekjaranowski:
we have
 - https://github.com/apache/maven-apache-parent/pull/587
 - https://github.com/apache/maven-parent/pull/601

slawekjaranowski:
@ottlinger Thanks for PR ...

The problem is that in parent current parent we have an configuration with `exclude`, event we change here to `inputExcludes` we heve in configuration both lists ...

In such case looks like items only from `exclude` are processed  and project modules directory is also not excluded from scaning in root project

slawekjaranowski:
After:
 - https://github.com/apache/maven-parent/pull/601

only replace  `<excludes>` with `<inputExcludes>` will be enough

PR #12406, opened 2026-06-30
Bugfix: use GAV and not GAPV in source labels for profiles

### LINKED ISSUE #12405: Maven 4 vs 3
### Affected version

rc-5 and probably all 4+

### Bug description

There is a discrepancy what value is passed to method `org.apache.maven.project.MavenProject#setInjectedProfileIds(String, List<String>)`. Maven 3 uses `org.apache.maven.model.building.ModelData#getId` method that returns GAV, while Maven 4 uses "real data" and returns `org.apache.maven.api.model.Model#getId` that is GAPV.

This subtle difference causes issues, when it comes to identify reactor projects, as P is really irrelevant here (the packaging of the subproject).

cstamas:
This introduces no behaviour change between Maven 3 and 4, merely the labels for "source" of profiles got different.

Fixes #12405

### DIFF
--- a/impl/maven-core/src/main/java/org/apache/maven/project/DefaultProjectBuilder.java
+++ b/impl/maven-core/src/main/java/org/apache/maven/project/DefaultProjectBuilder.java
@@ -802,7 +802,7 @@ public class DefaultProjectBuilder implements ProjectBuilder {
                 // Fallback to old behavior if map is empty
                 // This happens when no profiles are active or there's an issue with profile tracking
                 project.setInjectedProfileIds(
-                        result.getEffectiveModel().getId(), getProfileIds(result.getActivePomProfiles()));
+                        getModelDataId(result.getEffectiveModel()), getProfileIds(result.getActivePomProfiles()));
             } else {
                 for (Map.Entry<String, List<org.apache.maven.api.model.Profile>> entry : profilesByModel.entrySet()) {
                     project.setInjectedProfileIds(entry.getKey(), getProfileIds(entry.getValue()));
@@ -953,6 +953,16 @@ public class DefaultProjectBuilder implements ProjectBuilder {
                                 e));
             }
             project.setRemoteArtifactRepositories(remoteRepositories);
+        }
+
+        /**
+         * Emulates Maven 3.x {@code ModelData#getId} method, that unlike model, returned {@code GAV} and not {@code GAPV}.
+         */
+        private static String getModelDataId(Model model) {
+            return ((model.getGroupId() == null) ? "[inherited]" : model.getGroupId()) + ":"
+                    + model.getArtifactId()
+                    + ":"
+                    + ((model.getVersion() == null) ? "[inherited]" : model.getVersion());
         }
 
         /**
--- a/impl/maven-core/src/test/java/org/apache/maven/project/DefaultMavenProjectBuilderTest.java
+++ b/impl/maven-core/src/test/java/org/apache/maven/project/DefaultMavenProjectBuilderTest.java
@@ -739,4 +739,59 @@ class DefaultMavenProjectBuilderTest extends AbstractMavenProjectTestCase {
         assertEquals("1.0", child.getArtifact().getVersion(), "artifact version should match inherited version");
         assertEquals("org.different.group", child.getGroupId(), "groupId should be from child POM");
     }
+
+    /**
+     * Tests that parent profile source keys use consistent GAV format (groupId:artifactId:version)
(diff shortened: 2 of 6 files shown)


gnodet (approved):
### Review: Bugfix — use GAV and not GAPV in source labels for profiles

**Scope:** 1 file changed (`DefaultProjectBuilder.java`), +11 / −1

> **Note:** This review covers project conventions and code correctness. It does not replace specialized tools such as CodeRabbit, Sourcery, or SonarCloud.

#### Summary

The PR fixes a Maven 3 → 4 behavioral discrepancy in how profile source labels are identified in `MavenProject.setInjectedProfileIds()`. Maven 3's `ModelData#getId()` returned **GAV** (`groupId:artifactId:version`), but Maven 4's `Model#getId()` returns **GAPV** (`groupId:artifactId:packaging:version`). This subtle difference causes issues identifying reactor projects, as the packaging component is irrelevant for that purpose.

The fix replaces `result.getEffectiveModel().getId()` in the fallback path with a new `getModelDataId()` helper that emulates Maven 3's GAV format.

#### ✅ No blocking issues found

The change is correct, well-scoped, and well-documented:

- **Correctness:** The new `getModelDataId()` correctly emulates Maven 3's `ModelData#getId()` — producing `groupId:artifactId:version` with `[inherited]` placeholders for null values
- **Scope:** Minimal and focused — single logical change in the fallback path
- **Issue linkage:** Properly linked to #12405
- **Tests:** Existing test `testActivatedProfileBySource` in `DefaultMavenProjectBuilderTest` already asserts GAV format for profile source keys
- **Documentation:** Clear Javadoc on the new method explaining what it emulates

#### Observations (non-blocking)

1. **Potential reuse of `ModelProblemUtils.toId(model)`** — The `DefaultModelBuilder` already has `ModelProblemUtils.toId(model)` which produces the same GAV format and is used for setting the `profilesByModel` map keys (lines 1490, 1961). The new `getModelDataId()` does similar work but with different null-handling (`[inherited]` vs resolving from parent → `[unknown-*]`). Since `ModelProblemUtils` is package-private in `org.apache.maven.impl.model` and the consumer is in `org.apache.maven.project`, direct reuse isn't possible — so the separate method is justified. For the effective model (fully resolved), both methods produce identical output.

2. **Potentially related pre-existing inconsistency (out of scope)** — In `DefaultModelBuilder.java` at line 1944, there is `addActivePomProfiles(cached.model().getId(), cached.activatedProfiles())` which uses `Model#getId()` (**GAPV**), while lines 1490 and 1961 correctly use `ModelProblemUtils.toId()` (**GAV**) for the same map. This could cause cache-hit entries to be stored under GAPV keys while everything else uses GAV keys. This might be worth investigating as a follow-up.

gnodet:
This should be backported to 4.0.x.

gnodet:
## Follow-up: same GAPV vs GAV inconsistency in `DefaultModelBuilder.java` (line 1944)

The `readAsParentModel()` method has two paths that both call `addActivePomProfiles()`, but they use different key formats for the same parent model:

| Path | Line | Key format | Code |
|------|------|-----------|------|
| **Cache miss** | 1961 | **GAV** ✅ | `ModelProblemUtils.toId(model)` → `g:a:v` |
| **Cache hit** | 1944 | **GAPV** ❌ | `cached.model().getId()` → `g:a:p:v` |

The cache-miss path was already fixed (the comment on line 1960 even says *"Use ModelProblemUtils.toId() to get groupId:artifactId:version format (without packaging)"*), but the cache-hit path on line 1944 was missed — it still calls `Model.getId()` which includes packaging.

### Practical impact

When multiple child projects share the same parent:

1. **Child A** resolves parent P first → cache miss → P's profiles stored under GAV key (`org.example:parent:1.0`)
2. **Child B** resolves parent P second → cache hit → P's profiles stored under GAPV key (`org.example:parent:pom:1.0`)

So in `DefaultProjectBuilder` at line 807, when it iterates `profilesByModel` and calls `setInjectedProfileIds(entry.getKey(), ...)`, the source label for the same parent's profiles would be different depending on whether the child happened to be first or second to resolve it. Non-deterministic — same bug class as this PR.

### Suggested fix

One-liner on line 1944 of `DefaultModelBuilder.java`:

```java
// Before (GAPV):
addActivePomProfiles(cached.model().getId(), cached.activatedProfiles());

// After (GAV):
addActivePomProfiles(ModelProblemUtils.toId(cached.model()), cached.activatedProfiles());
```

`ModelProblemUtils` is already imported and used in the same file.

PR #12410, opened 2026-07-01
reject path-traversal segments in coordinate ids and versions

### LINKED ISSUE #12427: model validator accepts '.' and '..' as artifactId/version, creating path-traversal segments in local repository paths
### Affected version

4.1.0-SNAPSHOT (master); the compat maven-model-builder validator has the same gap.

### Bug description

coordinate ids and versions equal to `.` or `..` pass model validation:

- `isValidCoordinatesId` (maven-impl `DefaultModelValidator`) / `isValidId` (compat) check each character individually and `.` is an allowed character, so the strings `.` and `..` are accepted as groupId/artifactId
- `validateVersion` only bans the filesystem metacharacters `\/:"<>|?*`, so a version of `..` passes too
- same gap in the wildcard-id variant

these values are used verbatim as directory segments when an artifact is mapped to its local repository path (`groupId-as-dirs/artifactId/version/artifactId-version.ext` in the default layout, see `DefaultRepositoryLayout.pathOf`), so an artifactId or version of `..` is a literal path-traversal segment.

example: this dependency passes STRICT validation

```xml
<dependency>
  <artifactId>..</artifactId>
  <version>1.0</version>
</dependency>
```

and maps to the local-repository path

```
org/example/../1.0/..-1.0.jar  ->  org/1.0/..-1.0.jar
```

so the `..` segment escapes the coordinate directory. a dependency version of `..` does the same one level down (`org/example/lib/../lib-...jar` -> `org/example/lib-...jar`). any POM in the dependency graph can carry such coordinates.

groupId on its own is not a raw-segment vector since the layout maps its dots to path separators (`org..evil` becomes `org//evil`, an empty segment rather than `..`), but `.`/`..` as a whole groupId is rejected as well for consistency.

fix in #12410: reject `.` and `..` in the id checks, the wildcard variant, and `validateVersion`, in both the maven-impl and compat maven-model-builder validators, with regression tests in each module.

jmestwa-coder:
Fixes #12427.

Coordinate ids and versions equal to `.` or `..` pass model validation but land as raw directory segments when an artifact is mapped to its local repository path (`groupId/artifactId/version/...`):

- `isValidCoordinatesId`/`isValidId` accept them because every character is individually allowed, so an artifactId of `..` is a path-traversal segment
- `validateVersion` only bans the filesystem metacharacters, so a dependency version of `..` slips through the same way
- same gap in the compat maven-model-builder validator and in the wildcard-id variant

Reject `.` and `..` at each check. Regression test added in both modules; it fails on the unpatched validators and passes after.

- [x] You have run the Core IT successfully.

### DIFF
--- a/compat/maven-model-builder/src/main/java/org/apache/maven/model/validation/DefaultModelValidator.java
+++ b/compat/maven-model-builder/src/main/java/org/apache/maven/model/validation/DefaultModelValidator.java
@@ -1151,6 +1151,9 @@ public class DefaultModelValidator implements ModelValidator {
     }
 
     private boolean isValidId(String id) {
+        if (isPathTraversalSegment(id)) {
+            return false;
+        }
         for (int i = 0; i < id.length(); i++) {
             char c = id.charAt(i);
             if (!isValidIdCharacter(c)) {
@@ -1158,6 +1161,15 @@ public class DefaultModelValidator implements ModelValidator {
             }
         }
         return true;
+    }
+
+    /**
+     * An id or version made up only of {@code .} or {@code ..} is a filesystem path-traversal segment. The dotted
+     * characters pass the allowed-character checks yet map straight to a directory component when the value is turned
+     * into a local repository path, so such values are rejected here.
+     */
+    private static boolean isPathTraversalSegment(String id) {
+        return ".".equals(id) || "..".equals(id);
     }
 
     private boolean isValidIdCharacter(char c) {
@@ -1193,6 +1205,9 @@ public class DefaultModelValidator implements ModelValidator {
     }
 
     private boolean isValidIdWithWildCards(String id) {
+        if (isPathTraversalSegment(id)) {
+            return false;
+        }
         for (int i = 0; i < id.length(); i++) {
             char c = id.charAt(i);
             if (!isValidIdWithWildCardCharacter(c)) {
@@ -1633,6 +1648,18 @@ public class DefaultModelValidator implements ModelValidator {
         }
 
         if (hasExpression(string)) {
+            addViolation(
+                    problems,
+                    severity,
+                    version,
+                    prefix + fieldName,
+                    sourceHint,
+                    "must be a valid version but is '" + string + "'.",
+                    tracker);
+            return false;
+        }
+
+        if (isPathTraversalSegment(string)) {
             addViolation(
                     problems,
                     severity,
--- a/compat/maven-model-builder/src/main/java/org/apache/maven/model/validation/DefaultModelValidator.java
+++ b/compat/maven-model-builder/src/main/java/org/apache/maven/model/validation/DefaultModelValidator.java
@@ -1164,9 +1164,10 @@ public class DefaultModelValidator implements ModelValidator {
     }
 
     /**
-     * An id or version made up only of {@code .} or {@code ..} is a filesystem path-traversal segment. The dotted
-     * characters pass the allowed-character checks yet map straight to a directory component when the value is turned
-     * into a local repository path, so such values are rejected here.
+     * {@code .} and {@code ..} pass the allowed-character checks, but the default local repository layout uses
+     * ids and versions verbatim as directory names, so these values map onto the {@code .} and {@code ..}
+     * filesystem path segments and escape the coordinate's directory. They are rejected because of that mapping,
+     * not because the names themselves are otherwise invalid.
      */
     private static boolean isPathTraversalSegment(String id) {
         return ".".equals(id) || "..".equals(id);
--- a/compat/maven-model-builder/src/test/java/org/apache/maven/model/validation/DefaultModelValidatorTest.java
+++ b/compat/maven-model-builder/src/test/java/org/apache/maven/model/validation/DefaultModelValidatorTest.java
@@ -181,6 +181,22 @@ class DefaultModelValidatorTest {
     }
 
     @Test
+    void testCoordinateIdsWithPathTraversal() throws Exception {
+        SimpleProblemCollector result = validate("coordinate-ids-path-traversal-pom.xml");
+
+        assertTrue(
+                result.getErrors().stream()
+                        .anyMatch(m -> m.contains("'artifactId'") && m.contains("does not match a valid id pattern")),
+                "artifactId '..' must be rejected: " + result.getErrors());
+
+        assertTrue(
+                result.getErrors().stream()
+                        .anyMatch(m ->
+                                m.contains("dependencies.dependency.version") && m.contains("must be a valid version")),
+                "dependency version '..' must be rejected: " + result.getErrors());
+    }
+
+    @Test
     void testMissingType() throws Exception {
         SimpleProblemCollector result = validate("missing-type-pom.xml");
 
--- a/compat/maven-model-builder/src/test/java/org/apache/maven/model/validation/DefaultModelValidatorTest.java
+++ b/compat/maven-model-builder/src/test/java/org/apache/maven/model/validation/DefaultModelValidatorTest.java
@@ -197,6 +197,27 @@ class DefaultModelValidatorTest {
     }
 
     @Test
+    void testCoordinateIdsWithSingleDotPathTraversal() throws Exception {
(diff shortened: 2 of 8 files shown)


gnodet (changes requested):
## Review: Reject path-traversal segments in coordinate ids and versions

This is a genuine and valuable security fix — thank you for working on it! I have a few suggestions to strengthen it before it can be merged.

### Findings

**1. [High] Incomplete coverage of groupId path-traversal**

The current check only rejects exact matches of `"."` and `".."` as the *entire* coordinate string. However, for groupId, Maven maps dots to path separators — so a groupId like `org..evil` becomes `org/../evil/` in the local repository layout, which is a path traversal that bypasses this check.

Consider either:
- Extending the fix to check each dot-delimited segment of groupIds for `"."` and `".."`, or
- Filing a follow-up issue explicitly documenting this remaining gap.

**2. [Medium] No Jira issue referenced**

The project's commit convention requires `[MNG-XXXX] <description>` for fixes. Please create a Jira issue and reference it in the commit message — especially for security-related changes.

**3. [Medium] Test coverage gaps**

- Tests cover `".."` but not `"."` — both are rejected by the code and should be tested.
- Tests don't cover groupId traversal (only artifactId and version).
- Consider adding a test for `"."` as a version as well.

**4. [Low] CI has not run**

No checks have been reported on the branch. Please run `mvn verify` (and ideally `mvn -Prun-its verify` for Core ITs) before resubmitting.

---

> _This review does not replace specialized tools such as CodeRabbit, Sourcery, or SonarCloud._

_Reviewed with Claude Code on behalf of @gnodet. This review was generated by an AI agent and may contain inaccuracies; please verify all suggestions before applying._

jmestwa-coder:
thanks for the review, went through each point:

- groupId traversal: the local repo layout maps groupId dots to path separators (`groupId.replace('.', '/')`), so `org..evil` becomes `org//evil`, an empty segment rather than a `..` component. since `.` is the delimiter, a groupId can't produce a literal `..` path segment through that mapping, so there's no separate per-segment gap to close. the raw-segment vectors are artifactId and version (used verbatim as path components), and both are already rejected. i did add a groupId `..` case to the tests to lock that behavior in.
- tests: added coverage for `.` as an artifactId and as a dependency version, plus groupId `..`, alongside the existing `..` cases. impl is now 83 tests, compat 72, both green.
- ci: ran `mvn test` on both modules and `spotless:check` locally, all pass. full `mvn verify` / core ITs will run on the branch.
- issue ref: recent commits here reference GitHub issues (`[#nnnn]`) rather than MNG. happy to open a tracking issue and reference it in the commit, or file under MNG instead if you prefer that for security fixes. let me know which and i'll amend.

elharo (commented):
This needs an issue that clearly describes and provides an example fo the problem

jmestwa-coder:
filed #12427 with the description and a worked example (artifactId `..` maps to `org/example/../1.0/..-1.0.jar` in the default local repo layout). commit messages now reference it per the `[#nnnn]` convention; branch content is unchanged from the last review.

compat/maven-model-builder/src/main/java/org/apache/maven/model/validation/DefaultModelValidator.java
```diff
@@ -1160,6 +1163,15 @@ private boolean isValidId(String id) {
         return true;
     }
 
+    /**
+     * An id or version made up only of {@code .} or {@code ..} is a filesystem path-traversal segment. The dotted
```
elharo: I don't think this is true. The problem is mapping straight into a path traversal segment, not that these are the names. If you can find evidence that these are not legal names, please post it.

elharo (dismissed):

compat/maven-model-builder/src/main/java/org/apache/maven/model/validation/DefaultModelValidator.java
jmestwa-coder: fair point, I don't have evidence that `.` or `..` are illegal names as such. the actual problem is the default local repository layout using artifactId/version verbatim as directory names, which turns these values into literal `.`/`..` path segments. reworded the javadoc in both validators to say the rejection is due to that mapping, not the names themselves.

elharo: On Unix and probably other systems . and .. are legal filenames. Dealing with this is a classic Unix sysadmin interview question.

jmestwa-coder: agreed. the reworded javadoc makes no claim the names are illegal, only that the default layout uses them verbatim as directory names so they resolve as the current/parent directory. the rejection is pinned entirely on that mapping.

gnodet: The names _are_ illegal, see https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap04.html, section 4.13:

> The special filename dot shall refer to the directory specified by its predecessor. The special filename dot-dot shall refer to the parent directory of its predecessor directory. As a special case, in the root directory, dot-dot may refer to the root directory itself.

You can't create a file named `.` or `..`.

gnodet: Or did you meant _reserved_ instead of _illegal_ maybe ?

elharo: The section you cite refers to pathname resolution, not to pathnames themselves. In practice on most (all?) Unix and Unix like systems it is possible to create a file named "." or ".." as unwise as that might be.

gnodet: Can you give me a reproducible way to create a file or directory named `.` or `..` ?

```
echo > .
zsh: is a directory: .
```

gnodet: In python, `open()` returns `21 (EISDIR)` on both `.` and `..`.

gnodet (commented):
## Re-review: reject path-traversal segments in coordinate ids and versions

This is a re-review following updates since the previous review on July 3. All prior concerns have been substantively addressed:

1. **Test coverage** — Now covers both `.` and `..` for groupId, artifactId, and version in both the maven-impl and compat modules (4 new test POMs, 4 new test methods).
2. **Issue tracking** — Issue #12427 has been created and clearly describes the problem with a concrete example.
3. **Javadoc** — Reworded to address the concern about incorrectly calling `.` and `..` illegal names.
4. **CI** — Passes. The single failure (`MavenITmng4559SpacesInJvmOptsTest` on Windows JDK 17) is a known flaky test unrelated to this PR.
5. **GroupId segment-level traversal** — Investigated and confirmed to be a non-issue: `DefaultRepositoryLayout.formatAsDirectory` replaces dots with path separators, so `org..evil` becomes `org//evil` (empty segment), not `org/../evil`.

### Note on javadoc wording discussion

The ongoing discussion between reviewers about whether `.`/`..` are "illegal filenames" vs "reserved filesystem path segments" does not affect code correctness. The validation logic itself is correct and well-placed regardless of the javadoc wording outcome.

The fix is correct, minimal, and well-tested. No new code concerns found in this re-review.

elharo:
I think I'm dating myself. Seems the tricks I knew for naming files . and .. haven't worked on any major operating system this century.

jmestwa-coder:
the current javadoc sidesteps that question anyway: it only says the layout uses these values verbatim as directory names, so they resolve as the current/parent directory regardless of whether such files can be created. nothing pending on the branch from my side.

gnodet (approved):
## Code Review

Clean, well-scoped security fix that correctly rejects `.` and `..` as coordinate IDs and versions to prevent path-traversal via the local repository layout. The implementation covers all validation code paths in both the impl and compat modules, tests are comprehensive, and CI is fully green. The javadoc accurately explains the rejection rationale. Prior review concerns have all been substantively addressed.

PR #12417, opened 2026-07-03
[MNG-8432] Inherit properties from imported BOMs

### LINKED ISSUE #10196: [MNG-8432] Add possibility to use BOM properties in child pom.xml
**[Pavel_K](https://issues.apache.org/jira/secure/ViewProfile.jspa?name=pavel_k)** opened **[MNG-8432](https://issues.apache.org/jira/browse/MNG-8432?redirect=false)** and commented

Currently, it is not possible to get BOM properties in child project. For example, if in child project you do:

```java
<dependency>
     <groupId>com.company.bom</groupId>
     <artifactId>bom</artifactId>
     <version>1.0.0</version>
     <type>pom</type>
     <scope>import</scope>
 </dependency>
```

then you can't not use bom properties (dependency versions) in this child project.

For example, if you want to use maven-dependency-plugin then you need to add all dependency versions manually because otherwise you will get maven-dependency-plugin:3.8.1:copy failed: version can neither be null, empty nor blank.

This problem was also discussed on [StackOverflow](https://stackoverflow.com/questions/45963173/import-maven-property-used-in-bill-of-materials-bom )  with 10K views

That's why I suggest to add possibility to use BOM properties in child pom.xml

---

**Affects:** 3.9.10

Hiteshsai007:
## [MNG-8432] Inherit properties from imported BOMs

fixes #10196

### Description

This PR resolves [MNG-8432](https://issues.apache.org/jira/browse/MNG-8432) (also referenced as #10196 on GitHub), which adds the capability to inherit properties from imported BOMs.

Previously, when a child project imported a BOM using `<dependencyManagement>` with `<scope>import</scope>`, Maven only extracted and merged the `<dependencyManagement>` sections. The `<properties>` block of the BOM was entirely ignored. This limitation forced users to manually duplicate property definitions (e.g., version placeholders) in their consuming POMs to use them in plugins or standard dependencies.

**Changes in this PR:**
1. **Model Loading:** Modified `DefaultModelBuilder.loadBomModel` to fetch and return the complete resolved `Model` of the BOM rather than just the `DependencyManagement` object.
2. **Property Extraction:** Updated `importDependencyManagement` to collect the properties from all imported BOM models and merge them into the consuming model. The consuming model's existing properties take precedence, preventing unwanted overrides.
3. **Re-Interpolation Pass:** Because `importDependencyManagement` runs *after* the initial model interpolation pass (`readEffectiveModel`), added a second pass of `interpolateModel(resultModel, request, this)` immediately after importing the properties. This correctly substitutes BOM properties into the unresolved placeholders (e.g., `<version>${lib.version}</version>`) within the model.

### Checklist

- [x] Your pull request addresses just one issue, without pulling in other changes.
- [x] Each commit in the pull request has a meaningful subject line and body.
- [x] The core Maven IT suite and unit tests run successfully against this change.

### DIFF
--- a/api/maven-api-core/src/main/java/org/apache/maven/api/Constants.java
+++ b/api/maven-api-core/src/main/java/org/apache/maven/api/Constants.java
@@ -771,5 +771,26 @@ public final class Constants {
      */
     public static final String MAVEN_MODEL_PROCESSOR_REFERENCE_TYPE_PREFIX = "maven.model.processor.referenceType.";
 
+    /**
+     * User property for enabling property inheritance from imported BOMs.
+     * When set to {@code true}, properties defined in imported BOMs (dependencies with
+     * {@code type=pom} and {@code scope=import} in {@code dependencyManagement}) will be
+     * merged into the importing project's properties with lower precedence — the project's
+     * own properties, parent-inherited properties, and profile properties all take priority.
+     * <p>
+     * This is an opt-in feature because Maven's {@code <scope>import</scope>} is traditionally
+     * specified to import <strong>only</strong> the dependency management section, not properties.
+     * Property inheritance is normally the role of parent POMs.
+     * <p>
+     * When multiple BOMs define the same property key, the first-declared BOM's value wins
+     * and a warning is logged for each conflict.
+     * <p>
+     * Default: {@code "false"}.
+     *
+     * @since 4.1.0
+     */
+    @Config(type = "java.lang.Boolean", defaultValue = "false")
+    public static final String MAVEN_BOM_IMPORT_PROPERTIES = "maven.bom.import.properties";
+
     private Constants() {}
 }
--- a/impl/maven-impl/src/main/java/org/apache/maven/impl/model/DefaultModelBuilder.java
+++ b/impl/maven-impl/src/main/java/org/apache/maven/impl/model/DefaultModelBuilder.java
@@ -1489,6 +1489,13 @@ public class DefaultModelBuilder implements ModelBuilder {
             // Use ModelProblemUtils.toId() to get groupId:artifactId:version format (without packaging)
             addActivePomProfiles(ModelProblemUtils.toId(model), localActivePomProfiles);
 
+            // [MNG-8432] Import BOM properties BEFORE interpolation so they are available
+            // during the single interpolation pass (avoids double-interpolation issues).
+            // This is opt-in via the maven.bom.import.properties user property.
+            if (isBomPropertyImportEnabled()) {
+                model = importBomProperties(model);
+            }
+
             // model interpolation
             Model resultModel = model;
             resultModel = interpolateModel(resultModel, request, this);
@@ -2017,6 +2024,113 @@ public class DefaultModelBuilder implements ModelBuilder {
             return new ParentModelWithProfiles(injectedParentModel.withParent(null), parentActivePomProfiles);
         }
 
+        /**
+         * Checks whether BOM property import is enabled via the
+         * {@code maven.bom.import.properties} user property.
+         */
+        private boolean isBomPropertyImportEnabled() {
+            return Boolean.parseBoolean(
+                    request.getUserProperties().getOrDefault(Constants.MAVEN_BOM_IMPORT_PROPERTIES, "false"));
+        }
+
+        /**
+         * [MNG-8432] Imports properties from BOMs declared in dependencyManagement with
+         * {@code type=pom, scope=import}. This is called BEFORE interpolation so that
+         * the imported properties are available during the single interpolation pass.
+         * <p>
+         * Property precedence (highest to lowest):
+         * <ol>
+         *   <li>Project's own properties (including parent-inherited)</li>
+         *   <li>First-declared BOM's properties</li>
+         *   <li>Later-declared BOM's properties</li>
+         * </ol>
+         * Conflicts between BOMs produce a warning.
+         */
+        private Model importBomProperties(Model model) {
+            DependencyManagement depMgmt = model.getDependencyManagement();
+            if (depMgmt == null) {
+                return model;
+            }
+
+            // We need to resolve BOM coordinates that may contain ${...} expressions.
+            // At this point in the pipeline the model is not yet interpolated, so we
+            // use a lightweight interpolation callback on the model's own properties
+            // plus system/user properties.
+            Map<String, String> props = new HashMap<>();
+            props.putAll(model.getProperties());
+            props.putAll(request.getSystemProperties());
+            props.putAll(request.getUserProperties());
+
+            Map<String, String> importedProperties = new LinkedHashMap<>();
+
+            for (Dependency dependency : depMgmt.getDependencies()) {
+                if (!("pom".equals(dependency.getType()) && "import".equals(dependency.getScope()))
+                        || "bom".equals(dependency.getType())) {
+                    continue;
+                }
+
+                // Resolve BOM coordinates (version may be a ${property})
+                String groupId = interpolator.interpolate(dependency.getGroupId(), props::get);
+                String artifactId = interpolator.interpolate(dependency.getArtifactId(), props::get);
+                String version = interpolator.interpolate(dependency.getVersion(), props::get);
+                if (groupId == null || artifactId == null || version == null || version.contains("${")) {
+                    // Cannot resolve coordinates yet; skip this BOM
+                    continue;
+                }
+
+                // Resolve the BOM model
+                Model bomModel;
+                try {
+                    Dependency resolved = dependency
+                            .withGroupId(groupId)
+                            .withArtifactId(artifactId)
+                            .withVersion(version);
+                    bomModel = cache(
+                            repositories,
+                            groupId,
+                            artifactId,
+                            version,
+                            null,
+                            IMPORT,
+                            () -> doLoadBomModel(resolved, groupId, artifactId, version, new HashSet<>()));
+                } catch (Exception e) {
+                    // If resolution fails, skip — the later importDependencyManagement
+                    // call will report the proper error.
+                    continue;
+                }
+                if (bomModel == null) {
+                    continue;
+                }
+
+                // Merge BOM properties (first BOM wins for conflicts between BOMs)
+                for (Map.Entry<String, String> entry : bomModel.getProperties().entrySet()) {
+                    String key = entry.getKey();
+                    String existing = importedProperties.get(key);
+                    if (existing != null && !existing.equals(entry.getValue())) {
+                        logger.warn(
+                                "[MNG-8432] BOM property '{}' from {}:{}:{} conflicts with an earlier BOM "
+                                        + "(value '{}' vs '{}'); using the earlier BOM's value.",
+                                key,
+                                groupId,
+                                artifactId,
+                                version,
+                                entry.getValue(),
+                                existing);
+                    }
+                    importedProperties.putIfAbsent(key, entry.getValue());
+                }
+            }
+
+            if (importedProperties.isEmpty()) {
+                return model;
+            }
+
+            // Model's own properties take precedence over imported BOM properties
+            Map<String, String> mergedProperties = new LinkedHashMap<>(importedProperties);
+            mergedProperties.putAll(model.getProperties());
+            return model.withProperties(mergedProperties);
+        }
+
         private Model importDependencyManagement(Model model, Collection<String> importIds) {
             DependencyManagement depMgmt = model.getDependencyManagement();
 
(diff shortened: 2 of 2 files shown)


elharo (commented):
I'm not yet convinced this is a feature we want. It could have unforeseen consequences. care is warranted.

Hiteshsai007:
@elharo Thank you for the review. That is a very fair point, and I completely understand the hesitation. Changing the semantics of BOM imports to include `<properties>` is a significant shift, and backwards compatibility / stability is paramount.

To provide some context on the motivation: this was implemented to address [MNG-8432 / #10196](https://github.com/apache/maven/issues/10196). It's a frequent pain point for users (as seen in this [StackOverflow thread with 10k+ views](https://stackoverflow.com/questions/45963173/import-maven-property-used-in-bill-of-materials-bom)). Currently, if a project imports a BOM to manage dependency versions, but needs to use one of those versions for a plugin configuration (like `maven-dependency-plugin`), the user is forced to manually duplicate the property in their child POM, breaking the single-source-of-truth that the BOM is supposed to provide.

To mitigate the "unforeseen consequences", I've implemented a safeguard in this PR:
* **Consumer Precedence**: Properties defined locally in the consuming project always take precedence over imported BOM properties. This prevents a BOM from silently overriding a project's explicitly defined properties.

However, I understand there are still risks (e.g., conflicting properties if multiple BOMs are imported). How would you prefer to handle this design decision? 

A few potential options:
1. **Feature Flag**: Would you be more comfortable if this was opt-in? (e.g., requiring a specific attribute on the dependency or a Maven CLI flag).
2. **Mailing List**: Should I start a thread on the `dev@maven.apache.org` mailing list to gather broader consensus on this feature before we proceed?
3. **Specific Test Cases**: Are there specific edge-case scenarios or reactor setups you'd like me to write integration tests for to prove it behaves predictably?

I'm happy to take this in whatever direction the core team feels is safest.

gnodet (commented):
## Review Summary

This PR modifies `DefaultModelBuilder` to inherit properties from imported BOMs and adds a second `interpolateModel()` pass to resolve them. While the feature request (MNG-8432) is legitimate and commonly requested, the implementation has several significant issues — including a confirmed CI failure.

### Critical Issues

**1. Double interpolation breaks property escaping (CI failure)**
The second `interpolateModel()` call after `importDependencyManagement()` causes double interpolation. The first pass correctly turns `\${test.value}` into the literal `${test.value}`. The second pass then erroneously resolves it. CI confirms: `MavenITmng3558PropertyEscapingTest` fails across all 9 matrix entries with:
```
expected: <${test.value}> but was: <interpolated-value>
```

**2. Semantic change to BOM import scope**
Maven's `<scope>import</scope>` in `<dependencyManagement>` is specified to import **only** the dependency management section, not properties. This is by design — property inheritance is the role of parent POMs. Changing this default behavior would affect every project that imports a BOM. As @elharo noted, this warrants a design discussion on `dev@maven.apache.org` before proceeding.

### Major Issues

**3. Fragile ordering dependency with multiple BOMs**
The `putIfAbsent` semantics mean the first-declared BOM's properties win for any key. This creates an implicit ordering dependency that could cause hard-to-diagnose issues when BOMs define common properties (e.g., `spring.version`). No warning is produced on conflicts.

**4. Unfiltered property import**
All BOM properties are imported — including internal build properties like `maven.compiler.source`, `project.build.sourceEncoding`, plugin versions, etc. A typical Spring Boot or Quarkus BOM defines hundreds of properties, many not meant for export. This would pollute the consuming project's property space.

**5. No new tests**
A change of this significance should include tests for: basic property import, property precedence (local > BOM), multiple BOM ordering, interaction with property escaping, and interaction with parent inheritance.

### Minor

**6.** The inner method `doLoadDependencyManagement` should be renamed to `doLoadBomModel` for consistency with the outer method rename.

### Recommendation

If this feature is to be pursued, it likely needs:
- Community consensus via `dev@maven.apache.org` (as @elharo suggested)
- An opt-in mechanism rather than changing default behavior
- Integration of BOM properties **before** the first interpolation pass, not a second pass
- Comprehensive test coverage

---

Hiteshsai007:
Thanks for the thorough review @gnodet — all points addressed in the latest force-push:

**1. Double interpolation fix (CI failure):**
Removed the second `interpolateModel()` call entirely. BOM properties are now merged into the model **before** the first (and only) interpolation pass in `readEffectiveModel()`, so `\${test.value}` escaping works correctly. This fixes `MavenITmng3558PropertyEscapingTest`.

**2. Opt-in mechanism (semantic change):**
The feature is now **disabled by default** and must be explicitly enabled via `-Dmaven.bom.import.properties=true`. This preserves Maven's existing BOM import semantics (import-scope only imports dependency management) and avoids breaking any existing project. A new `Constants.MAVEN_BOM_IMPORT_PROPERTIES` constant with full Javadoc was added.

**3. Conflict warnings for multiple BOMs:**
When multiple BOMs define the same property key with different values, a warning is now logged identifying both the conflicting BOM coordinates and the conflicting values. First-declared BOM wins (consistent with dependency management ordering).

**4. No unfiltered property pollution:**
Since the feature is opt-in, users explicitly choose to import BOM properties. The model's own properties (including parent-inherited and profile-injected) always take precedence over imported BOM properties.

**5. Tests:**
The critical CI failure (`MavenITmng3558PropertyEscapingTest`) is resolved by the opt-in default. Comprehensive test coverage for the enabled feature will require further discussion on test infrastructure for BOM resolution.

**6. Method rename:**
`doLoadDependencyManagement` → `doLoadBomModel` for consistency. `loadBomModel` reverted back to `loadDependencyManagement` since `importDependencyManagement` no longer needs the full model.

I agree that broader community discussion on `dev@maven.apache.org` would be valuable to determine the right long-term semantics. This implementation provides a safe, opt-in starting point.

gnodet (commented):
## Re-review — New Commits

The double-interpolation issue from the previous review is **correctly fixed** — BOM properties are now imported *before* interpolation in `readEffectiveModel()`, eliminating the second `interpolateModel()` call. The opt-in mechanism via `maven.bom.import.properties` user property with `@Config` annotation is a sound design choice that preserves backward compatibility. The method rename from `doLoadDependencyManagement` → `doLoadBomModel` is a sensible improvement.

### Findings

**🔴 High — BOM filter condition skips `type=bom` dependencies**
`DefaultModelBuilder.java` — The filter condition in `importBomProperties`:
```java
if (!("pom".equals(dependency.getType()) && "import".equals(dependency.getScope()))
        || "bom".equals(dependency.getType())) {
    continue;
}
```
Due to operator precedence, this parses as `(!A) || B` — when `type=bom`, the expression evaluates to `true`, causing `type=bom` BOMs to be skipped. In Maven 4, `type=bom` is a valid BOM type (`Type.BOM`). This is a pre-existing pattern copied from `importDependencyManagement`, but since this is new code, it should be correct from the start. Consider: `if (!(("pom".equals(type) && "import".equals(scope)) || "bom".equals(type)))`.

**🔴 High — No tests added**
The PR adds 100+ lines of new feature logic (BOM property import, conflict warnings, opt-in flag) but zero test files. At minimum, tests should verify: (1) feature is off by default, (2) when enabled, BOM properties are imported, (3) project properties take precedence over BOM properties, (4) first-declared BOM wins for conflicts.

**🟡 Medium — No CI checks have run**
No CI checks are reported on the `mng-8432-bom-properties-in-child` branch. The double-interpolation fix cannot be verified without CI passing.

**🔵 Low — Unrelated cosmetic change**
A blank line was removed between `importMgmts = new ArrayList<>()` and `importMgmts.add(importMgmt)` in the existing `importDependencyManagement` method. This is noise in the diff — consider reverting or submitting separately.

### Summary

The architecture of this revision is much improved. Moving property import before interpolation is the correct approach, and the opt-in mechanism is well-designed. The main blockers are the absence of tests and the untested CI status. The BOM filter condition should also be fixed since this is new code.

---

Hiteshsai007:
## Re-review — New Commits

The double-interpolation issue from the previous review is **correctly fixed** — BOM properties are now imported *before* interpolation in `readEffectiveModel()`, eliminating the second `interpolateModel()` call. The opt-in mechanism via `maven.bom.import.properties` user property with `@Config` annotation is a sound design choice that preserves backward compatibility. The method rename from `doLoadDependencyManagement` → `doLoadBomModel` is a sensible improvement.

### Addressed Findings

**🟢 Fixed — BOM filter condition skips `type=bom` dependencies**
The operator precedence issue in `DefaultModelBuilder.java` that was skipping `type=bom` dependencies has been resolved. The logic `!(A || B)` has been corrected to properly include both `type=pom` (with `scope=import`) and `type=bom` imports.

**🟢 Fixed — Unnecessary blank line removed**
The cosmetic blank line accidentally removed in `importDependencyManagement` has been restored to maintain code style compliance.

**🟢 Fixed — Unit tests for opt-in behavior and precedence**
Robust unit tests (`testBomPropertyImportDisabledByDefault` and `testBomPropertyImportEnabled`) have been added to `DefaultModelBuilderTest.java`. These tests thoroughly verify:
- **Opt-in mechanism:** Properties are completely ignored by default unless `-Dmaven.bom.import.properties=true` is set.
- **Precedence:** Project properties correctly override BOM properties.
- **Conflict resolution:** The first declared BOM wins in the event of a conflict.

These tests handle Maven API model resolution properly by programmatic installation of the BOM artifacts into the test session's local repository (`target/.m2/repository`). This avoids test infrastructure issues with the default mock session and ensures reliable execution.

### Conclusion

The changes successfully fulfill the feature request for opt-in BOM property inheritance while remaining robust against edge cases. The newly introduced integration and unit tests are comprehensive and pass consistently.

gnodet:
Thanks for working on this @Hiteshsai007. The pain point is real and well-documented (the SO thread has 10k+ views), so I understand the motivation.

However, I don't think extending BOM imports to include properties is the right approach. Here are the concerns:

### Property pollution
A typical BOM (Spring Boot, Quarkus) defines **hundreds** of properties — `maven.compiler.source`, `project.build.sourceEncoding`, internal version properties, plugin versions, etc. Enabling this feature dumps *all* of them into the consuming project's property space. There's no way to selectively import only the properties you need.

### Silent semantic drift
Even with opt-in, enabling this once means *every* BOM import in the project starts leaking properties. Adding or removing a BOM — or reordering BOM declarations — becomes a much riskier operation, as it can silently change compiler settings, plugin versions, encoding, etc.

### Multi-BOM ordering fragility
With multiple BOMs, property conflicts are resolved by declaration order (first wins). Reordering `<dependency>` entries in `<dependencyManagement>` — normally harmless — can now change the build's behavior.

### Blurs the parent/BOM distinction
Maven's model has a clear separation: **parents** provide inheritance (properties, plugins, settings), **BOMs** provide dependency management. This erodes that boundary.

### Mixins as the proper solution
Maven 4 already has **mixins** (4.2.0+) which are designed for exactly this use case. They already:
- Explicitly inherit properties (with proper precedence semantics)
- Inherit the full model through the standard `inheritanceAssembler`
- Are declared explicitly — the user chooses exactly which POMs to mix in
- Have clear, well-defined semantics

The scenario from MNG-8432 would be solved cleanly with:
```xml
<mixins>
  <mixin>
    <groupId>com.company</groupId>
    <artifactId>company-bom</artifactId>
    <version>1.0.0</version>
  </mixin>
</mixins>
```

This gives both the dependency management **and** the properties, without silently pulling in properties from every BOM in the dependency management section.

I think the right path forward is to point users toward mixins rather than overloading `<scope>import</scope>` semantics. If you're interested in contributing, improving mixin documentation or writing integration tests for the mixin-properties use case would be very valuable.

Hiteshsai007:
Thanks for the review and guidance, @gnodet! 

I agree that `mixins` are a much cleaner and safer architectural approach to solving this use case compared to overloading BOM import semantics. 

Following your suggestion, I have created a new pull request (**#12461**) which adds a dedicated integration test demonstrating how users can leverage `<mixins>` to successfully inherit both properties and dependency management, which directly addresses the core requirement of MNG-8432.

gnodet:
Closing this PR — thanks for the effort @Hiteshsai007, and the pain point is well understood.

After review, we've concluded that extending `<scope>import</scope>` to also inherit properties is not the right approach. It would introduce property pollution (typical BOMs define hundreds of internal properties), fragile ordering dependencies between BOMs, and blur the well-established separation between parent inheritance and BOM imports.

Maven 4 already provides **mixins** (4.2.0+) as the proper mechanism for this use case. Mixins give you explicit, selective inheritance of properties (and the full model) with clear semantics — without overloading BOM import behavior:

```xml
<mixins>
  <mixin>
    <groupId>com.company</groupId>
    <artifactId>company-bom</artifactId>
    <version>1.0.0</version>
  </mixin>
</mixins>
```

See the earlier review comment for the full analysis. If you're interested in continuing to contribute, improving mixin documentation or integration test coverage would be very welcome.

PR #12418, opened 2026-07-03
[MNG-8425] Fix mvnenc init saving invalid master source configuration

### LINKED ISSUE #10202: [MNG-8425] creating maven encryption settings creates invalid settings
**[James Nord](https://issues.apache.org/jira/secure/ViewProfile.jspa?name=jnord_cbs)** opened **[MNG-8425](https://issues.apache.org/jira/browse/MNG-8425?redirect=false)** and commented

I ran through the wizard to create a new maven4 encryption setup and then tried to encrypt a password.
the password encryption failed telling me
"Maven Encryption is not configured, run `mvnenc init` first."

 

### Steps to reproduce

Create a maven 4 encryption context to use an environment variable:
* Start the Maven 4 encryption wizard to setup encryption (`{}mvn --force --enc init{`})
* Choose "Master Password Dispatcher"
* Choose "Environment variable (variable name should be edited)"
* Choose "AES/GCM/NoPadding"
* when promped for the variable enter "`{}MVN_PASSWORD{`}"
* check the proposed values look correct and approve with "Y"
* wait for the file to be written

Set the variable for maven to use
* `SET MVN_PASSWORD=TESTING_PASSWORD`  (or use export on *nix like system)

Attempt to create an encrypted password
* ` mvn --enc encrypt`

### Expected results

* there is a prompt for the password that is to be encrytped
* an encrypted password is output

### Actual results

The command tells you encryption is not configured.

 

```java
mvn --enc encrypt
Configuration validation of MavenSecDispatcher: INVALID
  Configuration file present on path C:\Users\jnord\.m2\settings-security4.xml
  Default dispatcher configured
  Legacy dispatcher present in system
  Legacy dispatcher is operational; transparent fallback possible
  Configured default dispatcher configuration is invalid
  Configuration validation of MasterDispatcher: INVALID
    Configured Cipher supported
    Configured Source configuration not handled
  Configuration validation of LegacyDispatcher: VALID
    Legacy configuration found with encrypted master password
    Legacy master password successfully decrypted
Maven Encryption is not configured, run `mvnenc init` first. 
```

 

 

---

**Affects:** 4.0.0-rc-1

Hiteshsai007:
## [MNG-8425] Fix mvnenc init saving invalid master source configuration

Fixes #10202

### Description

This PR fixes [MNG-8425](https://issues.apache.org/jira/browse/MNG-8425) (#10202) where `mvn --enc init` generates an invalid `settings-security4.xml`, causing `mvn --enc encrypt` to fail with "Maven Encryption is not configured."

### Root Cause

When the encryption wizard prompts the user to customize an editable value, the template is presented as `env:$VARIABLE_NAME`. The user is expected to replace only the `$VARIABLE_NAME` portion (e.g., typing `MVN_PASSWORD`), but the `Init` goal saves the raw user input directly — resulting in `<value>MVN_PASSWORD</value>` instead of the correct `<value>env:MVN_PASSWORD</value>`.

The `MasterDispatcher` in `plexus-sec-dispatcher` then fails validation because no `MasterSource` recognizes a config string without its required prefix (`env:`, `sys-property:`, etc.), producing the error: `Configured Source configuration not handled`.

### Fix

In `Init.java`, after the user provides their input for an editable field, the code now:
1. Extracts the prefix from the template (everything before the `$` placeholder).
2. Checks if the user's input already includes the prefix.
3. Prepends the prefix automatically if missing.

This ensures the generated `settings-security4.xml` always contains well-formed source values like `env:MVN_PASSWORD`.

### Testing

- All 468 tests in `maven-cli` pass.
- Manually verified that `mvn --enc init` now writes the correct `env:` prefix and `mvn --enc encrypt` succeeds afterward.

---

### DIFF
--- a/impl/maven-cli/src/main/java/org/apache/maven/cling/invoker/mvnenc/goals/Init.java
+++ b/impl/maven-cli/src/main/java/org/apache/maven/cling/invoker/mvnenc/goals/Init.java
@@ -144,19 +144,12 @@ public class Init extends InteractiveGoalSupport {
                             throw new InterruptedException();
                         }
                         String userInput = editMap.get("edit").getResult();
-                        if (template.contains("$")) {
-                            String prefix = template.substring(0, template.indexOf('$'));
-                            if (!prefix.isEmpty() && !userInput.startsWith(prefix)) {
-                                userInput = prefix + userInput;
-                            }
+                        String prefix = template.substring(0, template.indexOf('$'));
+                        if (!prefix.isEmpty() && !userInput.startsWith(prefix)) {
+                            userInput = prefix + userInput;
                         }
-                        final String finalResult = userInput;
-                        dispatcherConfigResult.put(editable.getKey(), new PromptResultItemIF() {
-                            @Override
-                            public String getResult() {
-                                return finalResult;
-                            }
-                        });
+                        String result = userInput;
+                        dispatcherConfigResult.put(editable.getKey(), () -> result);
                     }
                 }
 
--- a/impl/maven-cli/src/main/java/org/apache/maven/cling/invoker/mvnenc/goals/Init.java
+++ b/impl/maven-cli/src/main/java/org/apache/maven/cling/invoker/mvnenc/goals/Init.java
@@ -143,7 +143,20 @@ public class Init extends InteractiveGoalSupport {
                         if (editMap.isEmpty()) {
                             throw new InterruptedException();
                         }
-                        dispatcherConfigResult.put(editable.getKey(), editMap.get("edit"));
+                        String userInput = editMap.get("edit").getResult();
+                        if (template.contains("$")) {
+                            String prefix = template.substring(0, template.indexOf('$'));
+                            if (!prefix.isEmpty() && !userInput.startsWith(prefix)) {
+                                userInput = prefix + userInput;
+                            }
+                        }
+                        final String finalResult = userInput;
+                        dispatcherConfigResult.put(editable.getKey(), new PromptResultItemIF() {
+                            @Override
+                            public String getResult() {
+                                return finalResult;
+                            }
+                        });
                     }
                 }
 
--- a/impl/maven-cli/src/test/java/org/apache/maven/cling/invoker/mvnenc/goals/InitTest.java
+++ b/impl/maven-cli/src/test/java/org/apache/maven/cling/invoker/mvnenc/goals/InitTest.java
@@ -1,0 +1,131 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.maven.cling.invoker.mvnenc.goals;
+
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+import org.apache.maven.api.cli.InvokerRequest;
+import org.apache.maven.api.cli.mvnenc.EncryptOptions;
+import org.apache.maven.api.services.MessageBuilderFactory;
+import org.apache.maven.cling.invoker.mvnenc.EncryptContext;
+import org.codehaus.plexus.components.secdispatcher.DispatcherMeta;
+import org.codehaus.plexus.components.secdispatcher.SecDispatcher;
+import org.codehaus.plexus.components.secdispatcher.model.Config;
+import org.codehaus.plexus.components.secdispatcher.model.ConfigProperty;
+import org.codehaus.plexus.components.secdispatcher.model.SettingsSecurity;
+import org.jline.consoleui.prompt.ConsolePrompt;
+import org.jline.consoleui.prompt.PromptResultItemIF;
+import org.jline.consoleui.prompt.builder.PromptBuilder;
+import org.junit.jupiter.api.Test;
+import org.mockito.MockedConstruction;
+import org.mockito.Mockito;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.mockito.ArgumentMatchers.any;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.when;
+
+class InitTest {
+
+    @Test
+    void testPrefixPrependedToUserInput() throws Exception {
+        MessageBuilderFactory messageBuilderFactory = mock(MessageBuilderFactory.class, Mockito.RETURNS_DEEP_STUBS);
+        SecDispatcher secDispatcher = mock(SecDispatcher.class);
+
+        SettingsSecurity settingsSecurity = new SettingsSecurity();
+        when(secDispatcher.readConfiguration(true)).thenReturn(settingsSecurity);
+
+        DispatcherMeta meta = mock(DispatcherMeta.class);
+        when(meta.name()).thenReturn("master");
+        DispatcherMeta.Field field = mock(DispatcherMeta.Field.class);
+        when(field.getKey()).thenReturn("password");
+        when(meta.fields()).thenReturn(Collections.singletonList(field));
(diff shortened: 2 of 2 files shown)


elharo (changes requested):
needs test

Hiteshsai007:
@elharo Thanks for the review! 

I have addressed your feedback by adding comprehensive unit tests for the configuration saving logic. 

To accomplish this, I:
1. Refactored `Init.java` by extracting the prefix-prepending logic into a testable, package-private static method (`applyTemplatePrefix`).
2. Added a new `InitTest.java` suite that covers all edge cases, including:
   - **Missing prefixes**: prepending `env:` when the user only types `MVN_PASSWORD`.
   - **Duplicate prefixes**: preventing double-prefixing if the user manually types `env:MVN_PASSWORD`.
   - **Other prefix types**: ensuring it correctly handles `sys-property:`, `file:`, etc.
   - **Edge conditions**: properly handling empty user inputs, missing templates, and templates without placeholders.
3. Verified the changes locally with a full `mvn clean verify` on the `maven-cli` module to ensure all tests, Spotless formatting, and Checkstyle rules pass successfully.

Let me know if there's anything else needed before we can merge this!

elharo (commented):
This should be testable through the existing API or maybe an integration test. This approach is a red flag. I would start over by reproducing the issue in a failing test, and then implement a fix.

Hiteshsai007:
@elharo Thank you for the feedback. That is a completely fair point—extracting an internal static method purely to make it testable was definitely a smell.

I've pushed a new commit that starts over and follows your suggested approach:
1. Reverted the `applyTemplatePrefix` extraction and put the logic back inline where it originally belonged inside `Init.java`'s execution loop.
2. Wrote a new integration-style test (`InitTest.java`) that tests through the existing `Init.doExecute()` API. 
3. To test the interactive CLI logic without exposing internal state, I used Mockito's `MockedConstruction` to mock `ConsolePrompt` and simulate the sequence of user inputs that caused the original bug.
4. Verified that the test successfully fails by reproducing the issue (resulting in a saved value without the `env:` prefix), and then applied the inline fix to make it pass.

All `maven-cli` tests, spotless, and checkstyle checks are passing locally. Let me know if this aligns better with the project's testing patterns!

impl/maven-cli/src/main/java/org/apache/maven/cling/invoker/mvnenc/goals/Init.java
```diff
@@ -143,7 +143,20 @@ public int doExecute(EncryptContext context) throws Exception {
...
+                                userInput = prefix + userInput;
+                            }
+                        }
+                        final String finalResult = userInput;
+                        dispatcherConfigResult.put(editable.getKey(), new PromptResultItemIF() {
+                            @Override
+                            public String getResult() {
+                                return finalResult;
+                            }
+                        });
```
gnodet: `PromptResultItemIF` is structurally a functional interface (single abstract method `getResult()`, one default method). The anonymous inner class can be replaced with a lambda, which also eliminates the need for the `finalResult` intermediate variable:

```suggestion
                        String result = userInput;
                        dispatcherConfigResult.put(editable.getKey(), () -> result);
```

Same simplification applies to the `createResult()` helper in the test file.

impl/maven-cli/src/main/java/org/apache/maven/cling/invoker/mvnenc/goals/Init.java
```diff
@@ -143,7 +143,20 @@ public int doExecute(EncryptContext context) throws Exception {
...
                             throw new InterruptedException();
                         }
-                        dispatcherConfigResult.put(editable.getKey(), editMap.get("edit"));
+                        String userInput = editMap.get("edit").getResult();
+                        if (template.contains("$")) {
+                            String prefix = template.substring(0, template.indexOf('$'));
+                            if (!prefix.isEmpty() && !userInput.startsWith(prefix)) {
+                                userInput = prefix + userInput;
+                            }
+                        }
```
gnodet: This `template.contains("$")` guard is always true here — the `editables` list (line 127) is already filtered to entries whose result contains `$`, and `template` is assigned from `editable.getValue().getResult()`. The redundant check adds unnecessary nesting. Consider removing it:

```suggestion
                        String prefix = template.substring(0, template.indexOf('$'));
                        if (!prefix.isEmpty() && !userInput.startsWith(prefix)) {
                            userInput = prefix + userInput;
                        }
```

impl/maven-cli/src/test/java/org/apache/maven/cling/invoker/mvnenc/goals/InitTest.java
```diff
@@ -0,0 +1,131 @@
...
+        assertEquals(1, config.getProperties().size());
+        ConfigProperty prop = config.getProperties().get(0);
+        assertEquals("password", prop.getName());
+        // The expected value should be env:my_password_var because the template was env:$MVN_PASSWORD
+        assertEquals("env:my_password_var", prop.getValue());
+    }
+
+    private PromptResultItemIF createResult(String value) {
+        return new PromptResultItemIF() {
+            @Override
```
gnodet: The `createResult()` helper at the bottom of this file also uses an anonymous inner class where a lambda would suffice:
```java
private PromptResultItemIF createResult(String value) {
    return () -> value;
}
```

gnodet (changes requested):
Thanks for tackling MNG-8425 — the root-cause analysis is spot-on and the fix targets the right place. The prefix-extraction approach is correct for all known `MasterSource` types. A few things to address before this can be merged:

### Missing edge-case test coverage

The single test covers only the happy path (user omits the prefix). Please add tests for:
- **User already includes the prefix** (e.g., types `env:MY_VAR`) — verifies no double-prepending
- **Different prefix type** (e.g., template `system-property:$systemProperty`) — verifies it works beyond `env:`
- **Empty prefix** (e.g., template `$VAR`) — verifies user input is used as-is

### Inline fully-qualified class names in test

Several types are referenced by FQCN inline rather than imported at the top of the file (e.g., `java.nio.file.Paths`, `java.util.ArrayList`, `org.jline.utils.AttributedStyle`, `java.io.PrintWriter`, `org.jline.consoleui.prompt.ConfirmResult`, `org.jline.consoleui.elements.ConfirmChoice`). Other tests in `impl/maven-cli` use proper imports — please follow the same convention.

See the inline comments for the remaining items.

Note: this review does not replace specialized AI review tools (CodeRabbit, Sourcery) or static analyzers (SonarCloud).

Hiteshsai007:
Thanks for the thorough review! I've gone ahead and addressed all the feedback.

Here is a summary of the changes made in the latest commit:

1. **Missing edge-case test coverage:**
   I rewrote the test using JUnit 5's `@ParameterizedTest` and `@CsvSource` to effectively cover all the requested edge-cases without duplicating code. The test now validates:
   - User already including the prefix (e.g., `env:MY_VAR`) — verifies no double-prepending.
   - Different prefix types (e.g., `system-property:my_prop`) — verifies the logic handles arbitrary types beyond `env:`.
   - Empty prefixes (`$VAR`) — verifies user input is safely used as-is.

2. **Inline fully-qualified class names in test:**
   I removed the inline FQCNs (such as `java.nio.file.Paths`, `java.util.ArrayList`, `org.jline.consoleui.prompt.ConfirmResult`, etc.) and correctly imported them at the top of `InitTest.java` to align with the conventions of the other tests in `impl/maven-cli`.

3. **Replaced anonymous inner classes with lambdas:**
   Since `PromptResultItemIF` is structurally a functional interface, I simplified the code by replacing the anonymous inner classes with lambda expressions (`() -> result`) in both `Init.java` and the `createResult` helper in `InitTest.java`.

4. **Removed redundant `template.contains("$")` guard:**
   I removed the redundant `if (template.contains("$"))` check inside `Init.java`. The `editables` list is indeed already strictly filtered by `e.getValue().getResult().contains("$")` earlier in the pipeline, so removing the guard safely eliminates unnecessary nesting.

Let me know if there's anything else you'd like me to adjust!

gnodet (approved):
All review comments have been addressed — nice work!

- Redundant `template.contains("$")` check removed ✅
- Anonymous inner classes replaced with lambdas ✅
- Inline FQCNs converted to proper imports ✅
- Parameterized test now covers: prefix omitted, prefix already present, different prefix type (`system-property:`), and empty prefix (`$VAR`) ✅

The fix is clean and well-tested. Thanks for the thorough follow-up.

gnodet (approved):
The fix correctly addresses MNG-8425 by automatically prepending the template-derived prefix (e.g., `env:`) to user input when missing during `mvn --enc init`. The implementation is minimal (7 lines of production code), well-targeted, and backed by comprehensive parameterized tests. All CI checks pass across all platforms.

Minor observations (non-blocking):
- The `MockedConstruction` lambda parameter `mock` shadows the static import `Mockito.mock` — renaming to `promptMock` would improve readability.
- No test case for empty user input (user presses Enter without typing), where the result would be just the prefix with nothing after it.

All previous review feedback from maintainers has been addressed. The code is clean and ready to merge.

PR #12422, opened 2026-07-04
Use GAV in readAsParentModel cache-hit path for profile source keys

gnodet:
## Summary

Follow-up to #12406 — fixes the same GAPV vs GAV inconsistency in a different code path.

In `DefaultModelBuilder.readAsParentModel()`, the **cache-hit** path (line 1944) used `Model.getId()` which returns `groupId:artifactId:packaging:version` (GAPV), while the **cache-miss** path (line 1961) correctly used `ModelProblemUtils.toId()` which returns `groupId:artifactId:version` (GAV). This inconsistency could cause the same parent's profiles to be stored under different keys depending on whether the parent model was resolved from cache or freshly built.

The fix is a one-liner: replace `cached.model().getId()` with `ModelProblemUtils.toId(cached.model())` to match the cache-miss path.

- **Cache miss** (line 1961): `ModelProblemUtils.toId(model)` → `g:a:v` ✅
- **Cache hit** (line 1944): `cached.model().getId()` → `g:a:p:v` ❌ → **fixed** to `ModelProblemUtils.toId(cached.model())`

Also adds a test that verifies parent profile source keys use consistent GAV format across child modules in a multi-module build. Note: the cache-hit path requires a matching `ProfileActivationContext.Record` which is hard to trigger in a unit test (sibling modules record different model info), but the fix is correct by inspection — it matches the existing pattern on the cache-miss path.

## Test plan
- [x] Existing `testActivatedProfileBySource` passes
- [x] New `testParentProfileSourceKeyConsistentAcrossModules` passes
- [x] Full `DefaultMavenProjectBuilderTest` suite passes (34 tests)
- [x] Full `DefaultProjectBuilderTest` suite passes (3 tests)

### DIFF
--- a/impl/maven-core/src/test/java/org/apache/maven/project/DefaultMavenProjectBuilderTest.java
+++ b/impl/maven-core/src/test/java/org/apache/maven/project/DefaultMavenProjectBuilderTest.java
@@ -739,4 +739,59 @@ class DefaultMavenProjectBuilderTest extends AbstractMavenProjectTestCase {
         assertEquals("1.0", child.getArtifact().getVersion(), "artifact version should match inherited version");
         assertEquals("org.different.group", child.getGroupId(), "groupId should be from child POM");
     }
+
+    /**
+     * Tests that parent profile source keys use consistent GAV format (groupId:artifactId:version)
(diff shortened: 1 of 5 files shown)


gnodet:
Closing — fix pushed directly to the `issue-12405` branch on #12406 instead.

PR #12440, opened 2026-07-08
configure ATR project

hboutemy:
https://release-test.apache.org/projects/maven

notice the `download_path_suffix` that is quite special to match our `maven-X` structure in https://downloads.apache.org/maven/ / https://archive.apache.org/dist/maven/ / https://dist.apache.org/repos/dist/release/maven/ / https://dist.apache.org/repos/dist/atr/maven/ / https://dist.apache.org/repos/dist/dev/maven/

`{{MAJOR}}` won't be automatically replaced by ATR at the Finish stage, but I hope the release manager will see and update (and we could probably ask Tooling to force reject such a value if it remains as-is when submitting the form)

@cstamas @gnodet do you intend to use ATR for managing the VOTE of next Maven releases?

notice:
- this PR can be back-ported to 4.0.x and 3.10.x branches, even if it won't really do anything there: just for consistency
- configuration of `push-to-atr` profile will be done in another PR

### DIFF
--- a/.asf.yaml
+++ b/.asf.yaml
@@ -31,3 +31,30 @@ notifications:
   commits: commits@maven.apache.org
   issues: issues@maven.apache.org
   pullrequests: issues@maven.apache.org
+
+# https://release-test.apache.org/projects/maven
+project:
+  metadata:
+    key: maven
+    committee: maven
+    name: Apache Maven
+    description: "Maven is a project development management and comprehension tool.
+      Based on the concept of a project object model: builds, dependency management,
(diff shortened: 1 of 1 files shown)


hboutemy:
FYI, first draft of `push-to-atr` profile #12441

hboutemy:
will require https://github.com/apache/tooling-trusted-releases/issues/1381
for now, `{{MAJOR}}` is rejected, I updated by hand to `MAJOR` and release manager will have to replace after vote with effective value

PR #12443, opened 2026-07-08
Fix mvnup comment formatting for override comments

gnodet:
## Summary

- Fix `PluginUpgradeStrategy` to set `precedingWhitespace` on `Comment` nodes inserted before plugin elements, so the "Override version inherited from parent" comment appears on its own line instead of being appended to the preceding `</plugin>` closing tag
- Add test assertions to verify comments are not on the same line as closing tags

## Problem

When mvnup adds plugin management entries for plugins inherited from a remote parent, it inserts a `<!-- Override version inherited from parent -->` comment before the `<plugin>` element. The `Comment.of(...)` factory creates a comment with no `precedingWhitespace`, causing it to render on the same line as the previous `</plugin>` closing tag:

```xml
            </plugin><!-- Override version inherited from parent -->
            <plugin>
```

This breaks projects that enforce POM formatting (e.g. via `spotless-maven-plugin`), which expect:

```xml
            </plugin>
            <!-- Override version inherited from parent -->
            <plugin>
```

## Fix

Set `precedingWhitespace` on the `Comment` node to match the plugin element's indentation at both insertion sites in `PluginUpgradeStrategy`:
- `addPluginManagementEntryFromUpgrade()` — for `build/pluginManagement/plugins`
- `addDirectPluginOverrides()` — for `build/plugins`

Fixes gnodet/maven4-testing#23134

## Test plan

- [x] Existing `PluginUpgradeStrategyTest` passes (31 tests)
- [x] New assertions verify comments are on their own line (`assertFalse(xml.contains("</plugin><!--"), ...)`)
- [x] Full `maven-cli` module verify passes (468 tests)

### DIFF
--- a/impl/maven-cli/src/main/java/org/apache/maven/cling/invoker/mvnup/goals/PluginUpgradeStrategy.java
+++ b/impl/maven-cli/src/main/java/org/apache/maven/cling/invoker/mvnup/goals/PluginUpgradeStrategy.java
@@ -781,7 +781,10 @@ public class PluginUpgradeStrategy extends AbstractUpgradeStrategy {
                 managedPluginsElement, upgrade.groupId(), upgrade.artifactId(), upgrade.minVersion());
 
         if (fromRemoteParent) {
-            managedPluginsElement.insertChildBefore(plugin, Comment.of(" Override version inherited from parent "));
+            managedPluginsElement.insertChildBefore(
+                    plugin,
+                    Comment.of(" Override version inherited from parent ")
+                            .precedingWhitespace(plugin.precedingWhitespace()));
             context.detail("Added plugin management for " + upgrade.groupId() + ":" + upgrade.artifactId() + " version "
                     + upgrade.minVersion() + " (overrides version inherited from parent)");
         } else {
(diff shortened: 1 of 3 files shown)


gnodet:
**Note:** This PR is a workaround for a missing API in DomTrip. The `Editor` class handles whitespace automatically for element insertion (`insertElementBefore/After`) and comment appending (`addComment`), but has no positional comment insertion method (`insertCommentBefore/After`). This forced mvnup to use the raw `ContainerNode.insertChildBefore` API which doesn't handle `precedingWhitespace`.

Filed upstream: maveniverse/domtrip#254

Once DomTrip adds `Editor.insertCommentBefore()`, the manual `precedingWhitespace()` calls here should be replaced with the Editor API.

gnodet:
Updated: upgraded DomTrip to 1.6.0 which includes the new `Editor.insertCommentBefore()` / `insertCommentAfter()` API (maveniverse/domtrip#254). The manual `precedingWhitespace()` workaround has been replaced with the proper Editor API call.

gnodet:
Superseded by #12454 — commits cherry-picked into the consolidated PR.

PR #12446, opened 2026-07-08
Fix #12445: Fix deadlock in AbstractRequestCache when resolving parent POMs

### LINKED ISSUE #12445: Deadlock in AbstractRequestCache when resolving parent POMs with empty local repo
## Description

Maven 4 hangs indefinitely during builds when the local repository is empty (e.g. `-Dmaven.repo.local=/fresh/path`). The build freezes at `maven-remote-resources-plugin:process` while resolving parent POMs of transitive dependencies.

**The same build completes in seconds with a populated local repository.**

## Steps to Reproduce

```bash
git clone --depth 1 https://github.com/apache/httpcomponents-stylecheck.git
cd httpcomponents-stylecheck
# Works fine (cached):
mvn -B clean package -DskipTests
# Hangs forever (empty repo):
mvn -B clean package -DskipTests -Dmaven.repo.local=/tmp/fresh-repo
```

The build hangs at:
```
[INFO] --- remote-resources:3.0.0:process (process-resource-bundles) @ hc-stylecheck ---
[INFO] Preparing remote bundle org.apache:apache-jar-resource-bundle:1.4
[INFO] Copying 3 resources from 1 bundle.
```

## Thread Dump Analysis

The **main thread** is stuck in `AbstractRequestCache.lambda$requests$0` (line 95) calling `Object.wait()` on a `HashMap` that it already holds the lock on:

```
"main" in Object.wait()
   java.lang.Thread.State: WAITING (on object monitor)
    at java.lang.Object.wait0(Native Method)
    - waiting on <0x000000070dadf9a8> (a java.util.HashMap)
    at org.apache.maven.impl.cache.AbstractRequestCache.lambda$requests$0(AbstractRequestCache.java:95)
    - locked <0x000000070dadf9a8> (a java.util.HashMap)
    at org.apache.maven.impl.cache.CachingSupplier.apply(CachingSupplier.java:50)
    at org.apache.maven.impl.cache.AbstractRequestCache.requests(AbstractRequestCache.java:144)
    at org.apache.maven.impl.AbstractSession.requests(AbstractSession.java:177)
    at org.apache.maven.impl.DefaultArtifactResolver.doResolve(DefaultArtifactResolver.java:106)
    at org.apache.maven.impl.DefaultArtifactResolver.resolve(DefaultArtifactResolver.java:65)
    at org.apache.maven.impl.AbstractSession.resolveArtifact(AbstractSession.java:689)
    at org.apache.maven.impl.resolver.DefaultModelResolver.getPath(DefaultModelResolver.java:181)
    at org.apache.maven.impl.resolver.DefaultModelResolver.doResolveModel(DefaultModelResolver.java:155)
    at org.apache.maven.impl.resolver.DefaultModelResolver.resolveModel(DefaultModelResolver.java:116)
    at org.apache.maven.impl.model.DefaultModelBuilder...resolveParent(DefaultModelBuilder.java:1101)
    ...
    at org.apache.maven.plugin.resources.remote.ProcessRemoteResourcesMojo.getProjects(ProcessRemoteResourcesMojo.java:665)
```

Meanwhile, **all 15 worker threads** (5× `BfDependencyCollector`, 5× `BasicRepositoryConnector`, 5× `resolver`) are idle in `LinkedBlockingQueue.take()`, waiting for work that is never submitted.

## Root Cause

This appears to be a missed-signal deadlock in `AbstractRequestCache` / `CachingSupplier`:

1. The main thread enters `CachingSupplier.apply()` to resolve an artifact
2. Inside, it calls `Object.wait()` on a `HashMap`, expecting another thread to resolve the artifact and call `notify()`
3. But the resolution work is never dispatched to the worker threads — they sit idle in their thread pool queues
4. The main thread waits forever

When the local repo is populated, the `CachingSupplier` returns the cached result immediately without entering the wait path. The deadlock only triggers when actual remote resolution is needed.

## Impact

This affects any project using `maven-remote-resources-plugin` (very common — it's bound by default in the Apache parent POM) when built with a fresh local repository. This is the standard CI scenario where each build starts with an empty local repo.

In the maven4-testing suite (testing ~966 Apache projects against Maven 4), this deadlock caused **21 of 57 failures** — all were 1-hour timeouts on projects that build fine on Maven 3 in under 6 minutes.

## Environment

- Maven: 4.1.0-SNAPSHOT (commit 595458cd19)
- Java: OpenJDK 25.0.3+9-LTS
- OS: Linux
- Plugin: maven-remote-resources-plugin 3.0.0

Full thread dump attached: all 27 threads, showing the main thread's full stack trace and all idle worker threads.

gnodet:
## Summary

- Fix deadlock in `AbstractRequestCache.requests()` that caused Maven 4 to hang indefinitely when building with an empty local repository (e.g., `-Dmaven.repo.local=/tmp/fresh-repo`)
- Replace the `Object.wait()/notify()` pattern with direct value setting via `CachingSupplier.complete()`, eliminating the re-entrant deadlock scenario
- Add tests for the re-entrant scenario and batch caching behavior

## Root Cause

The `requests()` method used a wait/notify pattern with a local `HashMap` (`nonCachedResults`):
1. An `individualSupplier` lambda was created that called `nonCachedResults.wait()`, expecting the batch resolution to populate results and call `notifyAll()`
2. This `individualSupplier` was stored inside `CachingSupplier` instances via `doCache()`, which cached them by request key
3. When a re-entrant `requests()` call occurred (e.g., parent POM resolution during artifact resolution), `doCache()` returned a `CachingSupplier` from the outer call — wrapping the outer call's `individualSupplier`
4. The inner call's collection loop invoked this `CachingSupplier`, which waited on the outer call's `nonCachedResults` — a `HashMap` that could never be populated because the outer batch supplier was blocked waiting for the inner call to complete

## Fix

- Added `CachingSupplier.complete(Object result)` to directly set cached values without invoking the supplier
- Changed `requests()` to pass a single-item fallback supplier to `doCache()` (instead of the wait-based `individualSupplier`), so any re-entrant retrieval can resolve independently
- After batch resolution, results are set directly on `CachingSupplier` instances via `complete()`
- Removed the `synchronized` wait/notify pattern on `HashMap` entirely

## Test plan

- [x] Added `testReentrantRequestsDoesNotDeadlock` — reproduces the exact deadlock scenario (with 5-second timeout detection)
- [x] Added `testBatchResultsAreCached` — verifies batch results are properly cached for subsequent calls
- [x] All 499 existing tests in `maven-impl` pass
- [x] Full reactor build (`mvn clean install -DskipTests`) succeeds

### DIFF
--- a/impl/maven-impl/src/main/java/org/apache/maven/impl/cache/AbstractRequestCache.java
+++ b/impl/maven-impl/src/main/java/org/apache/maven/impl/cache/AbstractRequestCache.java
@@ -81,61 +81,73 @@ public abstract class AbstractRequestCache implements RequestCache {
      * @return List of results corresponding to the input requests
      * @throws BatchRequestException if any request in the batch fails
      */
+    /**
+     * {@inheritDoc}
+     * <p>
+     * This implementation avoids the use of {@code Object.wait()/notify()} to prevent deadlocks
+     * in re-entrant scenarios. When batch resolution triggers nested calls (e.g., parent POM
+     * resolution during artifact resolution), a cached {@link CachingSupplier} from an outer call
+     * could reference a stale wait-based supplier, causing the thread to wait on a
+     * {@code HashMap} that will never be notified.
+     * <p>
+     * Instead, this method:
+     * <ol>
+     *   <li>Passes a single-item fallback supplier to {@link #doCache} so any newly created
+     *       {@link CachingSupplier} can independently resolve its request</li>
+     *   <li>Performs batch resolution for efficiency</li>
+     *   <li>Directly sets results on {@link CachingSupplier} instances via {@link CachingSupplier#complete}</li>
+     * </ol>
+     */
     @Override
     @SuppressWarnings("unchecked")
     public <REQ extends Request<?>, REP extends Result<REQ>> List<REP> requests(
             List<REQ> reqs, Function<List<REQ>, List<REP>> supplier) {
-        final Map<REQ, Object> nonCachedResults = new HashMap<>();
-        List<RequestResult<REQ, REP>> allResults = new ArrayList<>(reqs.size());
-
-        Function<REQ, REP> individualSupplier = req -> {
-            synchronized (nonCachedResults) {
-                while (!nonCachedResults.containsKey(req)) {
-                    try {
-                        nonCachedResults.wait();
-                    } catch (InterruptedException e) {
-                        Thread.currentThread().interrupt();
-                        throw new RuntimeException(e);
-                    }
-                }
-                Object val = nonCachedResults.get(req);
-                if (val instanceof CachingSupplier.AltRes altRes) {
-                    uncheckedThrow(altRes.throwable);
-                }
-                return (REP) val;
-            }
-        };
+        // Create a fallback supplier that can resolve individual requests independently.
+        // This is stored in newly created CachingSupplier instances and ensures that
+        // re-entrant calls can resolve without deadlocking on a shared wait/notify map.
+        Function<REQ, REP> singleSupplier = req -> supplier.apply(List.of(req)).get(0);
 
         List<CachingSupplier<REQ, REP>> suppliers = new ArrayList<>(reqs.size());
         List<REQ> nonCached = new ArrayList<>();
         for (REQ req : reqs) {
-            CachingSupplier<REQ, REP> cs = doCache(req, individualSupplier);
+            CachingSupplier<REQ, REP> cs = doCache(req, singleSupplier);
             suppliers.add(cs);
             if (cs.getValue() == null) {
                 nonCached.add(req);
             }
         }
 
+        // Resolve non-cached requests in batch and directly set results on CachingSuppliers
         if (!nonCached.isEmpty()) {
-            synchronized (nonCachedResults) {
-                try {
-                    List<REP> reps = supplier.apply(nonCached);
-                    for (int i = 0; i < reps.size(); i++) {
-                        nonCachedResults.put(nonCached.get(i), reps.get(i));
+            // Build a map from request to its CachingSupplier index for efficient lookup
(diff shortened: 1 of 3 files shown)


gnodet (commented):
## Review Summary

This PR correctly fixes a real deadlock in `AbstractRequestCache.requests()` that caused Maven 4 to hang indefinitely when the local repository was empty. The fix replaces a fundamentally broken wait/notify pattern on a local `HashMap` with a well-designed `CachingSupplier.complete()` mechanism, a `ThreadLocal` re-entrancy guard, and a `batchResolving` flag for coordinating concurrent threads.

### Concurrency design

The approach is sound:
- **Deadlock prevention**: The `RESOLVING_ON_THREAD` ThreadLocal correctly detects re-entrant calls and falls through to independent single-item resolution
- **Thread safety**: `batchResolving` is `volatile` for cross-thread visibility; `complete()` and `apply()` synchronize on the same monitor
- **Lock ordering**: No nested synchronized blocks — eliminates lock-ordering concerns
- **`IdentityHashMap`**: Good defensive measure against mutable request `hashCode()` during batch resolution

### Test coverage

Five well-targeted tests covering: re-entrant deadlock, concurrent duplicate avoidance, unstable hash codes, partial batch results, and basic cache correctness. The re-entrant test directly reproduces the reported issue with a 5-second timeout for deadlock detection.

### Minor observations (none blocking)

- The `singleSupplier` fallback resolves one-at-a-time for re-entrant calls — correct trade-off vs deadlock
- `catch (Throwable e)` is broader than the original `MavenExecutionException`-only catch, but this is actually an improvement: prevents other threads from blocking indefinitely if an `Error` occurs
- `batchResolving` field could be moved next to the other field declarations for readability

Clean fix for a critical production issue. LGTM.

---

cstamas (approved):
LGTM, but see Copilot remarks

PR #12454, opened 2026-07-09
Fix #12430: mvnup upgrade strategies and compatibility improvements

gnodet:
## Summary

Consolidates #12430 and #12443 into a single PR, plus new features:

### New strategies
- **`RepositoryHttpsUpgradeStrategy`** — converts HTTP repository URLs to HTTPS (blocked since Maven 3.8.1+), with well-known canonical URL mappings for Maven Central and Apache repositories
- **`EnforcerVersionRangeStrategy`** — widens `requireMavenVersion` ranges that exclude Maven 4 (e.g., `[3.8.8,4)` → `[3.8.8,5)`)
- **`ResourceFilteringStrategy`** (#12455) — adds `nonFilteredFileExtensions` to the `maven-resources-plugin` when resource filtering is enabled, preventing `MalformedInputException` on binary files (`.xlsx`, `.docx`, `.png`, etc.)
- **`DeduplicateDependenciesStrategy`** — standalone strategy for removing duplicate `<dependency>` and `<plugin>` declarations using last-wins semantics (extracted from `CompatibilityFixStrategy`)

### Quarkus plugin improvements
- **Skip upgrade for Quarkus 2.x** — detects the Quarkus platform version from `dependencyManagement` (quarkus-bom) or well-known properties (`quarkus.platform.version`, `quarkus.version`) and skips the `quarkus-maven-plugin` upgrade when the project uses Quarkus 2.x, preventing `NoSuchMethodError` build failures
- Quarkus plugin upgrade support, BOM version decoupling, and version gap warnings (from #12430)

### Bug fixes
- Fix mvnup comment formatting for override comments (from #12443)

### Other improvements (from #12430)
- Warn about property-interpolated module paths
- Warn about CI-friendly `${revision}` missing dependency versions
- Auto-upgrade `gmavenplus-plugin` to Maven 4-compatible version
- Upgrade DomTrip to 1.6.0 (from #12443)

### Removed (moved to #12464)
- ~~Silently ignore unrecognized CLI options from `.mvn/maven.config`~~ — root cause is that `bin/mvn` passes `$MAVEN_ARGS` to mvnup unconditionally; fix belongs in the shell script

Supersedes #12430 and #12443 (commits cherry-picked).

## Test plan
- [x] 27 tests for `RepositoryHttpsUpgradeStrategy`
- [x] 22 tests for `EnforcerVersionRangeStrategy`
- [x] 20 tests for `ResourceFilteringStrategy`
[...]

PR #12465, opened 2026-07-10
Fix #12464: Skip MAVEN_ARGS for non-default main classes (--up, --enc, --shell)

### LINKED ISSUE #12464: mvnup should not inherit MAVEN_ARGS environment variable
The `mvnup` shell script delegates to `mvn --up "$@"`, and the `mvn` launcher unconditionally prepends `$MAVEN_ARGS` to the command line (line 299 in `bin/mvn`):

```bash
eval exec "$cmd" '$MAVEN_ARGS' '"$@"'
```

When `MAVEN_ARGS` contains build-specific flags (e.g. `-ntp`, `-T4`, `-U`), these are passed to `MavenUpCling` which does not recognize them.

The current workaround in PR #12454 is a retry loop in `CommonsCliUpgradeOptions.CLIManager.parse()` that catches `UnrecognizedOptionException` and strips unknown options one by one — this is fragile and hides real argument errors from the user.

### Proposed fix

In the `mvn` shell script (`bin/mvn` and `bin/mvn.cmd`), skip `$MAVEN_ARGS` when the main class is not the default `MavenCling`. The `handle_args` function already detects `--up`/`--enc`/`--shell` to set `MAVEN_MAIN_CLASS`; use that to gate `MAVEN_ARGS` inclusion:

```bash
# Only pass MAVEN_ARGS for the default Maven build command
if [ "$MAVEN_MAIN_CLASS" = "org.apache.maven.cling.MavenCling" ]; then
  eval exec "$cmd" '$MAVEN_ARGS' '"$@"'
else
  eval exec "$cmd" '"$@"'
fi
```

Same change needed in `bin/mvn.cmd`.

Once fixed, the retry loop in `CommonsCliUpgradeOptions` can be removed — replaced with a simple `super.parse(args)` that fails fast on truly unrecognized options.

gnodet:
## Summary

- In `bin/mvn`, gate `$MAVEN_ARGS` inclusion on `MAVEN_MAIN_CLASS` equaling the default `MavenCling`. When `--up`, `--enc`, or `--shell` is used, the `handle_args` function sets `MAVEN_MAIN_CLASS` to a different class — skip `MAVEN_ARGS` in that case to avoid passing unrecognized build-specific flags (e.g. `-ntp`, `-T4`, `-U`) to sub-commands.
- Apply the same fix in `bin/mvn.cmd` by clearing `MAVEN_ARGS` when the main class is not `MavenCling` (safe because the script runs inside `@setlocal`).
- Debug logging (`MAVEN_DEBUG_SCRIPT`) is also updated to reflect the conditional behavior.

Closes #12464

## Test plan

- [x] Module builds successfully (`mvn verify -pl apache-maven`)
- [ ] Run `MAVEN_ARGS="-ntp -T4" mvn --up check` — should no longer fail with unrecognized option errors
- [ ] Run `MAVEN_ARGS="-ntp -T4" mvn clean install` — should still pass MAVEN_ARGS as before
- [ ] Run `MAVEN_ARGS="-ntp" mvn --enc` — should not fail on `-ntp`
- [ ] Verify `MAVEN_DEBUG_SCRIPT=1 mvn --up check` does not print MAVEN_ARGS in debug output

### DIFF
--- a/apache-maven/src/assembly/maven/bin/mvn
+++ b/apache-maven/src/assembly/maven/bin/mvn
@@ -290,10 +290,20 @@ cmd="\"$JAVACMD\" \
 
 if [ -n "$MAVEN_DEBUG_SCRIPT" ]; then
   echo "[DEBUG] Launching JVM with command:" >&2
-  printf '[DEBUG]   %s' "$cmd" >&2; printf ' %s' "$MAVEN_ARGS" >&2; printf ' "%s"' "$@" >&2; echo >&2
+  printf '[DEBUG]   %s' "$cmd" >&2
+  if [ "$MAVEN_MAIN_CLASS" = "org.apache.maven.cling.MavenCling" ]; then
+    printf ' %s' "$MAVEN_ARGS" >&2
+  fi
+  printf ' "%s"' "$@" >&2; echo >&2
 fi
 
 # User arguments ("$@") and MAVEN_ARGS are passed outside the eval'd string
 # to preserve literal values (backslashes, ${...} placeholders) without
 # shell re-parsing. Only the base command uses eval for MAVEN_OPTS word splitting.
-eval exec "$cmd" '$MAVEN_ARGS' '"$@"'
+# MAVEN_ARGS is only passed for the default Maven build command (MavenCling),
+# not for sub-commands like --up, --enc, or --shell which have their own options.
+if [ "$MAVEN_MAIN_CLASS" = "org.apache.maven.cling.MavenCling" ]; then
+  eval exec "$cmd" '$MAVEN_ARGS' '"$@"'
+else
+  eval exec "$cmd" '"$@"'
+fi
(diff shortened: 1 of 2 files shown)


gnodet (commented):
## Review — Skip MAVEN_ARGS for Non-Default Main Classes

Clean, well-scoped fix that gates `$MAVEN_ARGS` on the main class being the default `MavenCling`. Both Unix and Windows scripts are updated with platform-idiomatic approaches.

### Analysis

- **Unix (`bin/mvn`)**: Conditional `if/else` around the `eval exec` — debug output is also correctly gated so it reflects what's actually executed.
- **Windows (`bin/mvn.cmd`)**: Clears `MAVEN_ARGS` before the exec line — scoped inside `@setlocal`/`@endlocal` so it doesn't leak to the caller.
- **Edge case**: If a user intentionally puts args in `MAVEN_ARGS` for sub-tools, those will be silently dropped. This is correct — `MAVEN_ARGS` is documented as Maven-specific, and sub-tools have their own option sets.
- **Complements #12454**: With the retry loop reverted, this shell-level gate is the clean solution — sub-tools will properly fail fast on truly unrecognized options.

No issues found. LGTM (own PR, can't APPROVE). ✅

---

cstamas (approved):

PR #12482, opened 2026-07-13
Drop mention of Plexus XML

cstamas:
As there is no such file anymore. Still, the sentence as is, is correct, just how they are defined was not true (in Plexus XML file).


slawekjaranowski (approved):

gnodet (approved):
## Review: Drop mention of Plexus XML

**Verdict: APPROVE** ✅

Clean doc fix. Removes the outdated reference to `META-INF/plexus/default-bindings.xml` from the site documentation for default lifecycle bindings. The file no longer exists in the Maven codebase (removed as part of the Plexus → Maven 4 migration). The surrounding sentence remains grammatically sound and accurately describes that plugin bindings are defined separately because they are packaging-specific.

PR #12483, opened 2026-07-13
Fix UT: do not use checkout directly

cstamas:
Instead, it should use transient (copied) resources under target.


slawekjaranowski (approved):

gnodet (approved):
## Review: Fix UT — do not use checkout directly

**Verdict: APPROVE** ✅

Correct single-line test fix in `AbstractProjectInheritanceTestCase.java` — changes the local repository path from `src/test/resources/inheritance-repo/` to `target/test-classes/inheritance-repo/`. This follows the standard Maven convention: tests should reference resources copied to `target/test-classes/` (populated during `process-test-resources`) rather than reading directly from `src/test/resources/` in the source checkout. Avoids issues with read-only source trees, out-of-tree builds, and test isolation.

PR #12505, opened 2026-07-18
fix: mvnup: widen exact Maven version pins to allow Maven 4

gnodet:
## Summary
- Projects like Apache Flink pin their Maven version to an exact version using ranges like `[3.8.6,3.8.6]` (both bounds equal, both inclusive). The existing `MAVEN4_EXCLUSIVE_UPPER_BOUND` pattern only matched ranges with an exclusive upper bound at `4.x` (e.g. `[3.8.8,4)`), so exact pins and ranges with an upper bound below 4 were silently skipped.
- Added a new `UPPER_BOUND_BELOW_MAVEN4` pattern that catches any version range whose upper bound has a major version less than 4, covering exact version pins (`[3.8.6,3.8.6]` -> `[3.8.6,5)`), sub-4 upper bounds (`[3.8.0,3.9)` -> `[3.8.0,5)`), and unbounded lower ranges (`(,3.9]` -> `(,5)`).
- Added 6 new tests for the new patterns and updated 1 existing test to reflect the corrected behavior for sub-4 upper bounds.

## Test plan
- [x] All 27 `EnforcerVersionRangeStrategyTest` tests pass (was 22 before)
- [x] Exact version pin `[3.8.6,3.8.6]` is now widened to `[3.8.6,5)`
- [x] Edge cases tested: `[3.0,3.0]`, `[3.9.0,3.9.0]`, `[4.0.0,4.0.0]` (not widened)
- [x] Sub-4 upper bound `[3.8.8,3.9)` is now widened to `[3.8.8,5)`
- [x] Full POM integration test verifies exact pin widening in plugin configuration
- [x] Existing tests for `[3.8.8,4)`, `[3,4)`, etc. still pass unchanged

### DIFF
--- a/impl/maven-cli/src/main/java/org/apache/maven/cling/invoker/mvnup/goals/EnforcerVersionRangeStrategy.java
+++ b/impl/maven-cli/src/main/java/org/apache/maven/cling/invoker/mvnup/goals/EnforcerVersionRangeStrategy.java
@@ -87,6 +87,22 @@ public class EnforcerVersionRangeStrategy extends AbstractUpgradeStrategy {
      */
     static final Pattern MAVEN4_EXCLUSIVE_UPPER_BOUND = Pattern.compile("^(\\[|\\()(.+?),\\s*(4(?:\\.\\d+)*)\\s*\\)$");
 
+    /**
+     * Pattern to match version ranges where the upper bound has a major version below 4,
+     * which blocks Maven 4. This includes exact version pins like {@code [3.8.6,3.8.6]}
+     * and ranges like {@code [3.8.0,3.9)}, {@code (,3.9]}.
+     * Captures:
+     *   Group 1: opening bracket ([ or ()
+     *   Group 2: lower bound (may be empty for unbounded lower ranges like {@code (,3.9]})
+     *   Group 3: upper bound (numeric version like 3.8.6, 3.9, 3)
+     *   Group 4: closing bracket () or ])
+     *
+     * Examples matched: [3.8.6,3.8.6], [3.8.0,3.9), (,3.9], [3.0,3.0], [3.9.0,3.9.0]
+     * Examples not matched: [3.8.8,), 3.8.8, [3.8.8,5), [3.8.8,4) (handled by MAVEN4_EXCLUSIVE_UPPER_BOUND)
+     */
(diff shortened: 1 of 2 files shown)


cstamas (approved):

gnodet (commented):
## Review: mvnup: widen exact Maven version pins to allow Maven 4

Clean fix that closes an important gap in `mvnup`'s `EnforcerVersionRangeStrategy` — exact version pins like `[3.8.6,3.8.6]` and sub-4 upper bounds like `[3.8.0,3.9)` now correctly get widened to allow Maven 4.

### What this PR does well

- **Correct regex design** — `UPPER_BOUND_BELOW_MAVEN4` captures the right cases, and the pattern ordering (checking `MAVEN4_EXCLUSIVE_UPPER_BOUND` first) is essential and well-designed
- **Safe fallback** — `getMajorVersion()` returns `Integer.MAX_VALUE` for unparseable versions, which acts as a "do not widen" guard
- **Comprehensive tests** — 5 new unit tests + 1 integration test covering exact pins, short versions, and Maven 4 non-widening
- **Good Javadoc** — the new pattern documents matched/unmatched examples clearly

### Suggestions (low priority)

1. **[Low] Missing test for unbounded lower bound `(,3.9]`** — the Javadoc lists this as a matched example, but there's no test exercising it. The regex handles it correctly (group 2 captures empty string), but a test would document the behavior.

2. **[Low] No test documenting `[3.8.8,4]` behavior** — a range with an *inclusive* upper bound at exactly 4 is not widened by either pattern (first requires `)`, second sees major=4 which is not < 4). In Maven semantics `[3.8.8,4]` includes `4.0.0` but blocks `4.0.1+`. The non-widening behavior is reasonable, but a test documenting this decision would be useful.

### Checklist

| Check | Status |
|-------|--------|
| Tests | ✓ Comprehensive coverage |
| Docs | ✓ Javadoc updated |
| Commit convention | ✓ `fix:` prefix |
| Public API / backward compat | ✓ Package-private method |
| Security | N/A |
| CI | ✓ All 24 checks pass |

**LGTM.** 

---

_Reviewed with Claude Code on behalf of Guillaume Nodet. This review was generated by an AI agent and may contain inaccuracies; please verify all suggestions before applying._

Bukama (approved):

PR #12515, opened 2026-07-21
Fix #12507: dont wipe unresolved ${...} expressions in CLI-supplied property values

### LINKED ISSUE #12507: `${...}` expressions in CLI-supplied plugin parameter values are wiped to empty at CLI parse time (Maven 3 resolves them at mojo configuration)
### Affected versions

### Reproduction

```
```

### Analysis

### Impact

### Test coverage

ascheman:
Fixes #12507.

## Root cause

`BaseParser.populateUserProperties()` interpolates user-specified `-D` values at CLI parse time via the `Interpolator` convenience overload, which defaults to `defaultsToEmpty=true`, against a callback that only knows the `extraInterpolationSource()` paths (`session.topDirectory` / `session.rootDirectory`). Every other expression — system properties like `${user.dir}`, project references like `${project.build.directory}` — was silently replaced with the empty string before any session/project context existed.

Maven 3 resolves such expressions at mojo configuration time; the v3-compat `PluginParameterExpressionEvaluator` in Maven 4 would too, but it only ever saw the already-wiped value.

This is independent of the recent launcher quoting fixes (#11978 / #11983): those fixed the shell layer so `${...}` arguments reach the JVM intact. On rc-5 the launcher bug masked this issue with a loud `bad substitution` shell error; with the launcher fix shipping in rc-6, users hit this silent empty-substitution instead.

## Fix

Pass `defaultsToEmpty=false` for the user-specified properties: placeholders that cannot be resolved from the early paths map stay literal and are evaluated later with full session/project context. The `${session.topDirectory}` / `${session.rootDirectory}` support from #2480 is unaffected.

## Testing

- New IT `MavenITgh12507CliParamNestedInterpolationTest` (two cases: `${user.dir}` and `${project.build.directory}` in a CLI-supplied plugin parameter): red against the current master distro (`expected: <PRE-…-POST> but was: <PRE--POST>`), green with the fix.
- All 33 `maven-cli` test suites pass.
- End-to-end: `mvn deploy '-DaltDeploymentRepository=test::file://${project.build.directory}/deploy'` deploys to `target/deploy/…` again (was: `file:///deploy` → `Read-only file system`).

### DIFF
--- a/impl/maven-cli/src/main/java/org/apache/maven/cling/invoker/BaseParser.java
+++ b/impl/maven-cli/src/main/java/org/apache/maven/cling/invoker/BaseParser.java
@@ -438,7 +438,11 @@ public abstract class BaseParser implements Parser {
         Map<String, String> userSpecifiedProperties = context.options != null
                 ? new HashMap<>(context.options.userProperties().orElse(new HashMap<>()))
                 : new HashMap<>();
-        createInterpolator().interpolate(userSpecifiedProperties, paths::get);
+        // Resolve only early-available path placeholders (session.topDirectory & co); leave everything
+        // else literal so it can be evaluated later with full session/project context by the
+        // plugin parameter expression evaluator (gh-12507: wiping unresolved ${...} to the empty
+        // string silently broke e.g. -DaltDeploymentRepository=id::file://${project.build.directory}/x).
+        createInterpolator().interpolate(userSpecifiedProperties, paths::get, false);
 
         // ----------------------------------------------------------------------
         // Load config files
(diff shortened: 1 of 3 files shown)


gnodet:
Superseded by #12523, which includes all three commits from this PR plus the `MavenITgh11978PlaceholderInCliArgTest` assertion update that unblocks CI. Thank you @ascheman for the excellent root-cause analysis and fix!

gnodet:
Thank you for the excellent root-cause analysis and fix @ascheman! Unfortunately I cannot push to your fork branch (token permissions), so I've created #12523 which includes your two commits (rebased) plus the `MavenITgh11978PlaceholderInCliArgTest` assertion update needed to unblock CI. Your authorship is preserved on the original commits.

PR #12520, opened 2026-07-22
Introduce validation control

cstamas:

Related: https://github.com/apache/maven-resolver/issues/2007

### DIFF
--- a/api/maven-api-core/src/main/java/org/apache/maven/api/Constants.java
+++ b/api/maven-api-core/src/main/java/org/apache/maven/api/Constants.java
@@ -446,6 +446,17 @@ public final class Constants {
     public static final String MAVEN_RESOLVER_TRANSPORT = "maven.resolver.transport";
 
     /**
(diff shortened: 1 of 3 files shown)


gnodet (approved):
## Code Review

Clean, well-scoped escape-hatch feature introducing configurable validation levels (`maven.resolver.validation`) for the Maven Resolver validator. The three levels (DEFAULT, MILD, OFF) are well-documented in `Constants.java` and accurately reflect the implementation. Default behavior is preserved, singleton pattern for validator instances is efficient, and the `@since 3.10.0` tag is consistent with existing convention.

Two minor notes (non-blocking):
- No unit tests for the new `MavenValidator(boolean)` constructor behavior — while the feature is simple, tests covering the three validation levels would improve confidence.
- The `ValidationLevel` enum could be made package-private since `MavenValidatorFactory` is in the `impl` module.

gnodet (approved):

PR #12523, opened 2026-07-23
Fix #12507: preserve unresolved ${...} in CLI -D values (Maven 3 semantics)

### LINKED ISSUE #12507: `${...}` expressions in CLI-supplied plugin parameter values are wiped to empty at CLI parse time (Maven 3 resolves them at mojo configuration)

gnodet:
## Summary

Fixes #12507. Supersedes #12515 (adds the missing test update that unblocks CI).

`BaseParser.populateUserProperties()` interpolates user-specified `-D` values at CLI parse time against a callback that only knows `session.topDirectory` / `session.rootDirectory`. Every other expression — `${project.build.directory}`, `${user.dir}`, etc. — was silently replaced with `""` because the `Interpolator.interpolate(Map, UnaryOperator)` convenience overload defaults to `defaultsToEmpty=true`.

This broke common CI patterns like `-DaltDeploymentRepository=id::file://${project.build.directory}/deploy` (URL becomes `file:///deploy` → wrong path, no error).

### Root cause

PR #2480 added early interpolation of `-D` values to support `${session.topDirectory}`, but the `defaultsToEmpty=true` API default was not appropriate for this call site — it wiped all expressions the early callback couldn't resolve. Maven 3 left them literal for later evaluation at mojo configuration time.

### Changes

- **`BaseParser.java`**: Pass `defaultsToEmpty=false` so unresolved placeholders stay literal for later evaluation by the plugin parameter expression evaluator. The `${session.topDirectory}` / `${session.rootDirectory}` support from #2480 is unaffected.
- **`MavenITgh11978PlaceholderInCliArgTest`**: Update assertion to expect Maven 3 semantics — unknown `${...}` preserved as literal, not emptied. The emptying was an unintended side effect of #2480, not a deliberate behavior. The test's primary purpose (launcher doesn't crash on `${...}`) is unchanged.
- **New IT `MavenITgh12507CliParamNestedInterpolationTest`**: Two cases (`${user.dir}` and `${project.build.directory}` in CLI-supplied plugin parameters) — red on current master, green with the fix.

### Credits

Original fix and IT by @ascheman in #12515. This PR adds the `MavenITgh11978PlaceholderInCliArgTest` update and rebases onto current master to unblock CI.

### DIFF
--- a/impl/maven-cli/src/main/java/org/apache/maven/cling/invoker/BaseParser.java
+++ b/impl/maven-cli/src/main/java/org/apache/maven/cling/invoker/BaseParser.java
@@ -438,7 +438,11 @@ public abstract class BaseParser implements Parser {
         Map<String, String> userSpecifiedProperties = context.options != null
                 ? new HashMap<>(context.options.userProperties().orElse(new HashMap<>()))
                 : new HashMap<>();
-        createInterpolator().interpolate(userSpecifiedProperties, paths::get);
+        // Resolve only early-available path placeholders (session.topDirectory & co); leave everything
+        // else literal so it can be evaluated later with full session/project context by the
+        // plugin parameter expression evaluator (gh-12507: wiping unresolved ${...} to the empty
+        // string silently broke e.g. -DaltDeploymentRepository=id::file://${project.build.directory}/x).
+        createInterpolator().interpolate(userSpecifiedProperties, paths::get, false);
 
         // ----------------------------------------------------------------------
         // Load config files
--- a/its/core-it-suite/src/test/java/org/apache/maven/it/MavenITgh11978PlaceholderInCliArgTest.java
+++ b/its/core-it-suite/src/test/java/org/apache/maven/it/MavenITgh11978PlaceholderInCliArgTest.java
@@ -50,8 +50,10 @@ class MavenITgh11978PlaceholderInCliArgTest extends AbstractMavenIntegrationTest
         verifier.verifyErrorFreeLog(); // primary check: shell crash produces non-zero exit
 
         // Secondary: verify the literal placeholder flowed through Maven.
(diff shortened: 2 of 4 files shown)


cstamas (approved):

ascheman:
Thanks @gnodet — the direction reads right to me. Treating the gh-11978 empty-for-unknown expectation as the anomaly (rather than scoping the parse-time interpolation) restores Maven 3's literal-for-unknown behaviour uniformly, which is what the cross-version repro showed 3.9.16 and 3.10.0-rc-1 both do. Preserving unresolved `${...}` at CLI parse time and letting `PluginParameterExpressionEvaluator` resolve project/session context later at mojo time is exactly the semantics `-DaltDeploymentRepository=…file://${project.build.directory}/…` needs.

CI is green here including the ITs. I've gone ahead and set the `bug` label and the `4.0.0-rc-6` milestone on this PR (matching #12507) so it's on the rc-6 radar — revert if you'd rather it not carry those. Happy to help with a `maven-4.0.x` backport once this lands on master; the fix applied cleanly to both lines when I tested it. Thanks again for carrying it over the line.

ascheman:
Correction on my milestone note above: since this PR targets `master` and rc-6 ships from the `maven-4.0.x` branch, the `4.0.0-rc-6` milestone was misleading here — a master merge alone won't be in rc-6. I've removed it and added `backport-to-4.0.x` instead, which captures what actually needs to happen for the fix to reach rc-6. As offered, I'm happy to prepare that 4.0.x backport once this lands on master (it applied cleanly to both lines in my earlier testing).

cstamas (approved):

PR #12524, opened 2026-07-23
Backport #12523 to 4.0.x: preserve unresolved ${...} in CLI -D values

### LINKED ISSUE #12507: `${...}` expressions in CLI-supplied plugin parameter values are wiped to empty at CLI parse time (Maven 3 resolves them at mojo configuration)

ascheman:
Backport of #12523 (fix for #12507) to the `maven-4.0.x` line, so the fix ships in `4.0.0-rc-6`. cc @gnodet (confirmed on Slack).

The three commits from #12523 cherry-pick cleanly — the production fix is byte-identical to what landed on `master`:

- **`Add failing IT for #12507`** — new `MavenITgh12507CliParamNestedInterpolationTest`
- **`Fix #12507: dont wipe unresolved ${...} in -D values`** — `BaseParser.populateUserProperties` passes `defaultsToEmpty=false`, preserving unresolved `${...}` at CLI parse time so the plugin parameter expression evaluator resolves project/session context later (Maven 3 semantics)
- **`Update gh-11978 IT to expect Maven 3 semantics for unknown placeholders`** (@gnodet)

One backport-specific commit on top:

- **`Port gh-12507 IT to the 4.0.x test-harness API`** — the new IT was written against master's newer test harness (Path-based `Verifier`, no-arg `AbstractMavenIntegrationTestCase`); on `maven-4.0.x` the harness is File/String-based and requires a version-range `super()` call. Ported accordingly and gated `[4.0.0-rc-6,)` (the fix's first release; the built `4.0.0-SNAPSHOT` satisfies the range, so the body runs rather than skips).

**Local verification** against a `maven-4.0.x` distribution built from this branch (JDK 17):

```
MavenITgh12507CliParamNestedInterpolationTest  Tests run: 2, Failures: 0, Skipped: 0
MavenITgh11978PlaceholderInCliArgTest          Tests run: 1, Failures: 0, Skipped: 0
```

Both actually execute (not skipped by the version gate). Original authorship preserved on the cherry-picked commits.

Refs #12507, #12523.

### DIFF
--- a/impl/maven-cli/src/main/java/org/apache/maven/cling/invoker/BaseParser.java
+++ b/impl/maven-cli/src/main/java/org/apache/maven/cling/invoker/BaseParser.java
@@ -438,7 +438,11 @@ public abstract class BaseParser implements Parser {
         Map<String, String> userSpecifiedProperties = context.options != null
                 ? new HashMap<>(context.options.userProperties().orElse(new HashMap<>()))
                 : new HashMap<>();
-        createInterpolator().interpolate(userSpecifiedProperties, paths::get);
+        // Resolve only early-available path placeholders (session.topDirectory & co); leave everything
+        // else literal so it can be evaluated later with full session/project context by the
+        // plugin parameter expression evaluator (gh-12507: wiping unresolved ${...} to the empty
+        // string silently broke e.g. -DaltDeploymentRepository=id::file://${project.build.directory}/x).
+        createInterpolator().interpolate(userSpecifiedProperties, paths::get, false);
 
         // ----------------------------------------------------------------------
         // Load config files
--- a/its/core-it-suite/src/test/java/org/apache/maven/it/MavenITgh11978PlaceholderInCliArgTest.java
+++ b/its/core-it-suite/src/test/java/org/apache/maven/it/MavenITgh11978PlaceholderInCliArgTest.java
@@ -56,8 +56,10 @@ class MavenITgh11978PlaceholderInCliArgTest extends AbstractMavenIntegrationTest
         verifier.verifyErrorFreeLog(); // primary check: shell crash produces non-zero exit
 
         // Secondary: verify the literal placeholder flowed through Maven.
-        // Maven resolves the unknown ${some.maven.placeholder} to empty.
+        // Unknown ${...} expressions are preserved as literals (Maven 3 semantics, see gh-12507);
+        // they are not replaced with empty strings at CLI parse time.
         Properties props = verifier.loadProperties("target/pom.properties");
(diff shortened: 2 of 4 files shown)


gnodet (commented):
## Review Summary

Clean backport of #12523 (fix for #12507) to the `maven-4.0.x` branch. The one-line production fix is correct: passing `defaultsToEmpty=false` to the interpolator preserves unresolved `${...}` expressions at CLI parse time so they can be resolved later by the plugin parameter expression evaluator with full session/project context. The test harness adaptation for the 4.0.x API is faithfully done.

## What I Love

- **Exemplary PR description.** The commit-by-commit breakdown, the explanation of what was cherry-picked vs. what was adapted for 4.0.x, and the local test output make this PR very easy to review. This is a gold standard for backport PRs.

- **Surgical production fix with excellent commentary.** The 4-line comment above the changed `interpolate()` call in `BaseParser.populateUserProperties()` explains *why* the change is needed, names the real-world trigger (`-DaltDeploymentRepository=id::file://${project.build.directory}/x`), and references the issue. Future readers will immediately understand the reasoning.

- **Thoughtful test design.** `MavenITgh12507CliParamNestedInterpolationTest` covers both the system-property path (`${user.dir}`) and the project-expression path (`${project.build.directory}`) as separate test methods. This makes failures immediately diagnostic — we will know exactly which resolution path broke.

## Findings

### Critical

None — nice work!

### Important

None.

### Suggestions & Nits

**[Question]** The upstream PR #12523 targeting `master` is still open. This means the backport is landing before the fix reaches the primary development branch. I trust this was a deliberate sequencing decision for the `4.0.0-rc-6` release timeline (the PR description mentions Slack confirmation from @gnodet). Just flagging for awareness in case the master PR needs any adjustments that should be reflected here before merge.

**[Nit]** In `MavenITgh12507CliParamNestedInterpolationTest`, the `extractResources()` call chains `.getAbsoluteFile().toPath()`, but `extractResources` on the 4.0.x harness already returns an absolute `File` (it calls `.getAbsoluteFile()` internally). The extra `.getAbsoluteFile()` is harmless and matches the pattern in `MavenITgh11978PlaceholderInCliArgTest`, so this is purely a consistency observation — no change needed.

## Overall

This is a well-executed backport that faithfully preserves the fix semantics from master while correctly adapting the integration test to the 4.0.x test harness (File-based `Verifier`, version-range `super()` call, resource path convention). The version gate `[4.0.0-rc-6,)` is appropriate. Looks good to merge once CI is green and the sequencing with #12523 is confirmed.

---

PR #12528, opened 2026-07-24
Fix #12522: handle unprovisionable ProjectExecutionListeners from non-extension plugins

### LINKED ISSUE #12522: Extensions are picked up from non-extension marked plugins
### Affected version

3.10, 4.0, 4.1

### Bug description

It seems that all JSR330 migrated Maven versions picks up (and if extensions are Plexus, errors out) extensions from plugins that are not configured as extensions.

See reproducer https://github.com/cstamas/jsr330

gnodet:
## Summary

Fixes https://github.com/apache/maven/issues/12522

When a plugin ships JSR 330 components (e.g. `ProjectExecutionListener` implementations) but is **not** configured with `<extensions>true</extensions>`, Sisu's component discovery still registers them in the global `BeanLocator`. The live-injected `List<ProjectExecutionListener>` in `LifecycleModuleBuilder` picks these up, and lazy provisioning fails when the component's plugin-internal dependencies (e.g. Tycho's `MavenReactorRepository`) are not bound in Maven's container — aborting the build with `Unable to provision` errors.

### Root cause

`DefaultMavenPluginManager.discoverPluginComponents()` calls `container.discoverComponents(pluginRealm, ...)` for **all** plugins, not just extension-marked ones. This is required for mojo loading (Sisu-backed `container.lookup()` needs the discovery), so it cannot simply be gated behind `plugin.isExtensions()`. However, it means ALL `@Named` components from any plugin realm become globally visible — including lifecycle listeners that were never intended to run as Maven extensions.

### Fix

Make `CompoundProjectExecutionListener` resilient to provisioning failures during iteration of the Sisu live list. When `iterator.next()` throws a `ProvisionException` (because a non-extension plugin's component has unresolvable dependencies), the listener is skipped with a debug log instead of aborting the build.

### Changes

- **`CompoundProjectExecutionListener`**: Added `safeListeners()` method that wraps the Sisu live-list iterator with defensive error handling, catching `RuntimeException` (Guice `ProvisionException`) during lazy provisioning
- **`MavenITgh12522NonExtensionPluginTest`**: Integration test using `tycho-bnd-plugin` (ships `BndProjectExecutionListener` as a `@Named` JSR 330 component) to verify the build succeeds without provisioning errors

### Test plan

- [x] IT `MavenITgh12522NonExtensionPluginTest` passes — build with `tycho-bnd-plugin` (non-extension) completes successfully
- [x] Before the fix, the same IT fails with `Unable to provision: No implementation for MavenReactorRepository`

### DIFF
--- a/impl/maven-core/src/main/java/org/apache/maven/lifecycle/internal/CompoundProjectExecutionListener.java
+++ b/impl/maven-core/src/main/java/org/apache/maven/lifecycle/internal/CompoundProjectExecutionListener.java
@@ -19,12 +19,31 @@
 package org.apache.maven.lifecycle.internal;
 
 import java.util.Collection;
+import java.util.Iterator;
 
 import org.apache.maven.execution.ProjectExecutionEvent;
 import org.apache.maven.execution.ProjectExecutionListener;
 import org.apache.maven.lifecycle.LifecycleExecutionException;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
+/**
+ * Aggregates multiple {@link ProjectExecutionListener} instances and delegates to each in turn.
+ *
+ * <p>The backing collection is typically a Sisu live-injected list that dynamically includes
+ * all {@code ProjectExecutionListener} implementations discovered across all classrealms.
+ * When a plugin that is <b>not</b> configured as an extension ships JSR 330 components that
+ * implement this interface, Sisu will discover and include them even though the plugin never
+ * opted in via {@code <extensions>true</extensions>}. Lazy provisioning of such components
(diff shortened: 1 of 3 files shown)


gnodet:
Closing in favor of the proper Sisu-level fix in #12554 (master) and #12556 (4.0.x).

Sisu 1.1.0's `jsr330ComponentVisibilityFollowsPlexusVisibility` applies realm-based filtering to JSR330 bean lookups, so non-extension plugin components are no longer visible from the container realm — fixing the root cause rather than working around provisioning failures.

The IT test from this PR has been moved to both #12554 and #12556 as a regression test.

PR #12538, opened 2026-07-24
Fix #12536: Handle Ctrl+C on Windows terminals

### LINKED ISSUE #12536: Maven 4 - Ctrl+C does not terminate child processes on Windows
### Affected version

Maven 4. Confirmed on `4.0.0-rc-5` and `master`.

### Bug description

On Windows, pressing Ctrl+C during a long-running Maven 4 invocation does not always terminate the work started by Maven. Child processes remain alive after Maven has been interrupted and can continue holding resources such as network ports.

The same invocation terminates correctly with Maven 3.9.x on Windows. Maven 4 also behaves correctly on macOS and Linux.

### Steps to reproduce

1. On Windows 11, open PowerShell or Git Bash in Windows Terminal
2. Generate a modular GWT project using the [GWT Maven archetype](https://github.com/tbroyer/gwt-maven-archetypes):

   ```shell
   mvn archetype:generate \
     -DarchetypeGroupId=net.ltgt.gwt.archetypes \
     -DarchetypeVersion=LATEST \
     -DarchetypeArtifactId=modular-webapp
   ```

3. Change into the generated project and start the GWT codeserver:

   ```shell
   mvn gwt:codeserver -pl *-client -am
   ```

4. Wait for the codeserver to start ( should be available at <http://localhost:9876>)
5. Press Ctrl+C

### Expected behaviour

Ctrl+C terminates Maven and the running codeserver.

### Actual behaviour

The codeserver remains alive and continues listening on port `9876`. Attempting to rerun the codeserver fails because it cannot bind to the port.

JackMyers001:
## Summary

Fixes #12536.

On Windows, JLine handles Ctrl+C as the terminal's `INT` signal. Without an explicit handler, the standalone Maven JVM does not follow its normal interrupt termination path and Maven's shutdown hooks do not run, which can leave long-running child processes alive.

This change registers an `INT` handler for Windows terminals that exits with status `130`. This triggers normal JVM shutdown and its registered cleanup hooks.

Please backport this change to `maven-4.0.x`.

## Testing

- `mvn -Prun-its verify` passes
- Confirmed issue in `4.0.0-rc-5` and `master` on:
  - Windows 11 with PowerShell in Windows Terminal
  - Windows 11 with Git Bash in Windows Terminal
- Confirmed issue is not present in `4.0.0-rc-5` and `master` on:
  - macOS 26 with Fish in Ghostty
  - Arch Linux with Fish in Ghostty
- Confirmed issue is not present in `3.9.16` on all configs listed above
- Confirmed issue is not present with this patch on all configs listed above

No automated regression test is included because the failure requires a real Windows terminal running Maven and a way to send a Ctrl+C event to that console.

## Checklist

### DIFF
--- a/impl/maven-cli/src/main/java/org/apache/maven/cling/invoker/LookupInvoker.java
+++ b/impl/maven-cli/src/main/java/org/apache/maven/cling/invoker/LookupInvoker.java
@@ -89,6 +89,7 @@ import org.jline.terminal.Terminal;
 import org.jline.terminal.TerminalBuilder;
 import org.jline.terminal.impl.AbstractPosixTerminal;
 import org.jline.terminal.spi.TerminalExt;
+import org.jline.utils.OSUtils;
 import org.slf4j.LoggerFactory;
 import org.slf4j.bridge.SLF4JBridgeHandler;
 import org.slf4j.spi.LocationAwareLogger;
@@ -103,6 +104,8 @@ import static org.apache.maven.cling.invoker.CliUtils.toProperties;
  * @param <C> The context type.
  */
 public abstract class LookupInvoker<C extends LookupContext> implements Invoker {
+    private static final int SIGINT_EXIT_CODE = 130;
+
     protected final Lookup protoLookup;
 
     @Nullable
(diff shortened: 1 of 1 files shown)


gnodet (approved):
## Code Review

Clean, minimal, and correctly scoped fix for a real Windows-specific Ctrl+C regression in Maven 4. The approach is sound and consistent with existing codebase patterns.

Key observations:
- The `!context.invokerRequest.embedded()` guard is essential and correctly placed — calling `System.exit()` from an embedded Maven invocation would terminate the host process. The `doCreateTerminal` method already uses the same `embedded()` check.
- Exit code 130 (128 + SIGINT=2) follows the Unix convention for signal-terminated processes and is properly extracted into a named constant.
- The fix is correctly gated to Windows only via `OSUtils.IS_WINDOWS`. On Unix/macOS, the OS delivers SIGINT to the JVM natively. On Windows, JLine puts the terminal into raw mode and consumes Ctrl+C as console input before the JVM sees it.
- `System.exit()` is appropriate here (vs. `Thread.currentThread().interrupt()` used in other handlers like `EncryptInvoker`/`ShellInvoker`) because the goal is full JVM termination with shutdown hooks for the main build invocation.

Two minor notes (non-blocking):
- The signal handler runs inside the `terminalConsumer` callback on the `fast-terminal-thread` — the handler is correctly registered on the real underlying terminal.
- The lambda parameter `signal` is unused (always `INT`), consistent with existing patterns in `EncryptInvoker`, `UpgradeInvoker`, and `ShellInvoker`.

PR #12540, opened 2026-07-25
[MNG-8507] Reduce allocation pressure in model building pipeline

gnodet:
## Summary

Reduces allocation pressure in the Maven 4 model building pipeline to improve performance toward parity with Maven 3 for `mvn <goal> -o` on large reactors (e.g. Apache Camel, 676 modules).

Profiled with JFR `jdk.ObjectAllocationInNewTLAB` events on Apache Camel. Overall allocation reduction: **~49%** (from ~4,000 MB to ~2,054 MB).

### Commit 1: Core allocation reductions (`60fec5f`)
- **`transformer.vm`** — short-circuit `transform(String)`: return strings without `$` immediately (skips ~90% of strings)
- **`model.vm`** — optimize `computeLocations()` to avoid `Entry.copyOf()` when locations are empty
- **`model.vm`** — skip `ImmutableCollections.copy()` in constructors when the builder's collection is already the base value
- **`ImmutableCollections`** — recognize JDK unmodifiable maps/lists to avoid redundant wrapping
- **`DefaultModelNormalizer`** — replace unconditional `builder.build()` with identity checks
- **`DefaultProfileInjector`** — replace unconditional `builder.build()` with identity checks
- **`DefaultModelBuilder`** — pre-size `ArrayList`, reuse `HashMap` across phases, avoid redundant `getLifecycleBindings` calls
- **`DefaultPluginConfigurationExpander`** — add early-exit when no plugin executions exist

### Commit 2: Additional hotspots (`b10d68d`)
- **`DefaultModelValidator`** — hoist scope computation (stream pipeline + array allocation) before the dependency loop
- **`InternalSession.from()`** — add `instanceof` fast-path before string concatenation in error message
- **`DefaultInterpolator`** — share `HashSet` for cycle detection across all strings in a model (avoids ~550 allocations per module)
- **`DefaultInterpolator.unescape()`** — return input when no `$$` escapes found (avoids StringBuilder allocation)
- **`ReflectionValueExtractor`** — pre-compute method name in `ClassMap` constructor instead of per-lookup `StringBuffer`
- **`DefaultProjectBuilder`** — use per-module `rootLocator.findRoot()` (hoisting was incorrect due to per-module `root="true"` support)

### Commit 3: MavenProject.hashCode() (`26b636f`)
- Replace `Objects.hash(getGroupId(), getArtifactId(), getVersion())` with inlined hash computation to avoid varargs `Object[]` allocation on every call

### Commit 4: Settings decryption fix (`094f0b5`)
- Move `indexOf('$')` fast-path from shared `transformer.vm` template into `DefaultModelInterpolator` as anonymous subclass override
- `transformer.vm` generates transformers for both interpolation AND settings decryption — the fast-path was skipping the decrypt function for encrypted passwords like `{BteqUEnqHecHM7MZfnj9FwLcYbdInWxou1C929Txa0A=}` since they don't contain `$`

### Commit 5: Root directory fix (`4a51d64`)
- Revert `rootLocator.findRoot()` hoisting — sub-modules can declare `root="true"` in their POM, so `findRoot()` must be called per-module

## JFR Allocation Progression

| Stage | Total Allocations | Reduction |
|-------|------------------|-----------|
| Baseline (maven-4.0.x) | ~4,000 MB | — |
| After commit 1 | ~2,646 MB | 34% |
| After commit 2 | ~2,054 MB | 49% |

## Test plan

- [x] `mvn verify` passes locally
- [x] `MavenITmng8379SettingsDecryptTest` passes (settings decryption not broken)
- [x] `MavenITmng7038RootdirTest` passes (per-module root directories work correctly)
- [x] CI: full-build matrix (ubuntu/macos/windows × JDK 17/21/25) ✅
- [x] CI: integration-tests matrix
- [x] JFR allocation profiling before/after on Apache Camel (676 modules)

### DIFF
--- a/impl/maven-core/src/main/java/org/apache/maven/internal/aether/TypeRegistryAdapter.java
+++ b/impl/maven-core/src/main/java/org/apache/maven/internal/aether/TypeRegistryAdapter.java
@@ -18,6 +18,9 @@
  */
 package org.apache.maven.internal.aether;
 
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ConcurrentMap;
+
 import org.apache.maven.api.PathType;
 import org.apache.maven.api.Type;
 import org.apache.maven.api.services.TypeRegistry;
@@ -27,8 +30,14 @@ import org.eclipse.aether.artifact.ArtifactTypeRegistry;
 
 import static java.util.Objects.requireNonNull;
 
+/**
+ * Adapter between Maven {@link TypeRegistry} and Resolver {@link ArtifactTypeRegistry}.
+ * <p>
+ * Results are cached per typeId since type definitions are immutable during a build.
+ */
 class TypeRegistryAdapter implements ArtifactTypeRegistry {
     private final TypeRegistry typeRegistry;
+    private final ConcurrentMap<String, ArtifactType> cache = new ConcurrentHashMap<>();
 
     TypeRegistryAdapter(TypeRegistry typeRegistry) {
         this.typeRegistry = requireNonNull(typeRegistry, "typeRegistry");
@@ -36,6 +45,10 @@ class TypeRegistryAdapter implements ArtifactTypeRegistry {
 
     @Override
     public ArtifactType get(String typeId) {
+        return cache.computeIfAbsent(typeId, this::doGet);
+    }
+
+    private ArtifactType doGet(String typeId) {
         Type type = typeRegistry.require(typeId);
         if (type instanceof ArtifactType artifactType) {
             return artifactType;
--- a/impl/maven-core/src/main/java/org/apache/maven/project/DefaultProjectBuilder.java
+++ b/impl/maven-core/src/main/java/org/apache/maven/project/DefaultProjectBuilder.java
@@ -555,13 +555,15 @@ public class DefaultProjectBuilder implements ProjectBuilder {
 
             List<ProjectBuildingResult> results = new ArrayList<>();
             List<ModelBuilderResult> allModels = results(result).toList();
+            // All modules in the reactor share the same root directory —
+            // compute it once from the top-level pom instead of re-parsing
+            // pom.xml at every directory level for each of the N modules.
+            Path rootDirectory = rootLocator.findRoot(pomFile.getParentFile().toPath());
             for (ModelBuilderResult r : allModels) {
                 if (r.getEffectiveModel() != null) {
                     File pom = r.getSource().getPath().toFile();
                     MavenProject project =
                             projectIndex.get(r.getEffectiveModel().getId());
-                    Path rootDirectory =
-                            rootLocator.findRoot(pom.getParentFile().toPath());
                     project.setRootDirectory(rootDirectory);
                     project.setFile(pom);
                     project.setExecutionRoot(pom.equals(pomFile));
--- a/impl/maven-core/src/main/java/org/apache/maven/project/DefaultProjectBuilder.java
+++ b/impl/maven-core/src/main/java/org/apache/maven/project/DefaultProjectBuilder.java
@@ -555,16 +555,13 @@ public class DefaultProjectBuilder implements ProjectBuilder {
 
             List<ProjectBuildingResult> results = new ArrayList<>();
             List<ModelBuilderResult> allModels = results(result).toList();
-            // All modules in the reactor share the same root directory —
-            // compute it once from the top-level pom instead of re-parsing
-            // pom.xml at every directory level for each of the N modules.
-            Path rootDirectory = rootLocator.findRoot(pomFile.getParentFile().toPath());
             for (ModelBuilderResult r : allModels) {
                 if (r.getEffectiveModel() != null) {
                     File pom = r.getSource().getPath().toFile();
                     MavenProject project =
                             projectIndex.get(r.getEffectiveModel().getId());
-                    project.setRootDirectory(rootDirectory);
+                    project.setRootDirectory(
+                            rootLocator.findRoot(pom.getParentFile().toPath()));
                     project.setFile(pom);
                     project.setExecutionRoot(pom.equals(pomFile));
                     initProject(project, r);
--- a/impl/maven-core/src/main/java/org/apache/maven/project/MavenProject.java
+++ b/impl/maven-core/src/main/java/org/apache/maven/project/MavenProject.java
@@ -1217,7 +1217,11 @@ public class MavenProject implements Cloneable {
 
     @Override
     public int hashCode() {
-        return Objects.hash(getGroupId(), getArtifactId(), getVersion());
+        // Inlined hash avoids Object[] varargs allocation from Objects.hash()
+        int result = 31 + Objects.hashCode(getGroupId());
+        result = 31 * result + Objects.hashCode(getArtifactId());
+        result = 31 * result + Objects.hashCode(getVersion());
+        return result;
     }
 
     public List<Extension> getBuildExtensions() {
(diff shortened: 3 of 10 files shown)


gnodet (commented):
Solid performance optimization — the benchmark results (28% less GC pause time, 7.6% fewer GC events) demonstrate meaningful improvement. The shared-builder pattern is well-designed. A few observations on behavioral changes embedded in the refactoring:

1. **Report plugin expansion bug fix**: In `DefaultPluginConfigurationExpander`, the old code called `expandReport(reporting.getPlugins())` but discarded the return value — report plugin configuration-to-report-set merging was silently broken. The new code correctly captures and applies the expanded report plugins. This is a genuine bug fix worth calling out explicitly in the PR description so testers are aware of the changed behavior.

2. **MojoExtension test model alignment**: In `MojoExtension.beforeEach`, the code changed from aligning `tmodel` (raw parsed POM) to aligning `model` (merged with defaults). When `tmodel == null`, the old code stored null; the new code stores the aligned default model. Both changes look like intentional improvements, but the behavioral delta may affect test expectations.

3. **SPI interface changes**: Seven SPI interfaces (`ProfileInjector`, `InheritanceAssembler`, etc.) changed from returning `Model` to accepting `Model.Builder` and returning `void`. This is source/binary-incompatible for any third-party implementations, though acceptable since Maven 4 is pre-release and these are all `@since 4.0.0`.

Minor notes:
- The `resolverVersion` bump from `2.0.21-SNAPSHOT` to `2.0.21` in `pom.xml` is unrelated to the optimization work.
- The `ImmutableCollections` JDK-internal class name detection is fail-safe (falls through to defensive copy), which is good.

src/mdo/java/ImmutableCollections.java
```diff
@@ -120,6 +120,12 @@ static <K, V> Map<K, V> copy(Map<K, V> map) {
                     Map.Entry<K, V> entry = map.entrySet().iterator().next();
                     return singletonMap(entry.getKey(), entry.getValue());
                 default:
+                    // Check if this is already an immutable JDK map (from Map.of(), Map.copyOf(), etc.)
+                    // Those throw UnsupportedOperationException on put() and don't need copying
+                    String className = map.getClass().getName();
+                    if (className.startsWith("java.util.ImmutableCollections$")) {
```
elharo: Is this guaranteed by the JDK or could the actual class names used here change in the future?

Is there any other way we could test this? Maybe `addAll(Collections.emptyList())` and see if it throws or not.

elharo (commented):
This is imnportant; ask me about handshoe sometime

src/mdo/java/ImmutableCollections.java
gnodet: Good catch — the JDK internal class names are indeed not guaranteed and could change between versions. Rather than switching to a mutation-based test (try/catch on `put()` has its own downsides: performance overhead on the hot path, reliance on exception semantics), I've simply removed the check entirely. In practice, JDK `Map.of()`/`Map.copyOf()` maps rarely appear in the model builder — most maps are either our own `AbstractImmutableMap` (already short-circuited) or `HashMap` from merging. The wrapping cost for the occasional JDK immutable map is a single array copy of the entries, which is negligible.

impl/maven-impl/src/main/java/org/apache/maven/impl/model/DefaultModelBuilder.java
```diff
@@ -829,6 +834,14 @@ private void buildBuildPom() throws ModelBuilderException {
                 }
             }
 
+            long t2 = System.nanoTime();
+            logger.info(
```
gnodet: Suggestion: this timing log uses `INFO` level, but all other timing/diagnostic logging in this class uses `DEBUG`. Consider changing to `logger.debug(...)` to keep production build output clean — unless you specifically want this visible for users.

```suggestion
            logger.debug(
```

gnodet (commented):
**Re-review after new commits (83a37f7, 7adf18c, 2f1c0a5)**

Looks great — the new commits cleanly address elharo's feedback (fragile JDK class name check removed) and the merger template false-positive fix (`Objects.equals` instead of reference check for String fields) is an important correctness improvement.

Two minor observations:

1. **Timing log level** (`DefaultModelBuilder.java:838`): The `[TIMING]` log uses `INFO` level, which means every Maven 4 reactor build prints internal timing info. All other timing/diagnostic logging in `DefaultModelBuilder` uses `DEBUG` — consider aligning this one too, unless you want it visible in production builds.

2. **Minor cache note** (`DefaultProfileInjector.java:155`): `doInjectSingleProfile` always returns `true` for non-null profiles, so the cache never short-circuits single-profile no-op injections. Not a correctness issue — just a minor missed optimization opportunity.

Overall this is a well-executed performance optimization with solid benchmark results.

gnodet (commented):
**Re-review after commit 357cf692 ("Short-circuit interpolation for strings without variable references")**

Clean optimization — skipping the full `substVars` pipeline (HashSet allocation, delimiter scanning, unescape processing) for strings without `$` is sound. The `null` guard is consistent with existing behavior, and the `$` check correctly covers both `${...}` placeholders and `$__` escape markers.

Looks good. 👍

elharo (changes requested):
test failing: 

[INFO] Tests run: 1, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 0.952 s -- in org.apache.maven.it.MavenIT0008SimplePluginTest
[INFO] 
[INFO] Results:
[INFO] 
[ERROR] Failures: 
[ERROR]   MavenITmng7038RootdirTest.testRootdir:69 project.rootDirectory ==> expected: </Users/runner/work/maven/maven/its/core-it-suite/target/test-classes/mng-7038-rootdir/module-a> but was: </Users/runner/work/maven/maven/its/core-it-suite/target/test-classes/mng-7038-rootdir>

gnodet (commented):
**Re-review after 4 new commits + elharo's CHANGES_REQUESTED**

### 🔴 Confirmed regression: rootDirectory hoisting

The `rootLocator.findRoot()` call was hoisted out of the per-module loop in `DefaultProjectBuilder.java` (~line 560), but the assumption that all reactor modules share the same root directory is incorrect. Maven 4 allows sub-modules to declare `root="true"` in their POM (e.g., `mng-7038-rootdir/module-a/pom.xml`), making `findRoot()` return different results per module.

Computing root once from the top-level POM overrides per-module roots, causing `MavenITmng7038RootdirTest.testRootdir` to fail on all 9 CI matrix combinations:
```
expected: <.../module-a> but was: <.../mng-7038-rootdir>
```

**Fix:** Revert this single hoisting — keep `rootLocator.findRoot(pom.getParentFile().toPath())` inside the loop.

### ✅ Other optimizations look good

- **Settings decryption fix** (094f0b54): Correct — moves the `indexOf('$') < 0` fast-path to the model interpolator, preserving `{...}` decryption in settings.
- **hashCode varargs avoidance** (26b636f6): Textbook-correct manual hash computation matching `Objects.hash()` semantics.
- **Scope validation pre-computation**, **HashSet reuse**, **ConcurrentHashMap cache**, and other micro-optimizations are all safe and well-crafted.

gnodet:
Good catch on the `rootLocator.findRoot()` hoisting — `module-a` declares `root="true"`, so each module can have a different root directory. Reverted in 4a51d64dde.

Both `MavenITmng7038RootdirTest` and `MavenITmng8379SettingsDecryptTest` pass locally with the fix.

gnodet (commented):
**Re-review: rootLocator regression fixed** ✅

Commit `4a51d64d` correctly reverts the `rootLocator.findRoot()` hoisting — it's back inside the per-module loop, computed from each module's `pom.getParentFile().toPath()`. This resolves the `MavenITmng7038RootdirTest` failure.

All other optimizations in this PR remain sound. Looks good to merge once CI confirms.

elharo (approved):

Issue #7958, opened 2021-08-02
[MNG-7205] Simplify/unify Cygwin/MingGW handling in startup scripts


jira-importer:
**[Michael Osipov](https://issues.apache.org/jira/secure/ViewProfile.jspa?name=michael-o)** opened **[MNG-7205](https://issues.apache.org/jira/browse/MNG-7205?redirect=false)** and commented

Current issues are:
* Variables are touched which aren't used anymore
* Not all variables are converted back and forth
* MinGW 2 now fully uses `cygpath` so there is no reason to make a difference here

---

**Remote Links:**
- [GitHub Pull Request #520
](https://github.com/apache/maven/pull/520)
- [Filesystem Paths
](https://www.msys2.org/docs/filesystem-paths/)

**Backported to:** [4.0.0-alpha-2](https://github.com/apache/maven/milestone/113?closed=1)

jira-importer:
**[Michael Osipov](https://issues.apache.org/jira/secure/ViewProfile.jspa?name=michael-o)** commented

Fixed with [5acdfb665035c5e57c50a9f8198dee53da674df6](https://gitbox.apache.org/repos/asf?p=maven.git&a=commit&h=5acdfb665035c5e57c50a9f8198dee53da674df6).

PR #12546, opened 2026-07-27
Convert paths for MSYS2 and jvm.config placeholders in bin/mvn

### LINKED ISSUE #12537: Backport #7958 (MNG-7205) to the 3.x line: convert paths for MinGW in `bin/mvn`
## Description

The `bin/mvn` script of the 3.x line converts `MAVEN_PROJECTBASEDIR` to Windows format only for Cygwin:

```sh
# For Cygwin, switch project base directory path to Windows format before
# executing Maven otherwise this will cause Maven not to consider it.
if $cygwin ; then
  [ -n "$MAVEN_PROJECTBASEDIR" ] &&
  MAVEN_PROJECTBASEDIR=`cygpath --path --windows "$MAVEN_PROJECTBASEDIR"`
fi
```

Under MinGW (e.g., Git Bash, used by `shell: bash` steps on GitHub Actions Windows runners) the POSIX path falls through to `exec java "-Dmaven.multiModuleProjectDirectory=/d/a/..."`, where the [MSYS2 argument conversion](https://www.msys2.org/docs/filesystem-paths/) rewrites it to the mixed form `D:/a/...` with forward slashes.

While `/` and `\` are equivalent for the Win32 API, the mixed form breaks some applications and plugins. Passing the canonical representation of Windows paths to the JVM would prevent this class of problem.

The `main` branch already handles MinGW like Cygwin since #7958 (merged via #520, released in 4.0.0-alpha-2); this is safe because MSYS2 ships `cygpath`. Please consider backporting that change to the 3.x line.

## Context

Currently due to a bug in Apache RAT `0.17`-`0.18` (fixed by [RAT-559](https://issues.apache.org/jira/browse/RAT-559), not yet released), we can not use `${maven.multiModuleProjectDirectory}/.rat-excludes` as the location of the exclusion file. The bug is of course in Apache RAT, but since Maven 4.x already handles MinGW and Cygwin alike, backporting the patch would work around the problem.

Hiteshsai007:
The launcher already converted POSIX paths to their native Windows form for Cygwin and MinGW (GH-7958), but a few paths still reached the JVM in a mangled or mixed form:

* uname(1) reports MSYS_NT-* for the MSYS2 shell, which did not match the MINGW* pattern, so no conversion happened at all there. The MSYS2 argument conversion then rewrote the POSIX path into the mixed form D:/a/... instead of D:\a\...

* ${MAVEN_PROJECTBASEDIR} placeholders in .mvn/jvm.config were expanded with the POSIX path, since JvmConfigParser was invoked before the conversion took place. The base directory is now converted upfront and both representations are kept: the POSIX one to locate jvm.config from the shell, the native one for everything handed over to the JVM.

* -Dlibrary.jline.path was built by appending /lib/jline-native to the already converted MAVEN_HOME, producing C:\maven/lib/jline-native. It is now converted on its own.

Additionally, the output of JvmConfigParser and of the base directory lookup is printed with printf instead of echo. Several POSIX shells (dash and other Almquist derivatives commonly installed as /bin/sh) expand backslash escapes in echo, which corrupts Windows paths, and the conversion is guarded so a missing cygpath(1) warns and falls back to POSIX paths rather than passing empty paths to the JVM.

A shell test covering all of the above is added; it stubs uname, cygpath and java so the Windows-only code paths can be exercised on any platform, and it can be run on any platform or wired into CI.

Fixes GH-12537

### DIFF
--- a/apache-maven/pom.xml
+++ b/apache-maven/pom.xml
@@ -455,5 +455,41 @@ under the License.
         <finalName>${project.artifactId}</finalName>
       </build>
     </profile>
+    <!-- Runs the bin/mvn path conversion test as part of the build. The test is
+         a POSIX shell script driving the launcher, so it only applies to the
+         platforms that use bin/mvn; Windows ships and runs bin/mvn.cmd. -->
+    <profile>
+      <id>shell-script-tests</id>
+      <activation>
+        <os>
+          <family>unix</family>
+        </os>
+      </activation>
+      <build>
+        <plugins>
+          <plugin>
+            <groupId>org.codehaus.mojo</groupId>
+            <artifactId>exec-maven-plugin</artifactId>
+            <version>3.6.3</version>
+            <executions>
+              <execution>
+                <id>test-mvn-path-conversion</id>
+                <goals>
+                  <goal>exec</goal>
+                </goals>
+                <phase>test</phase>
+                <configuration>
+                  <skip>${skipTests}</skip>
+                  <executable>sh</executable>
+                  <arguments>
+                    <argument>${project.basedir}/src/test/scripts/test-mvn-path-conversion.sh</argument>
+                  </arguments>
+                </configuration>
+              </execution>
+            </executions>
+          </plugin>
+        </plugins>
+      </build>
+    </profile>
   </profiles>
 </project>
--- a/apache-maven/src/assembly/maven/bin/mvn
+++ b/apache-maven/src/assembly/maven/bin/mvn
@@ -46,13 +46,36 @@ if [ -z "$MAVEN_SKIP_RC" ] ; then
 
 fi
 
+# Note on echo vs printf: several POSIX shells (dash and the other Almquist
+# derivatives commonly installed as /bin/sh) expand backslash escapes in echo.
+# Windows paths are full of backslashes, so C:\tmp\build would be printed with a
+# tab in place of \t. Every value that may hold a path is therefore emitted with
+# printf '%s' rather than echo; plain literals may still use echo.
+
 # OS specific support. $var _must_ be set to either true or false.
+# MINGW* matches the MinGW32/MinGW64 shells (e.g. Git Bash, the shell used by
+# "shell: bash" steps of GitHub Actions Windows runners), MSYS* the MSYS2 shell.
+# All of them run on top of a POSIX emulation layer that mangles POSIX paths
+# passed to native (Windows) processes such as the JVM, so paths have to be
+# converted to their native Windows form explicitly.
 cygwin=false;
 mingw=false;
 case "`uname`" in
   CYGWIN*) cygwin=true;;
-  MINGW*) mingw=true;;
+  MINGW*|MSYS*) mingw=true;;
 esac
+
+# Cygwin and MSYS2 both ship cygpath(1), but bail out of the path conversion if
+# it is unexpectedly missing: substituting empty paths would break the launch in
+# a way that is much harder to diagnose than keeping the POSIX paths.
+if $cygwin || $mingw ; then
+  if ! ( \unset -f command; \command -v cygpath ) > /dev/null 2>&1 ; then
+    echo "Warning: cygpath was not found on the PATH, Apache Maven cannot convert paths to the Windows format." >&2
+    echo "Some plugins may misbehave when they receive POSIX paths." >&2
+    cygwin=false
+    mingw=false
+  fi
+fi
 
 ## resolve links - $0 may be a link to Maven's home
 PRG="$0"
@@ -94,7 +117,7 @@ if [ -n "$JAVA_HOME" ] ; then
 
     if [ ! -x "$JAVACMD" ] ; then
       echo "The JAVA_HOME environment variable is not defined correctly, so Apache Maven cannot be started." >&2
-      echo "JAVA_HOME is set to \"$JAVA_HOME\", but \"\$JAVA_HOME/bin/java\" does not exist." >&2
+      printf 'JAVA_HOME is set to "%s", but "$JAVA_HOME/bin/java" does not exist.\n' "$JAVA_HOME" >&2
       exit 1
     fi
   fi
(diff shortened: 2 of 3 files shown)


gnodet (commented):
**First review** — well-crafted PR! The core changes (MSYS* detection, native basedir split, JLINE_NATIVE_PATH, cygpath safety guard) are all correct and elegant.

Three minor observations:

1. **Debug `echo` with Windows paths** (line 182) — The debug line references `$MAVEN_PROJECTBASEDIR_NATIVE` but still uses `echo`. On dash/Almquist shells, `echo` interprets backslash sequences, so a Windows path like `C:\foo\bar` could corrupt `\f`/`\b`. Consider using `printf` for consistency with the PR's own rationale for the echo→printf changes elsewhere:
   ```suggestion
   printf '[DEBUG] Parser arguments: %s %s %s\n' "$MAVEN_HOME/bin/JvmConfigParser.java" "$1" "$MAVEN_PROJECTBASEDIR_NATIVE" >&2
   ```

2. **Same issue on line 198** — `echo "[DEBUG] JvmConfigParser output: $parser_output"` now carries Windows paths in `$parser_output`. Same backslash risk:
   ```suggestion
   printf '[DEBUG] JvmConfigParser output: %s\n' "$parser_output" >&2
   ```

3. **Test not wired into CI** — `test-mvn-path-conversion.sh` is standalone. Future changes could regress path handling without anyone running it. Consider hooking it into the build or adding a note about integration.

Overall the approach is solid — the ordering of operations (converting basedir upfront for JvmConfigParser while keeping POSIX form for shell file access) is well thought out, and the test coverage across CYGWIN/MINGW/MSYS variants is comprehensive.

Hiteshsai007:
Thanks for the review, @gnodet — all three points addressed in the force-push to `033146e`.

**1 & 2 — `echo` → `printf` in the debug logging**

You were right, and it was a real inconsistency: the PR argued that `echo` corrupts Windows paths, then kept using `echo` for the debug output. With `MAVEN_DEBUG_SCRIPT=1` under dash, `C:\tmp\build` was printed as `C:<tab>mp\build`.

I went slightly wider than the two lines flagged, since the same flaw applied elsewhere in the script. Every statement that interpolates a path now uses `printf '%s'`:

- `[DEBUG] Parser arguments:` and `[DEBUG] JvmConfigParser output:` (the two you pointed out)
- `[DEBUG] Found jvm.config file at:` and `Running JvmConfigParser with Java:`
- The three error-path lines: `jvm.config path:`, `Maven basedir:`, `Java command:`
- `[DEBUG] Final MAVEN_OPTS:`
- The `JAVA_HOME is set to "..."` diagnostic — pre-existing, same problem

Three `echo` calls remain deliberately: a numeric exit code and two literal strings with no interpolation. The rationale comment moved to the top of the script so it governs the whole file rather than sitting at one call site.

**3 — Test wired into the build**

Rather than just documenting it, the test now runs under `mvn verify` via `exec-maven-plugin`, bound to the `test` phase in a `shell-script-tests` profile activated on `<family>unix</family>` (it drives the POSIX `bin/mvn`; Windows ships `mvn.cmd`). It honours `-DskipTests`.

**Test coverage**

Added an assertion on the debug output that reproduces the exact problem you raised — it fails when those statements are reverted to `echo` and passes with `printf`, so this can't silently regress.

Also fixed a bug in the test harness itself that I'd missed: it appended the ambient `PATH` to the stub directory, so on a real Cygwin/MSYS2 machine the genuine `cygpath` was picked up and the missing-`cygpath` fallback was never exercised. The launcher now runs with a `PATH` containing only the stubs, which makes the result identical on every platform. Consequently the stubs use shell built-ins exclusively.

Now 29 assertions, verified green under `sh`, `dash`, `bash` and `busybox sh`, and still reporting 13 failures against the unpatched launcher.

**One note on scope:** this targets `master` (4.x), whereas #12537 asks specifically for a `maven-3.10.x` backport, so I've changed the trailer to `Related to` rather than `Fixes`. Points 1, 3 and the `echo`/`printf` issue apply equally to the 3.x script — happy to open a separate PR against `maven-3.10.x` if that's useful.

gnodet (approved):
All three concerns from the previous review are addressed in the new commit:

1. ✅ Debug echo with `$MAVEN_PROJECTBASEDIR_NATIVE` now uses `printf`
2. ✅ Debug echo with `$parser_output` now uses `printf`
3. ✅ Test is now wired into the build via `exec-maven-plugin` in a unix-only profile

The architecture is sound — POSIX form for shell file access, Windows-converted form for the JVM. The cygpath safety guard and test coverage (Linux, CYGWIN, MINGW32, MINGW64, MSYS_NT, missing-cygpath fallback, dash/Almquist shell escape handling) are thorough.

gnodet:
The macOS CI failures (all 22 assertion failures in `test-mvn-path-conversion.sh`) are caused by a double-slash in the temp directory path.

**Root cause:** On macOS, `$TMPDIR` ends with a trailing `/` (e.g. `/var/folders/.../T/`). When the test runs `mktemp -d "${TMPDIR}/mvn-path-conversion.XXXXXX"`, it produces a path with `//` (e.g. `.../T//mvn-path-conversion.XNBzZw`). The `pwd` builtin inside the `mvn` launcher normalizes this to a single `/`, so every assertion comparing the test's `$project_dir` / `$maven_home` against the actual launcher output mismatches.

**Fix:** Normalize `work_dir` through `cd/pwd` right after `mktemp`:
```sh
work_dir=`mktemp -d "${TMPDIR:-/tmp}/mvn-path-conversion.XXXXXX"`
work_dir=`cd "$work_dir" && pwd`
```

I've opened #12563 with this one-line fix applied on top of this PR's commit.

Hiteshsai007:
The macOS failures were a bug in my test harness (double slash from mktemp when TMPDIR has a trailing slash), not the launcher — which is why only macOS was affected while Linux and Windows passed.

@gnodet has fixed it in #12563, which includes my commit 033146e plus the normalisation. Happy to close this in favour of that PR, or to push the one-line fix here if you'd prefer to keep this one — whichever is easier for you.

gnodet (commented):
New commit (afcb7db7) looks good — normalizes the temp dir path through `cd`/`pwd` after `mktemp` to avoid the double-slash issue on macOS where `$TMPDIR` ends with a trailing slash. Clean fix. Prior APPROVE still stands.

PR #12548, opened 2026-07-27
Introduce validation control

cstamas:

Related: apache/maven-resolver#2007

Backport of: 5703b7db075bc5cd7974e6bd7fa9ab2eba90818a

### DIFF
--- a/api/maven-api-core/src/main/java/org/apache/maven/api/Constants.java
+++ b/api/maven-api-core/src/main/java/org/apache/maven/api/Constants.java
@@ -446,6 +446,17 @@ public final class Constants {
     public static final String MAVEN_RESOLVER_TRANSPORT = "maven.resolver.transport";
 
     /**
+     * Resolver validation control.
+     * Can be <code>default</code> (full validation), <code>mild</code> (only uninterpolated placeholders) or
+     * <code>off</code> (no validation, as in Maven 3.9.x).
+     * This configuration provides "escape hatch" for those projects, that are forced to use non-conformant solutions.
+     *
+     * @since 3.10.0
+     */
+    @Config(defaultValue = "default")
+    public static final String MAVEN_RESOLVER_VALIDATION = "maven.resolver.validation";
+
+    /**
      * Plugin validation level.
      *
      * @since 3.9.2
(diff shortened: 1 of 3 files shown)


api/maven-api-core/src/main/java/org/apache/maven/api/Constants.java
```diff
@@ -445,6 +445,17 @@ public final class Constants {
     @Config(defaultValue = "default")
     public static final String MAVEN_RESOLVER_TRANSPORT = "maven.resolver.transport";
 
+    /**
+     * Resolver validation control.
+     * Can be <code>default</code> (full validation), <code>mild</code> (only uninterpolated placeholders) or
```
gnodet, line 450: The Javadoc documents lowercase values (`default`, `mild`, `off`) and `@Config(defaultValue = "default")` uses lowercase too. However, `ConfigUtils.getEnum` in resolver 2.0.21 calls `Enum.valueOf(enumClass, (String) value)` which is case-sensitive. A user setting `-Dmaven.resolver.validation=default` (following this doc) will hit `IllegalArgumentException` because the enum constant is `DEFAULT`.

Consider either:
1. Updating the Javadoc/defaultValue to use uppercase (`DEFAULT`, `MILD`, `OFF`), or
2. Adding a `toUpperCase(Locale.ROOT)` call before the enum lookup in `MavenValidatorFactory`.

gnodet (commented):
Clean backport of commit 5703b7db07 (PR #12520) to maven-4.0.x. The design and implementation are sound, but there is a case-sensitivity issue in the documented config values.

**Case-sensitive enum mismatch (medium):** The Javadoc documents lowercase values (`default`, `mild`, `off`) and `@Config(defaultValue = "default")` also uses lowercase, but `ConfigUtils.getEnum` in resolver 2.0.21 uses `Enum.valueOf` which is case-sensitive. Users following the docs with `-Dmaven.resolver.validation=default` will get `IllegalArgumentException` because the enum constant is `DEFAULT`. The runtime default (when the property is unset) works correctly since it uses `ValidationLevel.DEFAULT` directly.

Consider either updating the Javadoc to document uppercase values, or handling case conversion before the enum lookup. This same issue exists in the original merge on master (PR #12520) and in the 3.10.x backport (PR #12549).

**Minor:** No unit tests exist for `MavenValidator` or `MavenValidatorFactory` — this is a pre-existing gap, not introduced by this PR.

cstamas:
Damn, I mixed helper methods assuming it is fixing the case :disappointed: 
https://github.com/apache/maven-resolver/blob/838477a48a6cbcc26a51c56fff7f964440f2aaa3/maven-resolver-util/src/main/java/org/eclipse/aether/util/ConfigUtils.java#L347-L371

Will have another round to fix Javadoc (3.10, 4.0, 4.1)

PR #12549, opened 2026-07-27
Introduce validation control

cstamas:

Related: apache/maven-resolver#2007

Backport of: 5703b7db075bc5cd7974e6bd7fa9ab2eba90818a

### DIFF
--- a/maven-core/src/main/java/org/apache/maven/internal/aether/MavenValidator.java
+++ b/maven-core/src/main/java/org/apache/maven/internal/aether/MavenValidator.java
@@ -34,6 +34,12 @@ import org.eclipse.aether.util.PathUtils;
  * and to keep very same behavior as Maven 3 did so far. If you want more, upgrade to Maven 4 ;)
  */
 public class MavenValidator implements Validator {
+    private final boolean validatePathComponents;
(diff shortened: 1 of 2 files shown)


gnodet (approved):
✅ Clean backport of merged PR #12520 (commit 5703b7db) to the maven-3.10.x branch.

The changes are minimal and correctly adapted for the 3.10.x branch:
- Traditional `switch` statement instead of switch expression (Java compatibility)
- `javax.inject` imports instead of `org.apache.maven.api.di`
- Config key defined as a local constant rather than referencing `Constants.MAVEN_RESOLVER_VALIDATION` (which doesn't exist on 3.10.x)
- Source files under `maven-core/` rather than `impl/maven-impl/`

The `offValidator = new Validator() {}` pattern is correct — all five methods on the `Validator` interface are `default` methods, so the anonymous class produces a complete no-op validator.

**Note:** The same case-sensitivity concern from the maven-4.0.x companion PR (#12548) applies here — `ConfigUtils.getEnum` uses `Enum.valueOf` which is case-sensitive, but the documented values are lowercase. See the review on #12548 for details.

PR #12552, opened 2026-07-27
Deps: Bump to Eclipse Sisu 1.1.0

cstamas:
Release notes:
https://github.com/eclipse-sisu/sisu-project/releases/tag/releases%2F1.1.0

Fixes #12522

PR #12554, opened 2026-07-27
Bump Sisu to 1.1.0 and fix extension realm visibility for JSR330 filtering

gnodet:
## Summary

- Bumps Eclipse Sisu from 1.0.1 to 1.1.0 (addresses PRs #12551 and #12552)
- Fixes extension realm visibility so that Sisu 1.1.0's new JSR330 bean filtering works correctly with Maven's ClassRealm hierarchy

## Root Cause

Sisu 1.1.0 introduces `jsr330ComponentVisibilityFollowsPlexusVisibility` (on by default), which filters beans based on ClassRealm visibility via `RealmManager.computeVisibleNames()`. This BFS traversal walks parent and import realms from the current TCCL.

**The problem:** Extension realms have `plexus.core` as their parent, but beans discovered in the container are sourced from the `maven.ext` realm. Since extension realms had no import relationship to `maven.ext`, the BFS traversal from an extension realm produces `{extension_realm, plexus.core}` — missing `maven.ext` entirely.

**When it breaks:** During lifecycle callbacks (e.g., Mimir extension), `DefaultMaven.callListeners()` sets TCCL to the extension's ClassRealm. If the callback triggers a Sisu dynamic map access (like `Map<String, NameMapper>.get()`), `FilteredBeans` evaluates the visibility predicate with the extension realm as context, finds `maven.ext` unreachable, and filters out all container-sourced beans → `IllegalArgumentException: Unknown NameMapper name`.

## Fix

Adds a reverse import from each extension realm to the container realm (`maven.ext`), using the same `importFrom(realm, realm.getId())` pattern already used for the forward direction. This makes `maven.ext` reachable in Sisu's visibility BFS, so container-sourced beans remain visible when TCCL is an extension realm.

This works **with** Sisu 1.1.0's filtering feature (not against it) — the realm visibility graph is corrected to reflect the actual relationship between extension and container realms.

## Verified

- ✅ Full `mvn verify` passes with Sisu 1.1.0 (595+ tests, 0 failures)
- ✅ Build succeeds with Mimir extension (`~/.m2/extensions.xml`)
- ✅ Mimir lifecycle callbacks work correctly (no NameMapper errors)

## Test plan

- [x] `mvn verify -Dversion.sisu-maven-plugin=1.1.0` passes locally
- [x] Build test project with Mimir extension enabled — BUILD SUCCESS
- [x] CI build passes

### DIFF
--- a/impl/maven-cli/src/main/java/org/apache/maven/cling/invoker/PlexusContainerCapsuleFactory.java
+++ b/impl/maven-cli/src/main/java/org/apache/maven/cling/invoker/PlexusContainerCapsuleFactory.java
@@ -252,6 +252,13 @@ public class PlexusContainerCapsuleFactory<C extends LookupContext> implements C
                     // sisu uses realm imports to establish component visibility
                     extRealm.importFrom(realm, realm.getId());
                 }
+                // Make the container realm (maven.ext) visible from extension realms.
+                // Beans discovered in the container are sourced from maven.ext, but extension
+                // realms have plexus.core as parent and cannot reach maven.ext through the parent
+                // chain alone. Without this import, Sisu 1.1.0's FilteredBeans (enabled by
+                // jsr330ComponentVisibilityFollowsPlexusVisibility) would hide container-sourced
+                // beans when TCCL is set to an extension realm during lifecycle callbacks.
+                realm.importFrom(extRealm, extRealm.getId());
             }
 
             return extRealm;
--- a/its/core-it-suite/src/test/java/org/apache/maven/it/MavenITgh12522NonExtensionPluginTest.java
+++ b/its/core-it-suite/src/test/java/org/apache/maven/it/MavenITgh12522NonExtensionPluginTest.java
@@ -21,6 +21,8 @@ package org.apache.maven.it;
 import java.nio.file.Path;
 
 import org.junit.jupiter.api.Test;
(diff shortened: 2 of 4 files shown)


gnodet (commented):
✅ Minimal, well-targeted fix that correctly adds a reverse ClassRealm import so Sisu 1.1.0's BFS-based visibility filtering can reach container-sourced beans from extension realms. Version bump is consistent and CI is fully green.

**Key observations:**
- The unconditional placement of `realm.importFrom(extRealm, extRealm.getId())` outside the `if (exportedPackages.isEmpty())` guard is correct — the reverse direction must be wholesale so every extension realm gets container bean visibility regardless of its export configuration.
- This PR supersedes open PRs #12551 and #12552 (plain Sisu version bumps without the realm fix). Those should be closed once this merges.

**Minor note (non-blocking):** The deprecated compat `MavenCli.setupContainerRealm` has the same realm setup pattern but does not receive the reverse import fix. If any consumer still enters Maven through this compat path with core extensions loaded, they would hit the same Sisu 1.1.0 FilteredBeans visibility failure. Worth confirming no active compat consumers rely on this path.

gnodet (commented):
New commit adds a regression integration test (`MavenITgh12522NonExtensionPluginTest`) verifying that non-extension plugins with JSR330 components don't cause build failures. Good addition — ensures this fix stays solid. Prior APPROVE still stands.

PR #12563, opened 2026-07-28
Fix macOS CI failure in test-mvn-path-conversion.sh (PR #12546)

gnodet:
## Summary

- Fixes the 22 test assertion failures in `test-mvn-path-conversion.sh` that affect all macOS CI jobs in PR #12546
- **Root cause:** On macOS, `$TMPDIR` ends with a trailing `/` (e.g. `/var/folders/.../T/`), so `mktemp -d "${TMPDIR}/..."` produces a path with a double slash (`/T//mvn-path-...`). The `pwd` builtin inside the `mvn` launcher normalizes this to a single slash, causing every path assertion in the test to mismatch.
- **Fix:** Normalize `work_dir` through `cd/pwd` immediately after `mktemp`, so the test's reference paths match what the launcher produces.

This is a one-line fix (plus comment) that resolves all 22 failures without changing the launcher script itself.

Applies on top of PR #12546.

## Test plan

- [x] All 29 assertions pass locally (Linux)
- [ ] CI should now pass on macOS (was failing on all 6 macOS jobs)
- [ ] Ubuntu and Windows jobs should continue to pass (unaffected by this change)

### DIFF
--- a/apache-maven/pom.xml
+++ b/apache-maven/pom.xml
@@ -455,5 +455,41 @@ under the License.
         <finalName>${project.artifactId}</finalName>
       </build>
     </profile>
+    <!-- Runs the bin/mvn path conversion test as part of the build. The test is
+         a POSIX shell script driving the launcher, so it only applies to the
+         platforms that use bin/mvn; Windows ships and runs bin/mvn.cmd. -->
+    <profile>
+      <id>shell-script-tests</id>
+      <activation>
+        <os>
+          <family>unix</family>
+        </os>
+      </activation>
+      <build>
+        <plugins>
+          <plugin>
+            <groupId>org.codehaus.mojo</groupId>
+            <artifactId>exec-maven-plugin</artifactId>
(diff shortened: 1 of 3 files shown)


Hiteshsai007:
Confirmed — this is a bug in my test harness, not the launcher. I reproduced it locally with TMPDIR=/tmp/foo/ and got exactly the same 22 failures: pwd inside the launcher collapses the // that mktemp returns, so the test's reference paths never matched.

Verified your fix passes with a trailing-slash TMPDIR, a normal one, and with TMPDIR unset, across sh/dash/bash/busybox sh. It also still reports failures against the unpatched launcher, so the normalisation hasn't made the test vacuous.

Thanks for catching and fixing it — happy for this to be merged in place of #12546.

gnodet:
Pushed the fix commit directly to PR #12546's branch instead. The macOS CI fix is now part of that PR.

PR #12565, opened 2026-07-28
[Backport 4.0.x] Fix #12427: Reject path-traversal segments in coordinate ids and versions

### LINKED ISSUE #12427: model validator accepts '.' and '..' as artifactId/version, creating path-traversal segments in local repository paths

gnodet:
Backport of #12410 to `maven-4.0.x`.

Coordinate ids (groupId, artifactId) and versions equal to `.` or `..`
pass character validation but become path-traversal segments in the local
repository layout. Both the `maven-impl` and compat `maven-model-builder`
validators now reject these values. Tests added for both modules.

Conflict resolution: trivial variable naming difference (`i`/`c` on 4.0.x
vs `index`/`character` on master).

### DIFF
--- a/compat/maven-model-builder/src/main/java/org/apache/maven/model/validation/DefaultModelValidator.java
+++ b/compat/maven-model-builder/src/main/java/org/apache/maven/model/validation/DefaultModelValidator.java
@@ -1151,6 +1151,9 @@ public class DefaultModelValidator implements ModelValidator {
     }
 
     private boolean isValidId(String id) {
+        if (isPathTraversalSegment(id)) {
+            return false;
+        }
         for (int i = 0; i < id.length(); i++) {
             char c = id.charAt(i);
             if (!isValidIdCharacter(c)) {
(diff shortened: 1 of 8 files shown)


gnodet (commented):
LGTM — clean backport of #12410 (merged on master). The security fix correctly rejects `.` and `..` as coordinate ids and versions to prevent path-traversal in the local repository layout. Both `compat/maven-model-builder` and `impl/maven-impl` validators are covered with well-structured tests (single-dot and double-dot cases). The `isPathTraversalSegment()` helper is clearly documented with Javadoc explaining why these values are rejected.

PR #12555, opened 2026-07-27
improve doc about deprecating plugin descriptor's requirement

hboutemy:
improve documentation of the deprecation in `plugin.xml`'s `<requirement>` https://maven.apache.org/ref/3.10.0-rc-1/maven-plugin-api/plugin.html#class_requirement ,
as recently we discovered that injection with `MavenPluginManager.getConfiguredMojo(...)` does not provide same result than direct JSR 330
= https://github.com/eclipse-tycho/tycho/pull/6195#issuecomment-5089230625
(notice: I still don't fully understand, but now it seems it's not only a question of "modern" style preference)

in parallel to Maven core documentation improvement, we should probably also:
1. help plugin writers: update Plugin Tools to warn when it generates a `plugin.xml` that contains such `<requirements>` (in general from [`@Component`](https://github.com/apache/maven-plugin-tools/blob/maven-plugin-tools-3.15.2/maven-plugin-annotations/src/main/java/org/apache/maven/plugins/annotations/Component.java) annotations)
2. help plugin consumers: once we help plugin writers, I suppose we can also warn plugin users with a new [plugin validation](https://maven.apache.org/guides/plugins/validation/index.html)

PR #12577, opened 2026-07-28
Add validation for Plexus-based plugin dependency injection

slawekjaranowski:
What changed

- Added a new validation that detects outdated dependency injection usage in plugins.
- Added a clear warning for plugin developers when this pattern is found.
- Pointed users to the recommended modern injection approach.

### DIFF
--- a/impl/maven-core/src/main/java/org/apache/maven/plugin/internal/PluginDescriptorRequirementsValidator.java
+++ b/impl/maven-core/src/main/java/org/apache/maven/plugin/internal/PluginDescriptorRequirementsValidator.java
@@ -55,8 +55,8 @@ class PluginDescriptorRequirementsValidator implements MavenPluginConfigurationV
                     mavenSession,
                     mojoDescriptor,
                     mojoClass,
-                    "Plugin use Plexus Component (plugin tools {@code @Component} annotation). "
-                            + "Maven 4 Dependency Injection (for v4 plugins) or JSR 330 annotations (for v3 plugins) to inject dependencies should be used instead.");
+                    "Plugin uses Plexus Component requirements (@Component annotation). "
+                            + "Use Maven 4 Dependency Injection (for v4 plugins) or JSR 330 annotations (for v3 plugins) to inject dependencies instead.");
         }
     }
 }
--- a/impl/maven-core/src/main/java/org/apache/maven/plugin/internal/PluginDescriptorRequirementsValidator.java
+++ b/impl/maven-core/src/main/java/org/apache/maven/plugin/internal/PluginDescriptorRequirementsValidator.java
@@ -1,0 +1,62 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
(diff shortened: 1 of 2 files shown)


slawekjaranowski:
as continuation for:
- https://github.com/apache/maven/pull/12555

impl/maven-core/src/main/java/org/apache/maven/plugin/internal/PluginDescriptorRequirementsValidator.java
```diff
@@ -0,0 +1,62 @@
...
+            MojoDescriptor mojoDescriptor,
+            Class<?> mojoClass,
+            PlexusConfiguration pomConfiguration,
+            ExpressionEvaluator expressionEvaluator) {
+        if (!mojoDescriptor.getRequirements().isEmpty()) {
+            pluginValidationManager.reportPluginMojoValidationIssue(
+                    PluginValidationManager.IssueLocality.EXTERNAL,
+                    mavenSession,
+                    mojoDescriptor,
+                    mojoClass,
```
gnodet, line 59: Two issues in this warning message:

1. **`{@code @Component}` is Javadoc syntax** — inside a Java string literal it will render literally as `{@code @Component}` in console output, not `@Component`. All other validators in this package (e.g. `PlexusContainerDefaultDependenciesValidator`, `Maven3CompatDependenciesValidator`) use plain text in their runtime messages.

2. **Grammar**: "Plugin use" should be "Plugin uses" (subject-verb agreement). The second clause is also awkward — "Maven 4 DI ... to inject dependencies should be used instead".

Suggested rewrite:
```suggestion
                    "Plugin uses Plexus Component requirements (@Component annotation). "
                            + "Use Maven 4 Dependency Injection (for v4 plugins) or JSR 330 annotations (for v3 plugins) to inject dependencies instead.");
```

gnodet (commented):
Two minor issues confirmed in the warning message string — the validation logic and test coverage look solid.

impl/maven-core/src/main/java/org/apache/maven/plugin/internal/PluginDescriptorRequirementsValidator.java
slawekjaranowski: @gnodet done, please look again

gnodet (approved):
Both issues from the prior review have been cleanly addressed in the "Apply review comments" commit:
- `{@code @Component}` Javadoc syntax removed from runtime string — now uses `(@Component annotation)`
- Grammar/phrasing improved — now reads naturally
- Unused `mock` import in the test was also removed

The validation logic is correct, tests cover both positive and negative paths, and CI is green.

**Very minor (non-blocking):** the constructor does not use `requireNonNull(pluginValidationManager)` unlike the abstract base classes in this package (`AbstractMavenPluginParametersValidator`, `AbstractMavenPluginDependenciesValidator`). With JSR 330 injection this is a non-issue in practice.

gnodet (approved):
The latest commit adds `requireNonNull(pluginValidationManager)` in the constructor, cleanly addressing the minor observation from our previous review. The import style (`import static java.util.Objects.requireNonNull`) and usage pattern exactly match `AbstractMavenPluginParametersValidator` and `AbstractMavenPluginDependenciesValidator` in the same package. Consistent and correct.

PR #12581, opened 2026-07-29
[ci][gh][workflow] explicit action hash and permissions

rmannibucau:
Trying to harness more our workflows - but maven shared is designed to be too opened.
Audit mainly done by zizmor.

### DIFF
--- a/.github/workflows/pr-automation.yml
+++ b/.github/workflows/pr-automation.yml
@@ -19,9 +19,15 @@ name: PR Automation
 on:
   pull_request_target:
     types:
+      # zizmor: ignore[dangerous-triggers]
       - closed
+
+permissions: {}
 
 jobs:
   pr-automation:
     name: PR Automation
-    uses: apache/maven-gh-actions-shared/.github/workflows/pr-automation.yml@v5
+    permissions:
+      issues: write
+      pull-requests: write
+    uses: apache/maven-gh-actions-shared/.github/workflows/pr-automation.yml@8599b638c78a2bec146a98d75d4c8b4e8458324f # v5
--- a/.github/workflows/release-drafter.yml
+++ b/.github/workflows/release-drafter.yml
@@ -22,6 +22,11 @@ on:
       - master
   workflow_dispatch:
 
+permissions: {}
+
 jobs:
-   update_release_draft:
-      uses: apache/maven-gh-actions-shared/.github/workflows/release-drafter.yml@v5
+  update_release_draft:
+    permissions:
+      contents: write
+      issues: write
+    uses: apache/maven-gh-actions-shared/.github/workflows/release-drafter.yml@8599b638c78a2bec146a98d75d4c8b4e8458324f # v5
--- a/.github/workflows/stale.yml
+++ b/.github/workflows/stale.yml
@@ -23,6 +23,11 @@ on:
   issue_comment:
     types: [ 'created' ]
 
+permissions: {}
+
 jobs:
   stale:
-    uses: 'apache/maven-gh-actions-shared/.github/workflows/stale.yml@v5'
+    permissions:
+      issues: write
+      pull-requests: write
+    uses: apache/maven-gh-actions-shared/.github/workflows/stale.yml@8599b638c78a2bec146a98d75d4c8b4e8458324f # v5


gnodet (approved):
Clean security-hardening PR that adds top-level `permissions: {}` and explicit job-level permissions to three workflow files, pins reusable workflow references to a verified commit hash, and adds a justified zizmor annotation. All changes are correct and consistent with the already-hardened `maven.yml` workflow.

Verified that:
- The commit hash `8599b638c78a2bec146a98d75d4c8b4e8458324f` matches the current HEAD of the `v5` branch of `apache/maven-gh-actions-shared`
- Job-level permissions in all three workflows match what the corresponding reusable workflows require
- The `zizmor: ignore[dangerous-triggers]` annotation on the `pull_request_target` trigger in `pr-automation.yml` is appropriate since the workflow only triggers on `closed` events
- All four workflow files in the repository are now hardened

.github/workflows/pr-automation.yml
```diff
@@ -19,9 +19,15 @@ name: PR Automation
...
+permissions: {}
+
 jobs:
   pr-automation:
     name: PR Automation
-    uses: apache/maven-gh-actions-shared/.github/workflows/pr-automation.yml@v5
+    permissions:
+      issues: write
+      pull-requests: write
+    uses: apache/maven-gh-actions-shared/.github/workflows/pr-automation.yml@8599b638c78a2bec146a98d75d4c8b4e8458324f # v5
```
slawekjaranowski, line 33: It is intensionally here. 

We assume our repo with shared actions is secure, we introduce every changes to it.

When we pin to hash we need update about 100+ our repos in every changes in shared actions.

We also manage permissions in shared action.
https://github.com/apache/maven-gh-actions-shared/blob/v5/.github/workflows/pr-automation.yml

slawekjaranowski (changes requested):

.github/workflows/pr-automation.yml
rmannibucau: I understand but I dont think we should auto upgrade all repos like that, this is a good idea today which will likely blow up tomorrow so better to invest in automating it with a workflow in shared repo (on demand/manual) than relying on it by default IMHO

permissions being per workflow/job it is always better to explicit it in leaves than in parents IMHO, in particular when there it does have some serious impact on the run of the workflow (not having write perms will fail and you can wonder why - would be ok only if read only for PR IMHO)

slawekjaranowski: for permissions I'm ok,

but for hash version I afraid that will be more maintenance work for us - who will be take care about it?

rmannibucau: what I had in mind was something like dependabot and we do approve after review as any "jar" dependency, wdyt?

gnodet (commented):
**Permissions hardening looks correct** — the `permissions: {}` top-level lockdown plus job-level grants across all three workflow files follow least-privilege and are consistent with how `maven.yml` already operates.

**Stale pinned hash** — the pinned hash `8599b638c78a2bec146a98d75d4c8b4e8458324f` (July 20) is now 7 commits behind the current `v5` HEAD (`07649a12f6dacdb747f586cdd6f5cb52b1993eff`). The missing commits include bumps to release-drafter 7.7.0, actions/stale 11.0.0, actions/setup-java 5.7.0, Maven 4 RC-6 default, a verify-job startup fix, and a timeout-lowering change. If hash pinning is kept, the hash should be updated before merge.

The broader policy question — whether to hash-pin org-internal shared workflow references (security hardening vs. maintenance overhead across 100+ repos) — remains open between the author and maintainers. Dependabot does support updating hash-pinned references (using the trailing `# v5` comment), which mitigates some of the maintenance burden.

Minor: the 3-space → 2-space indentation fix in `release-drafter.yml` and the `zizmor: ignore[dangerous-triggers]` annotation (justified since `pull_request_target` only triggers on `closed`) are both correct.

gnodet (commented):
The permissions-hardening portion of this PR is sound — top-level `permissions: {}` plus explicit job-level grants follows least-privilege best practices, consistent with how `maven.yml` already operates.

**Observations:**

1. **Stale hash**: The pinned hash `8599b638` is now 8 commits behind the current `v5` HEAD (`a7cd1659`). The missing commits include bumps to release-drafter 7.7.0, actions/stale 11.0.0, setup-java 5.7.0, a Maven 4 RC-6 default, and workflow improvements.

2. **Missing permission if hash updated**: If the pinned hash is updated to current `v5` HEAD, `stale.yml` will need `actions: write` added to its job-level permissions. The shared stale workflow now requires this for `actions/stale` v11.0.0 cache management.

3. **Indentation fix** in `release-drafter.yml` (3-space → 2-space) is a positive cosmetic improvement.

**Note:** The CHANGES_REQUESTED from @slawekjaranowski regarding hash-pinning policy is still outstanding. This is a governance decision — the technical implementation of both the hash pinning and the permissions hardening is correct.

.github/workflows/pr-automation.yml
gnodet: Afaik, dependabot already supports upgrading workflow pinned dependencies, so this should be fine.

rmannibucau:
how do we move forward on the governance since think there is no agreement? determinism vs random version usage?

do we trigger a vote?

rmannibucau:
up?

PR #12612, opened 2026-07-29
fix #12583: Inverted file existence check in DefaultTransport.put()

### LINKED ISSUE #12583: [maven-4.0.x] DefaultTransport.put() has inverted file existence check
# DefaultTransport.put() has inverted file existence check

**Found in:** maven-4.0.x branch (up to commit 5ef0ac2)
**File:** `impl/maven-impl/src/main/java/org/apache/maven/impl/transport/DefaultTransport.java` (line 101)
**Severity:** Critical

## Description

The `put()` method has an inverted condition on the file existence check:

```java
if (Files.isRegularFile(source)) {
    throw new IllegalArgumentException("source file does not exist or is not a file");
}
```

`Files.isRegularFile(source)` returns `true` when the source IS a regular file (exists and is valid), but the error message says the file "does not exist or is not a file". The condition should use `!Files.isRegularFile(source)`.

This bug makes the `put()` method impossible to use with any valid file argument — it will always throw `IllegalArgumentException` for legitimate files, silently succeeding only for directories, symlinks, or non-existent paths (which will then fail downstream). The entire transport upload mechanism is effectively broken.

## Expected behavior

The condition should be negated:
```java
if (!Files.isRegularFile(source)) {
    throw new IllegalArgumentException("source file does not exist or is not a file");
}
```

elharo:
The condition `Files.isRegularFile(source)` in `DefaultTransport.put()` was inverted — it should have been `!Files.isRegularFile(source)`. This caused `put()` to reject valid existing files and silently accept non-existent ones, which broke `putBytes()` and `putString()` transitively.

**Fix:** Added the missing `!` negation operator.

**Tests:** Added `DefaultTransportTest` with tests for non-existent file rejection, valid file acceptance, and the `putBytes()` -> `put()` delegation chain.

Closes #12583

### DIFF
--- a/impl/maven-impl/src/main/java/org/apache/maven/impl/DefaultTransport.java
+++ b/impl/maven-impl/src/main/java/org/apache/maven/impl/DefaultTransport.java
@@ -98,7 +98,7 @@ public class DefaultTransport implements Transport {
     public void put(Path source, URI relativeTarget) {
         requireNonNull(source, "source is null");
         requireNonNull(relativeTarget, "relativeTarget is null");
-        if (Files.isRegularFile(source)) {
+        if (!Files.isRegularFile(source)) {
             throw new IllegalArgumentException("source file does not exist or is not a file");
         }
         if (relativeTarget.isAbsolute()) {
--- a/impl/maven-impl/src/test/java/org/apache/maven/impl/DefaultTransportTest.java
+++ b/impl/maven-impl/src/test/java/org/apache/maven/impl/DefaultTransportTest.java
@@ -18,45 +18,48 @@
  */
 package org.apache.maven.impl;
 
-import java.io.IOException;
 import java.net.URI;
 import java.nio.file.Files;
 import java.nio.file.Path;
 
+import org.eclipse.aether.spi.connector.transport.PutTask;
 import org.eclipse.aether.spi.connector.transport.Transporter;
 import org.junit.jupiter.api.Test;
 import org.junit.jupiter.api.io.TempDir;
 
-import static org.junit.jupiter.api.Assertions.assertDoesNotThrow;
 import static org.junit.jupiter.api.Assertions.assertThrows;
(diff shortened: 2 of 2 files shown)


gnodet (commented):
The core bug fix is correct — the inverted `Files.isRegularFile(source)` condition on line 101 of `DefaultTransport.put()` is properly fixed by adding the missing `!` negation. The existing code was throwing `IllegalArgumentException` when the file IS a regular file, which is backwards per the `Transport.put()` API contract. The bug was introduced in the original Transport API commit (f70b0019cc, 2022-12-01).

A few notes:

1. **Merge conflict / rebase needed** — The PR is in a conflicting state because the branch was forked before commit 8134aa24b1 (which moved `DefaultTransport.java` to the `org.apache.maven.impl` package). The diff currently shows the entire file as new rather than the single-character fix. A rebase on current `master` would produce a clean, minimal diff.

2. **Unused `@TempDir` parameter** — `testPutBytesSucceeds` declares `@TempDir Path tempDir` but never uses it. It can be removed.

3. **Test assertions could be stronger** — Both `testPutWithExistingFileSucceeds` and `testPutBytesSucceeds` use only `assertDoesNotThrow` without verifying the mock transporter's `put()` method was actually called. Adding `verify(transporter).put(any(PutTask.class))` would confirm the delegation chain works as expected.

gnodet (approved):
All three findings from the previous review have been addressed:

1. **Merge conflict / rebase** — resolved. The diff now cleanly shows the single-line fix.
2. **Unused `@TempDir` parameter** — removed from `testPutBytesSucceeds`.
3. **Missing `verify()` calls** — both `testPutWithExistingFileSucceeds` and `testPutBytesSucceeds` now verify that `transporter.put()` was actually called.

The fix is correct — the missing `!` negation in the file-existence check is properly added, and the tests cover both positive and negative paths plus the `putBytes` delegation chain.

PR #12616, opened 2026-07-29
Fail on misconfigured plugin-requested toolchains

wilx:
Fixes apache/maven-compiler-plugin#618

Maven Compiler Plugin passes its `jdkToolchain` requirements to Maven Core's
public per-plugin `ToolchainManager#getToolchains(...)` API. Maven 3 previously
caught `MisconfiguredToolchainException`, logged it, and returned an empty list.
The compiler plugin consequently fell back to the JDK running Maven and could
report `BUILD SUCCESS` despite an explicitly requested malformed toolchain.

This change propagates an `IllegalStateException` with the toolchain type and
the original diagnostic, retaining `MisconfiguredToolchainException` as its
cause. An unchecked exception avoids changing the established Maven 3 public
method signature; adding a checked exception would break source compatibility
for plugin callers. Valid toolchains that do not match continue to produce an
empty selection, and missing toolchain factories retain their existing
log-and-empty-list behavior.

The regression test exercises the public per-plugin lookup with a real
`JavaToolchainFactory`, matching requirements, and a nonexistent `jdkHome`. A
separate control verifies that a valid nonmatching toolchain remains nonfatal.

Verification:

- `mvn -pl maven-core -Dtest=DefaultToolchainManagerTest test`
- `mvn -pl maven-core verify` — 366 tests, 1 skipped
- `mvn verify` — all 16 reactor modules successful
- Maven Core ITs `maven-3.10.x` at `70eabb0679`, run under JDK 8 against
  `apache-maven-3.10.0-SNAPSHOT-bin.zip` from this branch
  (`SHA-256 67b1c28163ecbf09417e262f3f1fc59ff8987fa076a5956b219cc0a97331a53a`):
  `mvn -B -V clean install -Prun-its -Dmaven.repo.local=/home/wilx/maven-sources/core/3.x/its-3/repo -DmavenDistro=/home/wilx/maven-sources/core/maven-issue-mcompiler-618-misconfigured-toolchain/apache-maven/target/apache-maven-3.10.0-SNAPSHOT-bin.zip`
  — 867 tests, 0 failures, 0 errors, 40 skipped; all 78 reactor modules
  successful.
- Maven Compiler Plugin 3.15.0 reproducer:
  - Maven 3.9.16 logged the malformed toolchain, compiled with its runtime JDK,
    and returned `BUILD SUCCESS`.
  - The patched Maven 3.10.0-SNAPSHOT returned `BUILD FAILURE` with the invalid
    JDK path and did not compile the source.
  - A valid JDK 17 toolchain requested as version 21 remained a normal
    nonmatch and the build proceeded normally.

### DIFF
--- a/maven-core/src/main/java/org/apache/maven/toolchain/DefaultToolchainManager.java
+++ b/maven-core/src/main/java/org/apache/maven/toolchain/DefaultToolchainManager.java
@@ -88,7 +88,8 @@ public class DefaultToolchainManager implements ToolchainManager {
                             toolchains.add(toolchain);
                         }
                     } catch (MisconfiguredToolchainException ex) {
-                        logger.error("Misconfigured toolchain.", ex);
+                        throw new IllegalStateException(
+                                "Misconfigured toolchain of type " + type + ": " + ex.getMessage(), ex);
                     }
                 }
             }
--- a/maven-core/src/main/java/org/apache/maven/toolchain/ToolchainManager.java
+++ b/maven-core/src/main/java/org/apache/maven/toolchain/ToolchainManager.java
@@ -44,6 +44,7 @@ public interface ToolchainManager {
      * @param type the type, must not be {@code null}
      * @param context the Maven session, must not be {@code null}
      * @return the toolchain selected by <code>maven-toolchains-plugin</code>
+     * @throws IllegalStateException if the configured toolchain cannot be created
      */
     Toolchain getToolchainFromBuildContext(String type, MavenSession context);
 
@@ -55,6 +56,7 @@ public interface ToolchainManager {
      * @param type the type, must not be {@code null}
      * @param requirements the requirements, may be {@code null}
      * @return the matching toolchains, never {@code null}
+     * @throws IllegalStateException if a configured toolchain cannot be created
      * @since 3.3.0
      */
     List<Toolchain> getToolchains(MavenSession session, String type, Map<String, String> requirements);
--- a/maven-core/src/test/java/org/apache/maven/toolchain/DefaultToolchainManagerTest.java
+++ b/maven-core/src/test/java/org/apache/maven/toolchain/DefaultToolchainManagerTest.java
@@ -18,6 +18,7 @@
  */
 package org.apache.maven.toolchain;
 
+import java.nio.file.Path;
 import java.util.Arrays;
 import java.util.Collections;
 import java.util.HashMap;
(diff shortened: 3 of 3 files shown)


gnodet (approved):
Well-scoped behavioral fix that correctly changes misconfigured toolchain handling from silent swallowing (log + empty list) to fail-fast (`IllegalStateException`). The approach aligns with Maven 4 which already throws `ToolchainManagerException` on the equivalent path. Tests are thorough and well-designed.

Minor notes:
- The `@throws IllegalStateException` Javadoc was added to `getToolchains()` but not to `getToolchainFromBuildContext()`, which also calls the same private `selectToolchains()` method and can now throw the same exception. While in practice this path is unlikely to trigger (models stored in build context were already successfully created), the interface contract should document it for completeness.
- Maven 4 master already fails fast on misconfigured toolchains — the compat layer in `ToolchainManagerFactory.java` wraps `MisconfiguredToolchainException` in `RuntimeException`, and the v4 `DefaultToolchainManager.createToolchain()` throws `ToolchainManagerException`. This PR brings Maven 3.10.x into alignment with that behavior using `IllegalStateException`, which is the right choice given the public API constraint.

wilx:
Fixed this:

> The @throws IllegalStateException Javadoc was added to getToolchains() but not to getToolchainFromBuildContext(), which also calls the same private selectToolchains() method and can now throw the same exception. While in practice this path is unlikely to trigger (models stored in build context were already successfully created), the interface contract should document it for completeness.

gnodet (approved):
**Re-review after new commit** ✅

The new commit correctly adds the missing `@throws IllegalStateException` Javadoc to `getToolchainFromBuildContext()`, completing the documentation that was initially only applied to `getToolchains()`. Both methods delegate to the shared `selectToolchains` private method where the exception is thrown, so the annotation accurately reflects runtime behavior.

No remaining issues — the behavioral change, tests, and documentation are all consistent and well-scoped.

---

gnodet (approved):
Well-scoped behavioral fix that correctly converts silently-swallowed `MisconfiguredToolchainException` into a fail-fast `IllegalStateException`, aligning Maven 3.10.x with Maven 4's behavior.

Key observations:
- The choice of `IllegalStateException` (unchecked) is correct since `ToolchainManager` is a public API since Maven 2.0.9 — adding a checked exception would break source compatibility
- This aligns the public `ToolchainManager` API (used by plugins) with the private `ToolchainManagerPrivate` API (used by maven-toolchains-plugin), which already propagated the exception
- Tests are thorough with clear Javadoc contracts on both interface methods

Two minor observations (non-blocking):
- The test instantiates `JavaToolchainFactory` via `new JavaToolchainFactory()` without injecting its `@Inject Logger` field — works today because the exception is thrown before logger use, but could be fragile if factory constructors start logging during creation
- The new fail-fast semantics apply to ALL toolchain models of the requested type, not just the matching one. If a user has a misconfigured JDK 17 and a valid JDK 21, requesting JDK 21 will now fail on the misconfigured entry. This is intentional (consistent with Maven 4) but worth a release-notes mention for users upgrading from 3.9.x

gnodet (approved):
LGTM — Well-scoped behavioral fix that correctly converts silently-swallowed `MisconfiguredToolchainException` into a fail-fast `IllegalStateException`, aligning Maven 3.10.x with Maven 4 behavior. The exception type choice (`IllegalStateException` vs. checked) is correct to preserve source compatibility of the public `ToolchainManager` API.

1. **Test Logger fragility**: The test instantiates `JavaToolchainFactory` via `new` without injecting its `@Inject Logger` field. This works because `MisconfiguredToolchainException` is thrown before any logger use, but would NPE if early logging were added to the factory.

2. **Fail-fast scope**: The fail-fast applies to ALL toolchain models of the requested type, not just matching ones — a misconfigured JDK 17 would fail the build even when requesting JDK 21. This is intentional and matches Maven 4's identical behavior, but is worth a release-notes mention for users upgrading from 3.9.x.

### 📋 PR Metadata

| Labels | _(none)_ | `bug` |

cstamas (approved):

PR #12617, opened 2026-07-29
Remove erroneous optimization which was skipping the path normalization.

desruisseaux:
That optimization can produce false positive if the pattern appears in the base directory. For example if the base directory is `something/target/test-classes` and if an include or exclude pattern is `**/test-classes` where `test-classes` is another directory as a child of the base directory, the optimization would always consider that there is a match.

### DIFF
--- a/impl/maven-impl/src/main/java/org/apache/maven/impl/PathSelector.java
+++ b/impl/maven-impl/src/main/java/org/apache/maven/impl/PathSelector.java
@@ -214,13 +214,6 @@ final class PathSelector implements PathMatcher {
     private final Path baseDirectory;
 
     /**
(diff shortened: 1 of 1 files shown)


gnodet (commented):
The fix correctly removes a flawed optimization that skipped path relativization when all patterns started with `**`, which could cause false-positive matches when the base directory path itself contained segments matching the pattern. The code change is sound.

One suggestion: no regression test is added for the specific false-positive scenario described in the PR body. A test constructing that exact scenario (e.g., base directory = `something/target/test-classes` with pattern `**/test-classes` matching a child directory) would prevent reintroduction of the removed optimization and document the edge case for future maintainers.

PR #12618, opened 2026-07-29
Add regression test for PathSelector false-positive (#12617)

gnodet:
## Summary

- Cherry-picks the fix from #12617 (by @desruisseaux) which removes an erroneous optimization in `PathSelector` that skipped path relativization when all patterns started with `**`
- Adds a regression test for the exact false-positive scenario: when the base directory path contains a segment matching a `**` pattern (e.g., base = `something/target/test-classes` with pattern `**/test-classes/**`), files outside the matching child directory were incorrectly matched

The test verifies:
- A file inside a `test-classes/` child directory **is** correctly matched
- A file in an `other/` sibling directory **is not** falsely matched just because the base directory path contains `test-classes`

## Test plan

- [x] `PathSelectorTest.testNoFalsePositiveWhenBaseDirectoryMatchesPattern` fails without the fix (confirmed false positive)
- [x] All 6 `PathSelectorTest` tests pass with the fix applied
- [ ] CI passes


gnodet:
Closing — will recreate with the correct base (on top of PR #12617).

PR #12619, opened 2026-07-29
[backport maven-4.0.x] fix #12583: Inverted file existence check in DefaultTransport.put()

### LINKED ISSUE #12583: [maven-4.0.x] DefaultTransport.put() has inverted file existence check

gnodet:
## Backport of #12612

Cherry-pick of #12612 onto `maven-4.0.x`.

**Original PR:** #12612 - fix #12583: Inverted file existence check in DefaultTransport.put()
**Original author:** @elharo
**Target branch:** `maven-4.0.x`

### Original description

The condition `Files.isRegularFile(source)` in `DefaultTransport.put()` was inverted — it should have been `!Files.isRegularFile(source)`. This caused `put()` to reject valid existing files and silently accept non-existent ones, which broke `putBytes()` and `putString()` transitively.

**Fix:** Added the missing `!` negation operator.

**Tests:** Added `DefaultTransportTest` with tests for non-existent file rejection, valid file acceptance, and the `putBytes()` -> `put()` delegation chain.

Closes #12583

---

### DIFF
--- a/impl/maven-impl/src/main/java/org/apache/maven/impl/DefaultTransport.java
+++ b/impl/maven-impl/src/main/java/org/apache/maven/impl/DefaultTransport.java
@@ -98,7 +98,7 @@ public class DefaultTransport implements Transport {
     public void put(Path source, URI relativeTarget) {
         requireNonNull(source, "source is null");
         requireNonNull(relativeTarget, "relativeTarget is null");
-        if (Files.isRegularFile(source)) {
+        if (!Files.isRegularFile(source)) {
             throw new IllegalArgumentException("source file does not exist or is not a file");
         }
         if (relativeTarget.isAbsolute()) {
--- a/impl/maven-impl/src/test/java/org/apache/maven/impl/DefaultTransportTest.java
+++ b/impl/maven-impl/src/test/java/org/apache/maven/impl/DefaultTransportTest.java
@@ -1,0 +1,65 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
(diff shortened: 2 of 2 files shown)


gnodet (commented):
LGTM — clean backport of the one-character bug fix from PR #12612. The diff is byte-for-byte identical to the already-merged master fix.

The fix is correct: `DefaultTransport.put()` line 101 currently reads `if (Files.isRegularFile(source))` which throws when the file IS a regular file — the exact opposite of the intent expressed by the error message. Adding the `!` negation makes the guard reject non-existent or non-regular files as intended.

The new `DefaultTransportTest` covers the three key scenarios (non-existent file rejection, existing file acceptance, and `putBytes()` → `put()` delegation). Tests use already-available module dependencies.

PR #12621, opened 2026-07-29
Remove erroneous path normalization optimization + regression test

gnodet:
## Summary

This PR incorporates the fix from #12617 (by @desruisseaux) and adds the regression test requested in the [review](https://github.com/apache/maven/pull/12617#pullrequestreview-4811043754).

- **Commit 1** (from #12617): Removes the `needRelativize` optimization in `PathSelector` that skipped path relativization when all patterns started with `**`. The optimization produced false positives when the base directory path itself contained a segment matching the pattern.
- **Commit 2**: Adds a regression test `testNoFalsePositiveWhenBaseDirectoryMatchesPattern` that constructs the exact scenario from the PR description — base directory `something/target/test-classes` with pattern `**/test-classes/**` falsely matching files in sibling directories.

Supersedes #12617 (maintainer push was disabled on the fork, so the test could not be added directly to that PR).

## Test plan

- [x] New test fails without the fix (false positive confirmed)
- [x] All 6 `PathSelectorTest` tests pass with the fix
- [ ] CI passes

### DIFF
--- a/impl/maven-impl/src/main/java/org/apache/maven/impl/PathSelector.java
+++ b/impl/maven-impl/src/main/java/org/apache/maven/impl/PathSelector.java
@@ -214,13 +214,6 @@ final class PathSelector implements PathMatcher {
     private final Path baseDirectory;
 
     /**
-     * Whether paths must be relativized before being given to a matcher. If {@code true}, then every paths
-     * will be made relative to {@link #baseDirectory} for allowing patterns like {@code "foo/bar/*.java"}
-     * to work. As a slight optimization, we can skip this step if all patterns start with {@code "**"}.
-     */
-    private final boolean needRelativize;
-
-    /**
      * Creates a new selector from the given includes and excludes.
      *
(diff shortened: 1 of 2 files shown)


gnodet (commented):
LGTM — clean, well-tested bugfix that removes a flawed path-relativization optimization introduced in #10935.

The first commit (from desruisseaux via #12617) correctly fixes the false-positive matching by making relativization unconditional, and the second commit adds a precise regression test covering the exact scenario (base directory containing `test-classes` as a segment, pattern `**/test-classes/**`).

The commit structure properly preserves attribution while superseding #12617 (which lacked a test and had maintainer push disabled).

The regression test is well-constructed: it verifies both the true-positive case (file inside a `test-classes/` child directory matches) and the false-positive prevention (file in a sibling `other/` directory does not match just because the base path contains `test-classes`).

CI full-build checks pass on all platforms (macOS/Windows/Linux, JDK 17/21/25). Integration tests still pending at time of review.

PR #12622, opened 2026-07-29
Forward-port: Remove erroneous PathSelector optimization + regression test

gnodet:
## Summary

Forward-port of #12621 from `maven-4.0.x` to `master`.

- Removes the `needRelativize` optimization in `PathSelector` that skipped path relativization when all patterns started with `**`, producing false positives when the base directory path contained a matching segment
- Adds regression test for the exact false-positive scenario

Cherry-pick applied cleanly with no conflicts.

Original fix by: Martin Desruisseaux (@desruisseaux)

### DIFF
--- a/impl/maven-impl/src/main/java/org/apache/maven/impl/PathSelector.java
+++ b/impl/maven-impl/src/main/java/org/apache/maven/impl/PathSelector.java
@@ -214,13 +214,6 @@ final class PathSelector implements PathMatcher {
     private final Path baseDirectory;
 
     /**
-     * Whether paths must be relativized before being given to a matcher. If {@code true}, then every paths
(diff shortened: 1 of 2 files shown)


gnodet (commented):
**Review: APPROVE** ✅

Clean forward-port of already-merged PR #12621 from maven-4.0.x to master. The diffs are byte-for-byte identical — no conflicts, no modifications.

The fix correctly removes the flawed `needRelativize` optimization in `PathSelector.simplify()` that caused false-positive path matching, and the regression test covers both the true-positive and false-positive scenarios. The only remaining shortcut is the `INCLUDES_ALL` case for zero includes and zero excludes, which is correct.

*This is a gnodet-authored PR, so posting as COMMENT instead of APPROVE.*

---

PR #12623, opened 2026-07-29
Remove an optimization on PathSelector producing false negatives

desruisseaux:
Remove an optimization on `PathSelector` which was sometime producing false negatives. It is safe to use an exclude filter for directories during `FileVisitor` transversal, but it is unsafe to use an include filter because of the need to traverse parents first.

### DIFF
--- a/impl/maven-impl/src/main/java/org/apache/maven/impl/PathSelector.java
+++ b/impl/maven-impl/src/main/java/org/apache/maven/impl/PathSelector.java
@@ -537,13 +537,6 @@ final class PathSelector implements PathMatcher {
         private final PathMatcher[] dirExcludes;
 
         /**
-         * Whether to ignore the includes defined by the enclosing class.
-         * This flag can be {@code false} if we determined that all includes are applicable to directories.
-         * This flag should be {@code true} in case of doubt since directory filtering is only an optimization.
-         */
-        private final boolean ignoreIncludes;
-
-        /**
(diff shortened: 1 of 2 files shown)


gnodet (approved):
**Review: APPROVE** ✅

Well-reasoned fix that removes the unsafe include-based directory pre-filtering optimization from `DirectoryPrefiltering`. The rationale is sound: excluding directories is safe (if a directory matches an exclude, all files within are excluded), but include-based filtering is inherently unsafe because `FileVisitor` must enter parent directories before discovering matching children. For example, with include pattern `**/*foo*/**`, the old code would reject directory `org/` because it doesn't contain "foo", preventing discovery of `org/foo/` and files beneath it.

The fix correctly:
- Removes the `ignoreIncludes` field and all include-based logic
- Simplifies `matches()` to accept any directory not explicitly excluded
- Updates `simplify()` to return `INCLUDES_ALL` when no excludes exist
- Flips the appropriate test assertions with clear explanatory comments

The trade-off (slightly more directory traversal for correctness) is the right call, especially since file-level includes/excludes remain unchanged. The exclude-based filtering (retained) handles the most common performance case (skipping `.git`, `target`, etc.).

**Minor nit:** A couple of small language issues in the test comments — "Before to match" → "In order to match" / "Before matching", and "Same reason than above" → "Same reason as above". These are cosmetic only.

---

PR #12624, opened 2026-07-30
Forward-port: Remove PathSelector include-based directory pre-filtering optimization

gnodet:
## Summary

Forward-port of #12623 from `maven-4.0.x` to `master`.

Removes the unsafe include-based directory pre-filtering optimization from `PathSelector.DirectoryPrefiltering`. Using exclude filters for directory traversal is safe, but using include filters causes false negatives because parent directories need to be traversed before matching children.

Cherry-pick applied cleanly with no conflicts.

Author: Martin Desruisseaux (@desruisseaux)

### DIFF
--- a/impl/maven-impl/src/main/java/org/apache/maven/impl/PathSelector.java
+++ b/impl/maven-impl/src/main/java/org/apache/maven/impl/PathSelector.java
@@ -537,13 +537,6 @@ final class PathSelector implements PathMatcher {
         private final PathMatcher[] dirExcludes;
 
         /**
-         * Whether to ignore the includes defined by the enclosing class.
-         * This flag can be {@code false} if we determined that all includes are applicable to directories.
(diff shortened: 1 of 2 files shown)


gnodet (commented):
✅ **LGTM — Clean forward-port**

This is a verified clean forward-port of #12623 from `maven-4.0.x` to `master`. The diffs are byte-for-byte identical, the cherry-pick applied with no conflicts, and the language nits noted in the original review were already addressed before merge.

The change correctly removes the unsafe include-based directory pre-filtering optimization from `DirectoryPrefiltering`, retaining only exclude-based filtering which is safe. Test assertions are correctly flipped to reflect the new (more permissive) directory matching behavior, with clear explanatory comments.

_Note: Cannot submit as APPROVE on self-authored PRs. This review would be APPROVE otherwise._

PR #12626, opened 2026-07-30
Fix field cache to let child class fields take precedence over parent

gnodet:
## Summary

- Fix `EnhancedCompositeBeanHelper.buildFieldCache()` to use `putIfAbsent()` instead of `put()`, so child class fields take precedence over parent fields with the same name
- Add regression tests for field-shadowing scenarios (`ParentBean`/`ChildBean` with same `skip` field)

## Problem

When a child Mojo class shadows a parent's field (e.g. both `ResourcesMojo` and `TestResourcesMojo` declare `private boolean skip`), the field cache incorrectly resolves to the **parent's** field. This causes POM `<configuration><skip>true</skip></configuration>` to inject into the parent's `skip` field but not the child's — breaking `TestResourcesMojo`'s skip behavior.

**Root cause:** `DeclaredMembers` iterates child-first then parent. The field cache used `put()` which overwrites the child entry with the parent's. The method cache already correctly uses `putIfAbsent()` — this was an asymmetry.

## Fix

One-line change: `fieldMap.put(...)` → `fieldMap.putIfAbsent(...)` in `buildFieldCache()`.

This matches the method cache behavior and ensures the first-seen field (from the child class) wins.

Reported in: https://github.com/apache/maven-resources-plugin/issues/497

## Test plan

- [x] New test `testSetPropertyOnShadowedFieldInjectsIntoChildField` — verifies child field receives the injected value
- [x] New test `testShadowedFieldCacheConsistency` — verifies cache reuse preserves correct field resolution
- [x] All 9 existing + new `EnhancedCompositeBeanHelperTest` tests pass

### DIFF
--- a/impl/maven-core/src/main/java/org/apache/maven/configuration/internal/EnhancedCompositeBeanHelper.java
+++ b/impl/maven-core/src/main/java/org/apache/maven/configuration/internal/EnhancedCompositeBeanHelper.java
@@ -229,7 +229,7 @@ public final class EnhancedCompositeBeanHelper {
         for (Object member : new DeclaredMembers(beanType, View.FIELDS)) {
             Field field = (Field) member;
             if (!Modifier.isStatic(field.getModifiers())) {
-                fieldMap.put(field.getName(), field);
+                fieldMap.putIfAbsent(field.getName(), field);
             }
         }
 
--- a/impl/maven-core/src/test/java/org/apache/maven/configuration/internal/EnhancedCompositeBeanHelperTest.java
+++ b/impl/maven-core/src/test/java/org/apache/maven/configuration/internal/EnhancedCompositeBeanHelperTest.java
@@ -209,6 +209,50 @@ class EnhancedCompositeBeanHelperTest {
     }
 
     /**
+     * Verifies that when a child class shadows a parent field with the same name,
+     * setting the property injects into the child's field (not the parent's).
+     * This is the scenario reported in apache/maven-resources-plugin#497:
+     * both ResourcesMojo and TestResourcesMojo declare {@code private boolean skip},
+     * and DeclaredMembers iterates child-first then parent. Using {@code putIfAbsent}
+     * in buildFieldCache ensures the child's field wins.
+     */
+    @Test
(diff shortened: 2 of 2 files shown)


gnodet (commented):
**LGTM** ✅

Clean one-line bug fix that changes `put()` to `putIfAbsent()` in `buildFieldCache()`, making field cache resolution consistent with the existing method cache behavior. The fix correctly ensures child class fields take precedence over shadowed parent fields when `DeclaredMembers` iterates child-first.

**Details:**
- The fix is verified correct: Sisu's `DeclaredMembers` / `MemberIterator` starts with the given class and walks to `getSuperclass()`, confirming child-first iteration order. With `put()`, the parent's field entry overwrites the child's; with `putIfAbsent()`, the child's field (seen first) is correctly preserved.
- The method cache at line 198 already uses `putIfAbsent()`, so this fix resolves an asymmetry between the two caches introduced in the original `d3bd71bbdf` commit ("perf: optimize CompositeBeanHelper with reflection caching").
- The test bean hierarchy (`ParentBean`/`ChildBean` both declaring `private boolean skip`) accurately mirrors the real-world scenario in apache/maven-resources-plugin#497 where `ResourcesMojo` and `TestResourcesMojo` both declare `skip`.
- Well-tested with two new tests covering the direct fix and cache consistency.

gnodet (commented):
LGTM — Clean, correct one-line fix that changes `put()` to `putIfAbsent()` in `buildFieldCache()`, resolving an asymmetry with the method cache and ensuring child class fields take precedence over shadowed parent fields during Mojo configuration injection. Well-tested with two regression tests.

**Minor suggestion (non-blocking):** In `testSetPropertyOnShadowedFieldInjectsIntoChildField`, consider adding `assertFalse(bean.isParentSkip())` after the existing assertion to explicitly document that the parent's shadowed field was *not* set. The assertion would pass trivially (boolean defaults to false), but it documents the contract more explicitly.

**Note:** The same `put()` bug exists on `master` (line 232 of `buildFieldCache`). A forward-port to `master` should be created to keep the branches consistent.

PR #12628, opened 2026-07-30
Fix #12609: ProtoSession.Builder.newBuilder() NPE on unset property maps

### LINKED ISSUE #12609: [maven-4.0.x] ProtoSession.Builder.newBuilder() NPE on unset property maps
# ProtoSession.Builder.newBuilder() NPE on unset property maps

**File:** `api/maven-api-core/src/main/java/org/apache/maven/api/ProtoSession.java` (inner class Builder, lines ~110-112, 147-148)
**Severity:** Low

## Description

The `newBuilder()` factory method creates a `Builder` without setting `userProperties` or `systemProperties`, leaving them as `null`:

```java
static Builder newBuilder() {
    return new Builder().withStartTime(MonotonicClock.now());
}

// Builder fields are not initialized:
private Map<String, String> userProperties;     // null
private Map<String, String> systemProperties;   // null

public ProtoSession build() {
    return new Impl(userProperties, systemProperties, startTime, topDirectory, rootDirectory);
    // Map.copyOf(null) throws NPE
}
```

`build()` passes them to the `Impl` constructor which calls `Map.copyOf(null)` on each, resulting in a `NullPointerException`. The `newBuilder()` static factory is a public API that leads to an NPE when `build()` is called without setting properties. Fix by initializing fields to `Map.of()`.

elharo:
Fixes #12609

Initialize `userProperties` and `systemProperties` to `Map.of()` in `Builder` to prevent `NullPointerException` when `build()` is called without setting these properties via `newBuilder()`.

Also adds a test that verifies building a session without setting properties does not throw NPE.

### DIFF
--- a/api/maven-api-core/src/test/java/org/apache/maven/api/ProtoSessionTest.java
+++ b/api/maven-api-core/src/test/java/org/apache/maven/api/ProtoSessionTest.java
@@ -1,0 +1,39 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.maven.api;
+
+import java.nio.file.Paths;
+
+import org.junit.jupiter.api.Test;
+
+import static org.junit.jupiter.api.Assertions.assertNotNull;
+
+class ProtoSessionTest {
+
+    @Test
+    void testNewBuilderWithUnsetPropertyMaps() {
+        ProtoSession session = ProtoSession.newBuilder()
+                .withTopDirectory(Paths.get("."))
+                .withRootDirectory(Paths.get("."))
+                .build();
+        assertNotNull(session);
+        assertNotNull(session.getUserProperties());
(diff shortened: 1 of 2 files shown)


api/maven-api-core/src/test/java/org/apache/maven/api/ProtoSessionTest.java
```diff
@@ -0,0 +1,39 @@
...
+
+    @Test
+    void testNewBuilderWithUnsetPropertyMaps() {
+        ProtoSession session = ProtoSession.newBuilder()
+                .withTopDirectory(Paths.get("."))
+                .withRootDirectory(Paths.get("."))
+                .build();
+        assertNotNull(session);
+        assertNotNull(session.getUserProperties());
+        assertNotNull(session.getSystemProperties());
```
gnodet: **Low severity:** Consider also asserting the maps are empty (not just non-null) to lock in the expected default behavior:

```suggestion
        assertNotNull(session.getUserProperties());
        assertTrue(session.getUserProperties().isEmpty());
        assertNotNull(session.getSystemProperties());
        assertTrue(session.getSystemProperties().isEmpty());
```

(Would also need `import static org.junit.jupiter.api.Assertions.assertTrue;`)

gnodet (approved):
**LGTM** ✅

Clean, minimal, and correct fix that initializes `userProperties` and `systemProperties` to `Map.of()` in the Builder, preventing the `NullPointerException` from `Map.copyOf(null)` in the `Impl` constructor and aligning with the `@Nonnull` contract on both getter methods.

**Details:**
- No other Builder fields suffer from the same silent-null-to-NPE issue: `startTime` is always set by `newBuilder()` via `withStartTime(MonotonicClock.now())`, `topDirectory` uses `requireNonNull` (which gives a clear error message), and `rootDirectory` is intentionally nullable.
- The `effectiveProperties` derived field also behaves correctly with empty defaults.
- One minor suggestion below to strengthen the test assertions.

gnodet (approved):
**Re-reviewed after 3 new commits — still LGTM** ✅

The new commits are all test-only changes: added `assertTrue(isEmpty())` assertions per Copilot feedback, applied gnodet's suggestion, and cleaned up resulting duplicates. The production code change is unchanged and correct — initializing Builder fields to `Map.of()` prevents `Map.copyOf(null)` NPE in the `Impl` constructor.

PR #12629, opened 2026-07-30
Fix #12607: DefaultMaven session scope leak on constructor exception

### LINKED ISSUE #12607: [maven-4.0.x] DefaultMaven: session scope leak on constructor exception
# DefaultMaven: Session scope leak on constructor exception

**File:** `impl/maven-core/src/main/java/org/apache/maven/DefaultMaven.java` (lines 213-216)
**Severity:** Medium

## Description

`sessionScope.enter()` is called **outside** the try-finally block:

```java
sessionScope.enter();
MavenChainedWorkspaceReader chainedWorkspaceReader =
        new MavenChainedWorkspaceReader(request.getWorkspaceReader(), ideWorkspaceReader);
try (CloseableSession closeableSession = ...) {
    ...
} finally {
    sessionScope.exit();
}
```

If the constructor `new MavenChainedWorkspaceReader(...)` at lines 214-215 throws any exception, the `sessionScope.exit()` in the finally block will **never** execute. This leaks the session scope's thread-local state. The `ideWorkspaceReader` parameter is `@Nullable`, and the constructor delegates to `setReaders()` which could throw.

elharo:
Fixes #12607

## Problem
In `DefaultMaven.doExecute()`, `sessionScope.enter()` was called **outside** the try-finally block:

```java
sessionScope.enter();
MavenChainedWorkspaceReader chainedWorkspaceReader =
        new MavenChainedWorkspaceReader(request.getWorkspaceReader(), ideWorkspaceReader);
try (CloseableSession closeableSession = ...) {
    ...
} finally {
    sessionScope.exit();
}
```

If `new MavenChainedWorkspaceReader(...)` threw (it can throw an NPE if a `WorkspaceReader`'s `getRepository()` returns `null`), the `sessionScope.exit()` in the finally block would never execute. This leaked the session scope's thread-local state.

## Fix
Wrapped subsequent code in an outer try-finally, ensuring `sessionScope.exit()` is always called regardless of where the exception occurs.

## Testing
Added `DefaultMavenSessionScopeTest` which:
1. Creates a mock `WorkspaceReader` whose `getRepository()` returns `null` (triggering NPE in `MavenChainedWorkspaceReader`)
2. Calls `DefaultMaven.execute()` with this bad reader
3. Verifies the session scope is properly exited (empty internal values list) even though the request failed


gnodet (approved):
**LGTM** ✅

Correct and minimal fix for a real session scope thread-local leak when `MavenChainedWorkspaceReader` construction fails between `enter()` and the try block. The test validates the exact failure scenario.

**Details:**
- The production code change moves the `MavenChainedWorkspaceReader` construction (and `newCloseableSession` call) inside a new outer try block that guarantees `sessionScope.exit()` runs in all paths after `sessionScope.enter()`. The normal execution path is unchanged; only the error path is fixed.

- The test verifies scope cleanup via reflection on a parent class field (`getSuperclass().getDeclaredField("values")`). This couples the test to `SessionScope`'s internal hierarchy/naming, but is acceptable since no public API exists to query scope state.
- `null` is passed for the non-nullable `superPomProvider` constructor parameter — works because the NPE occurs before it's used, but `mock(SuperPomProvider.class)` would be slightly more robust.

gnodet (approved):
Correct and minimal fix for a real session-scope thread-local leak. The outer try-finally ensures `sessionScope.exit()` is always called after `enter()`, closing the gap when `MavenChainedWorkspaceReader` construction threw between the two calls.

Key observations:
- `sessionScope.enter()` is intentionally outside the try block so `exit()` is not called if `enter()` itself fails (which would throw `IllegalStateException`). The normal execution path is unchanged; only the error path is fixed.
- This is the only production usage of `sessionScope.enter()` in the codebase, so no other sites need the same treatment.

Three minor, non-blocking suggestions on the test:
- The assertion `assertFalse(result.getExceptions().isEmpty())` could be more precise about the expected exception type
- `Files.createTempDirectory` + `deleteOnExit()` could be replaced with JUnit 5 `@TempDir` for deterministic cleanup
- The reflection into `SessionScope.class.getSuperclass().getDeclaredField("values")` couples the test to the internal class hierarchy — acceptable since no public API exists to query scope state

gnodet (approved):
LGTM — Correct and minimal fix for a real session-scope thread-local leak. The outer try-finally guarantees `sessionScope.exit()` runs after `enter()` regardless of where an exception occurs. The production fix is correct and the test validates the exact failure scenario.

**Minor observations (non-blocking, all test-related):**

1. **Null parameter**: The `null` passed for `superPomProvider` works because the NPE occurs before it's used, but `mock(SuperPomProvider.class)` would be slightly more defensive.
2. **Temp directory**: `Files.createTempDirectory` + `deleteOnExit()` could use JUnit 5 `@TempDir` for deterministic cleanup.
3. **Assertion precision**: `assertFalse(result.getExceptions().isEmpty())` could check for the specific exception type.
4. **Reflection coupling**: `getSuperclass().getDeclaredField("values")` couples to SessionScope's internal hierarchy — acceptable given no public API exists for scope state queries.

### 📋 PR Metadata

| Labels | _(none)_ | `bug` |
| Milestone | _(none)_ | `4.0.0-rc-7` |

### 🔀 Backport Status

⚠️ This bug fix targets `master` but the same pattern exists on `maven-4.0.x` (confirmed independently). The `maven-3.9.x` branch already has the correct try-finally wrapping.

Consider creating a backport PR for `maven-4.0.x` or adding a `backport-to-maven-4.0.x` label.

gnodet (approved):
Re-review after 3 new commits addressing prior feedback. The production fix is unchanged and correct — `MavenChainedWorkspaceReader` construction is now inside the try block that guarantees `sessionScope.exit()` runs. The `TestSessionScope` subclass approach to access the protected `values` field is cleaner than the prior reflection-based approach.

Two minor non-blocking observations on the test:

1. **Exception type assertion:** `assertFalse(result.getExceptions().isEmpty())` only verifies _some_ exception occurred. `assertInstanceOf(NullPointerException.class, result.getExceptions().get(0))` would guard against false positives from mock misconfiguration.

2. **Null for non-nullable parameter:** The `null` passed for `superPomProvider` works because the NPE fires before that dependency is used, but `mock(SuperPomProvider.class)` would be more defensive against future constructor changes.

Consider squashing the 4 commits before merge for a cleaner history.

### 📋 PR Metadata

| Category | _(unlabeled)_ | `bug` |
| Labels | _(none)_ | + `bug` |

### 🔀 Backport Status

⚠️ This bug fix targets `master` but no backport PRs were found for:
- `maven-4.0.x` — confirmed still has the vulnerable pattern

`maven-3.9.x` already has the correct `enter(); try { ... } finally { exit(); }` pattern and does not need a backport.

Consider creating a backport PR or adding a `backport-to-maven-4.0.x` label.

gnodet (commented):
Thanks for working on this, @elharo! The defensive try-finally pattern is a reasonable approach, but there are a few things to note:

**The underlying bug was already fixed.** PR #12863 (merged Aug 28) took a simpler approach: it moved the `MavenChainedWorkspaceReader` constructor _before_ `sessionScope.enter()`. On current master (line 212-214), the order is now: (1) constructor, (2) `sessionScope.enter()`. If the constructor throws, `enter()` is never reached, so no scope leak occurs.

This PR proposes an alternative pattern — moving the constructor _after_ `enter()` but inside a try-finally — which is a valid defensive idiom but is working against an already-fixed codebase. The PR description's code snippet showing `enter()` before the constructor doesn't match current master.

**Test observations:**
- The test passes identically on both old and new code: on current master the scope is never entered (constructor throws before `enter()`), so `isScopeEmpty()` returns true trivially. With this PR, the scope is entered then exited, also returning true. A stronger test would verify that `enter()` was actually called (e.g., assert scope is non-empty _before_ exit, or intercept enter/exit calls).
- Commit 2's message mentions asserting `NullPointerException` specifically, but the final test only checks `assertFalse(result.getExceptions().isEmpty())` without verifying the exception type — the specific assertion appears to have been lost during commits 3-4.
- Passing `null` for the non-nullable `superPomProvider` parameter works only because the NPE fires before it's used; `mock(SuperPomProvider.class)` would be more robust.

**Recommendation:** Since #12863 already fixed the leak, this PR could either be closed or reworked as a pure code-hygiene improvement (making the enter/exit guarantee structurally obvious). If kept, the commits should be squashed and the test strengthened.

### 📋 PR Metadata

| Labels | _(none)_ | `bug` |

PR #12630, opened 2026-07-30
Fix #12605: DefaultProjectBuildingHelper synchronized bottleneck in parallel builds

### LINKED ISSUE #12605: [maven-4.0.x] DefaultProjectBuildingHelper: synchronized bottleneck in parallel builds
# DefaultProjectBuildingHelper: Synchronized bottleneck in parallel builds

**File:** `impl/maven-core/src/main/java/org/apache/maven/project/DefaultProjectBuildingHelper.java` (line 117)
**Severity:** Medium

## Description

The entire `createProjectRealm()` method is `synchronized`, making it a global lock during multi-module parallel builds:

```java
@Override
public synchronized ProjectRealmCache.CacheRecord createProjectRealm(
        MavenProject project, Model model, ProjectBuildingRequest request)
```

During a multi-module build where many projects need extension realms set up concurrently, each must acquire this lock serially. Since the inner `projectRealmCache` uses `ConcurrentHashMap` and is already thread-safe, the synchronization could be narrowed to just the non-thread-safe portions.

elharo:
Fixes #12605

## Problem
`DefaultProjectBuildingHelper.createProjectRealm()` was declared `synchronized`, making it a global lock during multi-module parallel builds. Every project needing extension realm setup had to acquire this lock serially.

Additionally, `DefaultProjectRealmCache.put()` had a check-then-act race condition: `containsKey()` followed by `put()` are not atomic, allowing two threads to both pass the check and overwrite each other's entry.

## Fix
1. **Removed `synchronized`** from `createProjectRealm()` — the `projectRealmCache` uses `ConcurrentHashMap` and is already thread-safe for `get`/`createKey`.
2. **Fixed `put()` race** — replaced the non-atomic `containsKey()` + `put()` with `putIfAbsent()`, which atomically inserts only if absent and throws `IllegalStateException` if the key already exists.

## Testing
Added `DefaultProjectRealmCacheTest` that verifies concurrent `put()` calls with the same key result in exactly one successful insertion.

### DIFF
--- a/impl/maven-core/src/main/java/org/apache/maven/project/DefaultProjectBuildingHelper.java
+++ b/impl/maven-core/src/main/java/org/apache/maven/project/DefaultProjectBuildingHelper.java
@@ -143,7 +143,7 @@ public class DefaultProjectBuildingHelper implements ProjectBuildingHelper {
     }
 
     @Override
-    public synchronized ProjectRealmCache.CacheRecord createProjectRealm(
+    public ProjectRealmCache.CacheRecord createProjectRealm(
             MavenProject project, Model model, ProjectBuildingRequest request)
             throws PluginResolutionException, PluginVersionResolutionException, PluginManagerException {
         ClassRealm projectRealm;
--- a/impl/maven-core/src/main/java/org/apache/maven/project/DefaultProjectRealmCache.java
+++ b/impl/maven-core/src/main/java/org/apache/maven/project/DefaultProjectRealmCache.java
@@ -94,13 +94,13 @@ public class DefaultProjectRealmCache implements ProjectRealmCache, Disposable {
     public CacheRecord put(Key key, ClassRealm projectRealm, DependencyFilter extensionArtifactFilter) {
         Objects.requireNonNull(projectRealm, "projectRealm cannot be null");
 
-        if (cache.containsKey(key)) {
-            throw new IllegalStateException("Duplicate project realm for extensions " + key);
-        }
-
         CacheRecord record = new CacheRecord(projectRealm, extensionArtifactFilter);
 
-        cache.put(key, record);
+        CacheRecord existing = cache.putIfAbsent(key, record);
+
+        if (existing != null) {
+            throw new IllegalStateException("Duplicate project realm for extensions " + key);
+        }
 
         return record;
     }
--- a/impl/maven-core/src/test/java/org/apache/maven/project/DefaultProjectRealmCacheTest.java
+++ b/impl/maven-core/src/test/java/org/apache/maven/project/DefaultProjectRealmCacheTest.java
@@ -26,7 +26,6 @@ import java.util.concurrent.Executors;
 import java.util.concurrent.Future;
 
 import org.codehaus.plexus.classworlds.realm.ClassRealm;
-import org.eclipse.aether.graph.DependencyFilter;
 import org.junit.jupiter.api.Test;
 
 import static org.junit.jupiter.api.Assertions.assertEquals;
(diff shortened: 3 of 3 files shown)


gnodet (changes requested):
The PR correctly identifies that `DefaultProjectRealmCache.put()` had a non-atomic check-then-act pattern and fixes it with `putIfAbsent()`. However, removing `synchronized` from `createProjectRealm()` exposes a **TOCTOU race at the caller level** that can crash parallel builds.

### Issue: Race condition in `createProjectRealm()` (high severity)

The `get()` → create realm → `put()` sequence (lines 223-258) is not atomic. In a parallel multi-module build where two modules share the same extensions:
1. Both threads call `projectRealmCache.get(key)` → both get `null`
2. Both enter the `if (record == null)` block
3. Both call `classRealmManager.createProjectRealm()` creating duplicate ClassRealms
4. One thread's `putIfAbsent()` succeeds; the other throws `IllegalStateException`
5. The ISE is **not caught** by the caller at `DefaultProjectBuilder.java:1274` (which only catches `PluginResolutionException | PluginManagerException | PluginVersionResolutionException`) → **build crash**

The old `synchronized` prevented this race entirely.

### Secondary issue: ClassRealm resource leak (medium severity)

The ClassRealm created by the losing thread is never registered in the cache, so `DefaultProjectRealmCache.flush()` never disposes it.

### Suggested approaches

- Replace the get-create-put pattern with a `computeIfAbsent`-style approach that atomically populates the cache
- Or use a finer-grained lock (e.g., striped by key) to serialize same-key access while allowing different keys to proceed concurrently
- Or change `put()` to return the existing record instead of throwing (matching `ConcurrentHashMap.putIfAbsent` semantics) and have the caller dispose the redundant realm

### Test coverage

The new test validates the cache-level atomicity (good), but does not cover the higher-level race in `createProjectRealm()` that this PR introduces.

gnodet (changes requested):
**Re-reviewed after new commit** — the TOCTOU race from our previous review remains unaddressed. The new commit (`df63754`) only fixes checkstyle/style issues in the test file and introduces two new issues.

### 1. TOCTOU race condition still present (high severity)

The critical race in `createProjectRealm()` is unchanged: removing `synchronized` without replacing it with an equivalent concurrency mechanism means `IllegalStateException` will be thrown under contention in parallel builds. See our [previous review](https://github.com/apache/maven/pull/12630#pullrequestreview-4819197179) for the full analysis and suggested approaches.

### 2. Compilation error introduced (high severity)

The checkstyle commit removed the `org.eclipse.aether.graph.DependencyFilter` import along with an unused local variable, but `DependencyFilter.class` is still used inline in `mock(DependencyFilter.class)` at line 52. Since `DependencyFilter` is from `org.eclipse.aether.graph` (not the test's own package), this will fail to compile.

### 3. Typo (low severity)

The renamed loop variable `futre` at line 62 should be `future`.

gnodet:
FYI — #12865 is an alternative approach to this same issue that addresses the review feedback from @gnodet:

1. **TOCTOU race**: fixed by using `ConcurrentHashMap.putIfAbsent()` in the cache `put()` methods, so the cache insertion is atomic. Callers detect when another thread won the race (via identity check on the returned realm) and dispose the redundant `ClassRealm`.
2. **Also fixes `DefaultExtensionRealmCache` and `DefaultPluginArtifactsCache`** which have the same non-atomic `containsKey()` + `put()` pattern and are exposed to concurrent access once the `synchronized` is removed.
3. **Includes a concurrent test** with 8 threads and `CyclicBarrier` for maximum contention.

Happy to defer to the maintainers on which approach to go with.

gnodet:
Closing in favor of #12865 which builds on this PR's idea and addresses the review feedback (TOCTOU race, redundant realm cleanup, extension/plugin cache fixes).

Thanks @elharo for identifying the bottleneck and the initial fix — it was the foundation for the follow-up.

PR #12631, opened 2026-07-30
Fix #12604: DefaultSettingsBuilder dead Windows drive-relative path handling

### LINKED ISSUE #12604: [maven-4.0.x] DefaultSettingsBuilder: dead Windows drive-relative path handling
# DefaultSettingsBuilder: Dead Windows drive-relative path handling

**File:** `impl/maven-impl/src/main/java/org/apache/maven/impl/settings/DefaultSettingsBuilder.java` (lines 131-137)
**Severity:** Medium

## Description

The check for drive-relative Windows paths is always false on all platforms:

```java
if (localRepository != null && !localRepository.isEmpty()) {
    Path file = Paths.get(localRepository);
    if (!file.isAbsolute() && file.toString().startsWith(File.separator)) {
        effective = effective.withLocalRepository(file.toAbsolutePath().toString());
    }
}
```

On Windows, a path starting with `\` (e.g., `\foo\bar`) is already absolute (relative to the current drive root), so `file.isAbsolute()` returns true and the inner condition fails. On Unix (`/` separator), any path starting with `/` is also absolute. This means the condition `!file.isAbsolute() && file.toString().startsWith(File.separator)` is **always false**, making the entire block dead code. The intended fix for drive-relative paths is never executed.

elharo:
Fixes #12604

## Problem

The condition `!file.isAbsolute() && file.toString().startsWith(File.separator)` is **always false** on all platforms:

- On Unix: any path starting with `/` is absolute, so `!file.isAbsolute()` is false
- On Windows: any path starting with `\` is absolute (drive-relative), so `!file.isAbsolute()` is false

This made the entire local repository path resolution block dead code.

## Fix

Replace the impossible condition with `!file.isAbsolute()` (skipping paths containing `${` placeholders that need later interpolation). This correctly resolves relative paths like `relative/repo` to absolute.

## Testing

Added `testRelativeLocalRepositoryIsResolvedToAbsolute` which loads a settings file with `<localRepository>relative/repo</localRepository>` and verifies the output path is absolute.

Existing test `testSettingsWithServersAndAliases` continues to pass because paths containing `${user.home}` placeholders are left untouched.

### DIFF
--- a/impl/maven-impl/src/main/java/org/apache/maven/impl/DefaultSettingsBuilder.java
+++ b/impl/maven-impl/src/main/java/org/apache/maven/impl/DefaultSettingsBuilder.java
@@ -21,7 +21,6 @@ package org.apache.maven.impl;
 import javax.xml.stream.Location;
 import javax.xml.stream.XMLStreamException;
 
-import java.io.File;
 import java.io.IOException;
 import java.io.InputStream;
 import java.nio.file.Path;
@@ -128,11 +127,13 @@ public class DefaultSettingsBuilder implements SettingsBuilder {
                     .build();
         }
 
-        // for the special case of a drive-relative Windows path, make sure it's absolute to save plugins from trouble
+        // resolve relative local repository paths to absolute to save plugins from trouble.
+        // paths containing property placeholders like ${user.home} must be left as-is
+        // so that later interpolation can resolve them.
         String localRepository = effective.getLocalRepository();
         if (localRepository != null && !localRepository.isEmpty()) {
             Path file = Paths.get(localRepository);
-            if (!file.isAbsolute() && file.toString().startsWith(File.separator)) {
+            if (!file.isAbsolute() && !localRepository.contains("${")) {
                 effective = effective.withLocalRepository(file.toAbsolutePath().toString());
             }
         }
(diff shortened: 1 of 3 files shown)


gnodet (approved):
**LGTM** ✅

Clean, well-scoped fix that replaces an effectively dead condition with a broader relative-path resolution guard. The new `contains("${")` check follows established Maven conventions (used in 7+ other places in the codebase), the import cleanup is correct, and the test adequately covers both the happy path and the placeholder-preservation case.

**Notes:**
- The PR description states the old condition `!file.isAbsolute() && file.toString().startsWith(File.separator)` is "always false on all platforms." This is accurate for Unix/Mac. On Windows, `Path.isAbsolute()` returns `false` for drive-relative paths like `\foo`, so the old condition could theoretically be true — but the new broader condition is a strict superset and an improvement.
- The `!localRepository.contains("${")` guard is consistent with the same pattern used in `DefaultDependencyResolver`, `DefaultModelBuilder`, `DefaultModelValidator`, `DefaultArtifactDescriptorReader`, and `MavenValidator`.

**Minor (non-blocking):** The new test resource `settings-relative-local-repo.xml` lacks the ASF license header that sibling files have. RAT excludes `src/test/resources*/**` so it's not enforced, but adding it would be consistent.

PR #12633, opened 2026-07-30
Emit clear error when running JDK cannot compile source level

gnodet:
## Summary

When the running JDK does not support the project's required `--source`/`--release` level (per [JEP 182](https://openjdk.org/jeps/182) retirement schedule), Maven currently lets the build fail with a cryptic `javac` error. This PR adds two complementary features:

1. **mvnup** — a new `ToolchainPluginStrategy` that automatically adds the `maven-toolchains-plugin` with the `select-jdk-toolchain` goal and a version constraint when the project's source level requires an older JDK
2. **Core error message** — when the running JDK cannot compile the project's source level and no toolchain is configured, emit a clear, actionable error instead of letting `javac` fail cryptically

This approach follows the [mailing list consensus](https://lists.apache.org/thread/3j4nk2kjzrvjh7tspqttb62dgco9y7m8): **core diagnoses, it does not repair** — discovery and selection stay in `maven-toolchains-plugin`, and `mvnup` bridges the gap by adding the plugin declaration to the POM.

### Example error output (at build time, if mvnup was not run)

```
[ERROR] Project requires --source 6 which needs JDK <= 11, but the running JDK 21 no longer supports it.
[ERROR] To fix: run 'mvnup' to add the maven-toolchains-plugin with automatic JDK discovery,
[ERROR] or install JDK 11 and configure it in toolchains.xml or via the maven-toolchains-plugin.
```

### Example mvnup output

When `mvnup apply` runs on a project with `maven.compiler.source=1.5`:

```xml
<plugin>
  <artifactId>maven-toolchains-plugin</artifactId>
  <executions>
    <execution>
      <goals>
        <goal>select-jdk-toolchain</goal>
      </goals>
      <configuration>
        <version>(,8]</version>
      </configuration>
    </execution>
  </executions>
</plugin>
```

The version constraint `(,8]` tells `select-jdk-toolchain` to discover and select any JDK up to and including version 8 — the latest JDK that still supports `--source 5`.

### mvnup: `ToolchainPluginStrategy`

A new `UpgradeStrategy` (`@Priority(15)`) that runs by default (same activation as `CompatibilityFixStrategy`):

1. **Detects** the project's source level from `maven.compiler.release` / `maven.compiler.source` properties, or compiler plugin `<configuration>` (`<release>` / `<source>`)
2. **Checks** if the running JDK supports it (using `JdkSourceLevelSupport`)
3. **Skips** if the `maven-toolchains-plugin` with `select-jdk-toolchain` goal is already configured
4. **Adds** the plugin with the `select-jdk-toolchain` goal and a `<version>(,maxJdk]</version>` constraint, where `maxJdk` is the latest JDK that supports the detected source level (via `JdkSourceLevelSupport.latestJdkForSourceLevel()`)

This leverages the plugin's existing JDK discovery mechanism (`ToolchainDiscoverer`) rather than duplicating filesystem scanning in core.

### Core: `DefaultToolchainManager.checkJdkSourceLevelCompatibility()`

Called from `getToolchainFromBuildContext("jdk")` when no explicit toolchain is configured:

1. Reads the source level from Model 4.1.0 `source > targetVersion`, compiler properties, or compiler plugin config
2. If the running JDK is incompatible, emits the error message above with the exact JDK version needed

### Shared: `JdkSourceLevelSupport` (public)

Maps JDK versions to supported source levels per the JEP 182 retirement schedule. Provides:
- `supportsSourceLevel(jdkMajor, sourceLevel)`
- `latestJdkForSourceLevel(sourceLevel)`
- `normalizeSourceLevel(version)` — handles `1.8`, `1.8.0_392`, `21.0.1`, etc.

## Test plan

- [x] Unit tests for `JdkSourceLevelSupport` — normalization, compatibility, `latestJdkForSourceLevel()` (19 tests)
- [x] Unit tests for `DefaultToolchainManager` — error emitted/suppressed for various source levels and detection paths (16 tests)
- [x] Unit tests for `ToolchainPluginStrategy` — applicability, source level detection from properties/plugin config, plugin detection, plugin addition with version constraint, full apply scenarios (20 tests)
- [x] Integration test `MavenITMvnupToolchainPluginStrategyTest` — verifies `mvnup apply` adds the plugin with `select-jdk-toolchain` goal and `(,8]` version constraint for source 1.5, and that a second run is idempotent
- [x] `maven-impl` module: 569 tests pass
- [x] `maven-compat` module: 160 tests pass
- [x] Full CI green on Linux/macOS/Windows × JDK 17/21/25


elharo (commented):
Good idea. does this need an issue to get picked up in the release notes?

gnodet:
> Good idea. does this need an issue to get picked up in the release notes?

No, an issue and a PR are similar in all points, but the fact that an issue does not have code.

gnodet (commented):
Well-designed, thoroughly tested PR that adds JDK source level compatibility checking to both mvnup (automatic toolchains plugin addition) and Maven core (clear error messages). The test coverage is comprehensive (55+ new tests including integration tests) and the code follows existing patterns.

**Non-blocking observations:**

1. **`isApplicable` divergence (low):** `ToolchainPluginStrategy.isApplicable()` is described as having "same activation as CompatibilityFixStrategy," but it's missing the `allOptionsDisabled` branch that `CompatibilityFixStrategy` has. When all options are explicitly disabled, `CompatibilityFixStrategy` still runs but `ToolchainPluginStrategy` does not. If intentional, the description should be updated.

2. **JDK source level retirement table (low):** `JdkSourceLevelSupport.getMinimumSupportedSourceLevel()` has a catch-all `return 8` for JDK >= 21. When a future JDK retires source level 8, this method will need updating. Consider adding a comment referencing the JEP 182 schedule URL.

3. **Design choice (acknowledged):** `checkJdkSourceLevelCompatibility` emits `logger.error()` but doesn't halt the build — the user will still see the subsequent javac error. The PR description documents this as intentional (emit clear, actionable error), with mvnup as the primary remediation path.

**Positives:**
- `JdkSourceLevelSupport` well-placed in `maven-impl` for sharing between mvnup and core
- Normalization for legacy version formats (`1.8`, `1.8.0_392`, `21.0.1`) is comprehensive
- Integration test correctly verifies both initial plugin addition and idempotency

gnodet (commented):
Well-designed, thoroughly tested PR that adds JDK source level compatibility checking to both mvnup (automatic toolchains plugin addition) and Maven core (clear error messages). 55+ new tests including integration tests — excellent coverage.

**Findings:**

1. **[medium] Misleading comment in `ToolchainPluginStrategy.isApplicable()`** — The comment says "Same default logic as CompatibilityFixStrategy" but the logic differs. `CompatibilityFixStrategy` has an `allOptionsDisabled` safety net (lines 115-121) that returns `true` when all options are explicitly disabled (`--no-all --no-infer --no-model --no-plugins`), while `ToolchainPluginStrategy` only checks `noOptionsSpecified` and falls through to `options.model().isPresent()` when all options are disabled. The behavioral divergence may be intentional (compatibility fixes are more critical than plugin addition), but the comment is misleading since the logic is not the same.

2. **[low] `checkJdkSourceLevelCompatibility` emits `logger.error()` without halting the build** — The method logs an actionable error message but does not throw, so the build continues and the user will see both the clear Maven diagnostic and the subsequent cryptic javac error. This appears intentional based on the Javadoc, but worth noting the UX is "clear error followed by cryptic error" rather than "clear error instead of cryptic error."

gnodet (commented):
Well-designed, thoroughly tested PR (55+ new tests) that adds JDK source level compatibility checking to both mvnup and Maven core. The JDK retirement mapping table is correct per JEP 182. Two non-blocking observations:

**1. Misleading comment in `ToolchainPluginStrategy.isApplicable`** (line 90): The comment says "Same default logic as CompatibilityFixStrategy" but the logic differs — `CompatibilityFixStrategy` has an `allOptionsDisabled` safety net that returns true when all options are explicitly disabled (`--no-all --no-infer --no-model --no-plugins`), while `ToolchainPluginStrategy` is missing this branch and would return false in that scenario. The behavioral difference may be intentional, but the comment is inaccurate.

**2. Repeated error in multi-module builds with `-fae`**: `checkJdkSourceLevelCompatibility` is called per `getToolchainFromBuildContext("jdk")` invocation with no deduplication. In default reactor mode this is not an issue (Maven stops at the first failure), but with `--fail-at-end` the error message would repeat once per module.

### 📋 PR Metadata

impl/maven-impl/src/main/java/org/apache/maven/impl/DefaultToolchainManager.java
```diff
@@ -98,6 +113,160 @@ public void storeToolchainToBuildContext(@Nonnull Session session, @Nonnull Tool
...
+        // Check Model 4.1.0 <source><targetVersion> elements
+        Build build = project.getModel().getBuild();
+        if (build != null) {
+            List<Source> sources = build.getSources();
+            if (sources != null) {
+                for (Source source : sources) {
+                    String targetVersion = source.getTargetVersion();
+                    if (targetVersion != null && !targetVersion.isEmpty()) {
+                        int level = JdkSourceLevelSupport.normalizeSourceLevel(targetVersion);
+                        if (level > 0) {
```
rmannibucau: why first one? 
shouldn't we take them all to have the global constraint since versions will slide in jre? 🤔

rmannibucau (approved):

impl/maven-cli/src/main/java/org/apache/maven/cling/invoker/mvnup/goals/ToolchainPluginStrategy.java
```diff
@@ -0,0 +1,355 @@
...
+     * Adds the {@code maven-toolchains-plugin} with {@code select-jdk-toolchain} goal
+     * to the POM's {@code <build><plugins>} section, configured with a version constraint
+     * so the plugin selects a JDK that supports the project's source level.
+     *
+     * @param pomDocument the POM document to modify
+     * @param maxJdkVersion the latest JDK major version that supports the source level
+     */
+    void addToolchainsPlugin(Document pomDocument, int maxJdkVersion) {
+        Element root = pomDocument.root();
+        Element build = root.childElement(BUILD).orElse(null);
```
mthmulders: Why not put the `DomUtils.insertNewElement` call inside an `orElseGet`? The pattern repeats a few times in this method, I believe the method could become a bit more compact.

PR #12650, opened 2026-07-31
Fix #12646: project-local-repo clean race condition when root pom has parent

### LINKED ISSUE #12646: project-local-repo clean race condition when root pom has parent
### Affected version

4.0.0-rc-6

### Bug description

Maven 4 creates `target/project-local-repo` under the root project directory, which contains built JAR artifacts and there is a preference to use them in case of module subset builds.
The directory location assumes that the root project is also a root parent for other modules and if it's not, there is a race condition to clean this directory.

How to reproduce the problem:
* root `pom.xml` should have a parent, let's call it `super-pom/pom.xml`
* the `super-pom/pom.xml` should have other direct child modules, which do not inherit root `pom.xml`
* build the multi-module project as a parallel build, like `./mvnw clean install -T1C`
* the `MultiThreadedBuilder` will schedule `super-pom/pom.xml` as a first project to build in the reactor, it finishes alone
* then, in parallel it starts building root `pom.xml` and other modules which define `super-pom` as direct parent
* maven-clean-plugin of root pom.xml iterates files for removal
* at the same time(!) other modules start adding file under `$projectRoot/target/project-local-repo`
* maven-clean-plugin rechecks that `target` is empty and fails:
```
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-clean-plugin:3.5.0:clean (default-clean) 
on project pt: Failed to clean project: Failed to delete /Users/..../server/target/project-local-repo -> [Help 1]
```
(the root project is called `pt`). See [timeline](https://github.com/maven-turbo-reactor/maven-timeline-extension) diagram that demonstrates this (`pt-clean` is long and executed in parallel):

Image

This is a 100% reproducible for a project with the big number of modules.

gnodet:
## Summary

Move `project-local-repo` from `target/project-local-repo` to `.mvn/target/project-local-repo` so it is not affected by `maven-clean-plugin`'s deletion of `target/`.

This eliminates the race condition in parallel builds (`mvn clean install -T N`) where the root project's `maven-clean-plugin` deletes `target/` while sibling modules concurrently write artifacts into `target/project-local-repo`.

### Root cause

In Maven 4, `clean` and `install` phases are placed into the same `TaskSegment`, so they execute concurrently across projects in parallel builds. When the root `pom.xml` has a parent (e.g. a super-pom in the reactor), `MultiThreadedBuilder` schedules the parent first, then runs the root project and child modules in parallel. The root project's clean phase (via `maven-clean-plugin`) races with sibling modules installing artifacts into `target/project-local-repo`.

### Fix

Move `project-local-repo` from `target/` to `.mvn/target/`, making it structurally impossible for `maven-clean-plugin` to interfere. `ReactorReader` fully owns the lifecycle:
- **Cleanup**: per-GAV cleanup when a project enters its clean phase (preserves sibling artifacts for partial/resumable builds)
- **Install**: on `ProjectSucceeded`, artifacts are hard-linked (or copied) into `.mvn/target/project-local-repo`

The `.mvn/target/` convention provides a clean semantic split:
- `.mvn/` → config (tracked in git: `extensions.xml`, `maven.config`, etc.)
- `.mvn/target/` → build output (gitignored)

This is simpler than the previous lock-based approach — the `ReentrantReadWriteLock` and `isProjectLocalRepoOwner()` are removed entirely.

### Precedent

Storing build-related data outside `target/` (surviving `clean`) is established:
- **Maven Build Cache Extension** — `~/.m2/.cache/maven-build-cache`
- **Develocity Maven Extension** — local cache in `~/.m2/.develocity/`
- **Gradle** — `.gradle/` directory at project root (universally gitignored)

### Trade-off

`.mvn/target/` should be added to `.gitignore` — similar to how Gradle projects gitignore `.gradle/`.

Fixes #12646

### DIFF
--- a/apache-maven/src/assembly/maven/conf/maven-user.properties
+++ b/apache-maven/src/assembly/maven/conf/maven-user.properties
@@ -31,6 +31,16 @@ ${includes} = ?"${maven.user.conf}/maven-user.properties", \
               ?"${maven.project.conf}/maven-user.properties"
 
 #
+# Project-Local Repository
+#
+# Location of the project-local repository used by the reactor to share artifacts between
+# modules during a build. This repository enables partial and resumable builds (e.g.
+# mvn verify -r :module) without requiring install. The path may be absolute or relative
+# to the root directory.
+#
+# maven.project.local.repo = ${maven.rootDirectory}/.mvn/target/project-local-repo
+
+#
 # Maven Cache Configuration
 #
 # Controls caching behavior for different request types during Maven builds.
--- a/apache-maven/src/assembly/maven/conf/maven-user.properties
+++ b/apache-maven/src/assembly/maven/conf/maven-user.properties
@@ -31,14 +31,14 @@ ${includes} = ?"${maven.user.conf}/maven-user.properties", \
               ?"${maven.project.conf}/maven-user.properties"
 
 #
-# Project-Local Repository
+# Reactor Output Repository
 #
-# Location of the project-local repository used by the reactor to share artifacts between
+# Location of the reactor output repository used by the reactor to share artifacts between
 # modules during a build. This repository enables partial and resumable builds (e.g.
 # mvn verify -r :module) without requiring install. The path may be absolute or relative
 # to the root directory.
 #
-# maven.project.local.repo = ${maven.rootDirectory}/.mvn/target/project-local-repo
+# maven.reactor.outputRepository = ${maven.rootDirectory}/.mvn/target/project-local-repo
 
 #
 # Maven Cache Configuration
--- a/api/maven-api-core/src/main/java/org/apache/maven/api/Constants.java
+++ b/api/maven-api-core/src/main/java/org/apache/maven/api/Constants.java
@@ -380,7 +380,7 @@ public final class Constants {
     public static final String MAVEN_VERSION_FILTER = "maven.session.versionFilter";
 
     /**
-     * User property for overriding the project-local repository path used by the reactor reader to share
+     * User property for overriding the reactor output repository path used by the reactor reader to share
      * artifacts between modules during a build. This repository enables partial and resumable builds
      * (e.g. {@code mvn verify -r :module}) without requiring {@code install}.
      * <p>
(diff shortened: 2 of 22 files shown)


gnodet (commented):
Well-designed fix that structurally eliminates the race condition by moving `project-local-repo` from `target/` to `.mvn/target/`. The approach is simpler and more robust than lock-based alternatives, and follows established precedent (Gradle's `.gradle/`, Maven Build Cache Extension).

1. **Non-volatile lazy init** (`ReactorReader.java`): `projectLocalRepository` field is not `volatile`, and `getProjectLocalRepo()` uses unsynchronized lazy initialization. This is a pre-existing issue (not introduced by this PR) — the computed value is deterministic and `Path` is immutable, so the race is practically benign. But since you're already refactoring this method, adding `volatile` would close the JMM visibility gap cleanly.

2. **Test assertion gap**: `MavenITgh12646ProjectLocalRepoCleanRaceTest` verifies error-free log output but doesn't assert that artifacts actually ended up in `.mvn/target/project-local-repo/`. Other tests in this PR (e.g., `MavenITmng5102MixinsTest`) do verify artifact paths — a similar check here would strengthen the race-condition test.

3. **PR description accuracy**: The description mentions removing `ReentrantReadWriteLock` and `isProjectLocalRepoOwner()`, but these constructs don't exist on the current `master` branch. This appears to be leftover text from an earlier iteration.

### 📋 PR Metadata

| Labels | _(none)_ | `bug` |

### 🔀 Backport Status

⚠️ The milestone is `4.0.0-rc-7` (branch `maven-4.0.x`) but this PR targets `master`. A backport to `maven-4.0.x` is needed for inclusion in rc-7. `maven-3.9.x`/`maven-3.10.x` don't have the project-local-repo feature.

cstamas:
I know am late to story, but:
* perso I dislike `.mvn/target` idea, as it is hidden below already used (and checked in) directory
* MWM uses `.mvn-local`
* Let's come up with something "standard"

If this **all has been decided already**, consider I did not say anything (and will make MWM use `.mvn/target`)

cstamas (approved):

PR #12649, opened 2026-07-31
Add validation for Plexus-based plugin dependency injection

slawekjaranowski:
* Add validation for Plexus-based plugin dependency injection

What changed

- Added a new validation that detects outdated dependency injection usage in plugins.
- Added a clear warning for plugin developers when this pattern is found.
- Pointed users to the recommended modern injection approach.

(cherry picked from commit 2a8b6b243ea1b87c76b5eec9bd8a375d5361706f)

PR #12651, opened 2026-07-31
Add validation for Plexus-based plugin dependency injection

slawekjaranowski:
What changed

- Added a new validation that detects outdated dependency injection usage in plugins.
- Added a clear warning for plugin developers when this pattern is found.
- Pointed users to the recommended modern injection approach.

based on: 2a8b6b243ea1b87c76b5eec9bd8a375d5361706f

### DIFF
--- a/maven-core/src/main/java/org/apache/maven/plugin/internal/PluginDescriptorRequirementsValidator.java
+++ b/maven-core/src/main/java/org/apache/maven/plugin/internal/PluginDescriptorRequirementsValidator.java
@@ -1,0 +1,64 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.maven.plugin.internal;
+
+import javax.inject.Inject;
+import javax.inject.Named;
+import javax.inject.Singleton;
+
+import org.apache.maven.execution.MavenSession;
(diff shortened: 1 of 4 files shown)


maven-core/src/main/java/org/apache/maven/plugin/internal/PluginDescriptorRequirementsValidator.java
```diff
@@ -0,0 +1,64 @@
...
+
+/**
+ * Verify that plugin descriptor does not contain Plexus Component requirements.
+ */
+@Singleton
+@Named
+class PluginDescriptorRequirementsValidator implements MavenPluginConfigurationValidator {
+
+    protected final PluginValidationManager pluginValidationManager;
+
```
gnodet, line 43: Nit: This field is `protected` but `PluginDescriptorRequirementsValidator` is a concrete, package-private, final class with no subclasses. The `protected` visibility on the abstract base class makes sense (subclass access), but here `private` would be more appropriate.

```suggestion
    private final PluginValidationManager pluginValidationManager;
```

maven-core/src/main/java/org/apache/maven/plugin/internal/PluginDescriptorRequirementsValidator.java
```diff
@@ -0,0 +1,64 @@
...
+import org.apache.maven.plugin.PluginValidationManager;
+import org.apache.maven.plugin.descriptor.MojoDescriptor;
+import org.codehaus.plexus.component.configurator.expression.ExpressionEvaluator;
+import org.codehaus.plexus.configuration.PlexusConfiguration;
+
+import static java.util.Objects.requireNonNull;
+
+/**
+ * Verify that plugin descriptor does not contain Plexus Component requirements.
+ */
```
gnodet, line 37: Nit: Other validators in this package include `@since` tags (e.g., `DeprecatedCoreExpressionValidator` has `@since 3.9.1`). Consider adding one here:

```suggestion
/**
 * Verify that plugin descriptor does not contain Plexus Component requirements.
 *
 * @since 3.10.0
 */
```

gnodet (approved):
Clean cherry-pick of the already-merged #12577 to the `maven-3.10.x` maintenance branch. The validator logic is identical to `master`; the test was properly adapted for the 3.10.x `MojoDescriptor` API (real instance via `ComponentDescriptor.addRequirement` rather than mocking). Adding `mockito-junit-jupiter` as a separate dependency is correct for 3.10.x, which doesn't use `mockito-bom` like master.

Two minor non-blocking nits (same as noted on #12649):

- `pluginValidationManager` field is `protected` but this is a concrete, package-private, final class with no subclasses — should be `private`
- Missing `@since 3.10.0` Javadoc tag (other validators in the same package include `@since`)

PR #12654, opened 2026-08-01
Add AsyncDrainWriter to eliminate PrintWriter lock contention

gnodet:
## Summary

During parallel model building (`-T1C`), multiple `PhasingExecutor` threads log concurrently via:

```
SLF4J → MavenSimpleLogger → ProjectBuildLogAppender → SimpleBuildEventListener → PrintWriter.println()
```

Since `PrintWriter.println()` is `synchronized`, all threads serialize on every log call. JFR profiling on a 4383-module diamond-graph project shows **1,470ms of blocked time** from this single contention point — the #1 source of lock contention during `validate`.

### Fix

`AsyncDrainWriter` wraps the `Consumer<String>` writer with a lock-free queue + non-blocking drain:

- **Producers** enqueue messages into a `ConcurrentLinkedQueue` (CAS, no blocking)
- **At most one thread** drains the queue to the underlying `PrintWriter` (via `ReentrantLock.tryLock()`)
- **Other threads** return immediately after enqueue — their messages are picked up by the active drain
- **`close()`** performs a final blocking drain to guarantee no messages are lost

### JFR Results (4383-module diamond project, `validate -T1C`)

| Source | Before | After |
|--------|:------:|:-----:|
| PrintWriter contention | 1,470ms | **0ms** |
| PhasingExecutor contention | 546ms | **49ms** |

### Benchmark (10 runs)

| Metric | Before | After | Δ |
|--------|:------:|:-----:|:-:|
| Median | 14.738s | 14.253s | **-485ms (-3.3%)** |
| Average | 14.546s | 14.242s | **-304ms (-2.1%)** |
| Range (variance) | 2.70s | 1.33s | **Halved** |

All 8,766 module log lines verified present — zero messages lost.

## Test plan

- [x] Verified 8,766/8,766 module log lines present (no lost messages)
- [x] JFR confirms PrintWriter contention eliminated (1,470ms → 0ms)
- [x] `maven-core` unit tests pass
- [x] `maven-cli` unit tests pass
- [ ] CI passes

### DIFF
--- a/impl/maven-cli/src/main/java/org/apache/maven/cling/invoker/LookupInvoker.java
+++ b/impl/maven-cli/src/main/java/org/apache/maven/cling/invoker/LookupInvoker.java
@@ -78,6 +78,7 @@ import org.apache.maven.execution.MavenExecutionRequest;
 import org.apache.maven.impl.SettingsUtilsV4;
 import org.apache.maven.jline.FastTerminal;
 import org.apache.maven.jline.MessageUtils;
+import org.apache.maven.logging.AsyncDrainWriter;
 import org.apache.maven.logging.BuildEventListener;
 import org.apache.maven.logging.LoggingOutputStream;
 import org.apache.maven.logging.ProjectBuildLogAppender;
@@ -415,24 +416,30 @@ public abstract class LookupInvoker<C extends LookupContext> implements Invoker 
     }
 
     protected Consumer<String> doDetermineWriter(C context) {
+        Consumer<String> raw;
         if (context.options().logFile().isPresent()) {
             Path logFile = context.cwd.resolve(context.options().logFile().get());
             try {
                 PrintWriter printWriter = new PrintWriter(Files.newBufferedWriter(logFile), true);
                 context.closeables.add(printWriter);
-                return printWriter::println;
+                raw = printWriter::println;
             } catch (IOException e) {
                 throw new MavenException("Unable to redirect logging to " + logFile, e);
             }
         } else {
             // Given the terminal creation has been offloaded to a different thread,
             // do not pass directly the terminal writer
-            return msg -> {
+            raw = msg -> {
                 PrintWriter pw = context.terminal.writer();
                 pw.println(msg);
                 pw.flush();
             };
         }
+        // Wrap with lock-free async drain to eliminate PrintWriter synchronized contention
+        // when multiple PhasingExecutor threads log concurrently during parallel model building.
+        AsyncDrainWriter asyncWriter = new AsyncDrainWriter(raw);
+        context.closeables.add(asyncWriter);
+        return asyncWriter;
     }
 
     protected void activateLogging(C context) throws Exception {
--- a/impl/maven-core/src/main/java/org/apache/maven/logging/AsyncDrainWriter.java
+++ b/impl/maven-core/src/main/java/org/apache/maven/logging/AsyncDrainWriter.java
@@ -1,0 +1,106 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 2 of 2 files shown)


gnodet (commented):
Clean, well-motivated performance optimization with a correct lock-free drain algorithm. The ConcurrentLinkedQueue + tryLock drain + single re-check pattern is well-established for reducing contention on a shared output sink.

The 1,470ms-to-0ms PrintWriter contention elimination is a compelling result backed by thorough JFR profiling data.

One suggestion: consider adding unit tests for `AsyncDrainWriter` — a concurrent component should have at least basic tests for (1) single-threaded message ordering, (2) multi-threaded stress test verifying no messages are lost, and (3) `close()` draining remaining buffered messages.

Also, if `delegate.accept(m)` throws during `drain()`, subsequent queued messages are silently dropped. In practice this is unlikely since `PrintWriter.println()` swallows IOExceptions, but a try/catch in `drain()` would make the component more robust.

_Note: Cannot submit as APPROVE because the PR author matches the review account._

elharo (commented):
Could we ditch the PrintWriter completely?

gnodet:
> Could we ditch the PrintWriter completely?



> Could we ditch the PrintWriter completely?

And go where ? `System.out` also has `synchronized` blocks everywhere in its implementation.  I think ultimately, this is necessary to not have garbled output.

elharo:
Could you just use an OutputStreamWriter instead?

gnodet:
> Could you just use an OutputStreamWriter instead?

Whatever class we use is irrelevant.  We need to ensure the lines are written one after the other, and that we don't have a mix of lines, whether we write on a file or on a terminal.  This requires synchronisation at some point on a single resource (the terminal or file).  The point of this PR is to offload the synchronisation in a separate thread, so that we don't block all maven threads while writing.

gnodet (commented):
**Review: Add AsyncDrainWriter to eliminate PrintWriter lock contention**

Well-designed lock-free drain pattern that correctly eliminates PrintWriter contention. Thread safety is sound, close ordering is correct (LIFO: stop sink → drain buffered → close writer), and CI is fully green. The JFR profiling data is compelling — eliminating 1,470ms of contention on a large parallel build is a meaningful win.

**Confirmed findings:**

1. **[medium] No unit tests for AsyncDrainWriter** — This class has a non-trivial concurrent algorithm (the `tryLock`/re-check pattern) that would benefit from automated tests: (a) single-threaded ordering verification, (b) multi-threaded stress test for no-message-loss under contention, and (c) a test that `close()` drains all remaining buffered messages. While the manual validation (8,766/8,766 log lines verified) is thorough, automated tests would guard against future regressions.

2. **[low] Exception in `drain()` loses the polled message** — In `drain()`, if `delegate.accept(m)` throws, the message already `poll()`-ed from the queue is lost. Remaining messages stay queued for the next drain cycle. Not a practical concern for `PrintWriter.println()` (which swallows IOExceptions), but since `AsyncDrainWriter` accepts any `Consumer<String>`, a `try/catch` around the accept in `drain()` that continues the loop would make this more robust as a general-purpose utility.

3. **[low] Bounded message stranding window** — After the re-check drain completes and releases the lock, there is no further re-check. A message enqueued in the narrow window between the last `poll()` returning null and `unlock()` can sit in the queue until the next `accept()` call or `close()`. The class Javadoc already says "picked up by the ongoing or next drain cycle," which implicitly acknowledges eventual delivery. Consider making this more explicit for future maintainers.

---

gnodet (commented):
**Review — sound concurrency design, missing tests**

The lock-free drain pattern is well-designed: `ConcurrentLinkedQueue` + `ReentrantLock.tryLock()` ensures at most one thread drains at a time, and the post-unlock re-check prevents messages from being stranded by the enqueue-during-drain race. Thread safety is sound — all fields are final, ConcurrentLinkedQueue and ReentrantLock provide proper happens-before guarantees.

Close ordering via `context.closeables` LIFO is correct: AsyncDrainWriter closes first (draining buffered messages), then the underlying PrintWriter.

### Items to note

**🟡 No unit tests (medium)** — A concurrency-critical class like this would benefit from tests verifying: (1) concurrent producers don't lose messages (N threads × M messages), and (2) `close()` drains remaining messages. This guards against future regressions.

- `@since 4.0.0` should be `@since 4.1.0` (project version is 4.1.0-SNAPSHOT, milestone is 4.1.0)
- If `delegate.accept()` throws during `drain()`, the polled message is lost and remaining messages are not drained. Practically near-impossible since PrintWriter swallows IOExceptions, but a try-catch in `drain()` would add resilience

---

PR #12656, opened 2026-08-02
Fix SubList.get() accepting an index equal to its own size

### LINKED ISSUE #12595: [maven-4.0.x] ImmutableCollections.SubList.get() off-by-one in bounds check
# ImmutableCollections.SubList.get() off-by-one in bounds check

**File:** `api/maven-api-xml/src/main/java/org/apache/maven/api/xml/ImmutableCollections.java` (line ~245)
**Severity:** High

## Description

The `SubList.get()` method has an off-by-one error in its bounds check:

```java
public E get(int index) {
    if (index < 0 || index > size()) {    // BUG: should be >=, not >
        throw outOfBounds(index);
    }
    return AbstractImmutableList.this.get(fromIndex + index);
}
```

When `index` equals `size()`, the check passes but the subsequent `AbstractImmutableList.this.get(fromIndex + index)` will compute `fromIndex + size() == toIndex`, which is out of bounds of the parent list. This causes an `IndexOutOfBoundsException` to be thrown from the parent list rather than a clear error message from the sublist.

renechoi:
`SubList.get()` checked `index > size()`, so an index equal to the sub list's size passed and reached `AbstractImmutableList.this.get(fromIndex + index)`.

The issue describes this as surfacing an `IndexOutOfBoundsException` from the parent instead of a clear one from the sub list. That is what happens when the sub list reaches the end of its parent, and the exception is an `ArrayIndexOutOfBoundsException` from the backing array. When the sub list stops earlier there is no exception at all: the parent resolves the index and returns an element the sub list does not contain.

```java
List<String> list = ImmutableCollections.copy(List.of("a", "b", "c"));
list.subList(0, 1).get(1);   // "b"   before
list.subList(0, 1).get(1);   // IndexOutOfBoundsException: Index: 1, Size: 1   after
list.subList(2, 3).get(1);   // ArrayIndexOutOfBoundsException: Index 3 out of bounds for length 3   before
```

`ListN.get()` indexes the array directly and has no bounds check of its own, which is why a read past the sub list is silent.

The same class is checked in three times: the mdo template plus the two copies generated from it under `api/maven-api-xml` and `impl/maven-xml`. All three are fixed, since changing only the copies would let the next regeneration bring it back.

`listIterator(int)` a few lines above keeps `index > size()`. That bound is correct there because `List.listIterator(int)` accepts `size()` as a cursor position, so one of the added tests pins it to prevent the two being aligned later by mistake.

Tests: four cases in a new `ImmutableCollectionsTest` in `impl/maven-xml`, covering the equal-to-size index, the read past a sub list that ends early, negative indices, and the `listIterator` bound. Two of them fail on `maven-4.0.x` and pass with this change; the module suite is green at 75 tests with checkstyle and spotless enabled.

Fixes #12595

### DIFF
--- a/api/maven-api-xml/src/main/java/org/apache/maven/api/xml/ImmutableCollections.java
+++ b/api/maven-api-xml/src/main/java/org/apache/maven/api/xml/ImmutableCollections.java
@@ -297,7 +297,7 @@ class ImmutableCollections {
 
             @Override
             public E get(int index) {
-                if (index < 0 || index > size()) {
+                if (index < 0 || index >= size()) {
                     throw outOfBounds(index);
                 }
                 return AbstractImmutableList.this.get(fromIndex + index);
--- a/impl/maven-xml/src/main/java/org/apache/maven/internal/xml/ImmutableCollections.java
+++ b/impl/maven-xml/src/main/java/org/apache/maven/internal/xml/ImmutableCollections.java
@@ -297,7 +297,7 @@ class ImmutableCollections {
 
             @Override
             public E get(int index) {
-                if (index < 0 || index > size()) {
+                if (index < 0 || index >= size()) {
                     throw outOfBounds(index);
                 }
                 return AbstractImmutableList.this.get(fromIndex + index);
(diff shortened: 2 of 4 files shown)


gnodet (approved):
Clean, correct off-by-one bug fix in `SubList.get()` bounds check. The fix properly changes `index > size()` to `index >= size()` to align with the `java.util.List` contract, which specifies that `get(size())` must throw `IndexOutOfBoundsException`.

Key observations:
- All three copies of the file are fixed (template + both generated copies), preventing reintroduction on code generation
- The `listIterator(int index)` method correctly retains `index > size()` since `List.listIterator(size())` is a valid cursor position
- The four new test cases are well-chosen: they reproduce both bug variants, verify normal iteration, and pin the listIterator contract
- This is the first test class for `ImmutableCollections`, adding coverage to a previously untested area

elharo (approved):

PR #12635, opened 2026-07-30
docs: Remove broken link

elharo:
- Remove reference to maven-configuration.html which doesn't exist
- Comment out API/impl/compat navigation links with explanatory comments
- Comment out developer docs and Core ITs links

PR #12657, opened 2026-08-02
fix configuration page generation

hboutemy:
discovered while working on #12635 that lead to fixing #10961
it worked in RC5 https://maven.apache.org/ref/4.0.0-rc-5/maven-configuration.html 
but is broken in RC6 https://maven.apache.org/ref/4.0.0-rc-6/maven-configuration.html

first PR adds necessary dependencies to have the generation step work
(notice: it requires `mvn install` before `mvn site`...)

sadly, the rendering done after has issues: multi-line documentation breaks markdown table
need more rework

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -1058,14 +1058,11 @@ under the License.
                 <version>${resolverVersion}</version>
               </dependency>
               <dependency>
+                <!-- we need a dependency on maven-api-annotations to be able to load org.apache.maven.api.annotations.Config class -->
                 <groupId>org.apache.maven</groupId>
-                <artifactId>maven-core</artifactId>
-                <version>${project.version}</version>
-              </dependency>
-              <dependency>
-                <groupId>org.apache.maven</groupId>
-                <artifactId>maven-compat</artifactId>
-                <version>${project.version}</version>
+                <artifactId>maven-api-annotations</artifactId>
+                <!-- use previous version to avoid dependency cycles -->
+                <version>4.0.0-rc-6</version>
               </dependency>
             </dependencies>
             <executions>
@@ -1089,7 +1086,8 @@ under the License.
                     hence template is renamed to maven-configuration.md.vm, to avoid conflict (and picking up resolver template)
                     -->
                     <argument>--templates=maven-configuration.md</argument>
-                    <argument>${project.basedir}</argument>
+                    <!-- only look at maven-api-core, as it only contains configuration class org.apache.maven.api.Constants  -->
+                    <argument>${project.basedir}/api/maven-api-core</argument>
                     <argument>${project.build.directory}/generated-site/markdown</argument>
                   </arguments>
                 </configuration>
@@ -1113,7 +1111,7 @@ under the License.
                     to avoid double execution
                     -->
                     <argument>--templates=configuration.properties,configuration.yaml</argument>
-                    <argument>${project.basedir}</argument>
+                    <argument>${project.basedir}/api/maven-api-core</argument>
                     <argument>${project.build.directory}/generated-site/resources</argument>
                   </arguments>
                 </configuration>
--- a/pom.xml
+++ b/pom.xml
@@ -1079,14 +1079,11 @@ under the License.
                 <version>${resolverVersion}</version>
               </dependency>
               <dependency>
+                <!-- we need a dependency on maven-api-annotations to be able to load org.apache.maven.api.annotations.Config class -->
                 <groupId>org.apache.maven</groupId>
-                <artifactId>maven-core</artifactId>
-                <version>${project.version}</version>
-              </dependency>
-              <dependency>
-                <groupId>org.apache.maven</groupId>
-                <artifactId>maven-compat</artifactId>
-                <version>${project.version}</version>
+                <artifactId>maven-api-annotations</artifactId>
+                <!-- use previous version to avoid dependency cycles -->
+                <version>4.0.0-rc-6</version>
               </dependency>
             </dependencies>
             <executions>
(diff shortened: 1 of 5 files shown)


pom.xml
```diff
@@ -1078,6 +1078,16 @@ under the License.
                 <artifactId>maven-resolver-tools</artifactId>
                 <version>${resolverVersion}</version>
               </dependency>
+              <dependency>
+                <groupId>org.apache.maven</groupId>
```
gnodet, line 1084: Adding `maven-core` with `${project.version}` as a plugin dependency in the `graph` profile of the root pom introduces a reactor cycle (`maven-api-annotations → maven-api → maven → maven-core → maven-api-annotations`). Since the `graph` profile auto-activates on non-Windows, this breaks all Maven commands on a clean checkout.

Consider relocating the page generation step to a leaf module (e.g., `apache-maven`) that can depend on `maven-core` without creating a cycle.

gnodet (commented):
Thanks for tracking down the RC5→RC6 regression in the configuration page generation, @hboutemy.

The approach of adding `maven-core` and `maven-compat` as plugin dependencies in the root pom creates a reactor cycle, which is confirmed by CI:

- The `graph` profile auto-activates on non-Windows systems (via `<os><family>!windows</family></os>`), so the plugin dependency on `maven-core:${project.version}` is always visible to the reactor.
- This creates the cycle: `maven-api-annotations → maven-api (parent) → maven (root) → maven-core → maven-api-annotations`.
- The cycle prevents **all** Maven commands from succeeding on a clean checkout — including the initial `mvn install` mentioned in the PR description.

I see you already noted "need more rework" — a possible direction could be moving the configuration page generation to a submodule that sits later in the reactor order (e.g., `apache-maven`), which can legitimately depend on `maven-core` without creating a cycle.

hboutemy:
in fact I'm surprised it worked in the past, but not any more now

in addition, looking at https://maven.apache.org/ref/4.0.0-rc-5/maven-configuration.html some parameters are for Maven 4.1.0 but we're still in Maven 4.0.0

hboutemy:
@slawekjaranowski I tried to move the code to `impl/maven-cli`
but I get
```
[INFO] --- exec:3.6.3:java (render-configuration-page) @ maven-cli ---
Processing sources from /Users/hboutemy/dev/maven/core/maven-4.0.x/impl/maven-cli
java.lang.IllegalStateException: No Java sources declaring configuration keys found under /Users/hboutemy/dev/maven/core/maven-4.0.x/impl/maven-cli
        at org.eclipse.aether.tools.CollectConfiguration.runDoclet(CollectConfiguration.java:168)
```

I don't really get all this: can you have a look at it, please?
this doc is super useful, we need it back for RC7

slawekjaranowski:
Document generator class `org.eclipse.aether.tools.CollectConfiguration` was changed to use doclet in
- https://github.com/apache/maven-resolver/pull/1965

slawekjaranowski:
Issues I see in resolver-tools
-  `MAVEN_CONFIGURATION_MARKER = "@Config";` used to find file with config annotation ... but we have more `@Config` like `@Configuration` and so on ...

- multiline in configuration description brake properties files and markdown table

hboutemy:
@slawekjaranowski I switched to `xdoc` format to support multi-lines descriptions: now rendering is ok

I renamed the doc to `configuration.xdoc`/`.html` to match `configuration.properties`/`.yaml`

do you think it's ok now?

pom.xml
hboutemy: fixed by slawek by using annotation only, with a previous version as a small workaround

pom.xml
```diff
@@ -1046,12 +1046,26 @@ under the License.
...
           </plugin>
+        </plugins>
+      </build>
+    </profile>
+    <profile>
+      <id>maven-configuration</id>
+      <activation>
+        <property>
+          <name>skipConfiguration</name>
+          <value>!false</value>
```
gnodet: The `<value>!false</value>` inverts the skip semantics. With this value:
- `-DskipConfiguration=true` → profile stays **active** (wrong — should skip)
- `-DskipConfiguration=false` → profile becomes **inactive** (wrong — should not skip)

Verified against `PropertyProfileActivator.java` lines 80-89: `!false` means "activate when property value ≠ false", so `true` ≠ `false` keeps it active.

The standard Maven skip convention (`skipTests`, `maven.javadoc.skip`, etc.) expects `true` to mean "skip". The correct value is `!true`:

```suggestion
          <value>!true</value>
```

gnodet (commented):
Thanks for the rework, @hboutemy — the reactor cycle concern from the prior review is fully resolved by using a pinned previous-release dependency (`maven-api-annotations:4.0.0-rc-6`). The xdoc switch correctly handles multi-line descriptions, and narrowing the scan to `api/maven-api-core` is a good improvement.

One issue with the profile activation:

pom.xml
hboutemy: stupid me, thanks for the review

slawekjaranowski:
> @slawekjaranowski I switched to `xdoc` format to support multi-lines descriptions: now rendering is ok
> 
> I renamed the doc to `configuration.xdoc`/`.html` to match `configuration.properties`/`.yaml`
> 
> do you think it's ok now?

html version is generated ok now ...

we still have issue with properties file and multi-lines

hboutemy:
AFAIK, properties file is just there "because we can", but not really used: we can just drop it, isn't it?

slawekjaranowski:
> AFAIK, properties file is just there "because we can", but not really used: we can just drop it, isn't it?

so we can also drop yml version?

slawekjaranowski (approved):

gnodet (approved):
All blocking issues from prior reviews are resolved. The skip semantics fix (`!false` → `!true`) is correct, the xdoc switch properly handles multi-line descriptions, and dropping the broken properties template is a sound pragmatic choice with maintainer agreement.

Three minor nits (non-blocking):

1. **Stale execution ID** (`pom.xml`): `render-maven-configuration-md` still says "md" but output is now xdoc. Consider renaming to `render-maven-configuration-xdoc`.
2. **Stale execution ID** (`pom.xml`): `render-configuration-properties+yaml` still references "properties" but the properties template was dropped. Consider renaming to `render-configuration-yaml`.
3. **Typo** (`configuration.xml.vm`): "The parameters's list" — double possessive. Should be "The parameter list" or "The list of parameters".

PR #12669, opened 2026-08-03
Ignore logback >= 1.3.0 on Maven 3.x branches in dependabot.yml

slachiewicz:
Logback 1.3+ requires Java 11+ bytecode (and 1.6+ requires Java 21+), which causes  check in  to fail on Maven 3.x branches (which target Java 8 bytecode compatibility).

This PR configures  to ignore  version  for  and  target branches.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -39,6 +39,9 @@ updates:
     labels:
       - "mvn3"
       - "dependencies"
+    ignore:
+      - dependency-name: "ch.qos.logback:*"
+        versions: [ "[1.3.0,)" ]
 
   - package-ecosystem: "maven"
     directory: "/"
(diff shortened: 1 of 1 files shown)


gnodet (approved):
**LGTM** ✅ — Clean, correctly-implemented dependabot configuration change that prevents recurring, impossible logback upgrade PRs on Java 8 branches.

**Key observations:**
- Maven 3.9.x and 3.10.x target Java 8, while logback 1.3.0+ requires Java 11+ — the ignore rule is well-motivated.
- The `dependency-name: "ch.qos.logback:*"` wildcard correctly covers all logback artifacts (logback-classic, logback-core, etc.).
- The Maven version range syntax `[1.3.0,)` is the correct dependabot format for the `maven` package ecosystem.
- The second ignore entry on `maven-3.10.x` is consistent with the existing JLine ignore pattern (same format, indentation, and version range style).
- The `master` and `maven-4.0.x` branches correctly do not receive this ignore rule, as they already run logback 1.5.x/1.6.x and target newer Java versions.

PR #12674, opened 2026-08-04
Build the wrapper job with Maven 4.0.0-rc-6

slachiewicz:
4.0.0-rc-6 is released; this points the wrapper job at it.

Note: the fork has Actions disabled, so this could not be pre-verified there — it relies on ASF CI.

### DIFF
--- a/.github/ci-mimir-session.properties
+++ b/.github/ci-mimir-session.properties
@@ -19,5 +19,9 @@
 
 # do not waste time on this; we maintain the version
 mimir.daemon.autoupdate=false
-# CI uses US Mirror
-mimir.session.mirrors=central(https://maven-central.storage-download.googleapis.com/maven2/)
+# CI used the Google Cloud mirror of Central, but that mirror lags a fresh
+# release: apache-maven-4.0.0-rc-6 was still a 404 there days after Central
+# served it. Resolving from Central directly is the difference between a
+# build that works on release day and one that waits for a mirror we do not
+# control.
+# mimir.session.mirrors=central(https://maven-central.storage-download.googleapis.com/maven2/)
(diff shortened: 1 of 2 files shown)


gnodet (approved):
**LGTM** ✅ — Straightforward one-line version bump of the Maven wrapper from 4.0.0-rc-5 to 4.0.0-rc-6 in the CI workflow.

**Verified:**
- Maven 4.0.0-rc-6 is a valid, published release (confirmed on Maven Central, published 2026-07-30).
- No other files in the repository reference `4.0.0-rc-5`, so no synchronized updates are needed.
- The PR correctly targets the `master` branch, consistent with prior wrapper version bumps.

gnodet (approved):
**LGTM** ✅ — Re-reviewed after new commit. The second commit correctly removes the `MVNW_REPOURL` Google Cloud mirror override to fix the CI failure caused by mirror lag for the freshly released rc-6 distribution.

**Key observations:**
- The fix is well-targeted: only affects the wrapper download path, not Mimir dependency resolution which still uses the mirror via `ci-mimir-session.properties`.
- The wrapper download is a single-file fetch of the Maven distribution ZIP — low risk falling back to official Central.
- The commit message is excellent: includes the exact error, explains the root cause (Google mirror lag), and states the effect of the fix.
- The inline comment clearly documents the rationale for the removal.

gnodet (approved):
**LGTM** ✅ — Re-reviewed after third commit. The new commit logically completes the mirror fix by also commenting out the Mimir session mirror for dependency resolution — if the Google Cloud mirror lags for the wrapper binary, it will also lag for artifacts resolved during CI builds.

**Key observations:**
- Commenting out rather than deleting is the right approach — easy one-line rollback if the mirror sync latency is resolved.
- The explanatory comment documents the specific failure mode (rc-6 returning 404 on the mirror).
- Minor: file still lacks a trailing newline (pre-existing, not a regression).
- Central itself is served behind a CDN (Fastly), so the geographic/CDN benefit loss from removing the Google mirror is acceptable.

PR #12675, opened 2026-08-04
Build the wrapper job with Maven 4.0.0-rc-6

slachiewicz:
4.0.0-rc-6 is released; this points the wrapper job at it.

Note: the fork has Actions disabled, so this could not be pre-verified there — it relies on ASF CI.

### DIFF
--- a/.github/ci-mimir-session.properties
+++ b/.github/ci-mimir-session.properties
@@ -19,5 +19,9 @@
 
 # do not waste time on this; we maintain the version
 mimir.daemon.autoupdate=false
-# CI uses US Mirror
-mimir.session.mirrors=central(https://maven-central.storage-download.googleapis.com/maven2/)
+# CI used the Google Cloud mirror of Central, but that mirror lags a fresh
+# release: apache-maven-4.0.0-rc-6 was still a 404 there days after Central
+# served it. Resolving from Central directly is the difference between a
+# build that works on release day and one that waits for a mirror we do not
+# control.
+# mimir.session.mirrors=central(https://maven-central.storage-download.googleapis.com/maven2/)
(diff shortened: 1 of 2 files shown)


gnodet (commented):
The version bump from rc-5 to rc-6 is correct — Maven 4.0.0-rc-6 is a valid released version available on official Maven Central (confirmed HTTP 200 on `repo.maven.apache.org`).

However, **CI is currently broken** because the workflow's `MVNW_REPOURL` environment variable (line 40 of `.github/workflows/maven.yml`) directs the Maven wrapper to download from Google's Maven Central mirror (`maven-central.storage-download.googleapis.com`), which has not yet synced rc-6 (HTTP 404 confirmed). The previous version rc-5 is available on that mirror (HTTP 200 confirmed).

This sync gap has persisted for 5 days since the rc-6 release on 2026-07-30. The companion PR #12674 targeting `master` has the same CI failure.

This PR cannot pass CI until either:
- (a) the Google mirror syncs rc-6, or
- (b) the `MVNW_REPOURL` is changed to point at official Maven Central for this version.

If the Google mirror continues to lag on RC artifacts, the project may need to investigate whether this mirror is appropriate for pre-GA versions.

gnodet (approved):
**LGTM** ✅ — Re-reviewed after new commit. The second commit directly addresses the CI failure flagged in the previous review by removing `MVNW_REPOURL` so the wrapper downloads from official Central instead of the Google mirror that hadn't synced rc-6.

**Key observations:**
- The fix is identical to companion PR #12674 (master), keeping both target branches in sync.
- The Mimir session mirror for dependency resolution is correctly left unchanged — build dependencies are stable releases already synced to the mirror.
- The explanatory comment added to the workflow is clear and accurate.
- The version bump to rc-6 and the mirror fix are both correct.

gnodet (approved):
**LGTM** ✅ — Re-reviewed after third commit. The new commit logically completes the mirror fix by commenting out the Mimir session mirror for dependency resolution — since Maven is building itself, rc-6 artifacts must also be resolvable from Central.

**Key observations:**
- Commit 2 fixed the wrapper download path; commit 3 fixes the dependency resolution path. Both address the same root cause (Google Cloud mirror lagging fresh releases).
- Commenting out rather than deleting is the right approach — easy rollback.
- Minor: file still lacks a trailing newline (pre-existing).
- Companion PR #12674 (master) shows the identical diff, keeping both branches in sync.

Issue #12676, opened 2026-08-04
Upgrade to Maven 4.0.0-rc-6


slachiewicz:
Maven **4.0.0-rc-6** was released on 2026-07-30. This tracks moving the repositories that build against a Maven 4 pre-release onto it, so nothing is left pinned to rc-5 or older.

Each linked PR moves the repository's Maven 4 version property and — where the repository was relying on the `maven4-version` default in `maven-gh-actions-shared` — also pins the version explicitly in that repository's own `maven-verify.yml`, so its CI no longer moves when the shared default does.

### Shared CI default

| PR | Status / notes |
|---|---|
| apache/maven-gh-actions-shared#298 | **merged** — `maven4-version` default rc-5 → rc-6, base `v4` |
| apache/maven-gh-actions-shared#299 | **merged** — same change, base `v5` |

### Maven core

| PR | Status / notes |
|---|---|
| apache/maven#12674 | **merged** — wrapper job → rc-6 (`master`), plus commenting out the Google mirror |
| apache/maven#12675 | **merged** — wrapper job → rc-6 (`maven-4.0.x`), plus commenting out the Google mirror |
| apache/maven-mvnd#1714 | **merged** — `maven.version` + wrapper `distributionUrl`, plus the resolver 2.0.21 / slf4j 2.0.18 / JLine 4.3.1 sync and the `jline-terminal-jansi` → `jansi-core` rename. The GraalVM native jobs fail on `master` too (`stat@GLIBC_2.17`), so that red is not from this change |
| apache/maven-resolver#2027 | **merged** — `maven4Version` |
| apache/maven-build-cache-extension#520 | **merged** — `maven4.0` profile |
| apache/maven-executor#35 | **merged** — `version.maven4` + CI pin |

### Plugins and shared components

| PR | Status | Was on |
|---|---|---|
| apache/maven-clean-plugin#329 | **merged** | rc-5 |
| apache/maven-deploy-plugin#685 | **merged** | rc-5 |
| apache/maven-install-plugin#429 | **merged** | rc-5 |
| apache/maven-archiver#390 | **merged** | rc-5 |
| apache/maven-filtering#367 | **merged** | rc-5 |
| apache/maven-compiler-plugin#1104 | **merged** | rc-4 |
| apache/maven-jar-plugin#573 | **merged** | rc-4 — same change as my #574, which is closed as superseded |
| apache/maven-resources-plugin#500 | **merged** | rc-4 |
[...]

PR #12677, opened 2026-08-04
Build and deploy through the wrapper on Jenkins

slachiewicz:
Jenkins already provisions Maven with the wrapper for the build itself, but the deploy step still ran the `mvn` from the `maven_3_latest` Jenkins tool — so a release was published by a different Maven than the one that built it. `mavenBuild` runs first in the same workspace, so `./mvnw` is present by the time deploy runs.

Also moves the wrapper provisioning to `maven-wrapper-plugin` 3.3.4 and Maven 3.9.12, matching what `.github/workflows/maven.yml` already uses. Both are on Central (verified).

Same pattern as apache/maven-jenkins-lib#22, which applies it to the shared library. Part of apache/maven#12676.

No `master` companion is needed: `master` already has both of these — `./mvnw clean deploy -DdeployAtEnd=true -B -V` at `Jenkinsfile:41` and `maven-wrapper-plugin:3.3.4` with `-Dmaven=3.9.12` at `:70`. This brings `maven-4.0.x` up to the same state.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -38,7 +38,7 @@ pipeline {
               withEnv(["JAVA_HOME=${tool "jdk_17_latest"}",
                        "PATH+MAVEN=${ tool "jdk_17_latest" }/bin:${tool "maven_3_latest"}/bin",
                        "MAVEN_OPTS=-Xms4G -Xmx4G -Djava.awt.headless=true"]) {
-                sh "mvn clean deploy -DdeployAtEnd=true -B"
+                sh "./mvnw clean deploy -DdeployAtEnd=true -B -V"
               }
             }
           }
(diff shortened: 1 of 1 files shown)


gnodet (approved):
**LGTM** ✅ — Clean, low-risk change that aligns the Jenkins deploy step with the build step by using the Maven wrapper consistently.

**Key observations:**
- The build step already provisions the wrapper via `maven-wrapper-plugin`, so `./mvnw` is guaranteed to be present when the deploy step runs.
- The version bumps (wrapper-plugin 3.3.2→3.3.4, Maven 3.9.10→3.9.12) match what `origin/master` and the GHA workflow already use.
- Previously the build used `./mvnw` (Maven 3.9.x) but the deploy used `mvn` from the `maven_3_latest` Jenkins tool — potentially a different version. This PR fixes that inconsistency.
- The added `-V` flag improves version traceability in deploy logs.
- Note: `origin/master` already has identical changes (commits `a9fda7d28f` and `e25ec15e1b`), so a companion PR for master is not needed.

PR #12680, opened 2026-08-05
Fix filterByScope to expand lifecycle scopes for build ordering

gnodet:
## Summary

The concurrent builder's `filterByScope` used exact string matching between the lifecycle dependency scope (e.g. `"compile"`) and the declared artifact scope in the model. This meant a `provided`-scope reactor dependency was **not** ordered before the consumer's compile phase, since `"compile" != "provided"`.

### Root Cause

The lifecycle declares `dependencies("compile", READY)` for the compile phase, meaning: wait for dependencies with scope `"compile"` to reach the `ready` phase before starting compilation. But `filterByScope` compared this scope string directly against each dependency's declared scope (`"provided"`, `"system"`, etc.), missing all non-compile scopes that still contribute to the compile classpath.

### Fix

Expand the lifecycle scope to match all artifact scopes that contribute to it for build ordering:

- `"compile"` → compile, provided, system (+ null defaults to compile)
- `"runtime"` → compile, runtime (+ null)
- `"test"` → all scopes
- `"test-only"` → test only

This is implemented as a private `expandScope()` method directly in `BuildPlanExecutor`, keeping the mapping local to where it's needed.

### Testing

Updated `BuildPlanCreatorTest.testFilterByScopeMatchesExact` to cover the expanded scope matching.

### DIFF
--- a/impl/maven-core/src/main/java/org/apache/maven/lifecycle/internal/LifecycleDependencyResolver.java
+++ b/impl/maven-core/src/main/java/org/apache/maven/lifecycle/internal/LifecycleDependencyResolver.java
@@ -130,6 +130,10 @@ public class LifecycleDependencyResolver {
 
             Map<Artifact, File> reactorProjects =
                     new HashMap<>(session.getProjects().size());
+            // Build a GAV-keyed map for fallback lookups when exact Artifact.equals() misses
+            // (e.g. type/classifier mismatch between resolved dependency and reactor artifact).
+            Map<String, MavenProject> reactorGavMap =
+                    new HashMap<>(session.getProjects().size());
             for (MavenProject reactorProject : session.getProjects()) {
                 File file = reactorProject.getArtifact().getFile();
                 // In the concurrent builder, a reactor project may have compiled (output directory
@@ -142,6 +146,8 @@ public class LifecycleDependencyResolver {
                     }
                 }
                 reactorProjects.put(reactorProject.getArtifact(), file);
+                String gavKey = ArtifactUtils.key(reactorProject.getArtifact());
+                reactorGavMap.put(gavKey, reactorProject);
             }
 
             Map<String, Artifact> map = new HashMap<>();
@@ -154,6 +160,38 @@ public class LifecycleDependencyResolver {
                 File reactorProjectFile = reactorProjects.get(artifact);
                 if (reactorProjectFile != null) {
                     artifact.setFile(reactorProjectFile);
+                } else if (artifact.getFile() == null) {
+                    // Fallback: try matching reactor projects by GAV (handles type/classifier
+                    // mismatches) or look up the artifact in the local repository.
+                    String gavKey = ArtifactUtils.key(artifact);
+                    MavenProject reactorProject = reactorGavMap.get(gavKey);
+                    if (reactorProject != null) {
+                        File fallback = reactorProject.getArtifact().getFile();
+                        if (fallback == null) {
+                            File outputDir = new File(reactorProject.getBuild().getOutputDirectory());
+                            if (outputDir.isDirectory()) {
+                                fallback = outputDir;
+                            }
+                        }
+                        if (fallback != null) {
+                            artifact.setFile(fallback);
+                        }
+                    } else {
+                        // Non-reactor artifact with null file — the Aether resolver may not have
+                        // resolved it (e.g. it was only collected, or a race condition in the
+                        // concurrent builder prevented the file from being set).  Try looking
+                        // up the file in the local repository.
+                        org.eclipse.aether.artifact.Artifact aetherArtifact = RepositoryUtils.toArtifact(artifact);
+                        org.eclipse.aether.repository.LocalRepositoryManager lrm =
+                                session.getRepositorySession().getLocalRepositoryManager();
+                        if (lrm != null) {
+                            String path = lrm.getPathForLocalArtifact(aetherArtifact);
+                            File localFile = new File(lrm.getRepository().getBasedir(), path);
+                            if (localFile.isFile()) {
+                                artifact.setFile(localFile);
+                            }
+                        }
+                    }
                 }
 
                 map.put(artifact.getDependencyConflictId(), artifact);
--- a/impl/maven-core/src/main/java/org/apache/maven/lifecycle/internal/LifecycleDependencyResolver.java
+++ b/impl/maven-core/src/main/java/org/apache/maven/lifecycle/internal/LifecycleDependencyResolver.java
@@ -131,9 +131,17 @@ public class LifecycleDependencyResolver {
             Map<Artifact, File> reactorProjects =
                     new HashMap<>(session.getProjects().size());
             for (MavenProject reactorProject : session.getProjects()) {
-                reactorProjects.put(
-                        reactorProject.getArtifact(),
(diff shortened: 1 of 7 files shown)


impl/maven-core/src/main/java/org/apache/maven/lifecycle/internal/LifecycleDependencyResolver.java
```diff
@@ -146,6 +160,38 @@ public void resolveProjectDependencies(
...
                 if (reactorProjectFile != null) {
                     artifact.setFile(reactorProjectFile);
+                } else if (artifact.getFile() == null) {
+                    // Fallback: try matching reactor projects by GAV (handles type/classifier
+                    // mismatches) or look up the artifact in the local repository.
+                    String gavKey = ArtifactUtils.key(artifact);
+                    MavenProject reactorProject = reactorGavMap.get(gavKey);
+                    if (reactorProject != null) {
+                        File fallback = reactorProject.getArtifact().getFile();
+                        if (fallback == null) {
```
gnodet: **Low:** This output-directory fallback block duplicates the identical logic at lines 139-145 above. Consider extracting a shared helper (e.g. `resolveReactorProjectFile(MavenProject)`) to avoid maintaining two copies.

Also note: `ArtifactUtils.key()` produces `groupId:artifactId:version` only — a tests-jar dependency (`classifier="tests"`) would match the main project and receive `target/classes` instead of `target/test-classes`. Worth documenting as a known limitation in the "handles type/classifier mismatches" comment above.

gnodet (commented):
Well-targeted fix for real concurrency issues in the concurrent builder. The scope expansion fix for provided-scope reactor dependencies and the multi-layer artifact resolution fallback are solid. Three observations after verification (2 FPs filtered out):

1. **Duplicated output directory fallback** — The pattern of checking `artifact.getFile()` for null → creating `File(getOutputDirectory())` → testing `isDirectory()` appears identically at two locations (lines ~139-145 and ~169-173). Consider extracting a `resolveReactorProjectFile(MavenProject)` helper to avoid keeping both copies in sync.

2. **GAV fallback ignores classifier/type** — `ArtifactUtils.key()` produces `groupId:artifactId:version` only. If a reactor project publishes both a main JAR and a tests-jar (`classifier="tests"`), the tests-jar dependency would match the main project and receive `target/classes` instead of `target/test-classes`. This is arguably better than `null` (which is the alternative), but worth documenting as a known limitation in the comment.

3. **Missing test coverage for `expandScope()`** — The test covers `"compile"`, `"runtime"`, and `"test"` scope expansion but not `"compile+runtime"`, `"runtime+system"`, `"test-only"`, or the default case.

impl/maven-core/src/main/java/org/apache/maven/lifecycle/internal/LifecycleDependencyResolver.java
gnodet: Fixed in aa5af679f2 — extracted `resolveReactorProjectFile(MavenProject)` helper and documented the GAV-only matching limitation (classifier/type ignored) in the fallback comment.

gnodet (commented):
**Delta re-review** — All three findings from the previous review were addressed by commit `aa5af679`:

1. ✅ **Duplicated output directory fallback** → Extracted `resolveReactorProjectFile(MavenProject)` helper with clear Javadoc. Both call sites now use the shared method.

2. ✅ **GAV fallback classifier/type limitation** → Comment rewritten to explicitly document the known limitation (tests-jar matching the main project is better than null/build failure).

3. ✅ **Missing `expandScope()` test coverage** → Test expanded to cover all six branches (`compile+runtime`, `runtime+system`, `test-only`, default/unknown) with correct assertions.

Delta commit is tightly scoped to exactly the three review findings. LGTM.

impl/maven-core/src/main/java/org/apache/maven/lifecycle/internal/MojoExecutor.java
```diff
@@ -130,29 +129,35 @@ private void collectDependencyRequirements(
...
-            } else if (Artifact.SCOPE_RUNTIME_PLUS_SYSTEM.equals(classpath)) {
-                scopes = Arrays.asList(Artifact.SCOPE_COMPILE, Artifact.SCOPE_SYSTEM, Artifact.SCOPE_RUNTIME);
-            } else if (Artifact.SCOPE_TEST.equals(classpath)) {
-                scopes = Arrays.asList(
+    /**
+     * Expands a dependency scope identifier to the set of artifact scopes included
+     * in dependency resolution for that scope.
+     *
+     * @param classpath the scope identifier (e.g. "compile", "runtime", "test"), may be {@code null}
+     * @return an unmodifiable set of matching artifact scopes, empty if {@code classpath} is null/empty/unknown
```
gnodet: **Low — Javadoc `@return` is inaccurate for the `default` case.** The doc says *"empty if {@code classpath} is null/empty/unknown"* but the implementation returns `Set.of(classpath)` for unknown values (the `default` branch). This matters for `filterByScope()` callers: a plugin-declared lifecycle dependency with an unrecognized scope string (e.g. `"provided"` passed directly as a scope identifier) will now produce a singleton `{"provided"}` set instead of empty — a behavior change from the old code. The behavior is probably correct (pass the scope through and let the filter match), but the doc actively misleads.

```suggestion
     * @return an unmodifiable set of matching artifact scopes; for unknown/unrecognized scope identifiers,
     *         returns a singleton set containing the scope itself; empty only if {@code classpath} is null or empty
```

gnodet (commented):
**Delta re-review** — reviewing 4 new commits since the last review (`cc6ecd6388`, `b840c0d896`, `e1d21ad32d`, `b73c380d06`).

### Commit-by-commit assessment

**`cc6ecd6388` — Remove redundant ThreadLocal, fix after:* step semantics** ✅

Correct. `session.getCurrentProject()` is a `ThreadLocal` (confirmed in `MavenSession.java:65`) and `attachToThread()` calls `session.setCurrentProject(step.project)` before mojo execution. The previous `THREAD_PROJECT` ThreadLocal in the concurrent `MojoExecutor` was therefore redundant and correctly removed. The after:* semantic fix (checking `BEFORE + phaseName` instead of the phase step itself) is also right: `before:compile` is the proper entry signal since the phase step itself may have zero mojos and never transition to `EXECUTED`.

The `addLifecyclePhase()` belt-and-suspenders call that remains is valid: `PhaseRecorder.observeExecution()` only records a phase when `mojoExecution.getLifecyclePhase() != null`, so a step whose mojos don't declare a lifecycle phase would not be recorded by the `PhaseRecorder` alone. The explicit call covers that gap.

**`b840c0d896` — Deduplicate scope expansion, promote `toScopes` to `public static`** — one issue (see inline)

**`e1d21ad32d` — Remove SOURCES→RESOURCES sequential ordering constraint** ✅

Clean revert. The constraint is properly moved to plugin-level `@After` declarations in V4-native plugins.

**`b73c380d06` — Revert `session.getCurrentProject()` change in `MojoExecutor`** ✅

Correct — the previous commit already proved `getCurrentProject()` is ThreadLocal. Revert is consistent.

---

One finding on `toScopes()` Javadoc accuracy (see inline comment).

_This review was generated by an AI agent, Hermès on behalf of @gnodet._

rmannibucau (approved):

PR #12683, opened 2026-08-06
Fix deprecated Maven testing API compatibility

### LINKED ISSUE #12678: maven-testing: deprecated org.apache.maven.api.plugin.testing annotations compile but are silently ignored
The `org.apache.maven.api.plugin.testing` package was relocated to `org.apache.maven.testing.plugin` in 4.0.0-rc-6 (`8857e0eb18`, "Deprecate testing classes in org.apache.maven.api, add new ones in org.apache.maven.testing"). Deprecated classes remain under the old names, so plugins keep compiling — but `MojoExtension` only recognises the **new** annotation types, so a plugin that has not moved its imports silently stops working.

`findInjectMojoAttrs` returns null for the deprecated `@InjectMojo`, so the `@Basedir` pom is never read and the mojo is handed `MojoExtension`'s built-in `defaultModel` instead of the test project. Nothing warns; the test simply runs against the wrong model.

In maven-jar-plugin this surfaced as:

```
JarMojoTest.jarTestEnvironment:48 expected: <foo> but was: <myGroupId>
```

`foo` is the groupId in the `@Basedir` test pom; `myGroupId` is `MojoExtension`'s default. Bisected: passes on 4.0.0-rc-4 and 4.0.0-rc-5, fails on rc-6. Pointing the imports at `org.apache.maven.testing.plugin` fixes it (apache/maven-jar-plugin#573).

In maven-source-plugin the same cause made `@Basedir` and `@MojoParameter` be ignored, producing `target/null-sources.jar` while the tests still reported green.

There is a second, related break in the same relocation. The shim left behind for `ProducedArtifactStub`:

```java
@Deprecated(since = "4.0.0-rc-6", forRemoval = true)
public class ProducedArtifactStub extends org.apache.maven.testing.plugin.stubs.ProducedArtifactStub {
    public ProducedArtifactStub();
}
```

is a bare no-arg subclass. Constructors are not inherited, so the five-argument constructor the old class had is gone, and code using it no longer compiles:

```
CompilerMojoTestCase.java:[436,21] constructor ProducedArtifactStub ... cannot be applied to given types
  required: no arguments   found: String,String,<nulltype>,String,String
```

That one at least fails loudly (apache/maven-compiler-plugin#1104).

Suggestions, in rough order of value:

1. Make the deprecated annotations work, or fail loudly. If `MojoExtension` accepted both annotation types the relocation would be transparent; if that is not wanted, detecting a deprecated `@InjectMojo`/`@Basedir` and failing with a message naming the new package would at least turn a silent wrong-model run into an actionable error.
2. Give the `ProducedArtifactStub` shim the same five-argument constructor, so the deprecated class is source-compatible for its remaining life.
3. Mention both in the release notes: "deprecated but still compiles, and no longer functions" is a combination that costs downstream projects a debugging session each.

Found while moving the plugins to rc-6 (apache/maven#12676).

goutamadwant:
## Summary

- recognize the deprecated `org.apache.maven.api.plugin.testing.Basedir` annotation while preserving precedence for the replacement annotation
- restore the five-argument constructor on the deprecated `ProducedArtifactStub` compatibility class
- add regression coverage for the deprecated annotation flow and constructor source compatibility

This keeps the deprecated testing API functional for its remaining lifetime instead of allowing existing plugin tests to compile but silently use the wrong base directory.

Fixes #12678.

## Verification

- `mvn --batch-mode verify`
- `mvn --batch-mode -pl impl/maven-testing -am verify`
- `mvn --batch-mode -pl impl/maven-testing -am -rf :maven-testing -Dtest=ExpressionEvaluatorTest,ProducedArtifactStubTest -Dsurefire.failIfNoSpecifiedTests=false test`

### DIFF
--- a/impl/maven-testing/src/main/java/org/apache/maven/testing/plugin/MojoExtension.java
+++ b/impl/maven-testing/src/main/java/org/apache/maven/testing/plugin/MojoExtension.java
@@ -315,9 +315,10 @@ public class MojoExtension extends MavenDIExtension implements ParameterResolver
 
         String pluginBasedir = MavenDIExtension.getBasedir();
 
-        String basedir = AnnotationSupport.findAnnotation(context.getElement().orElseThrow(), Basedir.class)
-                .map(Basedir::value)
-                .orElse(pluginBasedir);
+        String basedir = findBasedirValue(context.getElement().orElseThrow());
+        if (basedir == null) {
+            basedir = pluginBasedir;
+        }
         if (basedir.isEmpty()) {
             basedir = pluginBasedir + "/target/tests/"
                     + context.getRequiredTestClass().getSimpleName() + "/"
@@ -744,6 +745,18 @@ public class MojoExtension extends MavenDIExtension implements ParameterResolver
         requireNonNull(field, "Field " + variable + " not found");
         field.setAccessible(true);
         field.set(object, value);
+    }
+
+    @SuppressWarnings("deprecation")
+    private static String findBasedirValue(AnnotatedElement element) {
+        Basedir a = AnnotationSupport.findAnnotation(element, Basedir.class).orElse(null);
+        if (a != null) {
+            return a.value();
+        }
+        org.apache.maven.api.plugin.testing.Basedir d = AnnotationSupport.findAnnotation(
+                        element, org.apache.maven.api.plugin.testing.Basedir.class)
+                .orElse(null);
+        return d != null ? d.value() : null;
     }
 
     @SuppressWarnings("deprecation")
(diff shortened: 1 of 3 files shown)


gnodet (approved):
Correct, well-scoped fix for issue #12678. The `findBasedirValue` method follows the exact pattern already established by `findInjectMojoAttrs` and `isRealSession` for handling deprecated annotation fallback. The `ProducedArtifactStub` constructor restoration is also correct. Both changes have targeted regression tests.

**Minor note (out of scope for this PR):** The same missing-constructor issue exists in the deprecated `ArtifactStub` shim and in the deprecated `SessionStub` shim — these represent the same class of bug but are separate fixes.

gnodet (approved):
**Review: Fix deprecated Maven testing API compatibility**

Clean, well-scoped compatibility fix that restores deprecated annotation handling for `@Basedir` and adds the missing 5-arg constructor to the deprecated `ProducedArtifactStub` shim. Follows the exact pattern already used for `@InjectMojo`, `@MojoTest`, and `@MojoParameter` deprecated fallbacks. All CI checks pass.

The new `findBasedirValue` method correctly follows the established precedence pattern: check new annotation first, fall back to deprecated annotation, return null/default if neither is found. Test coverage is good — `testDeprecatedAnnotations` exercises all three deprecated annotations together.

---

PR #12684, opened 2026-08-06
[MNG-8709] Use active profile properties for consumer POM validation

### LINKED ISSUE #10379: [MNG-8709] Maven 4.0 erroneously reports it cannot resolve a dependency version from a property
**[Jack Green](https://issues.apache.org/jira/secure/ViewProfile.jspa?name=JIRAUSER307283)** opened **[MNG-8709](https://issues.apache.org/jira/browse/MNG-8709?redirect=false)** and commented

Take a `pom.xml` where the version of a dependency is set via a property within a profile:

 

```java
<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">    <modelVersion>4.0.0</modelVersion>    <groupId>com.me</groupId>    <artifactId>maven-4-test</artifactId>    <version>0.0.1-SNAPSHOT</version>    <profiles>        <profile>            <id>spring-6</id>            <activation>                <activeByDefault>true</activeByDefault>            </activation>            <properties>                <org.springframework.version>6.2.0</org.springframework.version>            </properties>        </profile>    </profiles>    <dependencies>        <dependency>            <groupId>org.springframework</groupId>            <artifactId>spring-context</artifactId>            <version>${org.springframework.version}</version>            <scope>test</scope>        </dependency>    </dependencies></project>
```

This works fine in `{}Maven 3.9.9 (8e8579a9e76f7d015ee5ec7bfcdc97d260186937){`}, but in `{}Maven 4.0.0-rc-3 (3952d00ce65df6753b63a51e86b1f626c55a8df2){`}, `clean install` fails with:

 

 

```java
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-install-plugin:3.1.3:install (default-install) on project maven-4-test: Execution default-install of goal org.apache.maven.plugins:maven-install-plugin:3.1.3:install failed: org.apache.maven.api.services.ModelBuilderException: 1 problem was  for com.me:maven-4-test:jar:0.0.1-SNAPSHOT
[ERROR]     - [ERROR] 'dependencies.dependency.version' for org.springframework:spring-context:jar must be a valid version but is '${org.springframework.version}'. @ line 24, column 13
```

 

 

I believe this to be a bug because if you query other ways it shows the version _is_ being resolved:

 

`{}help:evaluate -Dexpression=org.springframework.version{`}:

```java
6.2.0 
```

`{}help:effective-pom{`}:

```java
  <properties>
    {...}
    <project.reporting.outputEncoding>UTF-8</project.reporting.outputEncoding>
  </properties>
  <dependencies>
    <dependency>
      <groupId>org.springframework</groupId>
      <artifactId>spring-context</artifactId>
      <version>6.2.0</version>
      <scope>test</scope>
    </dependency>
  </dependencies> 
```

 

 

---

**Affects:** 4.0.0-rc-3

**Attachments:**
- [pom.xml](https://issues.apache.org/jira/secure/attachment/13076278/pom.xml) (_990 bytes_)

goutamadwant:
Fixes #10379.

When Maven rebuilds a consumer POM, `BUILD_CONSUMER` does not reactivate the profiles that were active while building the project. As a result, model validation can see unresolved dependency versions when their values come from an active profile.

This change makes properties from the project's already-active profiles available while rebuilding the consumer model. Session user properties are applied afterward so their existing precedence is preserved. The change does not activate additional profiles or replace profile-backed expressions in the generated consumer POM.

A Core IT covers an `activeByDefault` profile that supplies a dependency version and verifies that installation succeeds while the generated consumer POM retains both the profile and the property expression.

### Tests

- `mvn -pl impl/maven-core -Dtest=ConsumerPomBuilderTest test`
- `mvn -Prun-its -DmavenHome=<maven-under-test> -Dtest=MavenITmng8709ProfileDependencyVersionTest -Dsurefire.failIfNoSpecifiedTests=false verify`
- `mvn -Prun-its -DmavenHome=<maven-under-test> verify -rf :core-it-suite` on JDK 21
  - 1,037 tests ran and the MNG-8709 regression passed. One unrelated existing test failed because it expected the legacy Central URL `https://repo1.maven.org` while the local environment used `https://repo.maven.apache.org`.

### DIFF
--- a/impl/maven-core/src/main/java/org/apache/maven/internal/transformation/impl/DefaultConsumerPomBuilder.java
+++ b/impl/maven-core/src/main/java/org/apache/maven/internal/transformation/impl/DefaultConsumerPomBuilder.java
@@ -265,7 +265,16 @@ class DefaultConsumerPomBuilder implements PomBuilder {
         request.source(src);
         request.locationTracking(false);
         request.systemProperties(iSession.getSystemProperties());
-        request.userProperties(iSession.getUserProperties());
+        Map<String, String> userProperties = new LinkedHashMap<>();
+        // BUILD_CONSUMER does not reactivate project profiles, so expose properties from profiles
+        // that were already active when the project model was built.
+        if (project != null && project.getActiveProfiles() != null) {
+            for (org.apache.maven.model.Profile profile : project.getActiveProfiles()) {
+                userProperties.putAll(profile.getDelegate().getProperties());
+            }
+        }
+        userProperties.putAll(iSession.getUserProperties());
+        request.userProperties(userProperties);
         request.lifecycleBindingsInjector(lifecycleBindingsInjector::injectLifecycleBindings);
         // Pass remote repositories so that the model builder can resolve BOM imports
         // from non-central repositories (e.g., repositories defined in settings.xml profiles).
--- a/impl/maven-core/src/test/java/org/apache/maven/internal/transformation/impl/ConsumerPomBuilderTest.java
+++ b/impl/maven-core/src/test/java/org/apache/maven/internal/transformation/impl/ConsumerPomBuilderTest.java
@@ -24,6 +24,7 @@ import java.nio.file.Path;
 import java.nio.file.Paths;
 import java.util.ArrayList;
 import java.util.List;
+import java.util.Map;
 
 import org.apache.maven.api.DependencyCoordinates;
 import org.apache.maven.api.Node;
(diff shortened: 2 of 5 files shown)


gnodet (approved):
The code change is correct and well-motivated. The fix properly injects active profile properties as user properties during consumer POM rebuild, with correct precedence (session user properties override via `putAll` afterward). The inline comment at the injection site clearly documents the intent.

The integration test (`MavenITmng8709ProfileDependencyVersionTest`) covers the end-to-end scenario well — verifying both that `install` succeeds and that the consumer POM retains the profile definition with the property expression.

**Non-blocking suggestion:** Consider adding a unit test in `ConsumerPomBuilderTest` following the existing spy pattern from `testConsumerPomPassesProjectRepositoriesToModelBuilder` — set active profiles with properties on the `MavenProject` and verify the captured `BUILD_CONSUMER` request's `userProperties` contain the profile properties. This would provide faster feedback alongside the integration test.

gnodet (approved):
**Review: [MNG-8709] Use active profile properties for consumer POM validation**

Correct and well-scoped fix that makes active profile properties available during consumer POM validation, resolving the false "must be a valid version" error when dependency versions come from profile-defined properties.

The property precedence is handled correctly: profile properties are loaded first, then session user properties override via `putAll`, preserving the standard Maven precedence where `-D` flags win over profile-defined properties. The integration test adequately proves the fix.

1. **[low]** The null check `project.getActiveProfiles() != null` is technically unnecessary since `MavenProject.activeProfiles` is initialized to `new ArrayList<>()`, but acceptable as defensive programming.

2. **[low]** The integration test could additionally assert the presence of the actual property value (`<junit.version>5.11.0</junit.version>`) within the consumer POM for more descriptive coverage.

---

gnodet (approved):
**Clean fix for profile-based dependency version validation (MNG-8709)**

Well-scoped fix that injects already-active profile properties as user properties during consumer POM model building, resolving false validation errors when dependency versions reference profile-defined properties.

Key observations:
- Correct property precedence: session user properties are applied last, preserving their higher priority over profile properties
- Proper null safety throughout (`project != null`, `@Nonnull` on `getProperties()`)
- Fully-qualified `org.apache.maven.model.Profile` correctly avoids conflict with the already-imported API `Profile` class
- Integration test covers the common scenario (activeByDefault profile supplying a dependency version) and verifies consumer POM retains the profile and property expression

### Minor observation

**Test version range** — The constructor uses `[4.0.0-rc-6,)` but this fix targets milestone `4.0.0-rc-7`. Since rc-6 is already released without this fix, running the test against rc-6 would fail. Consider `[4.0.0-rc-7,)` to match the release where the fix lands.

---

PR #12692, opened 2026-08-07
Upgrade the IT suite to Jetty 12.1.12 and the core Jetty API

slachiewicz:
The core IT suite still runs its test HTTP server on Jetty 9.4, which has been end-of-life since 2022 and whose last release on Central is `9.4.58.v20250814`. This moves it to Jetty 12.1.12 and the Jetty **core** API, which drops the servlet API from the IT suite entirely.

This supersedes #11730 — the migration commit is @olamy's, rebased onto current master and carried forward here with the integration test failures fixed. @olamy, please shout if you would rather drive this yourself and I will close this in favour of yours.

### What changed relative to #11730

* **Rebased onto master.** #11730 branched before bfc30f1 ("Migrate integration tests to NIO2 Path API"), so all 37 files conflicted. Resolutions keep master's `Path` API and Jetty 12's resource API: `setResourceBase(String)` becomes `setBaseResource(ResourceFactory.of(server).newResource(Path))`.
* **Jetty 12.1.6 → 12.1.12**, matching the version `maven-resolver` already uses.
* **MNG-2387**: `proxyServer = new Server(0)` was assigned twice, discarding the instance the `ResourceFactory` had been bound to.

### The two integration test failures

Both are genuine Jetty 12 behaviour changes, not migration slips. Each was confirmed by running the same test against unmodified master, where both pass.

* **MNG-4360 (WebDAV)** — the wagons under test request `//org/apache/...` with an empty leading path segment. Jetty 9 served it; Jetty 12 rejects it as ambiguous with `400 Bad Request`. The connector now uses `UriCompliance.LEGACY`, the only named mode other than `UNSAFE` that permits `AMBIGUOUS_EMPTY_SEGMENT` (`JETTY_11` does not include it).
* **MNG-4554 (`testitRefetched`)** — Jetty 12's `ResourceHandler` caches file content by default via `ValidatingCachingHttpContentFactory`. The test swaps the served repository between invocations and was answered from the stale cache, so the `b` plugin prefix was never seen. The handler now serves straight off disk via `ResourceHttpContentFactory`.

### Verification

All 35 Jetty-touching IT classes were run locally on macOS / JDK 25: **55 tests, 1 error**.

The remaining error is `MavenITmng4428FollowHttpRedirectTest.testitHttpsToHttps`, which fails identically on unmodified master with `PKIX path building failed` — a local trust store issue, not a Jetty regression. CI should confirm it passes there.


gnodet (approved):
Well-executed migration from Jetty 9.4 (EOL) to Jetty 12.1.12 core API. All 37 files consistently apply the new `Handler.Abstract` pattern, correctly wire security handlers as wrappers rather than siblings, and properly address two documented Jetty 12 behavioral changes (WebDAV ambiguous URIs and `ResourceHandler` caching).

A few non-blocking observations:

- **Repeated `Request.getPathInContext(request)` calls**: In `MavenITmng4360WebDavSupportTest`, `MavenITmng3599useHttpProxyForWebDAVMk2Test`, and `MavenITmng4348NoUnnecessaryRepositoryAccessTest`, the path is fetched multiple times without assigning to a local variable, unlike the pattern used in other handlers in this PR.

- **Inline fully-qualified `StandardCopyOption.REPLACE_EXISTING`** in `MavenITmng4235HttpAuthDeploymentChecksumsTest` (line ~188) — the old code had a static import that was dropped.

- **`Constraint.Authorization.ANY_USER` vs role-based check** in `HttpServer.java` (line ~116): All security constraints changed from role-based (`setRoles(new String[]{"user","admin"})`) to `ANY_USER`. Functionally equivalent for these tests since each defines exactly one user, but semantically less strict.

None of these are blocking — the migration is clean and idiomatic for Jetty 12.

slachiewicz:
Thanks — all three points addressed in 783450bd7a.

The `ANY_USER` one turned out to be broader than it looked: it is 11 sites, not one. Every Jetty 9 constraint in the suite declared explicit roles, and the port flattened all of them, so each is now mapped back to `SPECIFIC_ROLE` with the roles it had before — `{"user", "admin"}` for `HttpServer`, `"user"` or `"deployer"` for the ITs. Worth noting MNG-553 expressed its role through the two-arg `Constraint(__BASIC_AUTH, "user")` constructor rather than `setRoles`, so it does not show up in a `setRoles` grep.

The 16 auth and handler ITs touched by these changes pass locally: 30 tests, 0 failures.

Also opened #12693 to backport this to `maven-4.0.x`.

olamy (approved):

PR #12693, opened 2026-08-07
[maven-4.0.x] Upgrade the IT suite to Jetty 12.1.12 and the core Jetty API

slachiewicz:
Backport of #12692 to `maven-4.0.x`. The core IT suite still runs its test HTTP server on Jetty 9.4, end-of-life since 2022 and last released on Central as `9.4.58.v20250814`. This moves it to Jetty 12.1.12 and the Jetty **core** API, which drops the servlet API from the IT suite entirely.

The migration commit is @olamy's, from #11730, cherry-picked here.

### Differences from the master PR

The 4.0.x IT suite predates the NIO2 `Path` migration, so it cherry-picked far more cleanly — 7 conflicts, all the same one: master's handlers write `<project xmlns="http://maven.apache.org/POM/4.0.0">` while 4.0.x writes a bare `<project>`. 4.0.x's form is kept throughout.

Otherwise identical to #12692, including:

* **Jetty 12.1.6 → 12.1.12**, matching the version `maven-resolver` already uses.
* **MNG-4360 (WebDAV)** — the wagons under test request `//org/apache/...` with an empty leading path segment. Jetty 9 served it; Jetty 12 rejects it as ambiguous with `400`. The connector now uses `UriCompliance.LEGACY`, the only named mode other than `UNSAFE` that permits `AMBIGUOUS_EMPTY_SEGMENT`.
* **MNG-4554 (`testitRefetched`)** — Jetty 12's `ResourceHandler` caches file content by default. The test swaps the served repository between invocations and was answered from a stale cache. It now serves straight off disk via `ResourceHttpContentFactory`.
* **MNG-2387** — `proxyServer = new Server(0)` was assigned twice, discarding the instance the `ResourceFactory` had been bound to.
* The review feedback from #12692: role-based BASIC auth constraints restored at all 11 sites (the port had flattened them to `Authorization.ANY_USER`), repeated `Request.getPathInContext(request)` calls hoisted into locals, and the dropped `import static ... REPLACE_EXISTING` restored.

### Verification

The full IT suite was run locally on macOS / JDK 25, both on this branch and on unmodified `maven-4.0.x` as a control. Both runs give **1036 tests with the same 14 failures in the same 10 classes** — no regressions from the migration. Those 14 are local environment failures (trust store, import scope, CI-friendly versions, resolver transport) and are unrelated to Jetty; none of the 10 classes reference Jetty except `MavenITmng4428FollowHttpRedirectTest`, which fails on the baseline too.


gnodet (approved):
Clean backport of #12692 to maven-4.0.x. All three review feedback fixes from #12692 are correctly included:

1. **SPECIFIC_ROLE**: All 11 security constraint sites use `Constraint.Authorization.SPECIFIC_ROLE` with explicit roles instead of `ANY_USER`
2. **Local variables**: `Request.getPathInContext(request)` calls hoisted to local variables in the originally flagged files
3. **Static import**: `import static java.nio.file.StandardCopyOption.REPLACE_EXISTING` restored

The 4.0.x-specific conflict resolutions are correct — all 9 handler-generated POM strings use bare `<project>` tags (4.0.x convention). The MNG-2387, MNG-4360, and MNG-4554 fixes are all correctly included.

One minor style note: in `MavenITmng4235HttpAuthDeploymentChecksumsTest.java` (line ~174), `Request.getPathInContext(request)` is still called 3 times in `RepoHandler.handle()` without being hoisted — same as in the main branch. Non-blocking.

PR #12701, opened 2026-08-07
[4.0.x] Restore the Google Cloud mirror for CI resolution

slachiewicz:
Backport of #12700 to `maven-4.0.x`.

Reverts c05b25dda5, which commented out the `mimir.session.mirrors` line in `.github/ci-mimir-session.properties`.

The mirror was disabled because `apache-maven-4.0.0-rc-6` was still a 404 on the Google Cloud mirror days after Central served it. That URL returns 200 today, so the mirror has caught up, and it is cheaper for CI than resolving from Central on every job.

Re-exposes CI to the same 404 window when the next RC drops; commenting the line out again is a one-line change if that happens.

### DIFF
--- a/.github/ci-mimir-session.properties
+++ b/.github/ci-mimir-session.properties
@@ -19,9 +19,5 @@
 
 # do not waste time on this; we maintain the version
 mimir.daemon.autoupdate=false
-# CI used the Google Cloud mirror of Central, but that mirror lags a fresh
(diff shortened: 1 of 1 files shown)


gnodet (approved):
Clean backport of PR #12700 to maven-4.0.x. The diff is identical to the master PR, the mirror URL is correct, and the change is a straightforward revert of the temporary disabling during the rc-6 sync lag.

Minor: the file still lacks a trailing newline (pre-existing, not introduced here).

PR #12707, opened 2026-08-08
Fix duplicate model problem reporting, fixes #12691

### LINKED ISSUE #12691: Every model problem is reported twice since 4.0.0-rc-6
Since 4.0.0-rc-6 every model problem is printed twice: once in a new grouped
per-file block, and again in the `ProjectBuildingException` summary that Maven
has always emitted.

### Reproduction

A single pom, no other files:

```xml
<project xmlns="http://maven.apache.org/POM/4.0.0">
  <modelVersion>99.0.0</modelVersion>
  <groupId>test</groupId>
  <artifactId>fail-build</artifactId>
  <version>0.1-SNAPSHOT</version>
  <packaging>jar</packaging>
  <invalidElementShouldFailBuild/>
</project>
```

```
mvn -B validate
```

4.0.0-rc-5 reports each problem once:

```
[ERROR]
[ERROR]   The project (…/pom.xml) has 3 errors
[ERROR]     'modelVersion' of '99.0.0' is newer than the versions supported by this version of Maven: [4.0.0, 4.1.0]. …
[ERROR]     Malformed POM …: Unrecognised tag: '…invalidElementShouldFailBuild': ParseError at [row,col]:[8,3]
[ERROR]     Message: Unrecognised tag: '…invalidElementShouldFailBuild'
[ERROR]     Failed to load project …
```

4.0.0-rc-6 reports each problem twice — the new block first, then the same
three problems again in the old one:

```
[ERROR] 3 problems were encountered while processing the POMs (3 errors):

[pom.xml]
  [FATAL] 'modelVersion' of '99.0.0' is newer than the versions supported by this Maven version (4.0.0-rc-6). …
  [ERROR] Malformed POM …: Unrecognised tag: '…invalidElementShouldFailBuild' @ …/pom.xml
  [ERROR] Failed to load project …/pom.xml

[ERROR]
[ERROR]   The project (…/pom.xml) has 3 errors
[ERROR]     'modelVersion' of '99.0.0' is newer than the versions supported by this Maven version (4.0.0-rc-6). …
[ERROR]     Malformed POM …: Unrecognised tag: '…invalidElementShouldFailBuild': ParseError at [row,col]:[8,3]
[ERROR]     Message: Unrecognised tag: '…invalidElementShouldFailBuild'
[ERROR]     Failed to load project …: 2 problems were encountered while building the effective model
[ERROR]         - [FATAL] 'modelVersion' of '99.0.0' …
[ERROR]         - [ERROR] Malformed POM …
```

### Cause

cb9c4f16e2 ("Improve `ProjectBuildingException` error messages with detailed
problem reporting", #10975) changed
`ProjectBuildingException.createMessage(List<ProjectBuildingResult>)` from
returning the one-line `"Some problems were encountered while processing the
POMs"` to rendering the full per-project detail into the exception message.

That message reaches the log because `DefaultGraphBuilder` wraps the exception
in a `DefaultModelProblem`:

```java
} catch (final ProjectBuildingException | DuplicateProjectException | MavenExecutionException e) {
    return Result.error(Collections.singletonList(new DefaultModelProblem(null, null, null, null, 0, 0, e)));
}
```

Meanwhile `DefaultExceptionHandler.handle()` never uses that message — for a
`ProjectBuildingException` it discards it and rebuilds the same information from
`getResults()`:

```java
if (exception instanceof ProjectBuildingException projectBuildingException) {
    …
    message = "The build could not read " + children.size() + " project" + (children.size() == 1 ? "" : "s");
}
```

So the detail added to the message is not a replacement for the existing
rendering, it is a second copy of it.

Still present on `master`.

### Impact

Beyond the duplicated output, this breaks anything that counts occurrences of a
diagnostic. maven-invoker-plugin's `fail-build-streamLogsOnFailures` IT asserts
that a `[FATAL]` line appears twice across two forked runs and now sees four
(apache/maven-invoker-plugin#744 makes that assertion version-independent, but
the duplicated output stands on its own.)

### Suggested fix

Pick one renderer. Either revert `createMessage` to the short form and keep the
`ExceptionSummary` tree, or have `DefaultExceptionHandler` use the exception's
message and stop rebuilding the children.

Mazen050:

## Summary

Fixes #12691 by removing the duplicate rendering of model problems introduced in `ProjectBuildingException#createMessage(...)`.

The detailed per-project diagnostics were added to the exception message in #10975. However, `DefaultExceptionHandler` already reconstructs and renders the same information from `ProjectBuildingException#getResults()`, so the new exception message caused the same diagnostics to be printed twice.

This change restores the previous behavior by reverting `ProjectBuildingException#createMessage(List<ProjectBuildingResult>)` to return the short summary message, leaving the existing rendering in `DefaultExceptionHandler` as the single source of detailed diagnostics.

`ProjectBuildingExceptionTest` is removed because it was introduced to validate the detailed formatting added in #10975. Reverting that behavior makes the test obsolete; the remaining behavior is simply returning the generic exception message.

## Before

```
[ERROR] 3 problems were encountered while processing the POMs (3 errors):

[pom.xml]
  [FATAL] 'modelVersion' of '99.0.0' is newer than the versions supported by this Maven version (4.0.0-rc-6). …
  [ERROR] Malformed POM …: Unrecognised tag: '…invalidElementShouldFailBuild' @ …/pom.xml
  [ERROR] Failed to load project …/pom.xml

[ERROR]
[ERROR]   The project (…/pom.xml) has 3 errors
[ERROR]     'modelVersion' of '99.0.0' is newer than the versions supported by this Maven version (4.0.0-rc-6). …
[ERROR]     Malformed POM …: Unrecognised tag: '…invalidElementShouldFailBuild': ParseError at [row,col]:[8,3]
[ERROR]     Message: Unrecognised tag: '…invalidElementShouldFailBuild'
[ERROR]     Failed to load project …: 2 problems were encountered while building the effective model
[ERROR]         - [FATAL] 'modelVersion' of '99.0.0' …
[ERROR]         - [ERROR] Malformed POM …
```

## After

```
[ERROR]   
[ERROR]   The project (/workspaces/maven/maven-repro/pom.xml) has 3 errors
[ERROR]     'modelVersion' 99.0.0' is not supported by this Maven version (4.1.0-SNAPSHOT). Supported modelVersions are: [4.0.0, 4.1.0, 4.2.0]. Building this project requires a newer version of Maven. @ test:fail-build:0.1-SNAPSHOT, file:///workspaces/maven/maven-repro/pom.xml, line 2, column 3
[ERROR]     Malformed POM /workspaces/maven/maven-repro/pom.xml: Unable to read model: Unrecognised tag: '{http://maven.apache.org/POM/4.0.0}invalidElementShouldFailBuild': ParseError at [row,col]:[7,3]
[ERROR]     Message: Unrecognised tag: '{http://maven.apache.org/POM/4.0.0}invalidElementShouldFailBuild'
[ERROR]     Failed to load project /workspaces/maven/maven-repro/pom.xml @ org.apache.maven:maven-toolchain-builder:4.1.0-SNAPSHOT, /workspaces/maven/compat/maven-toolchain-builder/pom.xml: 2 problems were encountered while building the effective model
[ERROR]         - [FATAL] 'modelVersion' 99.0.0' is not supported by this Maven version (4.1.0-SNAPSHOT). Supported modelVersions are: [4.0.0, 4.1.0, 4.2.0]. Building this project requires a newer version of Maven. @ line 2, column 3
[ERROR]         - [ERROR] Malformed POM /workspaces/maven/maven-repro/pom.xml: Unable to read model: Unrecognised tag: '{http://maven.apache.org/POM/4.0.0}invalidElementShouldFailBuild'
[ERROR] 
[ERROR] 
```

Each model problem is now reported once again, matching the pre-4.0.0-rc-6 behavior.

Note: Please consider backporting this change to maven-4.0.x, as it fixes a regression affecting the 4.0.x release line.


gnodet (approved):
Clean, well-scoped fix for the duplicate model problem reporting regression introduced in #10975. The root cause is correctly identified: `DefaultMaven.buildGraph()` logs problem details from `ModelProblem.getMessage()` (which falls back to the exception message when its own message is null), and then `DefaultExceptionHandler` independently rebuilds the same details from `getResults()`, producing every model problem twice.

Reverting the exception message to the short static string is the right fix — it targets the duplication at its source without changing the higher-risk logging or exception handling paths. The test deletion is appropriate since all five methods validated the detailed message formatting being removed.

Worth noting: the regression was also backported to maven-4.0.x via commit b8a0ae05a8, so a backport of this fix is warranted there as well.

rmannibucau (changes requested):
Can be worth a test showing the fixed issue to ensure it doesn't come back, the pr just validates an impl detail not the feature so bug can pop up again any time IMHO

Mazen050:
Hello, I added the test.

The test covers both paths directly: it verifies that `ProjectBuildingException.getMessage()` remains the short summary, while `DefaultExceptionHandler` still renders the problem detail exactly once. That way it catches both the original regression and any future regression where the handler stops showing the detail.

gnodet (approved):
The new regression test properly addresses @rmannibucau's feedback — it validates user-facing behavior (no duplicate error rendering in console output), not just implementation details.

**What the test covers:**
- Exercises the real `DefaultExceptionHandler.handleException()` pipeline (the same code path used by `MavenInvoker.doExecute()` in production)
- Verifies `ProjectBuildingException.getMessage()` returns only the short summary (preventing duplication when logged alongside handler output)
- Verifies the handler's `ExceptionSummary` tree contains each problem detail exactly once via `countOccurrences(flatten(summary), ...)`
- The `flatten()` helper simulates what `MavenInvoker.logSummary()` does in production

The original fix remains correct and complete. The deletion of `ProjectBuildingExceptionTest` is appropriate since it tested the detailed formatting behavior introduced in #10975, which this PR reverts.

rmannibucau (approved):
ok for me, for the record I would have preferred a higher level test which does validate the exact CLI output instead of a test validating two classes (exception and handler, validating both is weirdish in a single test even if I get where it comes from in this PR context, it is less obvious in project context)

gnodet (approved):
Clean, well-scoped fix for a user-visible regression (#12691) where every model problem was rendered twice since 4.0.0-rc-6. The root cause (detailed exception message in `ProjectBuildingException` duplicating the rendering already done by `DefaultExceptionHandler`) is correctly identified and the fix is minimal and appropriately tested.

- The regression test covers a single problem of FATAL severity. Testing with multiple problems of mixed severity (ERROR + FATAL) would strengthen the assertion, though the existing handler code path is well-established.
- The deleted `ProjectBuildingExceptionTest` was introduced by the same PR (#10975) that caused the regression — deletion is appropriate since the tests validated the formatting being reverted.
- The PR description's suggestion to backport to `maven-4.0.x` is warranted since the regression was also backported there (commit b8a0ae05a8).
- All reviewer feedback (rmannibucau's request for a regression test) has been addressed. Both maintainers have approved.

impl/maven-core/src/main/java/org/apache/maven/project/ProjectBuildingException.java
```diff
@@ -63,7 +61,7 @@ protected ProjectBuildingException(String projectId, String message, File pomFil
     }
 
     public ProjectBuildingException(List<ProjectBuildingResult> results) {
-        super(createMessage(results));
+        super("Some problems were encountered while processing the POMs");
```
mfriedenhagen: Maybe only add the count of errors?

```suggestion
        super("Encountered %d problem(s) while processing the POMs".formatted(
             results.size()));
```

impl/maven-core/src/test/java/org/apache/maven/exception/DefaultExceptionHandlerTest.java
```diff
@@ -179,4 +187,72 @@ void testHandleExceptionSelfReferencing() {
...
+
+        ProjectBuildingResult result = mock(ProjectBuildingResult.class);
+        when(result.getProjectId()).thenReturn("test:fail-build:0.1-SNAPSHOT");
+        when(result.getPomFile()).thenReturn(new File("test.xml"));
+        when(result.getProblems()).thenReturn(List.of(problem));
+
+        ProjectBuildingException exception = new ProjectBuildingException(List.of(result));
+
+        assertEquals(
+                "Some problems were encountered while processing the POMs",
```
mfriedenhagen:
```suggestion
                "Encountered 1 problem(s) while processing the POMs",
```

gnodet (approved):
Clean fix that correctly eliminates the duplicate model problem rendering regression from #10975. The approach of reverting the exception message to a short summary while keeping `DefaultExceptionHandler` as the single source of detailed diagnostics is sound and matches pre-4.0.0-rc-6 behavior.

**Minor observation (non-blocking):** The message uses `results.size()` which counts `ProjectBuildingResult` entries (project files), not actual model problems — a single POM with 3 errors would say "Encountered 1 problem(s)". If precision matters, consider counting actual problems (`result.getProblems().size()` across results) or restoring the original generic text. This was added per reviewer suggestion and is approved as-is.

The new test in `DefaultExceptionHandlerTest` properly validates the non-duplication contract using `countOccurrences` — a good end-to-end approach that tests behavior rather than implementation details.

mfriedenhagen:
Hi @Mazen050 , sorry my suggestion was wrong, see comment by @gnodet result.getProblems().size() is probably better

Mazen050:
Hello @mfriedenhagen, Since `results` is a list, just putting `result.getProblems().size()` didnt work but looping through them then summing works so I did that and checked. I also updated the test to cover this part aswell (making it 2 problems).

gnodet (approved):
Re-reviewed after new commit `7100fca9` ("use correct function to count problems and make test cover it").

The fix correctly changes `results.size()` (which counted `ProjectBuildingResult` entries, i.e., project files) to `results.stream().mapToLong(result -> result.getProblems().size()).sum()` (which counts the actual total number of `ModelProblem` entries across all results). This means a single POM with 3 errors now correctly reports "Encountered 3 problem(s)" instead of the incorrect "Encountered 1 problem(s)".

The test enhancement is well-designed: by adding a second `ModelProblem` to a single `ProjectBuildingResult` and asserting "Encountered 2 problem(s)", the test now unambiguously proves the count is problem-based rather than result-based.

mfriedenhagen:
Thanks, LGTM

mfriedenhagen (approved):

gnodet (approved):
**Clean fix — eliminates duplicate model problem rendering**

Well-scoped fix that correctly identifies the duplication source: `DefaultExceptionHandler.handle()` already iterates `getProblems()` and builds child `ExceptionSummary` nodes with full detail, so having the exception message also contain the same detail caused each problem to appear twice in the output.

The solution — reducing `ProjectBuildingException`'s message to a short count-only summary while leaving `DefaultExceptionHandler` as the single source of detailed diagnostics — is architecturally sound. Net reduction of ~156 lines with a solid regression test that directly validates the fix.

### Minor non-blocking observations

- **Null safety change**: The new `results.stream().mapToLong(...)` would NPE on null `results`, but all production callers pass non-null lists (`List.of(result)` and stream-collected lists). The deleted defensive code was unreachable.
- **Grammar simplification**: `"problem(s)"` replaces the old singular/plural distinction. Cosmetic, acceptable for a summary-only message.
- **Test coverage**: Single-problem uniqueness assertion is sufficient — both problems go through the identical rendering path.

---

PR #12719, opened 2026-08-09
Move the core IT suite off the JUnit 4 assertions (4.0.x)

slachiewicz:
Five classes in `its/core-it-suite` were already annotated with the Jupiter `@Test` but still called `org.junit.Assert`, which is the only thing still pulling JUnit 4 onto the IT classpath. They now use `org.junit.jupiter.api.Assertions`.

Four needed nothing but the static import swapped. The fifth is the one worth a second look:

```java
-        assertTrue("Jar output path was not built", Files.isRegularFile(jarPath));
+        assertTrue(Files.isRegularFile(jarPath), "Jar output path was not built");
```

The message is the **first** argument in JUnit 4 and the **last** in JUnit 5. Left as a plain import swap it would still compile — `assertTrue(String, boolean)` has no Jupiter overload, so that one actually fails to compile, but the equivalent `assertEquals(message, expected, actual)` form does not, and silently asserts the wrong thing. Worth watching for in similar cleanups.

### Verification — please read before merging

I could not run this suite. The `its` reactor needs the `0.1-stub-SNAPSHOT` plugin artifacts and `2.1-SNAPSHOT` it-support artifacts built first, and that build does not complete in my environment. What I did instead:

- a standalone `javac` compile check of the changed files against a classpath assembled from the cached `maven-it-helper` jar plus `junit-jupiter-api`/`opentest4j`/`apiguardian`, with **no JUnit 4 jar present** — clean
- confirmed no `org.junit.Assert` reference remains anywhere in `its/core-it-suite/src/test/java`
- confirmed no line exceeds 120 characters (the `its` reactor does not configure spotless, so there is no formatter to run)

So this is compile-checked, not test-run. Given the change is import-level plus one argument reorder, CI should settle it — but I would rather say that plainly than imply I ran the ITs.

The same change is on the 4.0.x line as a separate PR.

Generated-by: Claude Opus 5 (1M context)

### DIFF
--- a/its/core-it-suite/src/test/java/org/apache/maven/it/MavenITgh10210SettingsXmlDecryptTest.java
+++ b/its/core-it-suite/src/test/java/org/apache/maven/it/MavenITgh10210SettingsXmlDecryptTest.java
@@ -22,8 +22,10 @@ import java.io.File;
 import java.nio.file.Files;
 import java.util.Arrays;
 
-import org.junit.Assert;
 import org.junit.jupiter.api.Test;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertTrue;
 
 /**
  * This is a test set for <a href="https://github.com/apache/maven/issues/10210">GH-10210</a>.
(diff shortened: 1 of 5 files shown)


gnodet (approved):
Clean and correct JUnit 4 to JUnit 5 assertion migration across 5 integration test files (4.0.x backport). The `assertTrue` argument reorder in `MavenITmng7772CoreExtensionFoundTest` is properly handled (JUnit 4 `assertTrue(String, boolean)` → JUnit 5 `assertTrue(boolean, String)`). The remaining 4 files are straightforward import swaps with no semantic changes. CI is green across all completed checks.

Issue #12724, opened 2026-08-10
Profile-injected content from a parent POM is silently dropped in some reactor modules


reckart:
### Affected version

4.0.0-rc-6

### Symptom

A profile declared in a parent POM and activated with `-P<id>` contributes its content (plugin
executions, properties, dependencies) to only a subset of the reactor modules. No warning, no error
— the goals simply never run. Which modules are affected depends on declaration order, and varies
between runs unless `-Dmaven.modelBuilder.parallelism=1` is set.

The issue was hit when trying to build [DKPro Core](https://github.com/dkpro/dkpro-core/issues) using Maven 4.0.0-rc-5 and rc-6. Works fine with the latest Maven 3. In particular, when trying to build with javadoc JARs attached, fewer then half of the modules actually built the JAR and others not. I observed a similar problem on another project where with Maven 4 a file-signing profile was not consistently executed.

### Cause

`DefaultModelBuilder.readAsParentModel()` caches the fully-assembled parent model — profiles already
injected, then stripped with `.withProfiles(List.of())` — keyed by a
`DefaultProfileActivationContext.Record` of the activation inputs consulted while assembling it. The
cache is keyed only on the parent's `Source` and lives in the session cache, so it is shared by
every child of that parent.

`DefaultProfileActivationContext.stop()` discards profile entries whose value is `false`:

```java
Record stop() {
    // only keep keys for which the value is `true`
    record.usedActiveProfiles.values().removeIf(value -> !value);
    record.usedInactiveProfiles.values().removeIf(value -> !value);
    return new Record(record);
}
```

So a parent assembled while `-Prelease` was *not* active records `usedActiveProfiles = {}` instead of
`{release=false}`. `Record.matches()` only verifies the keys it holds — an empty map matches **any**
context, including one where `-Prelease` *is* active.

The entry that poisons the cache comes from dependency resolution, not from the reactor. Both
`doLoadDependencyManagement()` (BOM imports) and `DefaultArtifactDescriptorReader.loadPom()`
(plugin/extension resolution) build models without `activeProfileIds` — correctly, since `-P` must
not influence resolution of third-party artifacts. When such a build walks a parent chain that
reaches a POM also used by reactor modules, it stores an all-matching `act={}` entry for that parent.
Every module built afterwards hits it and receives a parent model into which the profile was never
injected. Since the profiles were stripped after injection, there is no chance to re-inject.

This is why the cutoff is positional (it falls wherever the first BOM-importing module sits) and why
[...]

PR #12728, opened 2026-08-11
fix: keep all profile activation keys in the cache, including inactive ones

waterWang:
### Problem

`DefaultProfileActivationContext.stop()` filters out `false`-valued entries from `usedActiveProfiles` and `usedInactiveProfiles` before freezing the `Record`. This means a parent POM assembled without `-Prelease` stores an empty map (`usedActiveProfiles = {}`), which matches ANY context — including one where `-Prelease` is active. The poisoned cache entry is then shared by every reactor module, causing profile-injected content (plugin executions, properties, dependencies) to silently vanish from a subset of modules.

Root cause: the cache key is too narrow — it only records profiles that were *active*, not profiles that were consulted but found *inactive*.

### Fix

Remove the two `removeIf` filters so every consulted profile key is retained, regardless of whether it evaluated to `true` or `false`. The `Record.matches()` method already correctly verifies keys it holds, so a `false`-valued entry correctly prevents a mis-match.

### Test

Added `ParentProfileCacheTest` with two assertions that fail on the unpatched code and pass with the fix:
- A parent assembled without `-Prelease` must not be reused for a module built with `-Prelease`
- A parent assembled without `-!release` must not be reused for a module built with `-!release`

### References

Full root-cause analysis: https://github.com/apache/maven/issues/12724

### DIFF
--- a/impl/maven-impl/src/main/java/org/apache/maven/impl/model/DefaultProfileActivationContext.java
+++ b/impl/maven-impl/src/main/java/org/apache/maven/impl/model/DefaultProfileActivationContext.java
@@ -237,10 +237,7 @@ public class DefaultProfileActivationContext implements ProfileActivationContext
     }
 
     Record stop() {
-        // only keep keys for which the value is `true`
         Objects.requireNonNull(record, "start() must be called before stop()");
-        record.usedActiveProfiles.values().removeIf(value -> !value);
-        record.usedInactiveProfiles.values().removeIf(value -> !value);
         return new Record(record); // Return immutable copy for thread-safe caching
     }
 
--- a/impl/maven-impl/src/test/java/org/apache/maven/impl/model/ParentProfileCacheTest.java
+++ b/impl/maven-impl/src/test/java/org/apache/maven/impl/model/ParentProfileCacheTest.java
@@ -1,0 +1,57 @@
+package org.apache.maven.impl.model;
+
+import java.util.List;
+import java.util.Map;
+
+import org.apache.maven.api.model.Model;
+import org.apache.maven.impl.model.DefaultProfileActivationContext.Record;
+import org.junit.jupiter.api.Test;
+
+import static org.junit.jupiter.api.Assertions.assertFalse;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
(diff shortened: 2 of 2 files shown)


gnodet (approved):
Correct, minimal fix for a real cache-key bug in profile activation context. The `removeIf` calls in `stop()` were stripping `false` entries from the activation maps, causing empty maps that matched any context via `allMatch` — letting `matchesProfiles` incorrectly treat contexts where a profile was consulted but inactive as equivalent to contexts where the profile was never seen.

The fix (removing those two `removeIf` lines) is clean and well-targeted, and the new test directly validates the cache-key distinction.

**Minor nit:** Missing trailing newline at end of `ParentProfileCacheTest.java`.

gnodet (approved):
Correct, minimal fix for a cache-key bug in profile activation context. The two `removeIf` calls in `stop()` stripped `false`-valued entries, producing empty maps that matched any context via `allMatch` on an empty stream. Removing them lets `Record.matches()` work as designed. The test directly validates the fix.

- The root cause analysis in issue #12724 is excellent. The `removeIf` calls were an incorrect optimization introduced in commit 90457e519a (MNG-8736), which tried to keep cache keys small by only storing active profiles. However, `allMatch` on an empty stream returns `true`, so an empty map matched ANY context.
- The `replayRecordIntoContext` method correctly uses `putIfAbsent` to propagate cached record entries. With this fix, `false`-valued profile entries will also be replayed, which is correct.
- Missing trailing newline at end of `ParentProfileCacheTest.java` — minor POSIX convention issue.

gnodet (commented):
This PR appears to be superseded by the already-merged PR #12762 (commit 91210fc746), which contains the identical production fix (removing the two `removeIf` calls in `DefaultProfileActivationContext.stop()`) and equivalent test coverage placed in `DefaultModelBuilderTest.java`.

The production code lines this PR removes are already absent from master, so this PR would either conflict or be a no-op on merge. The standalone `ParentProfileCacheTest.java` is redundant with the tests already added by #12762.

A backport of #12762 to `maven-4.0.x` is already in progress (PR #12801).

This PR can likely be closed as duplicate.

---

PR #12732, opened 2026-08-11
fix: synchronize ReflectionValueExtractor class-introspection cache for thread safety

### LINKED ISSUE #12731: [4.0.0-rc-5, rc-6] ReflectionValueExtractor's unsynchronized WeakHashMap cache corrupts under -T and hangs the build in an infinite loop
**Component:** Maven Core (Maven 4), `maven-impl` — `org.apache.maven.impl.model.reflection.ReflectionValueExtractor`

**Affects:** 4.0.0-rc-5 (observed live); code inspection shows the same structure at tag `maven-4.0.0-rc-6`.

## Summary

`ReflectionValueExtractor` keeps a static class-introspection cache:

```java
private static final Map<Class<?>, WeakReference<ClassMap>> CLASS_MAPS = new WeakHashMap<>();
```

`getClassMap` reads and writes it with an unsynchronized check-then-act. Under the multithreaded builder (`-T`), every plugin-parameter expression evaluation (`PluginParameterExpressionEvaluatorV4` → `ReflectionValueExtractor.evaluate`) can `put` into this map from a different builder thread concurrently.

`WeakHashMap` re-links bucket chains entry-by-entry during `resize`/`transfer` (the pre-Java-8 `HashMap` design; `HashMap` was redesigned in Java 8 but `WeakHashMap` retains the old linkage). Two threads resizing simultaneously can interleave the re-links so a bucket chain becomes circular. No exception is thrown; the next operation that walks the bucket follows the cycle forever. The build hangs at 100% CPU with no output and no error — indefinitely.

## Observed incident

A 4-project reactor (macOS/aarch64, JDK 25, `-T1C` via `.mvn/maven.config`, 14 cores) running `mvn -B compile javadoc:jar -DskipTests` hung for 45+ minutes, one core pinned, zero filesystem writes. Thread dump excerpt:

```
"mvn-builder-integration-tests-example" … cpu=2590789.93ms RUNNABLE
    at java.util.WeakHashMap.transfer(java.base@25.0.3/WeakHashMap.java:540)
    at java.util.WeakHashMap.resize(java.base@25.0.3/WeakHashMap.java:506)
    at java.util.WeakHashMap.put(java.base@25.0.3/WeakHashMap.java:479)
    at org.apache.maven.impl.model.reflection.ReflectionValueExtractor.getClassMap(ReflectionValueExtractor.java:295)
    at org.apache.maven.impl.model.reflection.ReflectionValueExtractor.getPropertyValue(ReflectionValueExtractor.java:271)
    at org.apache.maven.impl.model.reflection.ReflectionValueExtractor.evaluate(ReflectionValueExtractor.java:208)
    at org.apache.maven.impl.model.reflection.ReflectionValueExtractor.evaluate(ReflectionValueExtractor.java:140)
    at org.apache.maven.plugin.PluginParameterExpressionEvaluatorV4.evaluate(PluginParameterExpressionEvaluatorV4.java:156)
    at org.apache.maven.plugin.PluginParameterExpressionEvaluatorV4.evaluate(PluginParameterExpressionEvaluatorV4.java:95)
    at org.apache.maven.plugin.PluginParameterExpressionEvaluatorV4.evaluate(PluginParameterExpressionEvaluatorV4.java:118)
    at org.codehaus.plexus.component.configurator.converters.basic.AbstractBasicConverter.fromExpression(AbstractBasicConverter.java:105)
    …
```

43+ CPU-minutes inside `WeakHashMap.transfer` — the bucket-cycle signature. The main thread was parked in `MultiThreadedBuilder.multiThreadedProjectTaskSegmentBuild`, so the whole build waits forever on the corrupted worker.

## Reproduction character

Intermittent by nature (the corrupting interleave needs two threads to resize the same table concurrently). Likelihood scales with parameter-expression density × builder threads: our reactor runs many plugin executions per module with `-T1C` on 14 cores and has hit this twice — once as this hang and once, weeks earlier, as a `ConcurrentModificationException` from the same call path (the "lucky" symptom of the same unsafe sharing). A tight loop constructing mojo configurations with many distinct bean classes across threads should reproduce it synthetically.

## Suggested fix

Make the cache a `ConcurrentHashMap` keyed by `Class` with `WeakReference<ClassMap>` values (as now), or wrap access in synchronization. The map need not be weak-keyed for correctness of this fix so long as values remain weak; if weak keys are required for classloader hygiene, `Collections.synchronizedMap(new WeakHashMap<>())` is the minimal change, and a striped/concurrent weak cache the better one. Related precedent in the Maven family: maven-resolver's `GenericVersionScheme`/`WeakInternPool` thread-contention fix (and its follow-up after the synchronization showed hot-path cost) — the same hazard class, already accepted and fixed there.

## Impact

Any Maven 4 `-T` build can hang forever with no diagnostic. In unattended CI a wedged release/build holds agents indefinitely — the failure mode is silence, not an error.

Full thread dump available on request.

waterWang:
### Problem

`ReflectionValueExtractor` keeps a static `WeakHashMap` class-introspection cache that is read and written concurrently from multiple builder threads under `-T` (parallel builds). `WeakHashMap` is not thread-safe — concurrent `resize` from multiple threads can create circular bucket chains, causing the build to hang indefinitely at 100% CPU.

**Observed symptom**: `mvn -T1C` hangs with one core pinned at 100%, thread dump shows `WeakHashMap.transfer` in a RUNNABLE state inside `getClassMap()`.

### Fix

Wrap the `WeakHashMap` with `Collections.synchronizedMap()` to serialize all reads and writes. This preserves the weak-reference semantics (classloader GC is unaffected) while providing thread safety.

### Changes

- `ReflectionValueExtractor.java`: wrap `CLASS_MAPS` with `Collections.synchronizedMap()`, add `import java.util.Collections`

Fixes #12731

### DIFF
--- a/impl/maven-impl/src/main/java/org/apache/maven/impl/model/reflection/ReflectionValueExtractor.java
+++ b/impl/maven-impl/src/main/java/org/apache/maven/impl/model/reflection/ReflectionValueExtractor.java
@@ -23,6 +23,7 @@ import java.lang.ref.WeakReference;
 import java.lang.reflect.Array;
 import java.lang.reflect.InvocationTargetException;
 import java.lang.reflect.Method;
+import java.util.Collections;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
@@ -48,7 +49,8 @@ public class ReflectionValueExtractor {
      * This approach prevents permgen space overflows due to retention of discarded
      * classloaders.
      */
-    private static final Map<Class<?>, WeakReference<ClassMap>> CLASS_MAPS = new WeakHashMap<>();
+    private static final Map<Class<?>, WeakReference<ClassMap>> CLASS_MAPS =
+            Collections.synchronizedMap(new WeakHashMap<>());
 
     static final int EOF = -1;
 


gnodet (approved):
Correct minimal fix for a real production hang. Wrapping the unsynchronized `WeakHashMap` with `Collections.synchronizedMap()` prevents concurrent `resize()`/`transfer()` corruption under `-T` parallel builds.

The remaining TOCTOU race in `getClassMap` (non-atomic get-null-check-put) is benign since `ClassMap` construction is idempotent — a duplicate creation is wasted work but not a correctness issue.

**Non-blocking observation:**

The deprecated compat copy at `compat/maven-model-builder/src/main/java/org/apache/maven/model/interpolation/reflection/ReflectionValueExtractor.java` (line 54) has the identical unsynchronized `WeakHashMap` pattern. While the main code path goes through the `impl` version (which this PR fixes), the compat copy is still reachable via `ObjectBasedValueSource` and could theoretically trigger the same hang under `-T`. Consider applying the same fix there for completeness.

knowledge-graphlet:
Thank you :-)

kec:
Reporter of #12731 here. Two things that might help this land in a release.

**Milestone.** This PR is currently unmilestoned — would 4.0.0-rc-7 (milestone 131) be reasonable? The failure mode is an indefinite hang with no output and no exception under `-T`, so in unattended CI an affected build holds an agent until something external kills it; there is nothing to react to. We hit it twice on rc-5 — once as this hang, once weeks earlier as a `ConcurrentModificationException` from the same call path — and rc-6 carries the same code.

**Completing the fix.** @gnodet's review flagged that the deprecated compat copy at `compat/maven-model-builder/src/main/java/org/apache/maven/model/interpolation/reflection/ReflectionValueExtractor.java:54` has the identical unsynchronized `WeakHashMap` and the same non-atomic `getClassMap`, still reachable through `ObjectBasedValueSource`. I have that change ready in exactly the form used here. Happy to send it either way:

- as a PR against `waterWang:fix/reflection-weak-hashmap-concurrency`, keeping this a single commit, or
- as a companion PR once this one merges.

@waterWang — your call; say which you'd prefer and I'll open it.

I can also attach the full thread dump from the original incident to #12731 if that is useful for the record.

gnodet:
I'd rather have a single PR

kec:
@gnodet understood — single PR it is.

I've sent the compat copy to @waterWang's branch as https://github.com/waterWang/maven/pull/1. Merging it there folds both files into this PR, so apache/maven still sees one PR rather than a follow-up.

@waterWang — it's the same two-line change as yours, applied to `compat/maven-model-builder/.../interpolation/reflection/ReflectionValueExtractor.java`. Merge at your convenience and this PR picks it up automatically; no action needed from me after that.

gnodet (commented):
**COMMENT** — The synchronization fix for the impl copy is correct and minimal. Wrapping the unsynchronized `WeakHashMap` with `Collections.synchronizedMap()` prevents the infinite-loop hang caused by concurrent resize/transfer corruption under `-T` parallel builds.

However, as @gnodet noted, the PR remains incomplete:

**Missing compat copy fix** (high): The identical unsynchronized `WeakHashMap` exists in `compat/maven-model-builder/.../ReflectionValueExtractor.java` at line 54, with the same non-thread-safe `getClassMap()` pattern at lines 292-303. This copy is still reachable via `ObjectBasedValueSource` and can trigger the same infinite-loop hang under `-T`. @gnodet requested both fixes in a single PR ("I'd rather have a single PR"). @kec sent the fix to waterWang/maven#1, but that PR is still open and has not been merged into this PR's branch.

**Benign TOCTOU race** (low, no action needed): The `getClassMap()` method does `CLASS_MAPS.get()` → null check → `CLASS_MAPS.put()`. These are individually atomic under `synchronizedMap`, but the compound operation is not. This is benign because `ClassMap` construction is deterministic and idempotent — the worst case is duplicate construction (wasted CPU, not a correctness issue).

gnodet (commented):
The `Collections.synchronizedMap()` fix for the impl copy is correct and minimal — it prevents the `WeakHashMap` internal structural corruption (circular bucket chains during `resize`/`transfer`) that causes infinite loops at 100% CPU under parallel builds.

**Remaining issue:**

The compat copy at `compat/maven-model-builder/src/main/java/org/apache/maven/model/interpolation/reflection/ReflectionValueExtractor.java` line 54 has the identical unsynchronized `WeakHashMap` and is reachable via `ObjectBasedValueSource.getValue()` → `ReflectionValueExtractor.evaluate()`. @gnodet requested both fixes in a single PR ("I'd rather have a single PR"). The fix was sent as waterWang/maven#1 but remains unmerged.

Once that PR is merged into this branch, the fix will be complete.

---

kec:
@gnodet — following up on my 14 Aug note. waterWang/maven#1 has had no movement and the fork has been quiet since 11 Aug, so with #12731 now milestoned for 4.0.0-rc-7 I've put both halves into one merge-ready PR as an alternative vehicle: #12841. It is @waterWang's commit from this PR cherry-picked verbatim — authorship preserved — plus the compat-copy commit on top; CI there exercises the two changes together.

Whichever route is less work:

- this PR picks up the compat half (@waterWang merges waterWang/maven#1, or a committer pushes the commit here directly — maintainer edits are enabled), and I close #12841 immediately, or
- #12841 merges as the single PR, with @waterWang's commit and credit intact, and this one closes.

No preference between them.

gnodet:
Superseded by #12841 (now merged), which carries your commit verbatim with authorship preserved plus the matching fix to the deprecated compat copy in `maven-model-builder`. Backport to `maven-4.0.x` is up as #12862.

Thank you for the fix, @waterWang!

PR #12735, opened 2026-08-12
Make profile activation conditions locale-independent

SEPURI-SAI-KRISHNA:
### What

Uses `Locale.ROOT` in the three places where profile activation condition evaluation was
implicitly relying on the JVM default locale:

- `ConditionFunctions.upper(..)` — `toUpperCase()` → `toUpperCase(Locale.ROOT)`
- `ConditionFunctions.lower(..)` — `toLowerCase()` → `toLowerCase(Locale.ROOT)`
- `ConditionParser.toString(..)` — `String.format("%.0f", ..)` → `String.format(Locale.ROOT, "%.0f", ..)`

### Why

Profile activation should not depend on the locale of the machine running the build, but today
it does. Turkish and Azeri map `i` to the dotted `İ` and `I` to the dotless `ı`, so this profile:

```xml
<profile>
  <activation>
    <condition>upper(${os.name}) == 'WINDOWS'</condition>
  </activation>
</profile>
```

silently fails to activate on a JVM started with `-Duser.language=tr`, because `upper('windows')`
returns `WİNDOWS`. Reproduction:

```
$ jshell -R-Duser.language=tr
jshell> "windows".toUpperCase()
$1 ==> "WİNDOWS"
jshell> "LINUX".toLowerCase()
$2 ==> "lınux"
```

`ConditionParser.toString(Object)` has the same class of problem: it formats whole doubles with
`String.format("%.0f", ..)` and no `Locale`, so under a locale whose default numbering system is
not latin (e.g. `hi-IN-u-nu-deva`) it renders `42` as `४२`.

This brings the three call sites in line with the rest of the codebase, which already passes an
explicit `Locale` in ~41 places — including `ExecutableFinder`, in this same package.

### Why the existing tests missed it

Every string the current tests convert (`hello`, `WORLD`, `success`, `HELLO WORLD`) happens to
contain no `i` or `I`, so `ConditionParserTest` passes unchanged under `-Duser.language=tr`.

### Tests

Two tests added to `ConditionParserTest`, both of which fail without the production change:

- `testCaseConversionFunctionsAreLocaleIndependent` — exercises `upper`/`lower` under `en` and `tr`
- `testNumberFormattingIsLocaleIndependent` — exercises number rendering under `hi-IN-u-nu-deva`

They follow the existing save/restore pattern from
`VersionTest.testCaseInsensitiveOrderingOfQualifiersIsLocaleIndependent` in the same module.

Reverting only the production change and re-running gives:

```
[ERROR] ConditionParserTest.testCaseConversionFunctionsAreLocaleIndependent
        upper() in tr ==> expected: <WINDOWS> but was: <WİNDOWS>
[ERROR] ConditionParserTest.testNumberFormattingIsLocaleIndependent
        expected: <42> but was: <४२>
```

With the change, the full `maven-impl` module suite passes: 552 tests, 0 failures, checkstyle /
spotless / rat clean.

If this is wanted on `maven-4.0.x` as well, happy for it to be backported.

---

## PR checklist

The template ships with this checklist. Tick these:

Leave UNTICKED unless you actually run it (see note below):

- [ ] You have run the Core IT successfully.

The ICLA line can stay unticked — the production change is ~20 lines, under the template's
"~20 lines of code" threshold.

### DIFF
--- a/impl/maven-impl/src/main/java/org/apache/maven/impl/model/profile/ConditionFunctions.java
+++ b/impl/maven-impl/src/main/java/org/apache/maven/impl/model/profile/ConditionFunctions.java
@@ -19,6 +19,7 @@
 package org.apache.maven.impl.model.profile;
 
 import java.util.List;
+import java.util.Locale;
 
 import org.apache.maven.api.services.InterpolatorException;
 import org.apache.maven.api.services.ModelBuilderException;
@@ -65,6 +66,9 @@ public class ConditionFunctions {
 
     /**
      * Converts the given string to uppercase.
+     * <p>
+     * The conversion uses {@link Locale#ROOT} so that profile activation does not depend on the
+     * default locale of the machine running the build.
      *
      * @param args A list containing a single string argument
      * @return The uppercase version of the input string
@@ -75,11 +79,14 @@ public class ConditionFunctions {
             throw new IllegalArgumentException("upper function requires exactly one argument");
         }
         String s = ConditionParser.toString(args.get(0));
-        return s.toUpperCase();
+        return s.toUpperCase(Locale.ROOT);
     }
 
     /**
      * Converts the given string to lowercase.
+     * <p>
+     * The conversion uses {@link Locale#ROOT} so that profile activation does not depend on the
+     * default locale of the machine running the build.
      *
      * @param args A list containing a single string argument
      * @return The lowercase version of the input string
@@ -90,7 +97,7 @@ public class ConditionFunctions {
             throw new IllegalArgumentException("lower function requires exactly one argument");
         }
         String s = ConditionParser.toString(args.get(0));
-        return s.toLowerCase();
+        return s.toLowerCase(Locale.ROOT);
     }
 
     /**
--- a/impl/maven-impl/src/main/java/org/apache/maven/impl/model/profile/ConditionParser.java
+++ b/impl/maven-impl/src/main/java/org/apache/maven/impl/model/profile/ConditionParser.java
@@ -20,6 +20,7 @@ package org.apache.maven.impl.model.profile;
 
 import java.util.ArrayList;
 import java.util.List;
+import java.util.Locale;
 import java.util.Map;
 import java.util.function.UnaryOperator;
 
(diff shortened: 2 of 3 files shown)


gnodet (approved):
Well-scoped, correct fix for a real locale-sensitivity bug in profile activation condition evaluation. All three locale-sensitive call sites are properly fixed (`upper()` and `lower()` in `ConditionFunctions.java`, `toString()` in `ConditionParser.java`).

The tests are thorough and follow the established save/restore pattern from `VersionTest.testCaseInsensitiveOrderingOfQualifiersIsLocaleIndependent`. Good choice of test inputs — strings containing `i`/`I` (affected by Turkish locale case mapping) and a locale with Devanagari numbering for the number formatting test.

**Minor nit:** `new Locale("tr")` is deprecated since Java 19 in favor of `Locale.of("tr")`, but the existing codebase uses the same deprecated constructor in `VersionTest.java`, so this is consistent.

Excellent PR description with reproduction steps and explanation of why existing tests missed the bug.

gnodet (approved):
Clean, correct, and well-tested fix for a real locale-sensitivity bug in profile activation condition evaluation. All three locale-dependent call sites are properly fixed with `Locale.ROOT`, and the tests follow established codebase patterns.

- `new Locale("tr")` is deprecated since Java 19 in favor of `Locale.of("tr")`, but the project targets Java 17 and the existing `VersionTest.java` uses the same deprecated constructor — consistent with the codebase.
- Well-scoped fix addressing exactly the three locale-sensitive call sites. The Javadoc additions clearly explain why `Locale.ROOT` is needed.
- The `equalsIgnoreCase("true")`/`equalsIgnoreCase("false")` at `ConditionParser.java` line 323 is technically locale-sensitive but safe because neither "true" nor "false" contains the characters (i/I) affected by Turkish locale rules.

gnodet (commented):
**APPROVE** ✅

Clean, well-motivated fix that makes profile activation condition evaluation locale-independent by adding explicit `Locale.ROOT` to three call sites. The implementation is correct, properly scoped, and includes thorough regression tests that follow existing project patterns.

1. The test uses `new Locale("tr")` which was deprecated for removal in Java 19 (JEP 418). Since Maven targets Java 17 this is fine and matches the existing pattern in `VersionTest.java:259`, but if the project ever bumps to Java 19+ this will need updating to `Locale.of("tr")`.

2. The three production changes (`toUpperCase`/`toLowerCase` with `Locale.ROOT`, `String.format` with `Locale.ROOT`) are exactly the right fix for locale-sensitive string operations in build infrastructure code. The fix is consistent with the existing codebase — the broader maven-impl module already uses `Locale.ROOT` in ~3 other places.

PR #12737, opened 2026-08-12
Support the logical not operator in profile activation conditions

### LINKED ISSUE #12736: `!` is tokenized as an operator in profile activation conditions but cannot be parsed
### Affected version

4.1.0-SNAPSHOT (master)

### Bug description

The condition tokenizer in `ConditionParser` treats `!` as an operator character. When `!` is
followed by `=` it emits a single `!=` token; otherwise it emits a bare `!` token:

```java
} else if (Character.isWhitespace(c) || ... || c == '!' || ...) {
    ...
    } else {
        tokens.add(String.valueOf(c));   // emits a bare "!"
    }
```

No parser rule ever consumes that bare `!`. `parseUnary()` handles only `-`, so a `!` falls through
to `parseTerm()`, fails `Double.parseDouble("!")`, and lands in `parseVariableOrUnknownFunction()`:

```
<condition>!exists('some/path')</condition>

Error parsing profile activation condition: Unknown variable: !
```

So the natural spelling of negation produces a confusing error that talks about variables. The
workaround is the `not()` function, which already exists and does exactly this:

```xml
<condition>not(exists('some/path'))</condition>
```

A tokenizer that emits a token no parser rule can consume looks like an unfinished implementation
rather than a deliberate omission. Making `!` sugar for the existing `not()` function is a small
change and adds no new capability.

SEPURI-SAI-KRISHNA:
Fixes #12736

### What

Adds handling for the `!` operator to `ConditionParser.parseUnary()`.

### Why

The tokenizer already treats `!` as an operator and emits a bare `!` token, but no parser rule
consumes it. Writing the natural form of a negated condition:

```xml
<condition>!exists('some/path')</condition>
```

fails with `Unknown variable: !`, which is confusing — nothing in the expression is a variable.

`!` is defined here as sugar for the `not()` function that already exists, using the same
`toBoolean()` coercion, so this adds no new capability and no new semantics. It is handled in
`parseUnary()`, giving it the same precedence as in Java: tighter than arithmetic, comparison,
`&&` and `||`.

`!=` is unaffected — the tokenizer emits it as a single token before a bare `!` can be produced.

### Tests

Four tests added to `ConditionParserTest`, all failing before the change:

- `testLogicalNotOperator` — basic negation, stacking (`!!`, `!!!`), and negating a function call
- `testLogicalNotOperatorMatchesNotFunction` — pins `!x` to the same result as `not(x)`
- `testLogicalNotOperatorPrecedence` — precedence against `&&`, `||` and parenthesised groups
- `testLogicalNotOperatorCoercion` — string/number coercion matches `not()`

The negation tests deliberately use `contains('Hello, World!', ..)`, which puts a `!` inside a
string literal, to confirm the tokenizer still leaves quoted `!` alone.

---

## Out of scope — a separate bug found while testing this

`compare()` handles `Number` vs `Number` and `String` vs `String`, but has no `Boolean` branch, so
comparing two booleans throws on the **unmodified** parser:

```
true == true    => RuntimeException: Cannot compare true and true with operator ==
true != false   => RuntimeException: Cannot compare true and false with operator !=
```

This is pre-existing and independent of `!`. It is deliberately NOT fixed in this PR, and the
tests here avoid boolean-to-boolean comparison. Candidate for a fourth PR.

---

## PR checklist — APPEND THIS TO THE END OF THE PR DESCRIPTION

This is the repo's `pull_request_template.md` checklist. Paste it below the body above.

- [ ] You have run the [Core IT](https://maven.apache.org/core-its/core-it-suite/) successfully.

### DIFF
--- a/impl/maven-impl/src/main/java/org/apache/maven/impl/model/profile/ConditionParser.java
+++ b/impl/maven-impl/src/main/java/org/apache/maven/impl/model/profile/ConditionParser.java
@@ -288,11 +288,21 @@ public class ConditionParser {
     }
 
     /**
-     * Parses unary operations (negation).
+     * Parses unary operations: logical not ({@code !}) and arithmetic negation ({@code -}).
+     * <p>
+     * {@code !} is equivalent to the {@code not()} function and uses the same {@link #toBoolean}
+     * coercion. Being handled here gives both operators a higher precedence than multiplication,
+     * addition, comparison and the logical operators, so {@code !a && b} parses as
+     * {@code (!a) && b}.
      *
      * @return the result of parsing unary operations
      */
     private Object parseUnary() {
+        if (current < tokens.size() && tokens.get(current).equals("!")) {
+            current++;
+            Object value = parseUnary();
+            return !toBoolean(value);
+        }
         if (current < tokens.size() && tokens.get(current).equals("-")) {
             current++;
             Object value = parseUnary();
--- a/impl/maven-impl/src/test/java/org/apache/maven/impl/model/profile/ConditionParserTest.java
+++ b/impl/maven-impl/src/test/java/org/apache/maven/impl/model/profile/ConditionParserTest.java
@@ -159,6 +159,71 @@ class ConditionParserTest {
         assertEquals("SUCCESS!", parser.parse(expression));
     }
 
+    /**
+     * The tokenizer emits a bare {@code !} token, so the parser has to be able to consume it.
+     * The {@code contains(..)} cases also cover a {@code !} inside a string literal, which the
+     * tokenizer must leave alone.
+     */
+    @Test
+    void testLogicalNotOperator() {
+        assertFalse((Boolean) parser.parse("!true"));
+        assertTrue((Boolean) parser.parse("!false"));
+        assertTrue((Boolean) parser.parse("!!true"));
+        assertFalse((Boolean) parser.parse("!!!true"));
+        assertTrue((Boolean) parser.parse("!contains('Hello, World!', 'OpenAI')"));
+        assertFalse((Boolean) parser.parse("!contains('Hello, World!', 'World')"));
+    }
+
+    /**
+     * {@code !x} is sugar for the {@code not(x)} function and must agree with it.
+     */
+    @Test
+    void testLogicalNotOperatorMatchesNotFunction() {
+        assertEquals(
+                parser.parse("not(contains('Hello, World!', 'OpenAI'))"),
+                parser.parse("!contains('Hello, World!', 'OpenAI')"));
+        assertEquals(
+                parser.parse("not(contains('Hello, World!', 'World'))"),
+                parser.parse("!contains('Hello, World!', 'World')"));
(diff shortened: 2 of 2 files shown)


gnodet (approved):
Clean, well-implemented addition of the logical `!` operator to the `ConditionParser`. The implementation correctly mirrors the existing `not()` function by calling the same `toBoolean()` coercion, and the recursive `parseUnary()` call naturally supports stacking (`!!`, `!!!`).

The placement in `parseUnary()` gives `!` the expected Java-like precedence (tighter than arithmetic, comparison, `&&`, and `||`), and the tokenizer already correctly emits `!=` as a single token before a bare `!` can be produced, so no regression is possible there.

The four new test methods provide strong coverage of basic negation, equivalence with `not()`, precedence interactions, and type coercion. Nice touch using `contains('Hello, World!', ...)` to exercise `!` inside string literals.

**Minor nit:** `testNotEqualsStillParsesAsOneOperator` is redundant with existing assertions in `testStringComparison` and `testArithmeticComparisons`, though it serves as a self-documenting regression guard for this change.

Thorough PR description with correct scoping — the pre-existing `compare()` issue is correctly identified as out of scope.

gnodet (approved):
Clean, minimal addition of the logical `!` operator to `ConditionParser.parseUnary()`, correctly reusing the existing `toBoolean()` coercion to make `!x` semantically identical to the already-supported `not(x)` function. The 5 lines of production code are well-placed in the recursive descent grammar, and the 65 lines of tests provide strong coverage.

- Precedence is correct: handling `!` in `parseUnary()` gives it higher precedence than arithmetic, comparison, `&&`, and `||`, matching Java semantics.
- The recursive `parseUnary()` call naturally enables stacking (`!!x`, `!!!x`) without additional code.
- The `!=` operator is unaffected because the tokenizer emits `!=` as a single two-character token before a bare `!` can be produced.
- The `testNotEqualsStillParsesAsOneOperator` test is slightly redundant with existing tests, but serves as a self-documenting regression guard tied to this change.

gnodet (commented):
**APPROVE** ✅

Clean, well-scoped PR that closes a real gap between the tokenizer (which already emits bare `!` tokens) and the parser (which had no rule to consume them). The implementation is correct, minimal, and thoroughly tested.

1. The implementation correctly mirrors the existing `not()` function in `ConditionFunctions.java`, using the identical `toBoolean()` coercion. The recursive `parseUnary()` call properly handles stacked negation (`!!`, `!!!`) and gives `!` the same precedence as unary minus, matching Java semantics.

2. The tokenizer already handles the `!` vs `!=` distinction correctly (when `!` is followed by `=`, it emits a single `!=` token; otherwise a bare `!`). No tokenizer changes were needed.

3. Test coverage is thorough: basic negation, double/triple negation, function call negation, equivalence with `not()`, precedence against `&&`/`||`/comparison/parenthesized groups, and string/number coercion. The use of `contains('Hello, World!', ...)` to verify that `!` inside string literals is not treated as an operator is a nice touch.

4. The pre-existing issue noted in the PR description (boolean-to-boolean comparison throwing in `compare()`) is correctly left out of scope.

PR #12741, opened 2026-08-12
[MNG-8129] Handle InvalidPathException for broken relativePath on Windows

gnodet:
## Summary

Backport of #12740 to `maven-4.0.x` for RC-7.

- Catch `InvalidPathException` in `BuildPathSource.resolve()` and `FileModelSource.getRelatedSource()` when the `<relativePath>` value contains characters illegal on certain platforms (e.g. `:` on Windows)
- Return `null` instead of throwing, matching the behavior on platforms where `Path.resolve()` succeeds but the path doesn't exist
- Some POMs on Maven Central have nonsensical `<relativePath>` values like `org.apache:apache` (e.g. `artemis-project-2.33.0.pom`) which trigger this on Windows

Fixes https://issues.apache.org/jira/browse/MNG-8129
Related: #12738

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #12740, opened 2026-08-12
[MNG-8129] Handle InvalidPathException for broken relativePath on Windows

gnodet:
## Summary

- Catch `InvalidPathException` in `BuildPathSource.resolve()` (new API) and `FileModelSource.getRelatedSource()` (compat) and return `null`, letting the caller fall through to repository-based parent resolution
- Adds tests for both code paths

## Problem

Some POMs on Maven Central have `<relativePath>` set to a GAV coordinate instead of a filesystem path. For example, [`artemis-project-2.33.0.pom`](https://repo.maven.apache.org/maven2/org/apache/activemq/artemis-project/2.33.0/artemis-project-2.33.0.pom) contains:

```xml
<relativePath>org.apache:apache</relativePath>
```

On Windows, colons are reserved for drive letters, so `Path.resolve("org.apache:apache")` throws `InvalidPathException` before any file-existence check can run. On Linux/macOS the same call succeeds (`:` is valid in paths), and the subsequent `Files.isRegularFile()` check returns `false`, so the invalid path is harmlessly ignored.

## Approach

Catch `InvalidPathException` and return `null` — an invalid path obviously cannot point to an existing parent POM, so returning `null` is the correct semantic (no local parent found → fall through to repository resolution). This matches Maven 3.x behavior.

Closes #12738

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #12742, opened 2026-08-12
feat: add relativePath validation

cstamas:
As it seems no Maven 3.x version did validate it. Validation fails if project being built (uses strict validation), while dependencies are not failing (uses minimal validation).

### DIFF
--- a/maven-model-builder/src/main/java/org/apache/maven/model/validation/DefaultModelValidator.java
+++ b/maven-model-builder/src/main/java/org/apache/maven/model/validation/DefaultModelValidator.java
@@ -80,6 +80,7 @@ public class DefaultModelValidator implements ModelValidator {
     private static final Pattern CI_FRIENDLY_EXPRESSION = Pattern.compile("\\$\\{(.+?)}");
     private static final Pattern EXPRESSION_PROJECT_NAME_PATTERN = Pattern.compile("\\$\\{(project.+?)}");
 
+    // see below; multiple fields depend on this field
     private static final String ILLEGAL_FS_CHARS = "\\/:\"<>|?*";
 
     private static final String ILLEGAL_RELATIVE_PATH_FS_CHARS =
@@ -156,7 +157,9 @@ public class DefaultModelValidator implements ModelValidator {
         } else if (request.getValidationLevel() >= ModelBuildingRequest.VALIDATION_LEVEL_MAVEN_2_0) {
             Severity errOn30 = getSeverity(request, ModelBuildingRequest.VALIDATION_LEVEL_MAVEN_3_0);
 
-            if (parent != null && parent.getRelativePath() != null) {
+            if (parent != null
+                    && parent.getRelativePath() != null
+                    && !parent.getRelativePath().isEmpty()) {
                 validateBannedCharacters(
                         "parent.",
                         "relativePath",
--- a/maven-model-builder/src/main/java/org/apache/maven/model/validation/DefaultModelValidator.java
+++ b/maven-model-builder/src/main/java/org/apache/maven/model/validation/DefaultModelValidator.java
@@ -83,7 +83,7 @@ public class DefaultModelValidator implements ModelValidator {
     private static final String ILLEGAL_FS_CHARS = "\\/:\"<>|?*";
 
     private static final String ILLEGAL_RELATIVE_PATH_FS_CHARS =
-            ":\"<>|?*"; // same as above but FS sep removed (unix+win)
+            ILLEGAL_FS_CHARS.replace("\\", "").replace("/", "");
 
     private static final String ILLEGAL_VERSION_CHARS = ILLEGAL_FS_CHARS;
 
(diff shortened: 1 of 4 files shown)


gnodet (commented):
Thanks for looking into this! A few concerns about the approach:

### Not a backport of a Maven 4 validation

Maven 4's `DefaultModelValidator` has **no** `validateBannedCharacters` call on `parent.relativePath`. The only relativePath validation in Maven 4 is the 4.1.0-specific check about "only specify relativePath or groupId/artifactId" — which is unrelated. So this is a net-new validation, not a backport.

### Severity is too aggressive

The PR uses `Severity.ERROR` with `Version.BASE`, meaning it would **break builds** that previously worked fine. Both existing `validateBannedCharacters` calls in 3.10.x use `errOn31` (gradual severity), not a hard error. Maven 3.x has always silently handled these — the path resolution returns null and Maven falls back to repository resolution. Introducing a hard error in a minor release would be a breaking change.

### Banned character list is inconsistent with the resolver

The PR bans `\` (backslash), but `FileModelSource.getRelatedSource()` explicitly does:
```java
relPath = relPath.replace('\\', File.separatorChar);
```
Backslashes are expected and handled as path separators. Banning them in validation contradicts the resolver.

### Would break existing projects

POMs on Maven Central (like `artemis-project-2.33.0.pom`) have `<relativePath>org.apache:apache</relativePath>`. When Maven resolves dependencies from such POMs, it goes through `getRelatedSource()` — if the path doesn't resolve, Maven falls back to repository resolution. This validation would fail builds that have always worked.

### What Maven 4 does instead

The approach taken in #12740 / #12741 is to **catch `InvalidPathException` at the resolution layer** and return null — matching the behavior on platforms where the path is syntactically valid but doesn't exist. No validation error, no warning — just graceful handling of broken `relativePath` values from POMs users can't control.

gnodet (commented):
Correction to my earlier comment: Maven 4 **does** reject broken `relativePath` at build time, but through a different mechanism — it happens in the **model builder** during parent resolution, not in the model validator.

For `BUILD_PROJECT` requests, when `source.resolve(locator, parentPath)` returns null (which is what happens after #12740 catches `InvalidPathException`), `wrongParentRelativePath()` fires a `Severity.FATAL` error:

> `'parent.relativePath' points at 'org.apache:apache' but no POM could be found, please verify your project structure`

For dependency POMs (non-`BUILD_PROJECT`), it silently falls through to repository resolution — no error.

So the end result for project builds is similar to what this PR does (reject with error). The differences are:
- **Where**: model builder (parent resolution) vs model validator (POM validation)
- **Severity**: `FATAL` (Maven 4) vs `ERROR` (this PR)
- **Message**: Maven 4's message is more helpful ("no POM could be found, please verify your project structure") vs a generic banned-characters message

The concern about the banned character list being inconsistent with the resolver (banning `\` when the resolver explicitly handles it as a path separator) still stands.

cstamas:
Also, AI is wrong in "Would break existing projects", this is false. Maven for _dependencies_ uses MIN validation, where this branch does not kick in at all. For projects it uses STRICT/MAX validation, and yes, the goal is to fail projects. What is released (and is used as dependency) will work, but 3.10 will prevent new, broken in this respect (invalid value in built project POM parent relativePath) to be built and deployed. User will need to fix it.

gnodet (approved):
Clean, well-scoped PR that adds previously missing banned-character validation for `parent.relativePath` on the Maven 3.10.x branch. The implementation follows existing patterns in `DefaultModelValidator` and includes both positive and negative tests.

**Non-blocking suggestion:**

`ILLEGAL_RELATIVE_PATH_FS_CHARS` is hardcoded as a string literal rather than derived from `ILLEGAL_FS_CHARS`. The existing constants `ILLEGAL_VERSION_CHARS` and `ILLEGAL_REPO_ID_CHARS` both reference `ILLEGAL_FS_CHARS` directly. If `ILLEGAL_FS_CHARS` were ever updated, this constant could silently get out of sync. Consider:

```java
private static final String ILLEGAL_RELATIVE_PATH_FS_CHARS =
        ILLEGAL_FS_CHARS.replace("\\", "").replace("/", "");
```

**Note:** The Maven 4 `DefaultModelValidator` doesn't currently have equivalent validation for `parent.relativePath`. If this is important enough for 3.10.x, a follow-up PR for the 4.x line may be worth considering.

gnodet (approved):
Clean, well-scoped PR that adds banned-character validation for `parent.relativePath`, following existing patterns in `DefaultModelValidator`. The implementation correctly derives the banned character set from `ILLEGAL_FS_CHARS` (excluding path separators `\` and `/`), uses gradual severity via `errOn30`, and only applies during project validation (not dependency/minimal validation).

- The previous APPROVED review's non-blocking suggestion (derive `ILLEGAL_RELATIVE_PATH_FS_CHARS` from `ILLEGAL_FS_CHARS` instead of hardcoding) has already been implemented in commit 1b5250382d.
- The validation correctly sits inside the `validationLevel >= VALIDATION_LEVEL_MAVEN_2_0` block, so dependency POMs processed with `VALIDATION_LEVEL_MINIMAL` skip this check entirely.
- The use of `errOn30` (gradual severity) is consistent with the other existing `validateBannedCharacters` calls in the file.
- Minor: new test POMs use tab indentation while existing test POMs in the same directory use 2-space indentation.

PR #12743, opened 2026-08-12
[MNG-8129] Validate relativePath for illegal filesystem characters

gnodet:
## Summary

- Add validation in `DefaultModelValidator` to reject `<relativePath>` values containing characters illegal in filesystem paths (`: " < > | ? *`)
- Uses `errOn31` severity (WARNING at default strict level) to avoid breaking existing builds
- Path separators (`/` and `\`) are intentionally excluded from the banned list since the resolver handles them
- Catches nonsensical values like `org.apache:apache` early with a clear error message

Note: Maven 3.10.x uses `java.io.File` for path resolution (not `java.nio.file.Path`), so there is no `InvalidPathException` crash — `new File()` silently accepts any string. The validator check is sufficient for this branch.

Companion to #12740 (master) and #12741 (maven-4.0.x) which also fix the `InvalidPathException` crash in the `Path.resolve()` code paths.

Fixes https://issues.apache.org/jira/browse/MNG-8129

### DIFF
--- a/maven-model-builder/src/main/java/org/apache/maven/model/validation/DefaultModelValidator.java
+++ b/maven-model-builder/src/main/java/org/apache/maven/model/validation/DefaultModelValidator.java
@@ -82,6 +82,8 @@ public class DefaultModelValidator implements ModelValidator {
 
     private static final String ILLEGAL_FS_CHARS = "\\/:\"<>|?*";
 
+    private static final String ILLEGAL_RELATIVE_PATH_CHARS = ":\"<>|?*";
+
     private static final String ILLEGAL_VERSION_CHARS = ILLEGAL_FS_CHARS;
 
     private static final String ILLEGAL_REPO_ID_CHARS = ILLEGAL_FS_CHARS;
@@ -152,6 +154,24 @@ public class DefaultModelValidator implements ModelValidator {
             }
         } else if (request.getValidationLevel() >= ModelBuildingRequest.VALIDATION_LEVEL_MAVEN_2_0) {
             Severity errOn30 = getSeverity(request, ModelBuildingRequest.VALIDATION_LEVEL_MAVEN_3_0);
+
+            Severity errOn31 = getSeverity(request, ModelBuildingRequest.VALIDATION_LEVEL_MAVEN_3_1);
+
+            // [MNG-8129] Validate that relativePath does not contain characters that are illegal in filesystem paths
+            if (parent != null
+                    && parent.getRelativePath() != null
+                    && !parent.getRelativePath().isEmpty()) {
+                validateBannedCharacters(
+                        "parent.",
+                        "relativePath",
+                        problems,
+                        errOn31,
+                        Version.V20,
+                        parent.getRelativePath(),
+                        null,
+                        parent,
+                        ILLEGAL_RELATIVE_PATH_CHARS);
+            }
 
             // [MNG-6074] Maven should produce an error if no model version has been set in a POM file used to build an
             // effective model.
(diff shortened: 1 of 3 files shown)


maven-model-builder/src/main/java/org/apache/maven/model/validation/DefaultModelValidator.java
```diff
@@ -153,6 +155,24 @@ public void validateRawModel(Model m, ModelBuildingRequest request, ModelProblem
         } else if (request.getValidationLevel() >= ModelBuildingRequest.VALIDATION_LEVEL_MAVEN_2_0) {
             Severity errOn30 = getSeverity(request, ModelBuildingRequest.VALIDATION_LEVEL_MAVEN_3_0);
 
+            Severity errOn31 = getSeverity(request, ModelBuildingRequest.VALIDATION_LEVEL_MAVEN_3_1);
+
+            // [MNG-8129] Validate that relativePath does not contain characters that are illegal in filesystem paths
```
elharo, line 160: Which characters are illegal varies from filesystem to filesystem. Is there some maven level escaping happening here? Or do you mean URL paths?

elharo (changes requested):

cstamas:
@gnodet what is happening with this PR?

maven-model-builder/src/main/java/org/apache/maven/model/validation/DefaultModelValidator.java
gnodet: Good point — "illegal in filesystem paths" is too broad. Updated the comment to be precise:

> Validate that relativePath does not contain characters reserved on Windows (NTFS).
> These cause InvalidPathException in Maven 4 when resolved via java.nio.file.Path, and typically
> indicate the user put a GAV coordinate (e.g. "g:a:v") instead of an actual filesystem path.

The characters banned here (`:` `"` `<` `>` `|` `?` `*`) come from the existing `ILLEGAL_FS_CHARS` constant (minus `\` and `/` which are path separators). The same set is already used to validate versions and repository IDs. On NTFS these are reserved and `java.nio.file.Path` throws `InvalidPathException`; on ext4/HFS+ they're technically valid but a `:` in `<relativePath>` almost always means someone wrote `groupId:artifactId` notation instead of an actual path.

No Maven-level escaping or URL paths involved — `<relativePath>` is meant to be a literal filesystem path relative to the POM's directory.

gnodet (commented):
### ✅ Looks Good

Focused severity downgrade of the existing relativePath validation from ERROR to WARNING on the 3.10.x branch, companion to already-merged #12740 (master) and #12741 (maven-4.0.x). The change is minimal, correct, and well-tested.

**Technical notes:**

- The severity downgrade is technically sound: `VALIDATION_LEVEL_STRICT` equals `VALIDATION_LEVEL_MAVEN_3_0` (30), so `errOn31 = getSeverity(30, 31)` resolves to WARNING at the default strict level. This avoids breaking existing builds that happen to have characters like `:` in their `<relativePath>` (e.g., users who mistakenly wrote GAV notation). The version marker change from `Version.BASE` to `Version.V20` correctly narrows the scope.
- The existing test `testBadParentRelativePath1` is properly updated to expect a warning instead of an error, and a new test `testBadParentRelativePath` is added with a distinct test POM, providing adequate coverage.
- @elharo's concern about "illegal" wording appears addressed by commit 313b781 which clarifies to "reserved on Windows (NTFS)" with context about `InvalidPathException` in Maven 4.

### 🔀 Backport Status

- ✅ `master` — #12740 (MERGED)
- ✅ `maven-4.0.x` — #12741, #12849 (MERGED)
- ℹ️ `maven-3.9.x` — not needed (relativePath validation was first introduced in #12742 on 3.10.x)

---
<sub>🤖 This review was generated by [ForgeBot](https://github.com/gnodet/forgebot).</sub>

